PR #185, opened 2014-03-07
Add gloo.glplus namespace

almarklein:
Functions from OpenGL.GL except deprecated and what we use in
gloo.gl. Safes over half the functions on my machine. Injects
all constants for now.


Eric89GXL:
@almarklein how difficult would it be to resurrect this to get it working? Looks like an increasing number of PRs could potentially benefit from it.

almarklein:
Not too hard. Mostly a matter of some design choices. I will take some time to think about this some more soon.

Eric89GXL:
Tests pass over here, I'll merge once Travis is happy.

almarklein:
Travis is green. Names are now:

* gl2 - subset of desktop GL corresponding to ES 2.0
* es2 - real ES 2.0
* pyopengl2 - same as gl2, but using pyopengl underneath
* gl+ - same as pyopengl2, but with all constants and non-deprecated functions from PyOpenGL added in.

That way, we can support ES 3.0 at some point without having to rename.

Eric89GXL:
Excellent, thanks!

PR #206, opened 2014-03-24
Add flake_diff make target

lcampagn:
Idea test: have travis flake-check only code that is touched by the current PR. This allows us to keep stricter flake tests, with the same yellow-light warning. BUT: with this change, we have the option of admitting code that fails flake but nevertheless passes the eyeball test. After such code is committed to the repository, flake will no longer complain about it (unless it is touched again).

(not ready for review)


Eric89GXL:
@lcampagn I think the color and size changes have been merged elsewhere, okay to close this one?

PR #505, opened 2014-08-31
Refactored util/transforms.py translate, scale and rotate to just return

rreilink:
Refactored translate, scale and return to just return a matrix to be multiplied instead of the matrix multiplication being done in the method

Adapted scenen linear transforms as appropriate

Adapted examples as appropriate (except for ipython notebooks, cannot test those)


rossant:
Decision to make here...

Eric89GXL:
At the very least, this needs to be put in a merge-able state. Since merges were done before instead of rebases, I think it would need another merge with `master`. I'm not sure how difficult the merge might be, it could be easier to start from scratch. @rreilink do you have time to take a look? If not, I might also have time to take this over.

Eric89GXL:
Closing for upcoming PR

PR #612, opened 2014-11-08
WIP: Volume rendering & camera changes

### LINKED ISSUE #725: Image visual crashes when turntable camera is used.
**Issue**: Adding an Image visual to a scenegraph viewed by a turntable camera causes Vispy to throw a TypeError when the scene is drawn.

**Cause**: InverseTransform does not define the Linear/Orthogonal/NonScaling/Isometric flags correctly, and/or ChainTransform does not correctly handle cases where these flags are NoneType.

The Image visual uses the .Linear property of its transform as an optimization hint during its draw call. In this configuration of the scenegraph the ChainTransform passed to the draw call contains an InverseTransform, which does not define the Linear/Orthogonal/NonScaling/Isometric flags (instead inheriting the flags from BaseTransform, which are initialized to None). When evaulating the flag properties, ChainTransform.Linear '&='s the corresponding properties of its children, which for InverseTransform results in an illegal ```<bool> &= <NoneType>``` operation.

There are two possible solutions to this, but I'm not sure which would be preferred, which is why I'm opening an issue instead of making a pull request:
<ul>
  <li>Have InverseTransform propagate these flags from the inverted transform</li>
  <li>Have ChainTransform check for (and propagate?) None, and change the documentation of these flags to indicate they may be NoneType</li>
</ul> 

Example code (based on examples/basics/scene/image.py):
```python
import sys
from vispy import scene
from vispy import app
import numpy as np

canvas.size = 800, 600
canvas.show()

# Set up a viewbox to display the image with a turntable camera
view.set_camera('turntable')

# Create the image
img_data = np.random.normal(size=(100, 100, 3), loc=128,
                            scale=50).astype(np.ubyte)
image = scene.visuals.Image(img_data, parent=view.scene)

    app.run()
```

almarklein:
This PR implements volume rendering. It's basically a port of the technique that I use in Visvis.

* New Volume visual with two rendering styles (iso and mip). 
* Volume example in examples/scene/volume.py Try it, and use the keys to switch cameras etc.
* Refactored cameras a bit in general
* Refactored TurnTableCamera: no more mode and distance. You can just change the fov while getting the same zoom level. Setting fov to 0 means orthographic. Zooming is done by moving camera backward.
* New FlyCamera to fly around your data. 

This is about ready. There are more things to do, but that can be picked up in another PR. (I added todo items for this in volume.py)

I wish to do more with the camera's, but we should probably discuss this first. Will make a new issue for this.

todo:

* [x] avoid warning when swithching render styles
* [x] camera's

closes issue #725

volume

Closes #725


karlcz:
I thought I should continue a conversation here that @Eric89GXL started on #696... I have a working ray-cast volume renderer as a gloo application (not using any of the scene/visual stuff).  It's probably a long way from being a vispy visual class, but it might be relevant to your discussions about scope and target problems...?

I can try to sort out the open source license if the code might be of any use to you guys.  The biggest problem is that I may not have corresponding demo data to share in the near term.

It does the conventional 2-pass rendering:
1. render front and back faces with the volume unit-cube XYZ coordinates as RGB colors into high-precision FBO 2D textures
2. run the ray-cast 3D volume sampling and integration in a fragment shader for a simple full-viewport quad, using fragment coordinates to look up the corresponding ray entry/exit coordinates

It has a few useful features:
* Handle volumes that have different grid spacing and span on each axis
* A basic "crystal ball" user interface to do arbitrary rotations, zoom, and auto-rotation
* A generalized near-clipping plane solution to accommodate rotation/zoom where some voxels would fall "behind" the camera and to enable a user-controlled clipping tool to peel away the nearest voxels or to control a 2D slicing plane (as a separate shader program, basically just sampling the entry plane as it intersects the volume)
* A decent stochastic ray-caster (jittering the ray entry points)
* High-precision texture support to avoid spatial quantization with the two-pass algorithm and to allow GLSL transfer functions to work over high dynamic range volume data

It's still pretty primitive though, lacking proper widgets to control color transfer function etc. even though the GLSL code would easily allow other transfer functions.  The current app does a very basic linear transfer function with keyboard-based zero-intercept and slope increment/decrement controls.

almarklein:
@karlcz Thanks for the info. The approach in this PR calculates the ray direction and number of steps in the shader itself, and can therefore by applied in a single pass. It also has a means of clipping voxels behind the camera; flying around inside a CT/MRI scan is sweet!

It has some todo's w.r.t. anisotropic data (i.e. different grid spacing) and more efficiently transferring the data based on the required contrast limits. Also, I want to (allow it to) use a 2D texture, which should make it possible to run this in a browser. All this is for later.

> A decent stochastic ray-caster (jittering the ray entry points)

No idea what this is. Can you explain more?

karlcz:
Oh, I assumed this PR was blending a stack of grid-aligned textures because of the rendering artifact someone posted further up in the comments!  I wouldn't expect to see any gaps/slicing planes with ray-casting.  Out of curiosity, why do you want to avoid the two-pass method?

Also, do you mean it uses 2D textures to store the volume data now? Or you are considering such a change?  What sort of volume resolutions are you expecting to visualize in practice (without downsampling)?

Regarding the stochastic sampling process, it introduces a per-fragment randomization to jitter the entry-point along the calculated ray direction.  In other words, it takes a fractional-voxel step before starting the periodic integration loop.  The 3D texture is also sampled with linear interpolation in that loop.  The end result is a cheap kind of spatial dithering which makes the visualization a lot smoother under rotation and zoom. You almost always feel that you are viewing a continuous scalar field, rather than a grid of bricks.

almarklein:
@karlcz The slicing planes occur with the isosurface method, which stops the ray when a certain threshold is fount. When zooming in, due to discrete nature of stepping through the volume, the depth at which the surface is placed is discrete, and you get such artifacts. They are reduced by doing a refinement process, but to see them, you can just zoom-in a bit more. 

I added the pseudo-ramdom offset in the ray, and that makes artifacts like the above appear as dither, rather than an orientation-dependent striping. Much nicer indeed. thanks for the tip!

I use a 3D texture, but want to use a 2D texture (or the possibility to use one) to make it work on real ES 2.0 systems like webgl. I expect it to be slightly less efficient, because we need to the interpolation in z-direction in the GLSL (although I would not be surprised if some drivers/cards don't have their 3D interpolation in hardware anyway).

The volumes that you can visualize smoothly on a not-too-new machine is about 256**3. But with better cards you can go higher.

> why do you want to avoid the two-pass method?

I expect this to be be faster, since it's in one pass. Also easier to implement (once you've got the GLSL code right).

almarklein:
@lcampagn I tweaked the isosurface example. The viewbox error was already there in master. The problem with the grid example might be related to the change the viewbox camera. It might need a change in the gridlines visual to make it work again. I took a brief look, but was unable to fix it yet. Any ideas?

Also, the errors were very inconsistent, which is very annoying. Therefore I turned `Node._children` and `Node._parents` into lists. This did not remove the problems, but made them (much more) consistent.

Eric89GXL:
FYI @almarklein you can put `[size skip]` in your next commit message to skip the size checks, in case you want to see green.

Also, you need another rebase (sorry)

karlcz:
@almarklein and others: I got our basic volume rendering code factored out and posted, in case you are interested to look at it... https://github.com/informatics-isi-edu/volspy the sample file mentioned in the README is a 512x512x441 volume with two 32-bit floating point channels.  It renders around 10 fps on my GeForce GTX 760 graphics card when loaded into an 'rg32f' texture using the current vispy/master code with my merged texture internalformat PR.

almarklein:
Travis is green. Hooray!

I fixed the grid example; the PanZoomcamera now does not use of orthographic projection since that seems to make the inverse transformation ill-defined, breaking the GridLines... The viewbox example is broken. But this is also in Master. Let's fix that seperately.

Ready for merge AFAIC

Eric89GXL:
@campagnola you want to take a look at this one again?

I'll also take another look. Thanks for working through the issues on this one @almarklein.

campagnola:
Yesh, working on it.

vispy/scene/cameras/cameras.py
```diff
@@ -151,60 +428,61 @@ def _set_scene_transform(self, tr):
...
+        if not self._linked_cameras_no_update:
+            for cam in self._linked_cameras:
+                cam._linked_cameras_no_update = True
+                try:
+                    cam.set_state(self.get_state())
+                finally:
+                    cam._linked_cameras_no_update = False
 
     
 class PanZoomCamera(BaseCamera):
```
campagnola: My impression looking at this class is that its implementation was much simpler when it just used `self._rect` to keep track of its state, and the added complexity is a result of trying to force it to use scale_factor, center, and aspect ratio defined in BaseCamera. Are there new features here that I'm not accounting for?

vispy/scene/cameras/cameras.py
```diff
@@ -258,32 +518,75 @@ def rect(self, args):
...
         """
         if isinstance(args, tuple):
-            self._rect = Rect(*args)
+            rect = Rect(*args)
         else:
-            self._rect = Rect(args)
-        self._update_transform()
-
-    @property 
-    def invert_y(self):
```
campagnola: This is an important method.

vispy/scene/cameras/cameras.py
```diff
@@ -151,60 +428,61 @@ def _set_scene_transform(self, tr):
...
-        self._rect = Rect((0, 0), (1, 1))  # visible range in scene
-        self._invert = [False, True]
+        # todo: Y axis is flipped by default?
+        #self._aspect_ratio = self._aspect_ratio_n = 1.0, -1.0, 0.0
         self.transform = STTransform()
-        
-    def zoom(self, zoom, center):
-        """ Zoom the view around a center point.
+    
+    def zoom(self, scale_factor, center):
```
campagnola: Zoom should use a relative scale factor, not absolute.

vispy/scene/cameras/cameras.py
```diff
@@ -71,43 +197,191 @@ def interactive(self):
...
+        self._xlim, self._ylim, self._zlim = bounds_margins
+        # Store center location
+        self._center_loc = [(b[0] + r / 2) for b, r in zip(bounds, ranges)]
+        # Set given params (since _set_range() might derive other values)
+        for name, value in self._given_params.items():
+            setattr(self, name, value)
+        # Let specific camera handle it
+        self._set_range()
+        # Overwrite given params
+        for name, value in self._given_params.items():
```
campagnola: Can you explain what `_given_params` is for? This bit looks really awkward.

vispy/scene/cameras/cameras.py
```diff
@@ -71,43 +197,191 @@ def interactive(self):
...
-    def view_changed(self):
-        """ Called when this camera is connected to a new view.
+        E.g. (0.5, 1.0, 1.0) would make the x-dimension half as small.
+        Negative values can be used to flip dimensions, e.g. (1, -1)
+        is typically suited for visualizing images.
+        """
+        return self._aspect_ratio
+    
+    @aspect_ratio.setter
+    def aspect_ratio(self, value):
```
campagnola: I think `aspect_ratio` is the wrong name (if I understand its purpose correctly). Aspect ratio refers to the ratio of width to height of the view, which is usually independent of the ratio of scale factors for the camera. Maybe `scale_ratio` ?

vispy/scene/cameras/cameras.py
```diff
@@ -71,43 +197,191 @@ def interactive(self):
...
+        # Do we have a viewbox?
+        if self._viewbox is None:
+            return
+        self._resetting = True
+        # Size factor
+        size = self._viewbox.size
+        self._window_size_factor = size[1] / size[0]
+        # Get bounds
+        bounds = []
+        bounds_scene = self._viewbox.get_scene_bounds()
```
campagnola: Calculating bounds is very expensive. This should only be done for the axes that are not specified in the args.

vispy/scene/cameras/cameras.py
```diff
@@ -71,43 +197,191 @@ def interactive(self):
...
+        if ranges are not specified. The ranges should be 2-element
+        tuples specifying the min and max for each dimension.
+        """
+        # Do we have a viewbox?
+        if self._viewbox is None:
+            return
+        self._resetting = True
+        # Size factor
+        size = self._viewbox.size
+        self._window_size_factor = size[1] / size[0]
```
campagnola: The various uses of `size_factor` is really where `aspect_ratio` should be used..

vispy/scene/cameras/cameras.py
```diff
@@ -71,43 +197,191 @@ def interactive(self):
...
+        cam1._linked_cameras.append(cam2)
+        cam2._linked_cameras.append(cam1)
     
-    def disconnect(self):
-        self._viewbox.events.mouse_press.disconnect(self.view_mouse_event)
-        self._viewbox.events.mouse_release.disconnect(self.view_mouse_event)
-        self._viewbox.events.mouse_move.disconnect(self.view_mouse_event)
-        self._viewbox.events.mouse_wheel.disconnect(self.view_mouse_event)
-        self._viewbox.events.resize.disconnect(self.view_resize_event)
+    def view_changed(self):
```
campagnola: This was meant to be used when the camera is connected to a new view, not whenever the view range changes. Perhaps it needs a different name to make that clear..

vispy/visuals/line/line.py
```diff
@@ -92,6 +92,12 @@ def __init__(self, pos=None, color=(0.5, 0.5, 0.5, 1), width=1,
 
         self._changed = {'pos': False, 'color': False, 'width': False,
                          'connect': False}
+        
+        self._bounds = None
+        if pos is not None:
```
campagnola: DRY: bounds should be calculated in only one place, and that should be inside the bounds method (and the result should be cached).

vispy/scene/cameras/cameras.py
```diff
@@ -71,43 +197,191 @@ def interactive(self):
...
+    
+    def set_range(self, xrange=None, yrange=None, zrange=None, margin=0.05):
+        """ Set the range of the view region for the camera
+        
+        The view is reset to the given range or to the scene boundaries
+        if ranges are not specified. The ranges should be 2-element
+        tuples specifying the min and max for each dimension.
+        """
+        # Do we have a viewbox?
+        if self._viewbox is None:
```
campagnola: Should take effect even if there is no viewbox:

    cam = SomeCamera()
    cam.set_range(x=.., y=..)
    view.camera = cam

vispy/scene/cameras/cameras.py
almarklein: In particular aspect ratio is something that we want in the PanZoom camera.

vispy/scene/cameras/cameras.py
almarklein: I am not sure if I agree. The function looks to me as if calling it twice will yield the same result. I think because of the presense of the `center` parameter. You suggested a "function to allow relative zoom" earlier, but because of the presense of this function I decided that people could just do `camera.scale_factor *= 2.0`. If you know another solution I'd be happy to hear it.

vispy/scene/cameras/cameras.py
almarklein: Not anymore. Now we have `aspect_ratio`. Just do `cam.aspect_ratio = 1, -1` to flip the y-axis.

vispy/scene/cameras/cameras.py
almarklein: `_given_params` is a dict with parameters that were set upon initialization of the camera. When resetting the camera, we want to set the state back to the initial state, not some arbitrary default state. Maybe some extra comment?

vispy/scene/cameras/cameras.py
almarklein: `scale_ratio` seems appropriate. It also seems more consistent with `scale_factor`.

almarklein: Also `aspect_fixed` -> `fixed_ratio` or `fixed_scale_ratio` ?

vispy/scene/cameras/cameras.py
almarklein: I am not sure what you mean. `window_size_factor` (only used internally) applies to the window. `aspect_ratio` applies to the local coordinate frame. E.g. if the user resizes the window, the former is changed, but the latter should remain equal, and what was previously square should still be square.

vispy/scene/cameras/cameras.py
almarklein: Mmm. But if e.g. only `x` is specified, than we need to remember that (also *which* ranges were set), and take into account it is connected. Certainly possible, but code becomes a bit messy. Is that worth it do you think?

almarklein:
Thanks Luke. I will fix the points that I agree with now asap. Some other points need further discussion.

vispy/scene/cameras/cameras.py
almarklein: Done

vispy/scene/cameras/cameras.py
almarklein: Done. Though if you need all three bounds, its cheaper to get them in one go ...

vispy/scene/cameras/cameras.py
almarklein: `view_changed` is still a valid name. I updated the docstring though.

almarklein:
Ok, I addressed most of @campagnola 's comments. Discussion points that are more or less in the open:

*  PanZoomCamera does not have rect anymore. I did that to make it use scale_factor et. al instead, which is more consistent with the other camera's. Plus we now can specify the scale for each dimension, and whether a dimension is flipped.
* `PanZoomFactor.zoom()` -> should it do absolute or relative zooming?
* You cannot specify a distance with `TurnTableCamera`. I think this is a good thing. Rather than specifying a distance, you specify what you want to see (or have it done automatically) via `set_range()`. and the camera adjusts for you. If we do want a camera which allows setting a distance explicitly, we should make a new camera class IMO.

almarklein:
Further. WTF, Travis is spewing out all sorts of style errors which were previously fine. @Eric89GXL any idea?

Eric89GXL:
I'm on it

vispy/scene/cameras/cameras.py
campagnola: `zoom` is used here as a verb like "zoom in 2x", so the argument should be relative. If you wanted an absolute argument, then it should be `set_zoom` instead (but again the value would be arbitrary and the user would be forced to guess an appropriate value, which is my complaint about scale_factor as well). The problem with `cam.scale_factor *= x` is that it does not allow for the `center` argument.

campagnola: To clarify: `center` specifies the point that should remain stationary during the zoom (often the position of the mouse cursor).

vispy/scene/cameras/cameras.py
campagnola: It doesn't look like there should be any memory of the "initial state". When we call set_range, the result should just be a combination of the arguments given and the current state of the camera; no need to refer back to the initial state.

vispy/scene/cameras/cameras.py
campagnola: I just mean that the term "aspect ratio" is already defined as relating to the shape of the view. It is confusing to use it in reference to the internal coordinate system (I already made this mistake with pyqtgraph).

campagnola:
Sorry to drag this on.  `scale_factor` is not working for me. From #647:

> scale_factor seems to have two uses: setting the initial zoom level and
zooming interactively. For the first use, the user will probably prefer
something like camera.view(visual) or camera.set_range(limits). For the 
second use, I prefer something like camera.zoom(factor), which uses a
relative scale factor rather than an arbitrary absolute factor

I am concerned that enforcing the use of `scale_factor` will make the panzoom camera more difficult to work with (there is a similar class in pyqtgraph that took many, many iterations to get working correctly so I am particularly sensitive to this). I also think there is not a good argument for disallowing specification of turntable distance. 

Perhaps we need to back up and specify some real use cases and decide on the API before making further changes here.

vispy/scene/cameras/cameras.py
almarklein: Well, I suppose it aligns better with the pan method.

For the record, `scale_factor` maps directly to the amount of space that would be visible at the x-axis. I would not know how to make a less arbitrary scale measure.

almarklein:
Ok, `scale_factor` is a parameter that all camera's use internally as a measure of scale. Together with `center`, `aspect_ratio` and additional parameters for each camera, the final view is defined.

When `cam.set_range()` is called, the `aspect_ratio` is set to the appropriate value, so that the given range will appear visible. Same for `PanZoomCamera.rect()`. Users don't need to know about it, but it seems reasonable to expose it publicly (like all other parameters).

I shall also make `cam.zoom()` apply a relative zoom. This would make all methods (except `set_range()`) apply relative changes, whereas properties are all (naturally) absolute.

As for the `distance` parameter, I think it is very hard to guess what the distance should be to view a certain region of the scene, and it changes as the `fov` changes. Instead, the distance is determined automatically to make sure that the region specified in `set_range()` is visible. You can even adjust the `fov` from 0 (orthographic) to 90 and the same region is still always visible. I think this is a very useful feature.

A camera that can rotate around its center point at a certain distance is no doubt useful too, but targets other use-cases will be less easy to use. That's why I propose we make a new Camera class for it.

almarklein:
> It doesn't look like there should be any memory of the "initial state". When we call set_range, the result should just be a combination of the arguments given and the current state of the camera; no need to refer back to the initial state.

mmm, I suppose this is an artifact from when `set_range()` was still called `reset()`. Shall I just add a `reset()` method then?

campagnola:
Why do we need `reset()` at all? I don't think anything else in vispy has this behavior. If the user really wants such a feature, it is easy enough to save the initial camera state and restore it later on. It's a more flexible approach as well because it give the user the opportunity to configure the camera using its methods before storing the initial state.

almarklein:
> Why do we need reset() at all? 

MPL / Bokeh and others all have a little button to reset the view. In visvis I had it bound to double-click (which is arguably not the most convenient shortcut). It's quite handy when you're very much zoomed in and what to go back to the full view.

campagnola:
The "full view" might not have any relationship to the initial parameters used upon instantiation of the camera. Isn't it better to automatically choose some view that shows all of the data?

campagnola:
> . . . the distance is determined automatically to make sure that the region specified in set_range() is visible. You can even adjust the fov from 0 (orthographic) to 90 and the same region is still always visible. I think this is a very useful feature.

This I agree with (`set_range` is very useful) but I don't think the implementation needs to make it impossible to set the distance. We can probably come up with an API that allows both kinds of interaction.

almarklein:
Well I meant *initial view* which is usually the full view. 

> Isn't it better to automatically choose some view that shows all of the data? 

That's what `cam.set_range()` (without args) does. I imagine that `reset()` would call `set_range()` and then apply any arguments that were passed during initialization. But if we cannot agree on `reset` now, I'm happy to leave it for now and get back to the remaining issues.

almarklein:
Ok @campagnola , I added a `distance` property to TurntableCamera. By default this is None, which means that the best distance is calculated internally. If given, it is used. With distance given, zooming (with RMB) changes the distance. 

I also made `PanZoomCamera.zoom()` use a relative factor. Check the last 3 commits.

almarklein:
Ping @campagnola

examples/basics/scene/isocurve_for_trisurface.py
```diff
@@ -38,6 +37,8 @@
 # Add a 3D axis to keep us oriented
 scene.visuals.XYZAxis(parent=view.scene)
 
+view.camera = scene.TurntableCamera()
```
campagnola: I love the xy/z scaling in this example, but I overlooked it at first because I thought the scene was 2d / flat. Start with elevation=30 or so?

examples/basics/scene/one_scene_four_cams.py
```diff
@@ -0,0 +1,67 @@
...
+    image = scene.visuals.Image(im1, grid=(20, 20), parent=par)
+
+#vol1 = np.load(io.load_data_file('volume/stent.npz'))['arr_0']
+#volume1 = scene.visuals.Volume(vol1, parent=scenes)
+#volume1.transform = scene.STTransform(translate=(0, 0, 10))
+
+# Assign cameras
+vb1.camera = scene.BaseCamera()
+vb2.camera = scene.PanZoomCamera(fixed_ratio=True)
+vb3.camera = scene.TurntableCamera(fixed_ratio=True)
```
campagnola: Likewise here--this scene starts blank.

examples/basics/scene/one_scene_four_cams.py
```diff
@@ -0,0 +1,63 @@
...
+vb2.camera = 'panzoom'
+vb3.camera = 'turntable'
+vb4.camera = 'fly'
+
+# If True, show a cuboid at each camera
+if False:
+    cube = scene.visuals.Cube((3, 3, 5))
+    cube.transform = scene.STTransform(translate=(0, 0, 6))
+    for vb in (vb1, vb2, vb3, vb4):
+        vb.camera.parents = scenes
```
campagnola: I don't see the cameras anymore..

vispy/plot/plot.py
```diff
@@ -18,7 +19,10 @@ def plot(*args, **kwds):
     canvas.view = canvas.central_widget.add_view()
     canvas.line = visuals.LinePlot(*args, **kwds)
     canvas.view.add(canvas.line)
-    canvas.view.camera.auto_zoom(canvas.line)
+    if False:  # todo: of data-is-3D
```
campagnola: Can we just ask whether the bounds for the visual's z-axis is (0, 0) ?

vispy/plot/plot.py
```diff
@@ -31,15 +35,19 @@ def image(*args, **kwds):
     """
     canvas = SceneCanvas(keys='interactive')
     canvas.view = canvas.central_widget.add_view()
-    canvas.image = visuals.Image(*args, **kwds)
+    canvas.image = visuals.Image(parent=canvas.view.scene, *args, **kwds)
```
campagnola: I don't think this is necessary

vispy/plot/plot.py
```diff
@@ -31,15 +35,19 @@ def image(*args, **kwds):
...
     canvas.view.add(canvas.image)
     canvas.show()
-    canvas.view.camera.invert_y = False
-    canvas.view.camera.auto_zoom(canvas.image)
+    canvas.view.camera = cameras.PanZoomCamera(fixed_ratio=True,
+                                               scale_ratio=(1, -1, 1),
+                                               )
     plots.append(canvas)
+    # todo: (AK) I think this should return an image
```
campagnola: That's fine, so long as both the image and the canvas can be accessed somehow from the return value.

vispy/plot/plot.py
```diff
@@ -31,15 +35,19 @@ def image(*args, **kwds):
...
+    # Also, wtf: this creates an image in a viewbox in a widget in a canvas.
```
campagnola: What's the problem with that? Each object has its own purpose; that's separation of concerns.

examples/basics/scene/isocurve_for_trisurface.py
almarklein: Changed default value of elevation and azimuth of TurnTableCamera to 30.

examples/basics/scene/one_scene_four_cams.py
almarklein: Mmm, I still see them (when I set `False->True`).

vispy/plot/plot.py
almarklein: I think we might at some point place objects at nonzero z positions to achieve z ordering. E.g. move images slightly to the back so that lines/markers drawn on top are visible. In visvis I push existing images a bit backward each time `imshow()` gets called, so that the last image is always fully visible.

My proposal is that inside the "plot" functions here, we set the camera based on the given data. We could also check if the camera is already changed (from being the BaseCamera), and not set the camera in that case.

vispy/plot/plot.py
almarklein: Ah, I guess I overlooked the next line

vispy/plot/plot.py
almarklein: Let's discuss and apply in Paris :)

vispy/plot/plot.py
almarklein: Seems rather deep nesting. I can see the purpose of all, except of the widget. But we can look at this during the sprint.

almarklein:
@campagnola Thanks for the review. Addressed comments. Travis is green (except that I managed to misspell "size skip").

examples/basics/scene/one_scene_four_cams.py
campagnola: Ah, I missed the conditional this time. Let's just leave this on?

almarklein: Mmm, at this point performance gets pretty bad when turned on, and the little boxes look bad. I think it makes the demo less pretty, so I'd rather have it turned off by default.

PR #618, opened 2014-11-14
Removed closeEvent for qt backend

lcampagn:
Opening this PR for discussion:

When a canvas is closed using the Qt backend, it should be possible to show the canvas again by calling `show()` on it. It should only be closed permanently if the user requests that or if the canvas is collected.

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -317,11 +317,6 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
         g = self.geometry()
         return g.width(), g.height()
 
-    def closeEvent(self, ev):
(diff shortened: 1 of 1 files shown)


Eric89GXL:
This one looks like it needs responses from @lcampagn and @almarklein to move forward (or not)

lcampagn:
Restarted tests. Almar, any further concerns about this?

almarklein:
I think I am ok. With this PR, what happens when the user does `canvas.close()`? Will the widget be destroyed or only hidden?

lcampagn:
The only thing this PR changes is to remove the closeEvent() method; the behavior for close() is unchanged.

Eric89GXL:
Thanks for the fix.

PR #637, opened 2014-11-21
Scenegraph benchmark

lcampagn:
WIP: This PR adds a benchmark example for comparing scenegraph performance to similar optimal implementations. 

The intent is to add more benchmark examples and optimizations until we get comparable performance (or decide that we will never get comparable performance).

### DIFF
--- a/vispy/visuals/shaders/function.py
+++ b/vispy/visuals/shaders/function.py
@@ -18,6 +18,7 @@ trigger a recompile.
 """
 
 import re
+import logging
 import numpy as np
 
 from ...util.event import EventEmitter, Event
@@ -430,7 +431,7 @@ class Function(ShaderObject):
                     self.template_vars.add(var.lstrip('$'))
         
         self.changed(code_changed=True, value_changed=True)
-        if logger.level >= 10:
+        if logger.level == logging.DEBUG:
             import traceback
             last = traceback.format_list(traceback.extract_stack()[-2:-1])
             logger.debug("Assignment would trigger shader recompile:\n"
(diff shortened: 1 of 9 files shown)


vispy/visuals/shaders/function.py
```diff
@@ -430,7 +431,7 @@ def __setitem__(self, key, val):
                     self.template_vars.add(var.lstrip('$'))
         
         self.changed(code_changed=True, value_changed=True)
-        if logger.level >= 10:
+        if logger.level == logging.DEBUG:
```
Eric89GXL: I think this could be `<= logging.DEBUG` in case we ever make/use a custom logging level that is below DEBUG

Eric89GXL:
@lcampagn do you need any other info to progress on this PR, or is it ready (enough) for merge? It seemed like we were close to merge at some point, and there are some useful changes in here. WDYT?

lcampagn:
It was ready long ago.. we'll see how the tests work out now :)

Eric89GXL:
I exepect AppVeyor to fail, if you want to see full success go ahead and rebase, otherwise we can just ignore AppVeyor for now.

Eric89GXL:
You'll probably need a rebase to get the tests to run properly

lcampagn:
Don't the tests run on the merged code? If yes, then the tests should come out identical. If no, then there's obviously a big problem :)

Eric89GXL:
Unfortunately I don't think they do...

lcampagn:
If that were the case, then we would expect many PRs to show passing tests while open, but then break tests in master after merging.

Eric89GXL:
Looks like you're right, everybody is happy

Eric89GXL:
Thanks, one more PR cleared :)

Issue #551, opened 2014-09-09
Visuals should make use of document coordinate system


campagnola:
Most visuals make the assumption that the document and framebuffer coordinate systems are the same. This can be seen most easily by running `examples/basics/plot/export.pt`  (requires #352) -- the exported image should look roughly the same as the original canvas, but currently the line width and marker sizes are fixed to the framebuffer pixels, rather than to document pixels. 

We can write simple unit tests for this by rendering a scene once at native resolution, and again at 2x resolution, and then checking `image1 == image2[::2, ::2]`. (this requires antialiasing to be disabled to get a good match).

Eric89GXL:
@nippoo se also #99 and #639

PR #639, opened 2014-11-22
Fixing GLFW Tests & Support for Retina Devices

alexjc:
(This pull request is not ready to merge yet, but is intended to continue the discussion of #551 specifically for the GLFW use-case and fixing the tests.)

Currently with GLFW, the tests fail on Retina displays:
````
======================================================================
FAIL: Test drawing to a framebuffer
  File "vispy/gloo/tests/test_use_gloo.py", line 43, in test_use_framebuffer
    assert_equal(c.size, shape[::-1])
nose.proxy.AssertionError: Tuples differ: (600, 200) != (300, 100)

First differing element 0:
600
300

- (600, 200)
?  ^    ^
+ (300, 100)
?  ^    ^
````

Based on @lcampagn's comment: assuming `canvas.size` should return the window size, I tried changing the GLFW backend to return WindowSize instead of FrameBuffer size in both `_vispy_get_size` and `_on_resize`, that fixes all the tests.  (See the diff.)

Now, however, the rest doesn't work and the viewport (in particular) seems wrong.  This is the problem, I think:
````
vispy.gloo.set_viewport(0, 0, *event.size)
````

Should I pursue this direction?  How easy will it be to handle the glViewport sizing in a retina friendly way?  I'm starting to think it'd be easier for `canvas.size` to return the full retina framebuffer size and have a separate call for the window size.

### DIFF
--- a/vispy/app/backends/_glfw.py
+++ b/vispy/app/backends/_glfw.py
@@ -263,7 +263,7 @@ class CanvasBackend(BaseCanvasBackend):
 
         # Register callbacks
         glfw.glfwSetWindowRefreshCallback(self._id, self._on_draw)
-        glfw.glfwSetFramebufferSizeCallback(self._id, self._on_resize)
+        glfw.glfwSetWindowSizeCallback(self._id, self._on_resize)
         glfw.glfwSetKeyCallback(self._id, self._on_key_press)
         glfw.glfwSetMouseButtonCallback(self._id, self._on_mouse_button)
         glfw.glfwSetScrollCallback(self._id, self._on_mouse_scroll)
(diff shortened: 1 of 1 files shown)


nippoo:
Hey @alexjc, I was going to work on this today/tomorrow. Before I do, have you made any progress? Don't want to reinvent the wheel!

alexjc:
@nippoo No progress, sorry. It's been a busy start to the year and I didn't get around to it! Good luck :-)

Eric89GXL:
Merging for now so @nippoo can take it up, thanks for your help with the issues @alexjc

PR #654, opened 2014-12-01
Marker Visual

bdurin:
Remark: the purpose of this pull request is to prepare a much more interesting pull request where I implemented a simple abstract renderer for markers (which explain why I became maniac on marker and edge sizes)

- corrected GLSL function for markers where some key points are defined relative to fragment size instead of being
defined relative to marker size $v_size: the defaults were visible in the demo/gloo/how_markers.py example on the
clobber marker that was kind of dislocating when becoming too small; they could be seen also as deformation of
bars or cross markers when displayed at very different sizes.
- changed frag rendering so that the size of the marker is the guaranted size in pixel of the face. The edge is
now outside the face and if too small will be mixed with face antialiasing or not displayed if below 1/10th pixel wide
- added marker function rescaling so that the edge have the same width for the same edge_width parameter whatever
the marker (and as far as I checked the edge width is exactly edge_width wide in pixels) and whatever the part
of the marker (before tailed_arrow had a wider edge on the tail than on the arrow)
- added very common markers triangle up and down and star (at least in matplotlib) and changed the total size of the
fragment in vert and frag for the edge to be rendered correctly
- added a RescalingRelativeEdgeMarkerVisual that rescales as the scene is rescaled (see the added example below)
and for which edge is expressed as a fraction of the marker size
This new marker visual can be used with all markers and rescale the whole marker as the minimum of x and y rescaling.
- added a RescalingXYRelativeEdgeMarkerVisual that is similar to the previous one but rescale independently in the x
and y directions. It works only with specific markers called markersxy which are defined for arbitrary x and y sizes
($v_size becomes a vec2). Some of these markers keep a 1:1 aspect ratio (called square markers), using only the min
or the max of the x and y size. The others used both sizes (hbar, vbar, cross, x, rect and ellipse). Note that the
distinction between hbar and vbar does not much sense because of the 2 sizes but is kept for the moment.
Ellipse marker code may seem complicated but the key problem is to correctly rescale the function so that the edge
width is regular around the perimeter of the ellipse. The correct rescaling is given by the gradient of the level set
defining the marker (the marker function in the code) taken at the point on the ellipse for which the normal goes
through the point currently considered in the fragment shader. I could not find a way to keep the code of this
computation siple so I used an approximation: the point on the ellipse is the point corresponding to the hyperbola
constant coordinate in elliptic coordinate system (http://en.wikipedia.org/wiki/Elliptic_coordinate_system).
It's much simpler to compute and the result is nice looking (you can see some defaults by rescaling the ellipse so
that it has a large size and is almost flat).


vispy/visuals/markers.py
```diff
@@ -282,24 +514,36 @@ def set_data(self, pos=None, style='o', size=10., edge_width=1.,
...
+            The width of the symbol outline in px if larger than 1,
+            the width as a fraction of marker size if strictly smaller than 1
         edge_color : Color | ColorArray
             The color used to draw each symbol outline.
         face_color : Color | ColorArray
             The color used to draw each symbol interior.
+        scaling : bool
+            If set to True, marker scales when rezooming
+            (needs a viewbox to work).
+        small_edge_fade_out : bool
```
Eric89GXL: @bdurin so let's get rid of this option for now (so that the API / arguments do not grow). I think that's all that's left to do before we can merge.

vispy/visuals/markers.py
```diff
@@ -282,24 +514,36 @@ def set_data(self, pos=None, style='o', size=10., edge_width=1.,
             The array of locations to display each symbol.
         style : str
             The style of symbol to draw (see Notes).
-        size : float
+        size : float or array
             The symbol size in px.
         edge_width : float
-            The width of the symbol outline in px.
+            The width of the symbol outline in px if larger than 1,
+            the width as a fraction of marker size if strictly smaller than 1
```
Eric89GXL: Oh, this and the above parenthetical comment as well.

Eric89GXL:
@bdurin do you have time to make the final tweak to this code? If not, I can take over the PR and merge.

bdurin:
Hi Eric and all the team,
Sorry these last weeks I could not focus on the PR and next weeks also because I currently don't have access to the computer where my vispy and the rest fo the environment is set up.
So yes, Eric, if you can just remove the option for me, that would be nice and the PR could be closed.

Eric89GXL:
Thanks for the quick response, I'll open a new PR (your commits will be preserved)

PR #678, opened 2014-12-17
Fallback on xrandr to get linux DPI info.

anntzer:
xdpyinfo is not always available on Linux (even with a DE), whereas
xrandr is (AFAICT) at least dependency of KDE and XFCE.  But xrandr
fails when used remotely, so keep it too.


Eric89GXL:
@anntzer sorry to sit on this one for a while, is this ready for merge from your end?

PR #694, opened 2015-01-05
New wx_panel backend to embed a Canvas in a wx.Frame

robmcmullen:
NOTE: I don't actually expect this to be merged! I'm just showing what I had to do to get a vispy canvas embedded in a wx GUI.

But as I said on the vispy mailing list, I'm just starting out here and don't know much about vispy yet. But because the normal wx backend was subclassed from wx.Frame, it can't be embedded in a complex wx GUI because wx.Frame objects are always top level windows. Here, I subclass from wx.Panel which can then be included in a window with e.g. a menu bar.


Eric89GXL:
I wonder if we should just make the current behavior to be a `wx.Panel` instead of a `wx.Frame`. I programmed the `wx` backend but honestly I didn't know which to choose. Maybe `wx.Panel` is actually the right choice. @robmcmullen do you know what any drawbacks to using a `wx.Panel` would be to using a `wx.Frame`? If we instatiate an instance without a parent, would we lose window decorations or anything like that, or would it otherwise be equivalent to what we saw with `wx.Frame`?

robmcmullen:
`Wx` doesn't allow a `wx.Panel` without a parent. Maybe CanvasBackend could subclass from `wx.Panel` but if `parent is None` a containing `wx.Frame` could be created and the events that make sense only for a top-level frame could be additionally bound to that frame. I can take a stab at that if that sounds like a workable idea.

Eric89GXL:
+1 from me, this sounds like what I/we probably should have done from the beginning. @almarklein @lcampagn sound good to you?

robmcmullen:
Here's my first attempt at this: I removed the hack `wx_panel` backend and incorporated the code into the regular wx backend. It will create a `wx.Frame` if it isn't given a specific parent. The cleanup operations probably need work when deleting multiple views of the same context, but I haven't been able to create a test case yet.

examples/tutorial/app/simple_wx.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
```
Eric89GXL: Can you add this line:

https://github.com/vispy/vispy/blob/master/examples/tutorial/app/shared_context.py#L2

It tells Travis not to try running this example as a standard testable example.

vispy/app/backends/_wx.py
```diff
@@ -214,38 +214,44 @@ def __init__(self, *args, **kwargs):
...
-        
-        self._canvas = glcanvas.GLCanvas(self, wx.ID_ANY, wx.DefaultPosition,
-                                         wx.DefaultSize, 0, 'GLCanvas',
-                                         self._gl_attribs)
+        self._init = False
+        glcanvas.GLCanvas.__init__(self, parent, wx.ID_ANY, pos=position,
+                                   size=size, style=0, name='GLCanvas',
+                                   attribList=self._gl_attribs)
+            
+        self._canvas = self
```
Eric89GXL: I assume this is just for testing, so this is just a reminder that eventually we should get rid of this and just do `self.` instead of `self._canvas.`

Eric89GXL:
This looks reasonable to me. BTW there is already some shared context testing that occurs in `vispy/app/tests/test_context.py`, so if `py.test` / Travis are happy, then we should be okay. You can test locally by doing:
```Bash
larsoner@bunk:~/custombuilds/vispy$ _VISPY_TESTING_APP=wx nosetests vispy/app/tests/test_context.py
  File "/home/larsoner/custombuilds/vispy/vispy/app/backends/_wx.py", line 369, in on_close
  File "/usr/lib/python2.7/dist-packages/wx-2.8-gtk2-unicode/wx/_core.py", line 14615, in __getattr__
    raise PyDeadObjectError(self.attrStr % self._name)
wx._core.PyDeadObjectError: The C++ part of the CanvasBackend object has been deleted, attribute access no longer allowed.
  File "/home/larsoner/custombuilds/vispy/vispy/app/backends/_wx.py", line 369, in on_close
  File "/usr/lib/python2.7/dist-packages/wx-2.8-gtk2-unicode/wx/_core.py", line 14615, in __getattr__
    raise PyDeadObjectError(self.attrStr % self._name)
wx._core.PyDeadObjectError: The C++ part of the CanvasBackend object has been deleted, attribute access no longer allowed.
...
```
The tests pass for me, but we get some ugly warnings. Can you fix them?

robmcmullen:
I'll look into it: I see the same warnings that you do on my setup here. I may have to add an extra widget layer between the `wx.Frame` and the `GLCanvas` due to the way wx handles attributes of deleted objects

Eric89GXL:
It looks to me like a more general problem (beyond wx) where you're subclassing a C++ class in Python and the underlying C++ object has been deleted, but the Python object is in the process of being deleted. Maybe look at stackoverflow to see how people have dealt with it before.

robmcmullen:
wx has a special check to see if you can access attributes of an object, so added a check in the on_close method

Eric89GXL:
Cool, did the warnings disappear for you, then?

robmcmullen:
No warnings for me now. Do they still show up for you?

Eric89GXL:
Nope, looks good -- just wanted to get confirmation from you before pulling and checking :) 

Looking at one of the Travis outputs I see:
```
Failed strangely (<type 'exceptions.NameError'>): name 'glcanvas' is not defined
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/vispy-0.3.0-py2.7.egg/vispy/testing/_runners.py", line 309, in test
    run[0]()
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/vispy-0.3.0-py2.7.egg/vispy/testing/_runners.py", line 61, in _unit
    has, why_not = has_backend(mode, out=['why_not'])
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/vispy-0.3.0-py2.7.egg/vispy/testing/_testing.py", line 163, in has_backend
    mod = __import__('app.backends.%s' % module_name, globals(), level=2)
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/vispy-0.3.0-py2.7.egg/vispy/app/backends/_wx.py", line 198, in <module>
    class CanvasBackend(glcanvas.GLCanvas, BaseCanvasBackend):
NameError: name 'glcanvas' is not defined
```
In other words, the `glcanvas.GLCanvas` subclassing you do will break an import on any system without `wx` installed. We usually get around this by putting the `from wx.glcanvas import GLCanvas` in a `try` block with a dummy `GLCanvas` class put in the `except` block.

vispy/app/backends/_wx.py
```diff
@@ -195,7 +195,7 @@ def Skip(self):
         pass
 
 
-class CanvasBackend(Frame, BaseCanvasBackend):
```
Eric89GXL: Whatever we did to protect the `Frame` subclassing should work for your `GLCanvas` case.

robmcmullen:
Right, think I figured it out. Try the latest...

Eric89GXL:
LGTM. Any other changes you want to make or is this ready for review/merge from your end? I don't think the Travis / AppVeyor errors are related to your PR.

robmcmullen:
I'll check in one more fix to remove the ._canvas cruft left over from my hacking. Hopefully tonight.

almarklein:
Sweet. I think the wx backend was always a bit unstable (no offense Eric). Great to see it fixed. Do we still need the `simple_wx.py` once the real wx backend is fixed?

robmcmullen:
`simple_wx.py` shows how to embed a vispy widget in a larger wx GUI so it does show something different than `simple.py`. But maybe the location I put it in isn't the best place for it. I'm not sure where there are other examples showing UI-toolkit-specific stuff.

almarklein:
Sorry, my mistake; somehow I thought it was a new backend.

Eric89GXL:
@robmcmullen can you fix the PEP8 errors:

https://travis-ci.org/vispy/vispy/jobs/46030864#L1107

Then I think we're good to merge assuming you're done.

Eric89GXL:
Great, thanks for the contribution!

PR #695, opened 2015-01-06
FIX: Fix glfw on Travis

Eric89GXL:


Eric89GXL:
This *should* fix the problems we've been having with Travis and GLFW, we'll see.

Eric89GXL:
Turns out it was a `GLFW` bug. I talked to a dev on IRC and he pushed a fix, which appears to work for us. That fixed Travis.

AppVeyor builds seem to work about 25% of the time, with weird illegal instruction problems when using OpenGL. So for now I have made it so that it only runs the `nobackend` tests :(

Tests now pass, ready for review/merge from my end.

almarklein:
>  I talked to a dev on IRC and he pushed a fix, which appears to work for us. That fixed Travis.

Sweet! LGTM

Issue #610, opened 2014-11-07
Support different texture internal formats and precisions


karlcz:
I am not sure whether I've misunderstood how the gloo API is supposed to work, or future plans that might already be in flight... But, using the "bleeding edge" source distribution from the project homepage I find that all textures are being created as 8-bit per channel regardless of what numpy data type is used to supply pixel data.  

I was able to fix this for myself only through ugly hacking of gloo/glir.py and gloo/gl/_constants.py (yes, in spite of the warning not to edit the auto-generated file).  Essentially, the glTexImage2D and glTexImage3D calls require different enumerations for the internalFormat parameter, e.g. GL_RGBA16, GL_LUMINANCE16_ALPHA16, GL_LUMINANCE16, GL_ALPHA16, GL_RGB16. Using the old-fashioned formats without a precision, e.g. GL_RGBA or GL_LUMINANCE, seems to imply the 8-bit channel depth.

To make this proof-of-concept, I simply copied the enumerated values I could find in OpenGL.GL on my system, and introduced a typed-format mapping table in the GlirTexture class, mapping a pair of (format, dtype) into the real GL enumerations for internal storage formats.  I modified the glir code to call TexImage3D or TexImage2D during the set_data methods, rather than TexSubImage3D or TexSubImage2D. That way, I could set the internal format according to both the (format, gtype) information that glir already had available.

With this change, I was able to see twice the precision in my textures and framebuffer objects by creating gloo.Texture2D or gloo.Texture3D with np.uint8 vs np.uint16 input data.  I didn't try to do this for float16 or float32 yet, as I was not sure where to dig up the corresponding GL enumerations.

I think that gloo should either expose an internalFormat parameter so that an application can specify what precision they desire, or it should automatically map the numpy input data to the most equivalent iinternal format that is available.  The latter makes sense to me from a scientific viewpoint, since the application is already making data format and normalization decisions to prepare the texture data.

karlcz:
I started looking into this a bit... I think there are a few API style questions to address as well as back-end behavior.  Some approaches would be a lot simpler to implement than others.

[...]

PR #696, opened 2015-01-13
add optional 'internalformat' parameter to Texture2D and Texture3D

karlcz:
When constructing or resizing textures, allow internalformat to be specified if the application needs to control precision or representation of texture storage.

   gloo.Texture3D(..., internalformat='rgba8')
   tex.resize(..., internalformat='rgba8')

Accept lower-case strings similar to the existing 'format' argument, including the same generic values as 'format' ('luminance', 'alpha', 'luminance_alpha', 'rgb', 'rgba').

Allow these precision/type suffixes: '8', '16', '16f', '32f'. Allow these base formats with precision suffixes: 'r', 'rg', 'rgb', 'rgba'. For example, 'rgb32f' means GL_RGB32F.

When keyowrd argument is omitted, internalformat is chosen based on number of channels and precision is left unspecified so the backend or OpenGL driver can chose an appropriate default.


karlcz:
I'm not sure if this is really the way you guys would want to approach this, but it is one possible way to change the gloo API to support control of texture storage precision as discussed in #610

karlcz:
By the way, the two ugliest parts of this patch are the manual inclusion of internal format enumerants in the glir code, and replacing texture-resize glir command field 'format' with a pair of (format, internalformat).  I wasn't sure whether changing the width of the glir command tuple would break anything else, or I would have just added the extra internalformat field.  I'm happy to try to improve those parts if you have any preferred approach...

Eric89GXL:
@rossant can you comment on the GLIR queue issues? I have a feeling @almarklein might not have time...?

FYI the AppVeyor error is unrelated, if you rebase your PR to latest master it should fix itself.

rossant:
@karlcz for the enum is it possible to just use strings in the GLIR commands, and use the existing `as_enum()` function?

I would say it's okay to change the [GLIR spec](https://github.com/vispy/vispy/wiki/Spec.-GLIR) at this point, but the wiki page will have to be updated accordingly if we merge the PR.

I'll let @almarklein comment on the rest...

karlcz:
I'll try to rebase and look at adjusting the glir command format later today.  As for the enumerations, the problem is that these internal format enumerants are not in the current gloo/gl/_constants.py and I'm not sure how you would want to address that.  The codegen could be changed to expose the superset of constants instead of the intersection, but I got the feeling you guys were still debating how to expose such desktop API features.

rossant:
OK I see, we really need to make a decision now! (see also #185). @almarklein @lcampagn @Eric89GXL should we meet on gitter or something..?

karlcz:
These github processes are new to me... the wiki will have to be edited separately after a PR is merged?  Or should there be wiki content somewhere inside my fork and feature branch that I can edit now to match the code changes?

karlcz:
Also, I don't see where the other GLIR implementation lives (the Javascript WebGL one)... it will also have to be modified to support the new GLIR command format...?  I wonder if it should just silently ignore the extra internalformat argument, treating it as a mere hint.

Eric89GXL:
Yeah, the wiki needs to updated separately unfortunately. @rossant can answer the WebGL questions

rossant:
You'll find the WebGL code [here](https://github.com/vispy/vispy.js/blob/master/scripts/gloo.glir.js), it's still a bit WIP.

karlcz:
I guess I don't really understand the connection between vispy and that other code.  Can a vispy app written in Python somehow drive the GLIR queue on a browser?  Or would you really be rewriting an app in Javascript using the same gloo APIs?

karlcz:
Also, I think the Travis CI failures were a transient problem downloading data files for some of the tests.  I don't think there is anything actually broken in this PR at present.

rossant:
Yes, Python/VisPy generates GLIR commands, which are dynamically sent to the browser. See https://github.com/vispy/vispy/blob/master/vispy/app/backends/_ipynb_webgl.py

karlcz:
OK, so it looks like generic code sending the GLIR command tuples already prepared by the same gloo/glir.py code I have modified?

I think these tuples are mapped to arrays in the Javascript code.  Can it just ignore the extra internalformat slot on the end and not actually require any changes at this point?

lcampagn:
Just curious: is there any reason for the `internalformat` slot at all? Why not just allow 'rgb32f' on the existing `format` slot; the webgl side should then generate a reasonable error like "format not supported".

karlcz:
It made the code a lot simpler when I did it this way (I briefly looked at that other approach too). Format and internalformat do not accept the same values in OpenGL and it seems like it would become very messy to replicate all the decision logic in gloo to know which combinations make sense.

More importantly, I think it makes it easier to call this a precision/performance hint in gloo and just ignore it in a backwards-compatible manner on some backends.

karlcz:
Hmm, I suppose there should be a simple functional map from internalformat values to format values, and a simple test from that same mapping structure to decide which you get. So, you could overload 'format' and determine which pair of (format, internalformat) values to pass to OpenGL based on whether the app supplied a basic value suitable for 'format' or a precision-qualified format suitable for 'internalformat'.

I could prototype this alternate approach, overloading just the format field in the API.  Would you want to see a different PR with this alternative?  Would you expect the GLIR command to contain the two values as they should pass to OpenGL, or just the overloaded value which means each GLIR implementation would need to perform the same decoding and mapping?

The downside to this approach is that it creates a more gloo-specific concept instead of just exposing the underlying OpenGL concept as an optional parameter.  I feel that the optional parameter is closer to what one would expect if they have familiarity with the OpenGL texture API.

vispy/gloo/texture.py
```diff
@@ -176,7 +197,17 @@ def resize(self, shape, format=None):
             The format of the texture: 'luminance', 'alpha',
             'luminance_alpha', 'rgb', or 'rgba'. If not given the format
             is chosen automatically based on the number of channels.
-            When the data has one channel, 'luminance' is assumed.
+            When the data has one channel, 'luminance' is assumed.  
+        internalformat : str | enum
+
+            The internal (storage) format of the texture: 'luminance',
```
almarklein: Small style issue: no newline here

vispy/gloo/glir.py
```diff
@@ -18,6 +18,31 @@
 from ..ext.six import string_types
 from ..util import logger
 
+# TODO: expose these via an extension space in .gl?
+_internalformats = [
+    gl.Enum('GL_R', 8194), 
+    gl.Enum('GL_R8', 33321),
```
almarklein: I think specifying the enums here is alright for now, it's easily changed if necessary when we come up with a final solution.

almarklein:
I agree with @karlcz that since this is an "extra" argument that is beyond ES 2.0, and is not always supported, it makes sense to have a separate kwarg.

I am not sure about the place where the default value is calculated. Right now, the GLIR command must contain a sensible value for internalformat. I'd rather have that its "optional" for the GLIR command (i.e. can be set to None), and that the glir implementation either just uses FORMAT, or uses a good matching internal format.

karlcz:
I will try to look at these issues later this week.  I also have a few new bug-fix commits I'd like to test and get into this PR before you consider merging it.

One thing that is bothering me is that different versions of OpenGL documentation I have consulted seem to have minor differences in which enumerants are supported for internalformat (or possibly I have misread some of them before).  Also, there remains the possibility that an implementation will reject a request that the specification describes. For example, my NVIDIA drivers seem to reject GL_RG16F while accepting all the other 16F combinations.  Should gloo define a value set at all, or just pass through anything that converts into a GL enum and let the GL implementation validate it?

almarklein:
>  Should gloo define a value set at all, or just pass through anything that converts into a GL enum and let the GL implementation validate it?

I wonder how feasible it is to define a set of enums that "always work". If it is possible, it may be worth showing a warning when a "problematic" enum is used. But using the internalformat parameter in the first place, already makes the code less portable (e.g. no webgl), so users should already be more aware of what they're doing.

karlcz:
@almarklein: What about having the GLIR implementation catch the BAD_ENUM exception, print a warning that it might be due to format or internalformat, and re-raise the exception?

karlcz:
Again, the only failing test once again looks like a problem with the test host retrieving demo data from github.  Is this typical...?

almarklein:
@karlcz  that seems reasonable

karlcz:
@almarklein , I could not replicate the problem I saw before with GL_RG16F. All format combinations appear to be working as expected with the new unit tests I just added.  I may have simply confused myself with concurrent changes in my app.

karlcz:
OK, I think I've addressed all the review comments and my own bug-fixing and testing that I was worried about.  The only test failure in the last Travis run was another transient connection error.

vispy/gloo/glir.py
```diff
@@ -876,18 +905,22 @@ def set_interpolation(self, min, mag):
 class GlirTexture2D(GlirTexture):
     _target = gl.GL_TEXTURE_2D
     
-    def set_size(self, shape, format):
+    def set_size(self, shape, format, internalformat):
```
Eric89GXL: Should this be by default `internalformat=None`, or is it unnecessary?

vispy/gloo/texture.py
```diff
@@ -48,19 +48,37 @@ class BaseTexture(GLObject):
     _inv_formats = {
         'luminance': 1,
         'alpha': 1,
+        'red': 1,
         'luminance_alpha': 2,
+        'rg': 2,
         'rgb': 3,
-        'rgba': 4,
+        'rgba': 4
```
Eric89GXL: I have a slight preference for leaving the extra `,` in place so that if something is appended, you only see the added line instead of an added and a changed line, but I can live with the change :)

vispy/gloo/texture.py
```diff
@@ -176,7 +195,16 @@ def resize(self, shape, format=None):
             The format of the texture: 'luminance', 'alpha',
             'luminance_alpha', 'rgb', or 'rgba'. If not given the format
             is chosen automatically based on the number of channels.
-            When the data has one channel, 'luminance' is assumed.
+            When the data has one channel, 'luminance' is assumed.  
+        internalformat : str | enum
```
Eric89GXL: `str | enum | None` is more complete, then below you can say "if None" instead of "if not given"

Eric89GXL:
BTW can you think of a very simple example where setting the internal format is useful? It could go in `vispy/tutorial/gloo` to show people why it's nice to have. Also, I assume you've tested this and it solves your use case?

Other than these questions and my minor comments, LGTM. +1 for merge once these are addressed. @rossant do you need to do anything equivalently on the WebGL end, or do some special checks or anything? @almarklein do you have time to review and merge?

vispy/gloo/glir.py
karlcz: I think it is unnecessary unless there is some way that this GLIR implementation would be running with an older gloo that generates the older RESIZE command tuple.  The new code will always have the internalformat slot and it may hold a None value.

karlcz:
Yes, I have been updating and testing a ray-casting volume renderer with the changes in this PR.  It uses higher precision for both the 2D entry/exit map FBO textures and the 3D volume data. I do hope to get that code and some illustrative data released eventually, but it is probably a bit too complex to serve as a good demo app.

I'd have to think about some simpler demo. Do you need this before merging?  It would need to be something like a 2-pass rendering/lookup or some other GLSL calculations that benefit from the extra precision or dynamic range.

Eric89GXL:
Nah, if the example is not easy don't worry about it. If you can't do it in < ~100 lines of a gloo demo, it wouldn't serve the illustrative purpose I had naively hoped for in any case.

Regarding volume rendering, have you seen #612? If you know about volume rendering and want to have a look / try it out that would be great. If you have another better / different volume rendering technique, it would be great to incorporate that into the volume rendering visual once #612 is in.

karlcz:
@Eric89GXL yes I have seen #612 but I haven't dug into the code.  I suppose we should move this conversation over to that PR rather than go further afield here?

Let's not hold of this PR on a demo program, but I will let the idea roll around and see if I can come up with something small and illustrative...

Eric89GXL:
Cool, I'll let @almarklein take one more look if he has time, if not I'll merge

rossant:
I don't have time now to dig into the code unfortunately, and I don't know if there are things to do with WebGL. If so I can do it later after this is merged!

vispy/gloo/glir.py
almarklein: It depends on how we define the GLIR command. If we define it as allowing internalformat to be omitted, we need a default value. I think I prefer that we define commands as fully specified, with optional arguments set to None though. @rossant WDYT?

rossant: You mean the number of arguments of each GLIR command would be fixed, with possibly None values? If so, +1.

vispy/gloo/glir.py
```diff
@@ -876,18 +905,22 @@ def set_interpolation(self, min, mag):
...
         # Shape is height, width
         format = as_enum(format)
-        if (shape, format) != self._shape_format:
-            self._shape_format = shape, format
+        if internalformat is not None:
+            internalformat = as_enum(internalformat)
+        else:
+            internalformat = format
```
almarklein: `internalformat = format if internalformat is None else as_enum(internalformat)`

almarklein:
@rossant This PR may actually require a change to the JS gloo backend so that the precense of the internalformat does not break stuff. But I suspect that JS's argument handling will simply drop it (as we want).

@rossant do you have time to answer my question?

rossant:
I guess there will be a minimal change [here](https://github.com/vispy/vispy.js/blob/master/scripts/gloo.glir.js#L349-L367) (?)

karlcz:
My understanding of that gloo.glir.js function is that it treats the GLIR command arguments as an array and it should already simply ignore the additional internalformat argument on the end of the array, without any code changes.  This seems correct until a newer WebGL standard allows internalformat control.

rossant:
Ah yes that should work, provided this new argument is the last one.

karlcz:
@rossant Yes, this PR changes the GLIR command from ('SIZE', id, shape, format) to ('SIZE', id, shape, format, internalformat).  I think the _ipynb_webgl.py backend and gloo.glir.js have generic code to serialize/deserialize the commands without change.

My only hesitation is that gloo.glir.js comments suggest the second argument is a byte-count size rather than a pixel count shape tuple.  I'm not sure if this is just a stale comment or if there is another layer of code that remaps GLIR command content before it gets to this point...

Eric89GXL:
Looks like there is one last style thing:

https://travis-ci.org/vispy/vispy/jobs/47933721#L1110

BTW you can run `make flake` or `python make test flake` to find errors locally.

karlcz:
@Eric89GXL It seems emacs python-mode and flake8 disagreed about whitespace in this case... :)

On my system, the flake8 test prints many errors for other parts of the codebase, so it can be easy to overlook the one or two lines that also will be caught in the Travis build... I also have problems getting the full test suite to run locally in my Fedora 20 system, but I haven't had the time to try to find out what other system/environment dependency is causing the problems.

Eric89GXL:
Did you just run `flake8` or did you run `make flake` or `python make test flake`? The first one won't ignore the proper set of files, whereas the other two will.

karlcz:
I did 'make flake' and it shows errors in: vispy/util/config.py, vispy/visuals/line_plot.py, vispy/app/inputhook.py, make/make.py.

Eric89GXL:
Interesting, I wonder if you're using a newer version... what does `flake8 --version` say?

rossant:
I think there is a new version recently that gives different results than before (I've had failing tests in another project because of this).

Eric89GXL:
I just ran `pip install --upgrade` on flake8, pep8, and pyflakes, and still don't get any errors. I assume you're not using a dev version?

rossant:
That might not be relevant at all here, but in my other project I was using `conda update flake`.

Eric89GXL:
I checked and I seem to be running the same versions of pep8 and pyflakes that are currently distributed with conda. I'm confused...

karlcz:
% flake8 --version
2.0 (pep8: 1.5.6, pyflakes: 0.7.3, mccabe: 0.2.1)

karlcz:
In my case, it is the python-flake8 package that comes with Fedora 20.
I try pretty hard not to override distribution packages in my work
environments.

Eric89GXL:
I bet the newer version has disabled some things that it shows... it's likely that there are just some PEP8 types that we need to explicitly add to the ignore. Are the errors of the same type or a couple of types?

karlcz:
several different error codes:

vispy/util/config.py:326:1: F811 redefinition of unused 'atexit' from line 16
vispy/visuals/line_plot.py:28:1: F821 undefined name 'y_vals'
vispy/visuals/line_plot.py:29:1: F821 undefined name 'x_vals'
vispy/visuals/line_plot.py:29:1: F821 undefined name 'y_vals'
vispy/visuals/line_plot.py:30:1: F821 undefined name 'xy_vals'
vispy/app/inputhook.py:36:1: F999 syntax error in doctest
make/make.py:340:1: F811 redefinition of unused 'np' from line 24

Eric89GXL:
Looks like these are the sorts of things that come from running an older version, e.g.:

https://bugs.launchpad.net/pyflakes/+bug/1297161

We shouldn't disable these warnings, so you'll have to live with them unless you're willing to upgrade, unfortunately.

karlcz:
Hi, please hold off on merging this... I found that there was a regression in the internalformat handling and it is getting the default None applied in cases when it should have a specific one.

karlcz:
OK, I solved the regression and also added an example app you could use to quickly verify that high-precision textures are working on your installation.  Just run examples/tutorial/texture_precision.py and toggle the 'f' key to cycle through different precision formats...

If I ever finish chasing code-style issues and getting a clean build, I think it is OK to merge.

examples/tutorial/gloo/texture_precision.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env python
+
```
almarklein: Can you add your name as author here, so that you're properly acknowledged and we know who to contact in case this example breaks :)

almarklein:
LGTM, anyone else for a final check?

Eric89GXL:
Thanks for the example, it's neat and simple. @karlcz let me know when you've added your name to the top of that example and I'll merge.

karlcz:
OK, I should have thought about that first. I may have to add a university boilerplate notice and wait for them to agree to a particular open source license.  Will Apache 2.0 be OK if they do not want to go with BSD?

I already raised the issue to try to release my other gloo application, so I may just have to wait for that decision before we can add this.

Eric89GXL:
I don't think we want to mix licenses, so definitely try to get BSD clearance. Your university has a policy that restricts your ability to contribute to open-source BSD packages? I've never heard of such a thing, so I'm curious what the deal is...

karlcz:
@Eric89GXL  I think we'll get it right and be able to use BSD since you need that.

Mainly, these things get determined on a per-project basis here, and my recent work with vispy is a new offshoot project that hasn't already made any release decisions...  I only brought up Apache because I know we've been able to use that on many prior projects.

Eric89GXL:
Cool, let us know when we're good to go, then. Sorry for the hassle, but hopefully it will clear the way for future contributions, too.

Eric89GXL:
@karlcz ready to go from your end, assuming the CIs come back happy? If so I'll take one last look and merge.

karlcz:
I'm almost afraid to say yes, but yes let's go ahead and merge if you like. :-)

Witness the power of this fully armed and operational bureaucracy!

Eric89GXL:
Thanks for the contrib!

I hope you don't feel that the bureaucracy wasn't too much / too annoying. I actually think this is a pretty good prototype for how PRs should go -- nicely self-contained enhancement/fix, good Github + CI feedback loop, functionality shown with a nice example and tested with unit tests.

karlcz:
Ha, I was referring to the legal process on my end...

Eric89GXL:
Oh right, forgot about that -- I guess it's a good sanity check that vispy bureaucracy < university bureaucracy :)

PR #697, opened 2015-01-14
Add IPython version check in WebGL backend

rossant:
This fixes the WebGL backend which is currently broken with IPython 3.0.dev.

IPython 3.0 installs nbextensions in a system-wide directory by default. VisPy will install the nbextensions in a user directory to avoid permissions errors.

Ready for review/merge.


Eric89GXL:
LGTM, thanks for the fix.

Issue #607, opened 2014-10-30
Improve the WebGL backend


rossant:
This is a list of things to do to improve the WebGL backend once PR #592 is merged.
### JavaScript loading
- [x] Refactor vispy.js to use browserify instead of requireJS.
- [x] Do not install nbextension if not necessary. The required JS files (notably `vispy.min.js`) are copied into the IPython profile everytime the `ipynb_webgl` backend is used. We could do a simple check to only copy if necessary (unless IPython's `install_nbextension` already does that).
- [ ] Fail gracefully if WebGL is not available (and suggest the user to use the VNC backend)
### vispy.js
- [ ] Improve the GLIR implementation, notably:
  - in SHADERS, after attaching shaders, we should detach and destroy them
  - in TEXTURE, we should use the texture_number from the existing stored texture
  - try more complicated examples, notably with multiple textures
- [ ] Implement framebuffers in vispy.js (they are not supported currently).
### Misc
- [x] ~~Solve issue #606 so that all examples work in the notebook~~ tackle issue #615.
- [x] Use binary WebSockets (implemented in IPython master and upcoming IPython 3.0) to transfer NumPy arrays from Python to JavaScript.
- [ ] Refactor `ipynb_webgl` so that other notebook backends can reuse most of the code (and bring back static and vnc).
- [ ] Rename paint into draw
- [ ] Implement `VispyWidget(DOMWidget)` (see comment below)

rossant:
Note to self:

I would like to make the following possible. I create a VisPy SceneCanvas as usual (e.g. a line plot). Next, I want to easily create an IPython DOMWidget that:

* exposes several trait attributes, for example `line_color`
* displays the SceneCanvas using WebGL in a cell's widget area

Here is how it could work:

```python
# create the SceneCanvas
canvas = ...

class MyWidget(VispyWidget):
    """This class makes the bridge between the VisPy canvas and the WebGL widget.
    It exposes several trait attributes which notification callbacks can be customized.
    A VispyWidget is instanciated with a Canvas instance. Displaying the widget shows the
    WebGL version of the canvas.
    """

    # This trait attribute represent the line's color.
    line_color = 'blue'

    def on_trait_change(self, name, old, new):
        if name == 'line_color':
            self.canvas.myvisual.color = new
            
display(MyWidget(canvas))
```

lcampagn:
Isn't this already possible just by implementing properties on the subclass, like
```
@line_color.setter
def line_color(self, color):
[...]

PR #698, opened 2015-01-15
Improve WebGL backend

### LINKED ISSUE #615: Implement a server-side GLIR cache
This is backend-independent. The GL context needs to store a list of (almost) all GLIR commands emitted since initialization. Unnecessary GLIR commands (old commands superseded by new commands) are automatically discarded (pruning), so that this list should never be very long.

* [ ] implement the cache (easy)
* [ ] dynamically prune old GLIR commands that are superseded by new ones: `DRAW`, `UNIFORM`, `FUNC` (slightly less straightforward)
* completely get rid of `on_initialize()` (can be done in another PR)
* fix the `ipynb_webgl` backend so that the latest state is automatically reinitialized when the context is lost (can be done in another PR)

rossant:
Done:

* Improved JavaScript loading, using [browserify](http://browserify.org/) instead of requireJS
* Better handling of removed widgets
* VisPy timers are automatically stopped when a widget is removed

Caveats:

* Need to get rid of `on_initialize()` once for all, otherwise weird bugs occur with auto-starting timers in the notebook (I'll raise an issue)
* Doing `canvas.show()` several times with the same `Canvas` instance does not work. Users need to recreate the canvas explicitely. This needs to be clear in the documentation. Solving this would require a GLIR cache: possible in the future, but pretty low priority. The current behavior is good enough in my opinion (basically we have to make sure to create a new canvas in the same cell that we show it).

Tackle some of #607 (not all)
Close #615 (no need for a GLIR cache for now)


rossant:
Ready for review/merge once Travis is happy.

rossant:
AppVeyor fails, not sure why..

Eric89GXL:
Can you rebase against latest `master`? That should fix AppVeyor.

rossant:
Just did it, nothing happened.

Eric89GXL:
Oh -- if you merge #695 then rebase it should work :)

rossant:
OK done :)

vispy/app/backends/_ipynb_webgl.py
```diff
@@ -312,18 +312,30 @@ def _vispy_stop(self):
...
 # ---------------------------------------------------------- IPython Widget ---
+def _stop_timers(canvas):
+    """Stop all timers in a canvas."""
+    for attr in dir(canvas):
+        try:
+            attr_obj = getattr(canvas, attr)
+        except NotImplementedError:
+            attr_obj = None
+        if isinstance(attr_obj, Timer):
+            attr_obj.stop()
```
Eric89GXL: Can you mention why you needed to do try/except, and put it in the comments? If I came to this code to modify it later I would be confused...

vispy/html/static/js/webgl-backend.js
```diff
@@ -1,15 +1,5 @@
 // Load Vispy.js in the notebook.
 IPython.load_extensions("vispy.min");
-IPython.load_extensions("jquery.mousewheel.min");
-
-// HACK: this is UGLY but I didn't find a better way to do it
```
Eric89GXL: hooray, removing a HACK comment :)

Eric89GXL:
Other than my minor gripe, LGTM +1 for merge. I agree the multiple `.show()` case is just a limitation we can tell users about, I don't think it's really a show-stopper. I'll let @almarklein or @lcampagn have a look before merge, though, since I'm not too familiar with this stuff.

vispy/app/backends/_ipynb_webgl.py
rossant: Done!

rossant:
I had to rebase because of the example flake problems.

rossant:
Everybody is happy hurray :)

almarklein:
I don't fully understand every bit, but I could not detect anything bad happening :) LGTM

PR #701, opened 2015-01-15
Getting rid of on_initialize in the examples

### LINKED ISSUE #700: Get rid of on_initialize
I would like to get rid of the `initialize` event and the `on_initialize` callback in all VisPy. Having several ways of initializing the scene is quite confusing (constructor, `on_initialize`, first `on_draw`, etc.).

Is it possible to stick to a single option in the whole codebase + examples + documentation? If so, the first `on_draw` might be the way to go. I think the best way to do it would be to add an `is_first_draw` boolean in the `on_draw`'s `event` argument. Then, we would recommend users to use the following pattern:

```python
canvas = Canvas()

@canvas.connect
def on_draw(event):
    if event.is_first_draw:
        canvas.program = Program(...)
        ...
    canvas.program.draw()

```

and we would use this pattern in all `gloo` examples. An alternative solution could be:

```python
canvas = Canvas()

def initialize():
    canvas.program = Program(...)
    ...

@canvas.connect
def on_draw(event):
    if event.is_first_draw:
        initialize()
    canvas.program.draw()

```

I'm not sure how the scene stuff works but I suppose it is similar.

Doing this would also make the IPython WebGL backend more stable, cleaner, and it might even offer a way to fix the multiple-show problem.

WDYT?

rossant:
Close #700.

### DIFF
--- a/examples/basics/visuals/custom_visual.py
+++ b/examples/basics/visuals/custom_visual.py
@@ -10,7 +10,7 @@ from vispy import app
 from vispy import gloo
 from vispy.visuals.shaders import ModularProgram
 from vispy.visuals import Visual
-from vispy.visuals.transforms import (STTransform, LogTransform, 
+from vispy.visuals.transforms import (STTransform, LogTransform,
                                       TransformSystem, ChainTransform)
 
 
@@ -18,7 +18,7 @@ class MarkerVisual(Visual):
     # My full vertex shader, with just a `transform` hook.
     VERTEX_SHADER = """
         #version 120
-        
+
         attribute vec2 a_position;
         attribute vec3 a_color;
         attribute float a_size;
@@ -35,9 +35,9 @@ class MarkerVisual(Visual):
             v_antialias = 1.0;
             v_fg_color  = vec4(0.0,0.0,0.0,0.5);
             v_bg_color  = vec4(a_color,    1.0);
-            
+
             gl_Position = $transform(vec4(a_position,0,1));
-            
+
             gl_PointSize = 2.0*(v_radius + v_linewidth + 1.5*v_antialias);
         }
     """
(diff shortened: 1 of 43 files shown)


rossant:
Ready for review/merge.

Eric89GXL:
Did you mean to leave these in place?
```
$ git grep on_initialize
examples/demo/gloo/jfa/jfa_vispy.py:    def on_initialize(self, event):
examples/ipynb/webgl_example_1.ipynb:      "    def on_initialize(self, event):\n",
examples/ipynb/webgl_example_2.ipynb:      "def on_initialize(e):\n",
examples/tutorial/app/app_events.py:    def on_initialize(self, event):
examples/tutorial/gl/cube.py:    def on_initialize(self, event):
examples/tutorial/gl/fireworks.py:    def on_initialize(self, event):
examples/tutorial/gl/quad.py:    def on_initialize(self, event):
```

rossant:
I forgot jfa. The ipynb ones I'll tackle later in one of the WebGL PRs. The others I prefer to keep them as they're app/gl level, not gloo.

Eric89GXL:
LGTM -- this one is pretty straightforward, so I'll merge in a few hours unless @almarklein or @lcampagn you want to have a look.

rossant:
btw I manually tested every single example I modified.

Eric89GXL:
Awesome. I manually tested one and assumed / hoped it would generalize :) Thanks for being thorough -- I'll go ahead and merge, if others have fixes / comments we can do a subsequent PR

examples/basics/visuals/custom_visual.py
```diff
@@ -68,29 +68,29 @@ class MarkerVisual(Visual):
...
+        self._program = ModularProgram(self.VERTEX_SHADER,
                                        self.FRAGMENT_SHADER)
         self.set_data(pos=pos, color=color, size=size)
-        
+
     def set_options(self):
         """Special function that is used to set the options. Automatically
         called at initialization."""
-        gloo.set_state(clear_color=(1, 1, 1, 1), blend=True, 
+        gloo.set_state(clear_color=(1, 1, 1, 1), blend=True,
```
lcampagn: @almarklein, how will the current gloo/glir know to which context this call belongs? Does this need to be changed to `self.set_gl_state(...)`?

almarklein: This will be send to whatever canvas is current. The `gloo.x()` functions are mostly there for backward compatibility and are better replaced with `self.context.set_gl_state`

PR #708, opened 2015-01-20
Remove extra argument in _fetch_file's docstring

rossant:


Eric89GXL:
Thanks, I must have forgotten to remove that when I was cleaning up at some point.

PR #709, opened 2015-01-20
Update the ipynb examples to IPython 3.0

rossant:
* Updated the example notebooks from nbformat 3.0 to 4.0
* These examples require IPython 3.0

Ready to merge asap.


Eric89GXL:
The addition of whitespace makes it tough to see what was actually changed here, but I assume you did it right :) Just for my edification, are the changes backward-compatible with IPython 2? Did you just add a couple of fields that make things work better with IPython 3?

Eric89GXL:
Never mind, just read your description stating that they are 3.0-only. So there's no good way to have them be 3.0 and 2.0 compatible? Bummer...

rossant:
Which whitespaces?

No way of having them work with both versions. Just easier to enforce 3.0 (also no binary websocket in 2.0). There are also backward incompatible changes in the widgets in IPython 3.0.

Eric89GXL:
It looks like at least some of the changed lines are taking one space and turning it into three (at the beginning of each line)

PR #713, opened 2015-01-23
Add HSV<->RGB GLSL functions

### LINKED ISSUE #416: Implement reusable GLSL functions for color space transformations
At least RGB <-> HSV. Put them in `vispy.visuals.glsl`.

rossant:
Closes #416.

### DIFF
--- a/vispy/visuals/glsl/color.py
+++ b/vispy/visuals/glsl/color.py
@@ -38,3 +38,31 @@ vec3 get_color(int index) {
     return texture2D($colormap, vec2(x, .5)).rgb;
 }
 """
+
+
+# -----------------------------------------------------------------------------
+# Color space transformations
+# -----------------------------------------------------------------------------
+
+HSV_TO_RGB = """
+vec3 hsv_to_rgb(vec3 c)
+{
+    vec4 K = vec4(1.0, 2.0 / 3.0, 1.0 / 3.0, 3.0);
+    vec3 p = abs(fract(c.xxx + K.xyz) * 6.0 - K.www);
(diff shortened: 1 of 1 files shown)


Eric89GXL:
It might be good to check that these actually work. WDYT about doing some computations on the CPU using `Color`/`ColorArray`, then also plotting the result using a the simplest GLSL you can build around these functions, and checking that the results match?

Eric89GXL:
We can wait to do this later if you have other high-priority things to work on, but we should at least add a comment in the code with an `XXX` or `TODO` saying we should do it at some point.

rossant:
Hmm, that's a possibility, but no high priority IMO. I'm already using these functions and I know they work (and they are used quite a lot apparently). Unit testing them might not make much sense as they're never going to change...

Eric89GXL:
IMO it is not safe to assume that any code will stay fixed -- intentionally or unintentionally, even the simplest of code can get updated, modified, improved, extended, relocated, etc. That being said, I agree tests aren't a high priority here. Maybe I will tackle them later :)

rossant:
Hmm okay makes sense. This can be done later indeed. Shall I add a TODO?

Eric89GXL:
You've convinced me enough, so only bother if you're feeling bored -- otherwise I'll merge once Travis and AppVeyor are happy, even though it seems silly to wait

rossant:
Hold on I'll do it

rossant:
No prob with waiting for the green button

rossant:
Ready to merge

Eric89GXL:
Thanks for the changes and reminder, I got side-tracked

PR #714, opened 2015-01-27
Added panning to TurntableCamera with right click

jdreaver:
See the following image to visualize the vectors used in the code:

This works by computing the `up` and `right` vectors of the camera in world coordinates, and then mapping the change in mouse position from pixels to vectors in the world coordinates. The camera center is then translated using these vectors. This makes the motion fell very fluid, because a point under your mouse stays under your mouse.

Note that I can't figure out how to do this reliably with a perspective projection yet. The orthographic projection actually made this fairly easy, since `camera.width` gives you the screen width in world coordinates. (I also feel like the mouse wheel should change `distance` in a perspective view, and not `fov`. I get weird clipping when I zoom all the way in or all the way out. I know in pyqtgraph that `fov` would be changed if you hold down Ctrl and use the mouse wheel.)

I think we can abstract the panning calculation into a function, and maybe call it `pan_pixels`. The function would accept what is now `dx` and `dy` in the pull request code. I don't want users to confuse it with a function like `pan` that would simply modify the `center` coordinates directly using world coordinates.

### DIFF
--- a/vispy/scene/cameras/cameras.py
+++ b/vispy/scene/cameras/cameras.py
@@ -578,6 +578,38 @@ class TurntableCamera(PerspectiveCamera):
             d = p2c - p1c
             self.orbit(-d[0], d[1])
 
+        elif event.type == 'mouse_move' and 2 in event.buttons:
+            if self._mode != "ortho":
+                return
+
+            p1 = np.array(event.last_event.pos)[:2]
+            p2 = np.array(event.pos)[:2]
+            p1c = event.map_to_canvas(p1)[:2]
+            p2c = event.map_to_canvas(p2)[:2]
+            dx, dy = p2c - p1c
+            dy *= -1  # Makes y point up instead of down
+
+            # Vector `forward` goes from camera to center
+            azim = self.azimuth * np.pi / 180
+            elev = self.elevation * np.pi / 180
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks, this seems very useful! Note that the cameras (including this one) are changed in a pending PR (#612). You'll probably need a rebase once that's merged. One change is indeed that the distance is changed when zooming, rather than the fov.

jdreaver:
Great! I'll wait for that PR and then incorporate this change into it. The `fov` change is probably just what I need to get it to work for non-orthographic projections.

jdreaver:
Should I go ahead and modify this now that the camera changes are merged, or do you anticipate more camera changes during the sprint?

almarklein:
You probably want to check whether the current version of the camera's fulfill your needs; the panzoom camera has panning (shift + LMB). We are planning some more changes to the camera's, but I'm not sure if we'll get at it during the sprint. Let us know if you want to work on something, and we can hold those changes off.

jdreaver:
Oh, that would be great if the camera changes work well enough already! I am busy this week so I probably will not get to this before the sprint is over.

Eric89GXL:
I'm going to go ahead and close this, feel free to re-open or start a new PR if the current implementation is insufficient.

PR #717, opened 2015-01-29
ENH: Always on top, remove GLUT

### LINKED ISSUE #716: Always on top canvas
Add an `always_on_top` keyword when creating a canvas.

* Qt: `QtCore.Qt.WindowStaysOnTopHint`
* wx: `wx.STAY_ON_TOP`
* glfw: `GLFW_FLOATING`

Eric89GXL:
Adds always-on-top support. It also removes GLUT, which is nice because it cleans up a bunch of exceptions and not-well-tested code we had. I think we should be okay because we support many other backends, including pure-python solutions (e.g., Pyglet).

Closes #716.


rossant:
LGTM. +1 for removing GLUT

vispy/ext/bunch.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class SimpleBunch(dict):
+    """ Container object for datasets: dictionnary-like object that
+        exposes its keys as attributes.
+    """
+
+    def __init__(self, **kwargs):
+        dict.__init__(self, kwargs)
+        self.__dict__ = self
```
almarklein: +1 for putting the canvas arguments in a little namespace object.

Why is this in ext rather than util? It might even be small enough to just define it where its used.

almarklein:
LGTM except one comment.

vispy/ext/bunch.py
Eric89GXL: We might end up wanting it for other things, I'll move to util

Eric89GXL:
@almarklein done. Feel free to merge once it's green.

rossant:
thanks @Eric89GXL !

PR #719, opened 2015-02-06
Corrected improper use of Color.is_blank  (fixes scene/viewbox example)

campagnola:
Ready for review/merge.

### DIFF
--- a/vispy/color/color_array.py
+++ b/vispy/color/color_array.py
@@ -401,7 +401,10 @@ class Color(ColorArray):
     def lab(self):
         return super(Color, self).lab[0]
 
+    @property
     def is_blank(self):
+        """Boolean indicating whether the color is invisible.
+        """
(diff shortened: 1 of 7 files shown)


Eric89GXL:
Do you think it would be better to make this a property instead of method? I doubt you're the only person who will make such a mistake...

campagnola:
I was thinking the same thing.. not here, though.

campagnola:
I take it back--I'll do it now.

Eric89GXL:
Did the old `git grep` only reveal a few instances? :)

I'll merge once the green light pops up.

Eric89GXL:
Looks like at least for the first commit one of the examples was broken on Travis.

campagnola:
All is well.

Eric89GXL:
Great, thanks for the fix.

PR #720, opened 2015-02-10
FIX: Update for new pep8

Eric89GXL:
@almarklein ready for review/merge assuming Travis is happy. The PEP8 program was updated with new rules, and this ignores one of them and brings us into compliance with the others.

### DIFF
--- a/vispy/gloo/program.py
+++ b/vispy/gloo/program.py
@@ -301,7 +301,10 @@ class Program(GLObject):
             elif kind == 'attribute':
                 # Is this a constant value per vertex
                 is_constant = False
-                isscalar = lambda x: isinstance(x, (float, int))
+
+                def isscalar(x):
+                    return isinstance(x, (float, int))
+
                 if isscalar(data):
                     is_constant = True
                 elif isinstance(data, tuple):
--- a/examples/basics/scene/grid_large.py
+++ b/examples/basics/scene/grid_large.py
@@ -8,6 +8,7 @@
 Test automatic layout of multiple viewboxes using Grid.
 """
 
+import sys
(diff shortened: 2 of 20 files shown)


Eric89GXL:
@almarklein good to go, feel free to merge and re-trigger your Travis run

vispy/gloo/program.py
```diff
@@ -301,7 +301,10 @@ def __setitem__(self, name, data):
             elif kind == 'attribute':
                 # Is this a constant value per vertex
                 is_constant = False
-                isscalar = lambda x: isinstance(x, (float, int))
+
+                def isscalar(x):
+                    return isinstance(x, (float, int))
+
```
almarklein: why the F are lambdas bad according to flake? I personally find the lambda more pretty (and shorter!) in this case.

Perhaps also don't apply this rule?

almarklein: Are lambda's always considered bad? I cannot imagine this was the only lambda in our code base ...

almarklein:
Thanks for the quick fix by the way!

rossant:
I think `a = lambda...` is bad but not `f(lambda...)`

vispy/gloo/program.py
Eric89GXL: No, they aren't always bad, they are just not good when you are trying to assign a function to a variable name directly -- that is better done through `def`. See:

http://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def

almarklein: Still find lambda prettier in this case, but I'll merge for now :)

PR #722, opened 2015-02-13
Minor fixes for vispy.geometry

FiachAntaw:
Two minor fixes for vispy.geometry:
1) Changed create_cylinder and create_cone to explicitly set the face array datatype to np.uint32, matching the behavior of create_sphere (N.B. np.uint defaults to np.uint64 on some systems, which breaks these functions)

2) Expose the create_arrow as an external method of the geometry module, matching the behavior of the other .create_ methods.


Eric89GXL:
Looks great, ready to go from your end?

FiachAntaw:
Yeah, it's ready.

Eric89GXL:
Thanks for the fixes!

PR #723, opened 2015-02-13
ENH: Simple marker test

Eric89GXL:
This will help us know whether or not point sprites are working for people.

### DIFF
--- a/vispy/visuals/tests/test_markers.py
+++ b/vispy/visuals/tests/test_markers.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+from vispy.scene.visuals import Markers
+from vispy.testing import (requires_application, TestingCanvas,
+                           run_tests_if_main)
+from vispy.gloo.util import _screenshot
+
+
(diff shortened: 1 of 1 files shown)


Eric89GXL:
Travis and AppVeyor are happy, ready for review/merge from my end.

almarklein:
LGTM, but how does this *test* sprites? People can run this demo and see whether they see the markers or not?

Eric89GXL:
If the sprite doesn't draw, the test fails. This was confirmed in the other thread where I suspected the user's sprites weren't drawing.

Eric89GXL:
Maybe I don't understand your question, but I hope how the code works pretty straightforward as a standard drawing test:

1. Draw sprite visual

2. Capture screenshot

3. Check to make sure the screenshot contains something drawn (this could be made more comprehensive later by actually checking against a standard testing image, but I figured that was overkill for now)

People can run the code the standard way as part of the test suite, or they can execute it. If point sprites work, there will be no error. If point sprites fail to draw -- at least in one incarnation of the issue -- the assert will fail.

PR #726, opened 2015-02-18
More minor vispy.geometry fixes

FiachAntaw:
I ran into another issue with vispy.geometry on 64-bit python (specifically when creating a MeshData from unindexed vertices), so I went ahead and made all the implicit-width datatypes 32-bit (haven't run into any issues yet).

I also exposed the create_cone function, so all the .generation.create_ methods are now exposed in vispy.geometry with similar default parameters.

### DIFF
--- a/vispy/geometry/__init__.py
+++ b/vispy/geometry/__init__.py
@@ -8,13 +8,13 @@ This module implements classes and methods for handling geometric data.
 
 from __future__ import division
 
-__all__ = ['MeshData', 'PolygonData', 'Rect', 'Triangulation', 'create_cube',
-           'create_cylinder', 'create_sphere', 'create_arrow']
+__all__ = ['MeshData', 'PolygonData', 'Rect', 'Triangulation', 'create_arrow',
+           'create_cone', 'create_cube', 'create_cylinder', 'create_sphere']
 
 from .polygon import PolygonData  # noqa
 from .meshdata import MeshData  # noqa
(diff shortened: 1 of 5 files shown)


Eric89GXL:
It looks like this breaks a bit of code because of precision:

https://travis-ci.org/vispy/vispy/jobs/51164338

If you change the assert to be `np.allclose(number, 0)` or so instead of `number == 0`, it should fix it.

Don't worry about Travis being angry about style, that's unrelated to your changes.

FiachAntaw:
I tried weakening the assert, but other issues emerged that seemed to come from some precision issue deeper in the algorithm. Looking at the original, working code it seems that the datatype of the point list was getting silently changed from 32-bit to python's native double-precision float on line 84:

```python
self.pts = pts.view(np.float32).reshape(len(pts), 2)
...
newpts = np.empty((self.pts.shape[0]+2, 2), dtype=float)
...
newpts[2:] = self.pts
self.pts = newpts
```

The algorithm doesn't seem to work if newpts is switched to np.float32 or self.pts is initialized to np.float64 and I don't really want to mess with it without understanding the algorithm it's implementing, so I've reverting the change I made to triangulate.py.

Eric89GXL:
@FiachAntaw is this good to go from your end, then? LGTM

FiachAntaw:
Yeah, it's good to go.

Eric89GXL:
Great, thanks for the additional fixes.

PR #730, opened 2015-02-23
Add plot.scatter

brickgao:
Hello all,

I just made a prototype for scatter().

**Bug below has been fixed**

It works but there is still a problem that we need to zoom or change the view to see all points.

If you run the code below:

```bash
>>> import vispy.plot as pl
>>> pl.scatter([1, 2, 3, 4])
```

The pic that we want to get is:

but we get:

### DIFF
--- a/vispy/plot/plot.py
+++ b/vispy/plot/plot.py
@@ -142,11 +142,11 @@ def scatter(*args, **kwds):
         if isinstance(args[0], np.ndarray):
             _pos = args[0]
         else:
-            _pos[:, 0] = [np.float32(i) for i in range(len(args[0]))]
-            _pos[:, 1] = map(np.float32, args[0])
+            _pos[:, 0] = np.arange(len(args[0]))
+            _pos[:, 1] = np.array(args[0])
     elif len(args) == 2:
-        _pos[:, 0] = map(np.float32, args[0])
-        _pos[:, 1] = map(np.float32, args[1])
+        _pos[:, 0] = np.array(args[0])
+        _pos[:, 1] = np.array(args[1])
     else:
         raise TypeError('Arguments should be passed as ' +
                         '(Y,), (X, Y) or (np.array((X, Y)))')
--- a/vispy/plot/plot.py
+++ b/vispy/plot/plot.py
@@ -139,17 +139,17 @@ def scatter(*args, **kwds):
     canvas.view = canvas.central_widget.add_view()
     _pos = np.zeros((len(args[0]), 2))
     if len(args) == 1:
-        if isinstance(args[0], np.ndarray):
-            _pos = args[0]
-        else:
+        pos = np.asarray(args[0])
+        if pos.ndim == 1:
             _pos[:, 0] = np.arange(len(args[0]))
-            _pos[:, 1] = np.asarray(args[0])
+            _pos[:, 1] = pos
+        else:
+            _pos = pos
     elif len(args) == 2:
         _pos[:, 0] = np.asarray(args[0])
         _pos[:, 1] = np.asarray(args[1])
     else:
-        raise TypeError('Arguments should be passed as ' +
-                        '(Y,), (X, Y) or (np.array((X, Y)))')
+        raise ValueError('Invalid shape for position data')
     canvas.scatter = visuals.Markers()
     kwds['pos'] = _pos
     canvas.scatter.set_data(**kwds)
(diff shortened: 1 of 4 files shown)


vispy/plot/plot.py
```diff
@@ -103,3 +104,57 @@ def mesh(vertices=None, faces=None, vertex_colors=None, face_colors=None,
...
+    '''
+    canvas = SceneCanvas(keys='interactive')
+    canvas.view = canvas.central_widget.add_view()
+    _pos = np.zeros((len(args[0]), 2))
+    if len(args) == 1:
+        if isinstance(args[0], np.ndarray):
+            _pos = args[0]
+        else:
+            _pos[:, 0] = [np.float32(i) for i in range(len(args[0]))]
+            _pos[:, 1] = map(np.float32, args[0])
```
rossant: You should avoid Python loops on NumPy arrays. For example, you could use `np.arange()` here. Here is an useful reference: http://docs.scipy.org/doc/numpy/reference/routines.html

vispy/plot/plot.py
```diff
@@ -103,3 +104,57 @@ def mesh(vertices=None, faces=None, vertex_colors=None, face_colors=None,
...
+    elif len(args) == 2:
+        _pos[:, 0] = map(np.float32, args[0])
+        _pos[:, 1] = map(np.float32, args[1])
```
rossant: Same here: `map` should be avoided because it is unnecessary on NumPy arrays and very slow.

brickgao:
@rossant I have fixed it ;)

rossant:
Thanks, that's better! You can also use `asarray()` instead of `array()`: when the argument is already an array, no copy is performed, which is faster.

rossant:
@campagnola @Eric89GXL any comments on this feature?

vispy/plot/plot.py
```diff
@@ -103,3 +104,57 @@ def mesh(vertices=None, faces=None, vertex_colors=None, face_colors=None,
     # canvas.view.camera.auto_zoom(image)  # XXX Don't have this for Turntable
     plots.append(canvas)
     return canvas
+
+
+def scatter(*args, **kwds):
+    ''' Create a new canvas and make a scatter plots.
```
almarklein: plot (-s)

vispy/plot/plot.py
```diff
@@ -103,3 +104,57 @@ def mesh(vertices=None, faces=None, vertex_colors=None, face_colors=None,
...
+    canvas = SceneCanvas(keys='interactive')
+    canvas.view = canvas.central_widget.add_view()
+    _pos = np.zeros((len(args[0]), 2))
+    if len(args) == 1:
+        if isinstance(args[0], np.ndarray):
+            _pos = args[0]
+        else:
+            _pos[:, 0] = np.arange(len(args[0]))
+            _pos[:, 1] = np.asarray(args[0])
+    elif len(args) == 2:
```
almarklein: I'm not sure if this is quite right. I may want to provide a 1D array of y-values.  Suggestion:

    if len(args) == 1:
        # Either a y is given, or both x and y in one array
        pos = np.asarray(args[0])
        if pos.ndim == 1:
             _pos[:, 0] = np.arange(len(args[0]))
             _pos[:, 1] = pos
        elif pos.ndim == 2:
            _pos = args[0]
        else:
              raise ValueError('Invalid shape for position data')

brickgao:
@almarklein I get it, now `plot.scatter()` will work when pass these args: 

```bash
>>> import vispy.plot as pl
>>> import numpy as np
>>> pl.scatter([1, 2])
>>> pl.scatter([1, 2, 3, 4])
>>> pl.scatter([0, 1, 2, 3], [1, 2, 3, 4])
>>> pl.scatter(np.asarray([(0, 1), (1, 2), (2, 3), (3, 4)]))
>>> pl.scatter([(0, 1), (1, 2), (2, 3), (3, 4)])
```

brickgao:
@rossant Hi, I added `bounds()` to `vispy/visuals/markers.py`.

It fixed the bug that users need to zoom or change the view to see all points.

Now `vispy.plot.scatter()` works better.

Eric89GXL:
Can you add a simple example to the examples in `examples/basics/plotting/`? Otherwise the changes look reasonable to me. There are some minor issues with the docstring, but we can change those in the next month when we (hopefully) revamp the plotting system.

brickgao:
@Eric89GXL Ok, I have added it :)

Eric89GXL:
Great, thanks for the enhancement!

PR #731, opened 2015-02-23
Added TubeVisual for an extruded tube about a set of points

inclement:
This PR adds a TubeVisual, which takes a set of points and creates a tubular mesh about them.

I took the algorithm (and initial implementation) from three.js. It solves most of the problems that are naturally encountered, such as making sure the mesh parameterisation rotates neatly about the curve at intersections, and spreading accumulated twisting across the curve if it is closed.

There are still some problems and bugs - in particular, the color argument doesn't work (if I pass it to the constructor, I get gloo errors), nor does the face_colors (just seems to do nothing, maybe I don't understand how to use it), and some of the maths can be further vectorised. There are also questions about the API - I have in mind that I mostly want to plot a set of points and a colour at each point, hence the colors argument, but I'm not sure what the vispy philosophy is for it. Maybe it should be more mesh focused and the color stuff go in a higher level wrapper.

It seems to basically work though, so I'm creating this PR to hope for feedback on whether it's appropriate, and any major changes that should be made.

Edit: Here is a screenshot of the included example:

tube_example_reference2

### DIFF
--- a/examples/basics/visuals/tube.py
+++ b/examples/basics/visuals/tube.py
@@ -58,7 +58,7 @@ if __name__ == '__main__':
                             tube_points=8)
 
     l3 = scene.visuals.Tube(points3, colors=colors,
-                            shading='smooth',
+                            shading='flat',
                             tube_points=8,
                             closed=True)
 
--- a/examples/basics/visuals/tube.py
+++ b/examples/basics/visuals/tube.py
@@ -1,0 +1,77 @@
+"""
+Demonstration of Tube
+"""
+
+import sys
+from vispy import app, scene
+from vispy.scene.visuals import Tube
+
+from colorsys import hsv_to_rgb
+import numpy as np
+
+def torus_knot(p=3, q=4, num=100):
+    '''
+    Returns points in the p, q torus knot.
+
+    Note that not all combinations of p and q give knots;
+    some curves may pass through themselves.
+    '''
+    data = np.zeros((num, 3), dtype=np.float64)
+    ts = np.linspace(0, 2*np.pi, num)
+    rs = np.cos(q*ts) + 2
+    data[:, 0] = rs*np.cos(p*ts)
+    data[:, 1] = rs*np.sin(p*ts)
+    data[:, 2] = -1*np.sin(q*ts)
+    return data
+
+if __name__ == '__main__':
+    canvas = scene.SceneCanvas(keys='interactive')
+    canvas.view = canvas.central_widget.add_view()
+
+    points1 = torus_knot(5, 3)[:-1] * 5.
+    points1[:, 0] -= 20.
+    points1[:, 2] -= 20.
+
+    points2 = points1.copy()
+    points2[:, 2] += 41.
+
+    points3 = points1.copy()
+    points3[:, 0] += 41.
+    points3[:, 2] += 41
+
+    points4 = points1.copy()
+    points4[:, 0] += 41.
+
+    colours = np.linspace(0, 1, len(points1))
+    colours = np.array([hsv_to_rgb(c, 1, 1) for c in colours])
+
+    l1 = scene.visuals.Tube(points1, colors=colours,
+                            shading='flat',
+                            tube_points=8)
+
+    l2 = scene.visuals.Tube(points2, colors=colours,
+                            shading='smooth',
+                            tube_points=8)
+
+    l3 = scene.visuals.Tube(points3, colors=colours,
+                            shading='smooth',
+                            tube_points=8,
+                            closed=True)
+
+    l4 = scene.visuals.Tube(points4, colors=colours,
+                            shading='flat',
+                            tube_points=8,
+                            mode='lines')
+
+    canvas.view.add(l1)
+    canvas.view.add(l2)
+    canvas.view.add(l3)
+    canvas.view.add(l4)
+    canvas.view.set_camera('turntable', mode='perspective',
+                           up='z', distance=2.9*1.5*np.max(np.max(
+                               points1, axis=0)))
+    canvas.show()
+
+    if sys.flags.interactive != 1:
+        canvas.app.run()
+
--- a/examples/basics/visuals/tube.py
+++ b/examples/basics/visuals/tube.py
@@ -9,22 +9,6 @@ from vispy.geometry.torusknot import TorusKnot
 from colorsys import hsv_to_rgb
 import numpy as np
 
-
-def torus_knot(p=3, q=4, num=100):
-    '''
-    Returns points in the p, q torus knot.
-
-    Note that not all combinations of p and q give knots;
-    some curves may pass through themselves.
-    '''
-    data = np.zeros((num, 3), dtype=np.float64)
-    ts = np.linspace(0, 2*np.pi, num)
-    rs = np.cos(q*ts) + 2
-    data[:, 0] = rs*np.cos(p*ts)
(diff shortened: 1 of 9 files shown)


rossant:
image

Very cool, thanks!

examples/basics/visuals/tube.py
```diff
@@ -0,0 +1,86 @@
...
+
+    points3 = points1.copy()
+    points3[:, 0] += 41.
+    points3[:, 2] += 30
+
+    points4 = points1.copy()
+    points4[:, 0] += 41.
+
+    colors = np.linspace(0, 1, len(points1))
+    colors = np.array([hsv_to_rgb(c, 1, 1) for c in colors])
```
rossant: use [this](https://github.com/vispy/vispy/blob/master/vispy/color/color_space.py#L78) instead?

examples/basics/visuals/tube.py
```diff
@@ -0,0 +1,86 @@
...
+
+    vertex_colors = np.random.random(8 * len(points1))
+    vertex_colors = np.array([hsv_to_rgb(c, 1, 1) for c in vertex_colors])
+
+    face_colors = np.random.random(1568)
```
rossant: magic number?

inclement:
Thanks for the rapid feedback!

I'll work on these things and the other issues I'm aware of, and get back to you.

examples/basics/visuals/tube.py
```diff
@@ -0,0 +1,86 @@
...
+                            mode='lines')
+
+    canvas.view.add(l1)
+    canvas.view.add(l2)
+    canvas.view.add(l3)
+    canvas.view.add(l4)
+    canvas.view.set_camera('turntable', mode='perspective',
+                           up='z', distance=2.9*1.5*np.max(np.max(
+                               points1, axis=0)))
+    canvas.show()
```
Eric89GXL: Can you unindent all of the above and make it so that just the `.run()` and conditional below are nested? That will make it so that the example gets tested properly by the test suite.

examples/basics/visuals/tube.py
```diff
@@ -0,0 +1,86 @@
...
+    Note that not all combinations of p and q give knots;
+    some curves may pass through themselves.
+    '''
+    data = np.zeros((num, 3), dtype=np.float64)
+    ts = np.linspace(0, 2*np.pi, num)
+    rs = np.cos(q*ts) + 2
+    data[:, 0] = rs*np.cos(p*ts)
+    data[:, 1] = rs*np.sin(p*ts)
+    data[:, 2] = -1*np.sin(q*ts)
+    return data
```
Eric89GXL: Feel free to move this to `vispy.geometry` if you think it's general enough.

vispy/visuals/tube.py
```diff
@@ -0,0 +1,171 @@
...
+        for i in range(1, len(points)):
+            normals[i] = rotation_about_axis(
+                tangents[i], theta*i).dot(normals[i])
+
+    binormals = np.cross(tangents, normals)
+
+    return tangents, normals, binormals
+                
+
+def rotation_about_axis(axis, angle):
```
Eric89GXL: I think we already have something like this in `vispy.transforms`. Can you check? In any case, that's where this would belong if we don't have it.

vispy/visuals/tube.py
```diff
@@ -0,0 +1,171 @@
...
+                            color=color,
+                            shading=shading,
+                            mode=mode)
+                
+
+    def draw(self, transforms):
+        MeshVisual.draw(self, transforms)
+
+
+def frenet_frames(points, closed):
```
Eric89GXL: This should probably be private, so please add a leading underscore to the function name.

vispy/visuals/tube.py
```diff
@@ -0,0 +1,171 @@
+from __future__ import division
+
+from .mesh import MeshVisual
+import numpy as np
+from numpy.linalg import norm
+
+
+class TubeVisual(MeshVisual):
+    """Displays a tube by extruding a circle along a piecewise-linear
+    path.
```
Eric89GXL: Initial description should be a single line

vispy/visuals/tube.py
```diff
@@ -0,0 +1,171 @@
...
+    ----------
+    points : ndarray
+        An array of (x, y, z) points describing the path along which the
+        tube will be extruded.
+    radius : float
+        The radius of the tube. Defaults to 1.0.
+    tube_points : int
+        The number of points in the approximate circle of the tube's
+        cross section. Defaults to 8.
+    colors : ndarray
```
Eric89GXL: ndarray | None

vispy/visuals/tube.py
```diff
@@ -0,0 +1,171 @@
...
+        An array of colors at each point of the points array; every
+        vertex of the tube mesh around this point will take the given
+        color. Must be the same length as the points array, or None
+        to color only by the `color` argument. Defaults to None.
+    shading : str
+        Same as for the `MeshVisual` class. Defaults to 'smooth'.
+    vertex_colors: ndarray or None
```
Eric89GXL: s/or/|

Eric89GXL:
This does indeed look cool, thanks for the PR.

inclement:
@Eric89GXL I couldn't find a separate existing function for just a rotation matrix, the existing transforms function constructed it internally. I moved the implementation to a different function, but I'm not sure if this is what you had in mind - maybe there is a better place for it?

inclement:
@Eric89GXL Likewise with the TorusKnot geometry I added - maybe it should just be a function (perhaps in generation.py), though the current functions there create meshes rather than in this case just the points of a 1D curve?

vispy/visuals/tube.py
```diff
@@ -0,0 +1,153 @@
...
+    ----------
+    points : ndarray
+        An array of (x, y, z) points describing the path along which the
+        tube will be extruded.
+    radius : float
+        The radius of the tube. Defaults to 1.0.
+    tube_points : int
+        The number of points in the approximate circle of the tube's
+        cross section. Defaults to 8.
+    colors : ndarray | None
```
Eric89GXL: Can you check what we do with other visuals? I think we usually make it so that `colors` can be `ColorArray | Color`, and the visual then deals with it appropriately (an array of colors or a single color).

Eric89GXL:
Regarding the rotation matrix stuff -- I just want to make sure we keep things DRY. I thought that somewhere we already computed a rotation about an axis. Looks like you found it and made that instance use your new function, which is great. LGTM other than my last comment about having `colors` and `color` in the API. This isn't very clean, so it would be nice if we could unify them to a single argument, and triage the behavior based on whether one color or a set of colors was given.

@campagnola do you have time to take a look?

rossant:
> my last comment about having colors and color in the API. This isn't very clean, so it would be nice if we could unify them to a single argument, and triage the behavior based on whether one color or a set of colors was given.

I may be wrong, but isn't that something to be handled by the collections system?

Eric89GXL:
Eventually, yes, but for now I just want to make sure the API doesn't get polluted unnecessarily with two color-like arguments

inclement:
I've changed to make use of ColorArrays (which are a very nice API), but I'm not sure if some of the details are what you're after - since I'm internally converting the color to the vertex_colors array, I don't pass a color argument to mesh, and I'm not sure if that's ever likely to be important.

Eric89GXL:
Great, thanks for the addition!

PR #737, opened 2015-03-07
Implement a 1d sampler using pyopengl

bollu:
So, I implemented a 1d sampler similar to the already existing 3d sampler in the codebase by accessing ```pyopengl```. There a lot of layers and moving parts, so I'm not sure if I missed something in my implementation.

I also had to manually extract the constants ```GL_TEXTURE_1D``` and ```GL_SAMPLER_1D```. I'm assuming that they were not listed in ```_constants.py``` since they're not ```GLES2.0```? Either way, I'm not too fond of the current __extremely__ hacky solution I came up with. Would love a nicer solution

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -433,6 +433,7 @@ class GlirProgram(GlirObject):
         'mat2': 'glUniformMatrix2fv',
         'mat3': 'glUniformMatrix3fv',
         'mat4': 'glUniformMatrix4fv',
+        'sampler1D': 'glUniform1i',
         'sampler2D': 'glUniform1i',
         'sampler3D': 'glUniform1i',
     }
@@ -744,6 +745,8 @@ class GlirProgram(GlirObject):
         gl.glBindTexture(gl.GL_TEXTURE_2D, 0)
         if USE_TEX_3D:
             gl.glBindTexture(GL_TEXTURE_3D, 0)
+            gl.glBindTexture(GL_TEXTURE_1D, 0)
+
         #Deactivate program - should not be necessary. In single-program
         #apps it would not even make sense.
         #self.deactivate()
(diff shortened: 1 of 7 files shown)


Eric89GXL:
This is a good start. You followed the Texture3D code, which I think is the right approach.

However, the code needs tests. You can take a look in our test suite to see where `Texture3D` gets tested, and make similar tests with appropriate modifications. If you do that and fix the PEP8 errors (https://travis-ci.org/vispy/vispy/jobs/53483908) we should be good to go.

bollu:
ah, sure. I'll do that now

Eric89GXL:
cool, ping me when you're done and if it's all good I'll merge

vispy/gloo/glir.py
```diff
@@ -894,13 +898,57 @@ def set_wrapping(self, wrapping):
             gl.glTexParameterf(self._target, GL_TEXTURE_WRAP_R, wrapping[0])
         gl.glTexParameterf(self._target, gl.GL_TEXTURE_WRAP_S, wrapping[-2])
         gl.glTexParameterf(self._target, gl.GL_TEXTURE_WRAP_T, wrapping[-1])
-    
+
     def set_interpolation(self, min, mag):
         self.activate()
         min, mag = as_enum(min), as_enum(mag)
         gl.glTexParameterf(self._target, gl.GL_TEXTURE_MIN_FILTER, min)
         gl.glTexParameterf(self._target, gl.GL_TEXTURE_MAG_FILTER, mag)
```
bollu: I found this to be _very icky_. Is there a neater way that we could generate these constants? I'm ready to write code to automate this if there's any non-insanely complex way to automate this

vispy/gloo/glir.py
```diff
@@ -894,13 +898,57 @@ def set_wrapping(self, wrapping):
...
 
+# these should be auto generated in _constants.py. But that doesn't seem 
+# to be happening. TODO - figure out why the C parser in (createglapi.py)
+# is not extracting these constanst out.
+# found the constant value at:
```
bollu: I found this to be _very icky_. Is there a neater way that we could generate these constants? I'm ready to write code to automate this if there's any non-insanely complex way to automate this

Eric89GXL: No, not yet. We have a PR open already that does generate them automatically. Right now, our code only populates the constants from the OpenGL ES 2.0 spec. We want to expand it to optionally include those from the desktop spec as well (incl. the 1D and 3D texture constants), but haven't done it yet. If you are feeling adventurous, you can look at #185. We hope to fix it in the next few weeks. In the meantime your hack (following the 3D case) should be sufficient.

Eric89GXL:
Good to go from your end? Travis and happy and LGTM, so I'll merge if you're done.

Eric89GXL:
Ahh, saw your `gitter` response, merging.

Eric89GXL:
Thanks for the enhancement!

Issue #677, opened 2014-12-14
Axes, ticks, labels, grid visuals


rossant:
* Labels: we already have TextVisual, but we might want a specific visual
* Ticks: either Lines, or a big quad with a fragment shader (@rougier?)
* Grid: fragment shader (see [this paper](https://hal.inria.fr/hal-01081592/document) for example)
* Axes: lines

rossant:
related to #549

PR #738, opened 2015-03-07
Galaxy demo

bollu:
Ported the [Galaxy demo from glumpy](https://github.com/glumpy/glumpy/blob/master/examples/galaxy.py) to ```VisPy```

This depends on the 1D sampler implemented [in this pull request](https://github.com/vispy/vispy/pull/737). So please make sure to pull that first before merging this example change.

I've tried to comment the example as meaningfully as possible so that the code is easy to follow.

Images:

### DIFF
--- a/examples/demo/gloo/galaxy/galaxy.py
+++ b/examples/demo/gloo/galaxy/galaxy.py
@@ -209,7 +209,7 @@ class Canvas(app.Canvas):
         gloo.clear(color=True, depth=True)
         self.program.draw('points')
 
-if __name__ == '__main__':
+if __name__ == '__main__' and sys.flags.interactive == 0:
     c = Canvas()
     c.show()
     app.run()
--- a/examples/demo/gloo/galaxy/galaxy.py
+++ b/examples/demo/gloo/galaxy/galaxy.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 import numpy as np
+import sys
 
 from vispy.util.transforms import perspective
 from vispy.util import transforms
@@ -102,9 +103,7 @@ for i in range(n):
 # load the PNG that we use to blend the star with
 # to provide a circular look to each star.
 def load_galaxy_star_image():
-    from vispy.io import load_data_file
-
-    fname = load_data_file('galaxy/star-particle.png')
+    fname = io.load_data_file('galaxy/star-particle.png')
     raw_image = io.read_png(fname)
 
     return raw_image
--- a/examples/demo/gloo/galaxy/galaxy.py
+++ b/examples/demo/gloo/galaxy/galaxy.py
@@ -1,0 +1,216 @@
+#!/usr/bin/env python
+import numpy as np
+
+from vispy.util.transforms import perspective
+from vispy.util import transforms
+from vispy import gloo
+from vispy import app
+from vispy import io
+
+import galaxy_specrend
+from galaxy_simulation import Galaxy
+
+VERT_SHADER = """
+#version 120
+uniform mat4  u_model;
+uniform mat4  u_view;
+uniform mat4  u_projection;
+
+//sampler that maps [0, n] -> color according to blackbody law
+uniform sampler1D u_colormap;
+//index to sample the colormap at
+attribute float a_color_index;
+
+//size of the star
+attribute float a_size;
+//type
+//type 0 - stars
+//type 1 - dust
+//type 2 - h2a objects
+//type 3 - h2a objects
+attribute float a_type;
+attribute vec2  a_position;
+//brightness of the star
+attribute float a_brightness;
+
+varying vec3 v_color;
+void main (void)
+{
+    gl_Position = u_projection * u_view * u_model * vec4(a_position, 0.0,1.0);
+
+    //find base color according to physics from our sampler
+    vec3 base_color = texture1D(u_colormap, a_color_index).rgb;
+    //scale it down according to brightness
+    v_color = base_color * a_brightness;
+
+
+    if (a_size > 2.0)
+    {
+        gl_PointSize = a_size;
+    } else {
+        gl_PointSize = 0.0;
+    }
+
+    if (a_type == 2) {
+        v_color *= vec3(2, 1, 1);
+    }
+    else if (a_type == 3) {
+        v_color = vec3(.9);
+    }
+}
+"""
+
+FRAG_SHADER = """
+#version 120
+//star texture
+uniform sampler2D u_texture;
+//predicted color from black body
+varying vec3 v_color;
+
+void main()
+{
+    //amount of intensity from the grayscale star
+    float star_tex_intensity = texture2D(u_texture, gl_PointCoord).r;
+    gl_FragColor = vec4(v_color * star_tex_intensity, 0.8);
+}
+"""
+
+galaxy = Galaxy(10000)
+galaxy.reset(13000, 4000, 0.0004, 0.90, 0.90, 0.5, 200, 300)
+# coldest and hottest temperatures of out galaxy
+t0, t1 = 200.0, 10000.0
+# total number of discrete colors between t0 and t1
+n = 1000
+dt = (t1 - t0) / n
+
+# maps [0, n) -> colors
+# generate a linear interpolation of temperatures
+# then map the temperatures to colors using black body
+# color predictions
+colors = np.zeros(n, dtype=(np.float32, 3))
+for i in range(n):
+    temperature = t0 + i * dt
+    x, y, z = galaxy_specrend.spectrum_to_xyz(galaxy_specrend.bb_spectrum,
+                                              temperature)
+    r, g, b = galaxy_specrend.xyz_to_rgb(galaxy_specrend.SMPTEsystem, x, y, z)
+    r = min((max(r, 0), 1))
+    g = min((max(g, 0), 1))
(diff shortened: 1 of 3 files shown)


bollu:
what's next when it comes to this?

rossant:
Cool, thanks! we'll review this soon. Note that you have some failing tests on travis, not exactly sure what's happening ( @Eric89GXL ? )

Another glumpy example that would be very nice in VisPy is [this](https://github.com/vispy/vispy/issues/677#issuecomment-66922363). Once we have a working example of axes and ticks with VisPy, it will be very easy to create the associated visuals. For now we can focus on cartesian coordinate systems, we'll deal with other systems later.

bollu:
Ooh, that does look slick. Although I assume that will take some _serious_ anti aliasing to look super pretty

Eric89GXL:
There are style errors that are hopefully pretty easy to see on Travis. You can check locally using `python make test flake` from the `vispy` repo root.

Then there are example errors. This is a hint:
```
RuntimeError: Bad example formatting: fix or add `# vispy: testskip` to the top of the file.
```
In other words, the files that are not meant to be run as standalone examples (e.g., `galaxy_simulation.py`) should have that line added toward the top as in e.g. `demo/gloo/unstructured_2d.py`. You can check that locally using `python make test examples`.

bollu:
oh, thanks! I did not know that :)

Eric89GXL:
I assume closing the PR was a mistake...?

bollu:
Yes, It was. I'm sorry.

Eric89GXL:
Let's get the 1D sampler PR merged, then move on to this one.

bollu:
hm, right. This build will fail till the sampler code has been merged upstream.

bollu:
I have a feeling I messed the rebase up. Please do tell me if I did it right!

Eric89GXL:
That doesn't look quite right. You might be better off starting from a new branch based on the latest upstream master, and using `git cherry-pick`ing your subsequent commits.

bollu:
rebased it right this time, hopefully :)

Eric89GXL:
Rebase looks good, but you need to add the `vispy: testskip` comment to a couple of files, see:
```
https://travis-ci.org/vispy/vispy/jobs/53814246
```
You can test locally using `python make test examples`.

bollu:
Is it possible to test one particular example? because running through 114 of them is kind of annoying when you want to test only one

Eric89GXL:
Not currently. It wouldn't be too terrible to modify `make.py` and the test suite to only run a particular example, but it's probably overkill for your testing right now. You should really just need to add that `vispy: testskip` line at the top of those two files that are not meant to be run on their own...

bollu:
yep, I did that and pushed the changes. but, for future reference. Since I have a feeling that I'll be porting quite a few examples :)

bollu:
is it okay if I add a bit of code to selectively run an example? It should be simple enough I think. Is that okay?

Eric89GXL:
In general you either want one of two things, and you've actually done both here: you either tell Travis to skip an example, or you ensure it runs without warnings or errors.

In testing the latter case, you can usually just run the script locally to test that it runs cleanly. So far this has been sufficient for testing examples locally. This is similar to what the test suite does, it just automates the process. So in my mind it's overkill to add a condition to selectively run a single example (since you can just do it manually easily), but I won't stop you from adding it if you're curious to dig into our test internals :)

bollu:
cool, thanks :D
And I might as well take a look at the testing framework seeing as I'll be working on this project over the summer hopefully :)

Eric89GXL:
Some hints:

1. You probalby want to modify `vispy/make/make.py` so you can do `python make test examples examples/demo/gloo/whatever.py` or so.

2. This will eventually wrap to `_examples()` in `_runners.py`, which currently doesn't accept any arguments but could be extended to take a list of filenames to run (which, if `None`, runs all examples):

    https://github.com/vispy/vispy/blob/master/vispy/testing/_runners.py#L193

bollu:
cool, thanks!

bollu:
is this ready for a merge?

examples/demo/galaxy/galaxy.py
```diff
@@ -0,0 +1,212 @@
...
+    r = min((max(r, 0), 1))
+    g = min((max(g, 0), 1))
+    b = min((max(b, 0), 1))
+    colors[i] = galaxy_specrend.norm_rgb(r, g, b)
+
+
+# load the PNG that we use to blend the star with
+# to provide a circular look to each star.
+def load_galaxy_star_image():
+    raw_image = io.read_png("star-particle.png")
```
Eric89GXL: Right now the example must be run from this directory to work -- if you put this as the following, it won't:
```
raw_image = io.read_png(op.join(op.dirname(__file__), 'star-particle.png'))
```
However, we usually put binary data in the `vispy-data` repo instead of this one in order to keep this repo clean. I've uploaded the data here:
```
https://github.com/vispy/demo-data/blob/master/galaxy/star-particle.png
```
You should be able to do something like this to get the file auto-downloaded:
```
from vispy.io import load_data_file
fname = load_data_file('galaxy/star-particle.png')
```

...also, this example directory should live under `examples/demo/gloo`, not `examples/demo`.

Eric89GXL:
Once you have `git rm`'ed the `.png` file, please squash your commits for this PR so that the file is removed from the history (keep repo size low).

bollu:
ooh, thanks! and right. realized too late that I should not have checked in a binary blob

Eric89GXL:
Let me know when you've squashed the commits and the CIs are happy, and I'll merge.

Eric89GXL:
something went wrong with the commit history (see above). It might be easiest to start with a fresh branch based off of current upstream `master` and do a `git cherry-pick 81bda68`.

bollu:
whoops. that's interesting. didn't realize that happened

bollu:
okay, clean history this time

examples/demo/gloo/galaxy/galaxy.py
```diff
@@ -0,0 +1,216 @@
...
+    r = min((max(r, 0), 1))
+    g = min((max(g, 0), 1))
+    b = min((max(b, 0), 1))
+    colors[i] = galaxy_specrend.norm_rgb(r, g, b)
+
+
+# load the PNG that we use to blend the star with
+# to provide a circular look to each star.
+def load_galaxy_star_image():
+    from vispy.io import load_data_file
```
Eric89GXL: Keep imports at the top unless they need to be nested (this one doesn't need to be)

Eric89GXL: ...and actually you alread have `from vispy import io` up there, so this can just be removed and below do `io.load_data_file(...)`

examples/demo/gloo/galaxy/galaxy.py
```diff
@@ -0,0 +1,216 @@
...
+        self.program.bind(self.data_vbo)
+
+        # clear the screen and render
+        gloo.clear(color=True, depth=True)
+        self.program.draw('points')
+
+if __name__ == '__main__':
+    c = Canvas()
+    c.show()
+    app.run()
```
Eric89GXL: We usually protect this with a `sys.interactive` check, see other examples

bollu:
done. I think this is ready to be merged?

examples/demo/gloo/galaxy/galaxy.py
```diff
@@ -0,0 +1,215 @@
...
+        self.program.bind(self.data_vbo)
+
+        # clear the screen and render
+        gloo.clear(color=True, depth=True)
+        self.program.draw('points')
+
+if __name__ == '__main__' and sys.flags.interactive == 0:
+    c = Canvas()
+    c.show()
+    app.run()
```
Eric89GXL: Sorry to be a pain, but only the `app.run()` should go in a conditional about `sys.flags.interactive`. Try running your example with `python` and `python -i` to see the idea.

bollu: yeah, so I'm making sure that it's not interactive aren't I?
```
sys.flags.interactive == 0:
```

checks if the shell is interactive right?

jdreaver: I think Eric is saying the code should be like this:

```python
canvas = Canvas()
canvas.show()

    app.run()
```

Note how just `app.run()` is inside the conditional. This way, you can run this interactively in IPython and access `canvas`.

Eric89GXL: Pretty much, yeah. Basically the only thing that needs to be in the conditional about `sys.flags.interactive` is the `app.run()`. Putting the other two lines (creating and showing the canvas) in there will make the code not work if run interactively as `python -i file.py`, which isn't what we want. So this is what we usually do:

```
if __name__ == '__main__':
    canvas = Canvas()
    canvas.show()
    if sys.flags.interactive == 0:
        app.run()
```
But what @jdreaver suggested would also work, and is used in some of the `scene` examples.

bollu: oh, okay. That makes sense. I didn't understand the rationale before. Will change

bollu: Done. hope this is okay now

Eric89GXL:
Great, thanks for another one!

PR #740, opened 2015-03-10
add support to run individual examples into the testing framework

bollu:
you can invoke it in two ways:

##### this will invoke all the examples
```
python make test examples
``` 

##### this will invoke only the 3 examples passed
``` 
python make test examples examples/demo/gloo/mandelbrot.py  examples/demo/gloo/graph.py examples/demo/scene/magnify.py 
````

Hope this is okay.
If there are any changes that need to be done, please do tell me

### DIFF
--- a/vispy/testing/_runners.py
+++ b/vispy/testing/_runners.py
@@ -190,8 +190,14 @@ with canvas as c:
 """
 
 
-def _examples():
-    """Run all examples and make sure they work
+def _examples(individial_file_paths):
+    """Run examples and make sure they work.
+
+    Parameters
+    ----------
+
+    individial_file_paths: str
+        can be "" or a space separated list of paths to examples to test
     """
     root_dir, dev = _get_root_dir()
     reason = None
@@ -206,9 +212,20 @@ def _examples():
         msg = 'Skipping example test: %s' % reason
         print(msg)
         raise SkipTest(msg)
-    fnames = [op.join(d[0], fname)
-              for d in os.walk(op.join(root_dir, 'examples'))
-              for fname in d[2] if fname.endswith('.py')]
+
+    fnames = []
+
+    # if we're given individual file paths, then just use the given fnames
+    # otherwise, use the full example paths that have been
+    # passed to us
+    if individial_file_paths:
+        fnames = individial_file_paths.split(' ')
+
+    else:
+        fnames = [op.join(d[0], fname)
+                  for d in os.walk(op.join(root_dir, 'examples'))
+                  for fname in d[2] if fname.endswith('.py')]
+
     fnames = sorted(fnames, key=lambda x: x.lower())
     print(_line_sep + '\nRunning %s examples using %s backend'
           % (len(fnames), backend))
@@ -278,6 +295,7 @@ def test(label='full', extra_arg_string=''):
     label = 'pytest' if label == 'nose' else label
     known_types = ['full', 'unit', 'lineendings', 'extra', 'flake',
                    'nobackend', 'examples']
+
     if label not in known_types + backend_names:
         raise ValueError('label must be one of %s, or a backend name %s, '
                          'not \'%s\'' % (known_types, backend_names, label))
@@ -292,7 +310,8 @@ def test(label='full', extra_arg_string=''):
     elif label in backend_names:
         runs.append([partial(_unit, label, extra_arg_string), label])
     if label in ('full', 'examples'):
-        runs.append([_examples, 'examples'])
+        runs.append([partial(_examples, extra_arg_string),
+                    'examples'])
     if label in ('full', 'unit', 'nobackend'):
         runs.append([partial(_unit, 'nobackend', extra_arg_string),
                      'nobackend'])
--- a/vispy/testing/_runners.py
+++ b/vispy/testing/_runners.py
@@ -190,14 +190,15 @@ with canvas as c:
 """
 
 
-def _examples(individial_file_paths):
+def _examples(fnames_str):
     """Run examples and make sure they work.
 
     Parameters
     ----------
 
-    individial_file_paths: str
-        can be "" or a space separated list of paths to examples to test
+    fnames_str : str
+        Can be a space-separated list of paths to test, or an empty string to
+        auto-detect and run all examples.
     """
     root_dir, dev = _get_root_dir()
     reason = None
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: vispy/testing/_runners.py [hunk 1]
```diff
@@ -190,8 +190,14 @@ def _check_line_endings():
...
 
-def _examples():
-    """Run all examples and make sure they work
+def _examples(individial_file_paths):
+    """Run examples and make sure they work.
+
+    Parameters
+    ----------
+
+    individial_file_paths: str
```
Eric89GXL: let's go with a shorter name like `paths` (fewer chars to type), and numpy-style formatting has an extra space as `paths : str`

Eric89GXL: that is:
```
paths : str
```

vispy/testing/_runners.py
```diff
@@ -190,8 +190,14 @@ def _check_line_endings():
...
+        can be "" or a space separated list of paths to examples to test
```
Eric89GXL: Proper caps and punctuation:
```
        Can be a space-separated list of paths to test, or an empty string to
        auto-detect and run all examples.
```

vispy/testing/_runners.py
```diff
@@ -206,9 +212,20 @@ def _examples():
         msg = 'Skipping example test: %s' % reason
         print(msg)
         raise SkipTest(msg)
-    fnames = [op.join(d[0], fname)
-              for d in os.walk(op.join(root_dir, 'examples'))
-              for fname in d[2] if fname.endswith('.py')]
+
+    fnames = []
```
Eric89GXL: you don't need this, since you immediately overwrite it below

vispy/testing/_runners.py
Eric89GXL: ...and actually for simplicity, this should just be `fnames`, so we get to re-use the var name below

vispy/testing/_runners.py
```diff
@@ -292,7 +310,8 @@ def test(label='full', extra_arg_string=''):
     elif label in backend_names:
         runs.append([partial(_unit, label, extra_arg_string), label])
     if label in ('full', 'examples'):
-        runs.append([_examples, 'examples'])
+        runs.append([partial(_examples, extra_arg_string),
```
Eric89GXL: Don't pass `extra_arg_string` if `label == 'full'`...

vispy/testing/_runners.py
```diff
@@ -292,7 +310,8 @@ def test(label='full', extra_arg_string=''):
...
+                    'examples'])
     if label in ('full', 'unit', 'nobackend'):
         runs.append([partial(_unit, 'nobackend', extra_arg_string),
```
Eric89GXL: because in that case, it should be interpreted as being passed to `_unit` not to `_examples`

Eric89GXL:
Other than my minor comments, this looks like the right approach!

vispy/testing/_runners.py
bollu: right, but it "feels" nicer to show that the variable exists. Will remove it

vispy/testing/_runners.py
bollu: ah, right. missed that. thanks!

vispy/testing/_runners.py
bollu: oh, didn't know that ```space-separated``` should be hyphenated :)

vispy/testing/_runners.py
bollu: maybe this is a Rust thing, but I personally prefer long, descriptive names over short ones.
But very well, I'll change that.

And, I hadn't know the ```paths : str``` style. Will change!

bollu: I'm also not a fan of re-using variables. The two are conceptually different - why reuse the same variable name?

Eric89GXL: In my mind they're basically the same thing -- a list of paths. It's just that one is space-separated, and the other is a list of strings. Two advantages to re-using the name in Python: 1) it keeps the namespace tidier, and 2) automatic GC can take over the no-longer-needed name.

Eric89GXL: no-longer-needed data is-nasre, not name (that will be re-used)

bollu: compromise - name it ```fname_string``` ? Since it makes it clear that it's supposed to be a string, which is then converted into a list by splitting it up?

vispy/testing/_runners.py
```diff
@@ -190,8 +190,15 @@ def _check_line_endings():
...
 
 
-def _examples():
-    """Run all examples and make sure they work
+def _examples(fnames_str):
+    """Run examples and make sure they work.
+
+    Parameters
+    ----------
+
```
Eric89GXL: Name is fine by me, but under `numpy` docstring spec this newline also shouldn't be there :)

Eric89GXL:
And one last thing -- did you actually try this out to make sure it works as expected?

bollu:
yeah, I tested it. it works. The example strings I wrote in the top comment are what I ran in my shell.

vispy/testing/_runners.py
bollu: haha, okay.
I'm wondering why sublime text's ```flake8``` plugin doesn't lint docstrings too. Or is that not part of the same set of standards?

Eric89GXL: different. PEP8 is a Python-wide standard. Numpy/scipy have their own set of specifications for how docstrings should be formatted for consistency and proper `sphinx` (doc-building) purposes. There are just a few rules, so it shouldn't take too long to get them right.

bollu:
Here's some sample output:
```
 bollu  cantordust  ~  prog  gsoc-vispy  vispy  test-individual-files +  $  
python make test examples examples/demo/gloo/graph.py examples/demo/gloo/mandelbrot.py
Running 2 examples using pyqt4 backend
..
Success: 0 failed, 2 succeeded, 0 skipped in 1.0 seconds
Passed

Testing succeeded (0 failed, 0 skipped) in 1.387 seconds
```

Eric89GXL:
Excellent. I'll merge once the CIs come back happy.

bollu:
cool :)

Hope this is useful to someone else too

Eric89GXL:
We have a problem:

https://travis-ci.org/vispy/vispy/jobs/53870399

FYI Travis runs the `.travis.yml` script, and AppVeyor runs `appveyor.yml` to test the code.

Eric89GXL:
You can probably fix it most easily with `fname_str=''` to give it a default arg

vispy/testing/_runners.py
```diff
@@ -292,7 +309,10 @@ def test(label='full', extra_arg_string=''):
     elif label in backend_names:
         runs.append([partial(_unit, label, extra_arg_string), label])
     if label in ('full', 'examples'):
-        runs.append([_examples, 'examples'])
+        runs.append([partial(_examples, ""), 'examples'])
+    if label == "examples":
+        runs.append([partial(_examples, extra_arg_string),
+                    'examples'])
```
Eric89GXL: I think the logic here is still incorrect. `if label == 'examples'`, then `runs` will have `partial(_examples, "")` appended *and* `partial(_examples, extra_arg_string)`, right?

bollu: ah, right. it should be exclusive. thanks for the catch

Eric89GXL:
Great, thanks for the enhancement!

Issue #581, opened 2014-09-30
realtime_signals.py scale


hidoba:
The image is being scaled around the center, but it should scale around the right end of the signal.
suggested fix:
realtime_signals.py Line 79
vec2 position = vec2(x - (1 - 1 / u_scale.x), a_position);

bollu:
[added a pr](https://github.com/vispy/vispy/pull/745)

PR #745, opened 2015-03-15
fix scaling - scale around center, not right end of signal

bollu:
fix (issue #581)


Eric89GXL:
The title of the PR is backward (this makes it scale holding the right edge at the right edge instead of the center at the center), but otherwise it works :)

Eric89GXL:
Thanks for another fix!

PR #751, opened 2015-03-17
Fix vispy.plot.image fails on float64 textures

### LINKED ISSUE #750: vispy.plot.image fails on float64 textures

brickgao:
Fix #750

### DIFF
--- a/vispy/visuals/image.py
+++ b/vispy/visuals/image.py
@@ -28,8 +28,8 @@ class ImageVisual(ModularMesh):
 
             * 'subdivide': ImageVisual is represented as a grid of triangles
               with texture coordinates linearly mapped.
-            * 'impostor': ImageVisual is represented as a quad covering the 
-              entire view, with texture coordinates determined by the 
+            * 'impostor': ImageVisual is represented as a quad covering the
+              entire view, with texture coordinates determined by the
               transform. This produces the best transformation results, but may
               be slow.
 
--- a/vispy/visuals/image.py
+++ b/vispy/visuals/image.py
@@ -47,7 +47,6 @@ class ImageVisual(ModularMesh):
 
         self._texture = None
         self._interpolation = 'nearest'
-        data = np.array(data, copy=False)
         self.set_data(data)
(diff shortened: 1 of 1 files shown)


vispy/plot/plot.py
```diff
@@ -32,7 +32,8 @@ def image(*args, **kwds):
     """
     canvas = SceneCanvas(keys='interactive')
     canvas.view = canvas.central_widget.add_view()
-    canvas.image = visuals.Image(*args, **kwds)
+    _args = (args[0].astype(np.ubyte),) + args[1:]
```
rossant: I'm wondering whether this should be here of in `visuals.Image`.
Also it may not be correct to cast to ubyte: the user could provide a floating-point texture, in which case we need to cast from float64 to float32. @campagnola or @almarklein may have a better idea of how to implement this correctly.

brickgao:
@rossant I fix it in `visual.image` is more appropriate and I cast data from float64 to float32.

vispy/visuals/image.py
```diff
@@ -47,6 +47,8 @@ def __init__(self, data, method='subdivide', grid=(10, 10), **kwargs):
 
         self._texture = None
         self._interpolation = 'nearest'
+        if data.dtype == np.float64:
+            data = data.astype(np.float32)
```
almarklein: I think it would be better to do this conversion inside `set_data` Maybe we also need to first do something like `data = np.array(data, copy=False)` to convert whatever is given into a numpy array.

brickgao:
@almarklein Thanks for your idea, I made it in `set_data`.

vispy/visuals/image.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, data, method='subdivide', grid=(10, 10), **kwargs):
 
         self._texture = None
         self._interpolation = 'nearest'
+        data = np.array(data, copy=False)
```
almarklein: hehe, I would put this in set_data too :P

brickgao:
@almarklein I get it ;)

Eric89GXL:
Thanks @brickgao!

PR #756, opened 2015-03-22
Added an implicitly connected mesh Visual

inclement:
This visual takes x, y and z coordinates of the surface as separate arguments, and assumes they are vertices of a Cartesian lattice mesh. It is intended to make it simple to plot meshes based on e.g. the output of numpy.meshgrid, as is demonstrated in the included example using the visual to plot spherical harmonics. This is inspired by, though simpler than, Mayavi's [spherical harmonics example](http://docs.enthought.com/mayavi/mayavi/auto/example_spherical_harmonics.html).

I hope I didn't miss this functionality being already somewhere in vispy, but I couldn't find it. I think it's ultimately likely to be useful in a higher level plotting interface - I use this functionality quite frequently in mayavi.

There are some questions about the implementation, particularly how to interpret colors. My view is coloured by ultimately wanting a colormap system as in mayavi, but I guess this would again be part of a higher level plotting function. Anyway, I'm submitting it now to see if you think it's suitable for vispy.

Below is a screenshot of the included example. The lighting seems a little off, but I don't think it's a problem with the mesh. It also takes a couple of seconds to initialise, but (although I haven't looked in detail) the delay doesn't seem to be in the mesh construction but sometime after the window initialisation - is there a way to improve this?

implicit_mesh_example2

### DIFF
--- a/vispy/ext/six.py
+++ b/vispy/ext/six.py
@@ -14,6 +14,7 @@ from distutils.version import StrictVersion
 
 
 _SIX_MIN_VERSION = StrictVersion('1.4.1')
+_SIX_SEARCH_PATH = ['vispy.ext._bundled.six', 'six']
 
 
 def _find_module(name, path=None):
@@ -28,32 +29,42 @@ def _find_module(name, path=None):
             path = [path]
 
         fh, path, descr = imp.find_module(part, path)
+        if fh is not None and part != parts[-1]:
+            fh.close()
 
     return fh, path, descr
 
 
-for mod_name in ['vispy.ext._bundled.six', 'six']:
-    try:
-        mod_info = _find_module(mod_name)
-    except ImportError:
-        continue
+def _import_six(search_path=_SIX_SEARCH_PATH):
+    for mod_name in search_path:
+        try:
+            mod_info = _find_module(mod_name)
+        except ImportError:
+            continue
 
-    mod = imp.load_module(__name__, *mod_info)
+        try:
+            mod = imp.load_module(__name__, *mod_info)
+        finally:
+            if mod_info[0] is not None:
+                mod_info[0].close()
 
-    try:
-        if StrictVersion(mod.__version__) >= _SIX_MIN_VERSION:
-            break
-    except (AttributeError, ValueError):
-        # Attribute error if the six module isn't what it should be and doesn't
-        # have a .__version__; ValueError if the version string exists but is
-        # somehow bogus/unparseable
-        continue
-else:
-    raise ImportError(
-        "Vispy requires the 'six' module of minimum version {0}; "
-        "normally this is bundled with the astropy package so if you get "
-        "this warning consult the packager of your Vispy "
-        "distribution.".format(_SIX_MIN_VERSION))
+        try:
+            if StrictVersion(mod.__version__) >= _SIX_MIN_VERSION:
+                break
+        except (AttributeError, ValueError):
+            # Attribute error if the six module isn't what it should be and doesn't
+            # have a .__version__; ValueError if the version string exists but is
+            # somehow bogus/unparseable
+            continue
+    else:
+        raise ImportError(
+            "Vispy requires the 'six' module of minimum version {0}; "
+            "normally this is bundled with the Vispy package so if you get "
+            "this warning consult the packager of your Vispy "
+            "distribution.".format(_SIX_MIN_VERSION))
+
+
+_import_six()
 
 if PY3:
     file_types = (io.TextIOWrapper, io.BufferedRandom)


examples/basics/visuals/implicitmesh.py
```diff
@@ -0,0 +1,54 @@
+"""
+Demonstration of an implicitly connected mesh.
+"""
+
+import sys
+from vispy import scene
+from vispy.color import get_colormap
+
+import numpy as np
+from scipy.special import sph_harm
```
Eric89GXL: This `scipy` import is what's breaking the Travis build. Can you use `git grep` to find the other example where `scipy` is used? If we skip that example, please also skip this one. I think we usually also add some docstring at the top saying what the non-standard dependencies are, like `scipy`.

Eric89GXL:
This does indeed look useful!

Often with visuals the variables associated with them are not uploaded to the GPU until the first draw, which would explain why the delay is not on initialization, but rather on first draw.

What do you mean about the issue with colors? BTW we do have some support for colormaps in `vispy` if you wanted that functionality.

inclement:
My issue with the colors is that there's a particularly large number of ways you could in principle specify per-vertex colours for this visual - what I've added for now does some basic mangling to convert a 2+1d color array (1 color per 2d grid position) into a normal 1+1d one if necessary before interpreting as a normal ColorArray, but it seems a bit messy.

I've deliberately stayed away from colormaps for now because I'm not sure what API is intended for their use in visuals, and I haven't thought of anything that's not a bit messy, but I'd really like to add this functionality - I originally thought maybe it could be left to a higher level plotting interface, but I read some discussion on another issue that suggested such an interface should remain as close as possible to the actual visuals, which seems sensible.

In the interests of proposing something, how about a 'colormap' argument to specify the colormap and a 'scalars' argument to specify the indices? In this case the scalars argument would be expected to be a 2d array with the same shape as the xs, ys and zs arrays. The scalars would then be normalised to lie in the range 0 -> 1 (or maybe this should have a separate toggle option, mayavi normalises by default but this is sometimes annoying) and used to construct the vertex_colors. I think this convention would extend well to other visuals, but it has the disadvantage of introducing a second color argument type and the need to define some precedence for how the colormap scalars override or otherwise interact with (e.g. multiplying?) the other color or vertex_color arguments.

campagnola:
Looks good! This seems to be a superset of the functionality provided by SurfacePlotVisual, so I recommend taking cues from there about the way the API should look, especially regarding the use of colors--if the user specifies vertex coordinates as (M,N) arrays, then colors should be given as (M,N,3), (M,N,4) or an (M,N) array of Color instances (@Eric89GXL is it possible to make an N-dimensional ColorArray?). For now I would avoid implementing anything with ColorMap because we have not worked out how to do this with any mesh yet.

vispy/geometry/generation.py
```diff
@@ -289,3 +289,56 @@ def create_arrow(rows, cols, radius=0.1, length=1.0,
         faces = np.vstack((faces, md_cyl.get_faces()+nbr_verts_con))
 
     return MeshData(vertices=verts, faces=faces)
+
+
+def create_implicit_mesh(xs, ys, zs):
```
campagnola: `SurfacePlotVisual` has a vectorized method for generating the mesh vertices and indices; you should be able to adapt that for `create_implicit_mesh()`. (Ideally, `SurfacePlotVisual` could be reimplemented as a very lightweight subclass of `ImplicitMeshVisual`)

vispy/visuals/implicitmesh.py
```diff
@@ -0,0 +1,66 @@
+
+from .mesh import MeshVisual
+from ..geometry import create_implicit_mesh
+from ..color import ColorArray
+
+import numpy as np
+
+
+class ImplicitMeshVisual(MeshVisual):
```
campagnola: "implicit" is not descriptive enough, as there are many ways a mesh could be implicitly constructed around an array of vertices. I would recommend using "grid" instead--`GridMeshVisual`,  `create_grid_mesh()`, etc.

vispy/visuals/implicitmesh.py
```diff
@@ -0,0 +1,66 @@
...
+        this has the same shape as the xs, ys and zs arrays and gives
+        one colour per vertex.
+    shading : str | None
+        Same as for the `MeshVisual` class. Defaults to 'smooth'.
+    vertex_colors: ndarray | None
+        Same as for the `MeshVisual` class. Defaults to None. Overrides
+        the color argument if set.
+    face_colors: ndarray | None
+        Same as for the `MeshVisual` class. Defaults to None.
+    mode : str
```
campagnola: This is not needed--the visual generates the mesh indices and therefore knows exactly what triangle mode should be used.

inclement: I was using this to switch between lines and triangles, rather than alternative triangle modes, but I guess that's not important either.

Eric89GXL:
At this point ColorArray is only of shape (N, 4), not (N, M, ..., 4). We could extend it to support that case if we want, I think we could make those changes in a backward-compatible manner. It would be nice if we didn't need to, but I could imagine it making the coding of other objects like this cleaner.

campagnola:
I'd like to have ND color arrays at some point (I have used them frequently in the past) but it's not a priority right now.

rossant:
> I'd like to have ND color arrays at some point

+1

Eric89GXL:
Sorry that this one fell to the back burner @inclement -- where are we here? It looks like you took care of most of the issues except for the API one brought up by Luke [up here] (https://github.com/vispy/vispy/pull/756#issuecomment-84628547). Can you rebase and take a crack at addressing that?

Eric89GXL:
Oh actually you do have a commit to deal with color handling -- so maybe you took care of it but didn't ping for another review?

inclement:
@Eric89GXL I think the color api is more or less right after that commit, but I didn't ping because I was intending to come back to it and do it it more like the surface visual - as I remember, it makes use of some automatic conversion in the MeshData that I didn't want to reinvent. I also wanted to use set_data properly as part of it.

 I didn't get around to that, but I'll try to do so in the next couple of days. Sorry, it fell on my back burner too.

Eric89GXL:
@inclement was closing this intentional?

inclement:
No, I somehow messed up a rebase and pushed an empty branch so github closed it automatically. I'm just trying to work out what I did wrong, but I'll push here again when I fix it.

Eric89GXL:
Ahh, cool. Feel free to open another PR if it's easier.

PR #777, opened 2015-03-25
Convert wx DClick into vipsy_mouse_press events

StephenWest:
Wx converts the second mouse-press into a Click event.

This change converts these DClick events into standard vispy_mouse_press events, so that mouse press events are not skipped


Eric89GXL:
LGTM. Have you tested that this works in your code?

almarklein:
LGTM, but I'm wondering: what need we do if we want double clicks at some point?

Eric89GXL:
We can just modify this to go to our new double-click event as well, assuming that would make it consistent with other backends (sending both a mouse press/release and doubleclick event on the second click).

PR #781, opened 2015-03-25
Proposing a bgcolor property for SceneCanvas

inclement:
This makes it easier to dynamically change the bgcolor.

Changing one line to two lines in the `__init__` isn't so nice...it could be left as it was and work fine but with very minor code redundancy.

### DIFF
--- a/vispy/scene/canvas.py
+++ b/vispy/scene/canvas.py
@@ -87,10 +87,11 @@ class SceneCanvas(app.Canvas):
         self._fb_stack = []  # for storing information about framebuffers used
         self._vp_stack = []  # for storing information about viewports used
         self._scene = None
-        self._bgcolor = Color(kwargs.pop('bgcolor', 'black')).rgba
         
         # A default widget that follows the shape of the canvas
         self._central_widget = None
+
+        self._bgcolor = Color(kwargs.pop('bgcolor', 'black')).rgba
 
         app.Canvas.__init__(self, *args, **kwargs)
         self.events.mouse_press.connect(self._process_mouse_event)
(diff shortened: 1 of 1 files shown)


vispy/scene/canvas.py
```diff
@@ -87,17 +87,20 @@ def __init__(self, *args, **kwargs):
         self._fb_stack = []  # for storing information about framebuffers used
         self._vp_stack = []  # for storing information about viewports used
         self._scene = None
-        self._bgcolor = Color(kwargs.pop('bgcolor', 'black')).rgba
```
Eric89GXL: can't it just be `self.bgcolor = kwargs.pop('bgcolor', 'k')` to keep it as one line?

inclement: The splitting over two lines is because the `self.update()` in the setter must be called after the super `__init__`, but the argument must be popped from the kwargs before that happens. I could add a check before calling self.update() instead, I think, would that be better?

Eric89GXL: ahh it makes sense to split it then, didn't notice the super class init

Eric89GXL: actually, it *could* potentially cause a problem if something later (like a draw handler) gets called by the superclass init and expects `self._bgcolor` to already exist... so in this case I would say go ahead and violate DRY and just have the `Color` interpreting done in two places -- here and the setter.

inclement: Got it, I've changed it back.

Eric89GXL:
Thanks @inclement!

PR #793, opened 2015-03-26
ENH: Unify transforms

### LINKED ISSUE #97: Improve transforms.py
See discussion https://github.com/vispy/vispy/commit/76a4e924de4ece56658eb24f8d1d3a1636a8274d

The transforms now do the operation in-place *and* return the result. Better would be to avoid in-place operations and return a new matrix.

Also see http://glm.g-truc.net/0.9.5/index.html

### LINKED ISSUE #507: Inconsistency in use of transformation matrices (transposed)
In vispy, transformation matrices in util/transforms.py (created by translate etc) have the form
```
( r r r 0)
( r r r 0)
( r r r 0)
( tx ty tx 1)
```
I.e. with the translation as the last row. This is the convention in DirectX if I am correct. However, in OpenGL the convention is to have the translation in the last column:
```
( r r r tx)
( r r r ty)
( r r r tz)
( 0 0 0 1)
```

These are mathematically equivalent ways to store the data, (the matrix is just transposed). In vispy, this is converted in vispy/gloo/variable.py. However, this does cause an issue: with the transposition, the order of multiplication is reversed: `A.T * B.T = (B*A).T`. Thus, lets say we have two matrices A and B, and in the shader language we have two mat4 uniforms `u_x` and `u_y` Now if in the shader we have a statement `u_x` * `u_y`, Then the following two cases have a different effect:
* `u_x = A, u_y = B`
* `u_x = eye(4), u_y = A*B` (A and B ar numpy matrices so * denotes dot product)

The first case would be equivalent to `u_x = eye(4), u_y = B*A`, which is counter-intuitive: in shader language A*B is equivalent to B*A in Python. This file demonstrates the issue: https://gist.github.com/rreilink/9e787d0fa7c939519652 . In the on_timer handler, the uniforms `u_a` and `u_b` are for each frame changed between the two cases named above. This clearly shows that the two cases, which should be equivalent, are not.

I think this is an issue, since the order of multiplication of transformation matrices/ the order of applying operations is a very easy thing to get wrong. Luckily, the solution is not very difficult. 
* Change utils/transformations.py to return matrices according to OpenGL convention
* Change vispy/gloo/variable.py to not transpose matrices when setting the uniforms
* Change scene/transforms/linear.py to use post-multiply (`self.matrix*=translate(...)`) instead of pre-multiply (`self.matrix[...] = translate(...) * self.matrix`) which is currently used in my latest pull-request.

If you agree to this I'd be happy to implement the proposed solution

Eric89GXL:
Closes #97.
Closes #507.

Not ready for merge since it doesn't seem to work quite right yet...

### DIFF
--- a/vispy/scene/cameras/cameras.py
+++ b/vispy/scene/cameras/cameras.py
@@ -13,12 +13,12 @@ from ...util.quaternion import Quaternion
 from ...util import keys
 from ..node import Node
 from ...geometry import Rect
-from ...visuals.transforms import (STTransform, PerspectiveTransform, 
+from ...visuals.transforms import (STTransform, PerspectiveTransform,
                                    NullTransform, AffineTransform,
                                    TransformCache)
 
 
-def sind(q): 
+def sind(q):
     return math.sin(q*math.pi/180)
 
 
@@ -32,7 +32,7 @@ def cosd(q):
 
 def make_camera(cam_type, *args, **kwargs):
     """ Factory function for creating new cameras using a string name.
-    
+
     Parameters
     ----------
     cam_type : str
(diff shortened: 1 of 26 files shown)


Eric89GXL:
@almarklein ready for review/merge, tests pass over here. You will probably need to manually test.

vispy/util/tests/test_transforms.py
```diff
@@ -14,17 +13,19 @@ def test_transforms():
     """Test basic transforms"""
     xfm = np.random.randn(4, 4).astype(np.float32)
 
-    for rot in [xrotate, yrotate, zrotate]:
-        new_xfm = rot(rot(xfm, 90), -90)
-        assert_allclose(xfm, new_xfm)
-
-    new_xfm = rotate(rotate(xfm, 90, 1, 0, 0), 90, -1, 0, 0)
+    # Do a series of rotations that should end up into the same orientation
+    # again, to ensure the oreder of computation is all correct
```
rossant: order

rossant:
LGTM

almarklein:
Tested that all exampled that were modified still work. However, some scene/plotting examples don't. Basically everything that uses a 3D camera is broken.

vispy/scene/cameras/cameras.py
```diff
@@ -630,7 +631,7 @@ def viewbox_mouse_event(self, event):
         BaseCamera.viewbox_mouse_event(self, event)
         
         if event.type == 'mouse_wheel':
-            center = self._scene_transform.imap(event.pos)
+            center = np.array(self._scene_transform.imap(event.pos))[0]
```
Eric89GXL: In addition to changing what `*` means from broadcast-multiply to dot (under most circumstantces), I found another thing we must keep in mind when using `matrix` instead of `ndarray`. If you take a dot product between a matrix and a 1D vector, you get a matrix (2D) back:
```
>>> import numpy as np
>>> np.dot(np.matrix(np.ones((2, 2))), np.ones(2))
matrix([[ 2.,  2.]])
```
We'll have to keep exceptions like this in mind now that we have adopted use of the `matrix` class in a subset of our operations.

PR #794, opened 2015-03-26
Fix mpl export and camera details

almarklein:


examples/basics/plotting/mpl_plot.py
```diff
@@ -14,6 +14,7 @@
 import numpy as np
 import sys
 
+# import matplotlib.pyplot as plt
```
Eric89GXL: Did you need to comment this out? Might as well delete it if we don't want it...

vispy/scene/widgets/viewbox.py
```diff
@@ -157,8 +159,10 @@ def get_scene_bounds(self):
         for axis in (0, 1, 2):
             if np.inf in [np.abs(x) for x in bounds[axis]]:
                 bounds[axis] = -1, 1
-        
-        return bounds
+        if dim:
```
Eric89GXL: Couldn't dim be zero, so this should be `if dim is not None`?

examples/basics/plotting/mpl_plot.py
almarklein: No, I added it to show that you can just replace the import and it will run in vispy instead of MPL. Maybe clarify that a bit?

vispy/scene/widgets/viewbox.py
almarklein: good call

examples/basics/plotting/mpl_plot.py
Eric89GXL: ahh, no it's fine then

Eric89GXL:
Good deal, thanks

PR #800, opened 2015-03-26
WIP: Collection transforms

campagnola:
This PR is to merge the vispy transform system functionality into collections.


campagnola:
Ready for review @rossant @rougier

examples/collections/choropleth.py
```diff
@@ -9,17 +9,13 @@
 from vispy import app, gloo
 from vispy.util import load_data_file
 from vispy.visuals.collections import PathCollection, PolygonCollection
-from panzoom import PanZoom
+from panzoom import PanZoomTransform
```
rougier: I would prefer something like:
```
from vispy import transforms
...
panzoom = transforms.PanZoom()
```

rossant: +1: better if PanZoom is not in examples/

examples/collections/choropleth.py
```diff
@@ -41,6 +37,16 @@ def add(P, color):
...
 
+# Create the canvas.
+canvas = app.Canvas(size=(800, 800), keys='interactive')
+gloo.set_viewport(0, 0, canvas.size[0], canvas.size[1])
+gloo.set_state("translucent", depth_test=False)
+
+panzoom = PanZoomTransform(canvas)
+paths = PathCollection(mode="agg+", color="global", transform=panzoom)
+polys = PolygonCollection("raw", color="local", transform=panzoom)
+paths.update.connect(canvas.update)
```
rossant: why not the same for polys?

examples/collections/tiger.py
```diff
@@ -10,8 +10,9 @@
 from vispy.util import load_data_file
 from vispy.util.svg import Document
 from vispy.visuals.collections import PathCollection, PolygonCollection
-from panzoom import PanZoom
-
+from panzoom import PanZoomTransform
+import faulthandler
+faulthandler.enable()
```
rossant: what is that? to debug segfaults? I suppose you didn't intend to keep that.

FWI I also had some segfaults with this example, they might be related to the *triangles* library..?

examples/collections/choropleth.py
rougier: Paths are all black while color are local to each poly.
Note that in glumpy I get different colors for each counties (not state). There may be a bug in parsing the geo file (I used a topojson file in glumpy).

rossant:
in choropleth, I think the aspect ratio should be kept fixed

rossant:
@rougier or @campagnola could you benchmark tiger or choropleth before/after the PR (with no vsync) ?

examples/collections/tiger.py
campagnola: Yeah, forgot to take that out. It was triangles for me too.

examples/collections/choropleth.py
campagnola: It's not necessary here because they share the same transform.

campagnola:
I get ~44fps for both branches.

examples/collections/choropleth.py
campagnola: Moved to `visuals/transforms/interactive.py`

rougier:
44 fps is not good...

rossant:
How much do you get? I get 52 FPS on both of my GPUs (which is a bit weird since one is supposed to be much better...)

rougier:
Using glumpy, for a 1000x1000 window, I get ~200 fps for `tiger.py`.

rossant:
According to Luke this is not caused by the present PR. @campagnola any idea about the performance gap between glumpy and VisPy for the same example?

rougier:
One "problem" is that vispy uses vsync, this means you're stuck at 60 fps anyway. But if you're below 60fps, that's seems a bit low for the tiger or chloropeth. What abut the collection point example ?

rossant:
Can't we disable vsync with VisPy?

Idem: ~ 51 FPS.

rossant:
I'm at 60 FPS on gloo demos.

rougier:
Can yout try a 512x512 window with a single point in the collection (and not moving the mouse) and every attribute to be "local" or "global" (no shared). How much fps do you get from the gloo-point-cloud example ?

rossant:
> How much fps do you get from the gloo-point-cloud example ?

60

rossant:
> Can yout try a 512x512 window with a single point in the collection (and not moving the mouse) and every attribute to be "local" or "global" (no shared).

60

Eric89GXL:
The canvas has a vsync flag you can set to False. IIRC it is not respected
by all backends on all platforms, but it should work with a majority of
them (including whichever it works for on Glumpy; we don't do anything
special).

Eric89GXL:
Or maybe it must be passed as part of the context config... in any case it
can be disabled.

campagnola:
If I add `canvas.update()` to the draw callback in `tiger.py` then I get 112 fps on both branches. Probably my 40fps was limited by the rate of mouse events.

rossant:
@campagnola how do you disable vsync? `Canvas(vsync=False)` doesn't work for me.

campagnola:
Use a DE with broken vsync :)

Eric89GXL:
@rossant I think pyqt4 on Linux might be one of the ones that doesn't
respect the parameter, if that's what you're on. Try a different backend.
Also make sure your driver isn't set to force vsync.

campagnola:
If there are no other comments, this should be ready to merge. Let's do benchmarking elsewhere.

rossant:
LGTM then!

examples/collections/tiger.py
```diff
@@ -10,8 +10,9 @@
 from vispy.util import load_data_file
 from vispy.util.svg import Document
 from vispy.visuals.collections import PathCollection, PolygonCollection
-from panzoom import PanZoom
-
+from vispy.visuals.transforms import PanZoomTransform
+import faulthandler
+faulthandler.enable()
```
Eric89GXL: did you mean to get rid of this?

campagnola: I thought I did?? Pushing again..

campagnola: Done; thanks for looking carefully.

Eric89GXL:
Hooray, a PR with more red than green! :)

LGTM as well, I'll merge once I see green unless someone else beats me to it.

Eric89GXL:
Great, thanks for moving us toward collection visuals @campagnola. You planning to keep working on that bit?

campagnola:
Eventually. I'm working on a long list of related things that need to change to move the scenegraph forward. The collection visuals could probably be dealt with independently, though, if you'd like to give that a go.

Eric89GXL:
Okay, I'll ping on the issue if I start. We'll see who gets there first :)

bollu:
how would this need to be changed to we can integrate this fully? I need pointers on how the merge of the collections with the SceneGraph is going to happen :) could someone point me to the relevant code?

campagnola:
I've been thinking about this a lot lately. I'm working on an overhaul of the visual system (#895) that will hopefully get us closer to making this happen. It probably doesn't make sense to start on collections until that PR is complete.

Eric89GXL:
@bollu so perhaps it makes sense for now to help @campagnola with getting other aspects of #894 done. Let's discuss more over there.

PR #804, opened 2015-03-27
ENH: Plotting

### LINKED ISSUE #443: FIX: Visual vs ModularVisual
@almarklein @lcampagn do we have consensus on this yet? It would be good to unify these before release, if possible. It would mean converting a few visuals over to one system on another. WDYT?

### LINKED ISSUE #598: visual.Image, and vispy.plot.image, not working on Linux with Mesa driver
This is really weird. I have a fresh installation of Linux Mint (which should behave as Ubuntu). I've got the latest VisPy from Git, and the examples involving images (basics/visuals/image_visual.py, basics/scene/image.py, and basics/plotting/image.py) all don't work. Everything else, including the gloo/animate_image.py, seems to work fine.

This seems like there is a bug in the Image visual. I also just tried the latest VisPy on my Mac, and the Image visual there is fine.

Edit: Using Python 2.7.

### LINKED ISSUE #676: Add colormap support to ImageVisual
Suggest that a similar interface to matplotlib imshow colormap support be provided in Vispy - see examples [here](http://matplotlib.org/users/image_tutorial.html).

### LINKED ISSUE #680: Histogram visual
I would need to display many histograms in a single pass. A visual that only supports a single histogram might be sufficient if we have collections.

Features:

* with/without border
* ticks
* tick labels
* baseline

[Here is an implementation based on Galry](https://github.com/klusta-team/klustaviewa/blob/master/klustaviewa/views/correlogramsview.py)

Screenshot:

Histograms

### LINKED ISSUE #749: Improve vispy.plot interface and implementation
This issue regroups a few thoughs and suggestions about the plot interface.

* Allow marker_size argument in LinePlot
* in `vp.plot()` markers should be optional, and off by default
* All public functions should have a full docstring
* The plot interface should be as close as possible as the scene interface. A user who wants to go from the plot interface to the scene interface for more customization and interactivity options should not need to learn yet another way of doing the same thing.
* We keep a stack of all opened figures (=canvases). When issuing a new plot command, the last opened canvas is used, or a new one is created.
* `vp.figures()` returns the list of opened figures.
* `vp.figure()` creates a new figure and returns it.
* `vp.figure(index=2)` returns the 3 opened figure, or raises an error if it doesn't exist.
* `vp.figure(index='last')` or `index=-1` returns the last opened figure.
* Plot functions return a list of visuals that have been added to the current figure.
* All plot functions will share some set of keyword arguments, and the implementation should factor out that reusable set of functionality among the functions. Example:
  * background color
  * camera
* Subplot API
* Grid, axes, legend, labels...

### LINKED ISSUE #827: memory leak in vispy
Below is a minimal example. When pressing the r key a random is image is loaded as an ImageVisual.
The memory leak is still present after recent fixes #784, #790.

    from __future__ import division
    
    #%% Load packages
    import os,sys
    import numpy as np
    import vispy.app
    from vispy import gloo
    from vispy import visuals
    from vispy.visuals.transforms import STTransform, TransformSystem
    from PySide import QtGui, QtCore
    
    class DosWindow(QtGui.QMainWindow):
        verbose=1
        def __init__(self, image):
            super(DosWindow, self).__init__()
            self.setGeometry(0, 0, 800, 600 )
            self.canvas = Canvas(image)
            self.image = image.copy()
            layout = QtGui.QVBoxLayout()
            layout.addWidget(self.canvas.native)
            layout.setContentsMargins(QtCore.QMargins(0, 0, 0, 0))
            central_widget = QtGui.QWidget()
            central_widget.setLayout(layout)
            self.setCentralWidget(central_widget)
            self.show()
    
        def keyPressEvent(self, event):
            if event.key() == QtCore.Qt.Key_Q:
                exit(0)
            if event.key() == QtCore.Qt.Key_R:
                newimage=(0.4*np.random.rand()) + self.image            
                self.canvas.setImage( newimage )
                self.canvas.update()
                print('reload image')
    
    class Canvas(vispy.app.Canvas):
        def __init__(self, image):
            super(Canvas, self).__init__()
    
            self.size = ((image.shape[1]) , image.shape[0] )
            self.imsize=self.size
            self.setImage(image)
    
            # Create a TransformSystem that will tell the visual how to draw
            tr_sys = TransformSystem(self)
            tr_sys.visual_to_document = self.image_transform
            self.transform_system=tr_sys
    
        def setImage(self, image):
            self.image = visuals.ImageVisual(image, method='subdivide')
            self.image_transform=STTransform(scale=(1, 1), translate=(0, 0))
    
        def on_draw(self, ev):
            gloo.clear(color='white', depth=True)
            gloo.set_viewport(0, 0, *self.size)
            self.image.draw(self.transform_system)
    
    if __name__ == '__main__':
        import sys
    
        im=np.random.rand( 2000,5000)
        print('loaded data...')
        if QtGui.QApplication.startingUp():
            app = QtGui.QApplication(sys.argv)
        else:
            app = QtGui.QApplication.instance()
            
        win = DosWindow(im)
        if sys.flags.interactive != 1 :
            app.exec_()

Eric89GXL:
Closes #827.
Closes #749.
Closes #680.
Closes #676.
Closes #598.
Closes #443.


vispy/scene/widgets/grid.py
```diff
@@ -56,7 +57,9 @@ def add_widget(self, widget=None, row=None, col=None, row_span=1,
 
         _row = self._cells.setdefault(row, {})
         _row[col] = widget
-        self._grid_widgets[widget] = row, col, row_span, col_span
+        self._grid_widgets[self._n_added] = (row, col, row_span, col_span,
```
campagnola: dick + n_added looks like a roundabout way of making a list.

QuLogic:
Re: copyright lines, it's 2015 now. ;)

Eric89GXL:
@campagnola since it seems like some things depend on the changes I've made here, do you think it makes sense for me to polish what is currently here so we can do some rounds of review and merge something basic? It might also reduce future review burden to have it done in a couple of passes.

campagnola:
FYI I am working on images in a sub-PR.

Eric89GXL:
Cool, I'll wait to work until you make your PR, assuming it doesn't take too long. WDYT about my question above about trying to merge a simple version soon so that this doesn't hold up other progress?

campagnola:
Let's wait for my changes since this breaks some examples. I should be done soon. If that turns out to be a lie, then we can reconsider :)

Eric89GXL:
Yeah I think I might have accidentally rotated the image for example...

I mean, once your PR into my branch is merged, WDYT about my making a push to get a simple version of the new `vispy.plot` interface merged? For example, it won't be feature / plot-type complete and won't have axes, but we could work on those in subsequent PRs.

campagnola:
I'm fine with that. It'll give people an early chance to play with the API.

Eric89GXL:
Sounds good. Are you working on #827 (memory leak) at all in your PR, or just fixing image support?

campagnola:
It appears to be magically fixed in my PR.

Eric89GXL:
Awesome, I'll merge your PR into my branch and hopefully find time to clean this PR soon

Eric89GXL:
@campagnola I still see a memory leak running code from #827, do you not see it at your end?

figure_1

Eric89GXL:
@campagnola can you remind me how to do circular reference checking? It's a submodule in PyQtGraph for now, right?

campagnola:
Check it out:
https://github.com/campagnola/reftracker

Do something like:
```
$ python -i leaktest.py 
>>> ot = reftracker.ObjTracker()
>>> # press 'r' a few times
>>> ot.diff()
>>> # press 'r' a few times
>>> ot.diff()
>>> # print reference tree for the largest persistent array (or whatever type you are looking for):
>>> arrays = ot.findPersistent('ndarray')
>>> arrays.describe(0)
```

I see some leaked shader functions, but they are relatively small. I also notice that the behavior of the test script is quite different on your branch than it is in master..

vispy/scene/widgets/grid.py
Eric89GXL: yeah, I know -- it's more for future compatibility in case we want to be able to delete axes and such. This way we can `.pop` without needing to update index numbers or anything.

Eric89GXL:
Okay, I think this is ready for a first round of testing and review.

In the future, we might want to refactor the classes a bit, but this seems to work decently for now.

Eric89GXL:
Sorry for all the lines changed. Quite a few are removing `ModularVisual` code and adding `cubehelix` to `vispy.ext`, so it's not quite as bad as it looks, though.

examples/basics/plotting/scatter_histogram.py
```diff
@@ -0,0 +1,24 @@
...
+color = (0.8, 0.25, 0.)
+n_bins = 100
+
+fig = vp.Fig(show=False)
+fig[0:4, 0:4].plot(data, width=0, face_color=color + (0.05,), edge_color=None)
+fig[4, 0:4].histogram(data[0], bins=n_bins, color=color, orientation='h')
+fig[0:4, 4].histogram(data[1], bins=n_bins, color=color, orientation='v')
+
+if __name__ == '__main__':
+    fig.show(run=True)
```
rossant: The scatter plot is not clipped when panning -- is that normal?

examples/basics/plotting/scatter_histogram.py
```diff
@@ -0,0 +1,24 @@
...
+# -*- coding: utf-8 -*-
+# vispy: gallery 1
+
+"""
+A scatter plot of 2D points with matching histograms.
+"""
+
+import numpy as np
+
+from vispy import plot as vp
```
rossant: or `import vispy.plot as vp`?

examples/basics/plotting/quick.py
```diff
@@ -0,0 +1,30 @@
...
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
+Example of plotting multiple things using the quick interface.
+"""
+
+import numpy as np
+
+from vispy.io import load_data_file
+import vispy.plot.q as vpq
```
rossant: am I the only one to find this hideous? :)

examples/basics/plotting/scatter_histogram.py
```diff
@@ -0,0 +1,24 @@
...
+
+n = 100000
+data = np.random.randn(2, n)
+color = (0.8, 0.25, 0.)
+n_bins = 100
+
+fig = vp.Fig(show=False)
+fig[0:4, 0:4].plot(data, width=0, face_color=color + (0.05,), edge_color=None)
+fig[4, 0:4].histogram(data[0], bins=n_bins, color=color, orientation='h')
+fig[0:4, 4].histogram(data[1], bins=n_bins, color=color, orientation='v')
```
rossant: I like this way of specifying subplots

rossant:
from a quick look it seems great, thanks @Eric89GXL (and bonus point for closing many issues without adding many lines of code :) )

examples/basics/plotting/quick.py
campagnola: You are not alone. Here's an idea: what if we expose the CLI plotting interface via a completely separate top-level package?
```
import vplot as vp
vp.plot(...)
```
I seem to recall matplotlib doing something similar with `pylab`, is that right?

rossant: Looks good. Also, is it conceivable to have the same namespace, and toggle interactive mode with something like `vp.enable_interactive()`?

vispy/plot/_fig.py
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
```
campagnola: No reason to hide this module..

vispy/plot/_fig.py
```diff
@@ -0,0 +1,60 @@
...
+                                  show=show, size=size)
+        self._grid = self.central_widget.add_grid()
+        self._plot_widgets = []
+
+    @property
+    def plot_widgets(self):
+        """List of the associated PlotWidget instances"""
+        return tuple(self._plot_widgets)
+
+    def __getitem__(self, idxs):
```
campagnola: Can this be implemented on `Grid` instead?

examples/basics/plotting/scatter_histogram.py
campagnola: It is normal. The compound visuals do not yet support the `Visual.attach` API, which is necessary for clipping. That's in my current list of projects.

vispy/plot/_plotwidget.py
```diff
@@ -0,0 +1,254 @@
...
+        -------
+        spec : instance of Image
+            The spectrogram.
+
+        See also
+        --------
+        Image
+        """
+        # XXX once we have axes, we should use the "fft_freqs", too
+        data = stft(x, n_fft, step, fs)
```
campagnola: Ideally, this should create a `Spectrogram` instance which would be a thin wrapper around `Image`. Ok to leave that for later, though..

campagnola:
This looks good to me. All of the issues raised above look like they can be taken care of in future PRs.

campagnola:
`examples/basics/visuals/line_plot.py` appears to be broken: `TypeError: Too many dimensions given (max is 2).` but this was not picked up by the example tester..?

I think we want to allow 3D data in line_plot anyway.

examples/basics/plotting/quick.py
Eric89GXL: I don't think the "quick" plotting methods should live in the same namespace as the figure-based methods. Users of other plotting libraries e.g. matplotlib or MATLAB will be confused about the differences in behavior, e.g. why their plotting functions return figures/canvases instead of a visual, and why calling `plot` more than once causes multiple windows to show up.

I am -1 on having an entirely different top-level package, either, but I can live with it if other people want it.

examples/basics/plotting/scatter_histogram.py
Eric89GXL: thank @almarklein :)

Eric89GXL:
I'll fix the other comments tonight hopefully, I don't think they should be too bad

vispy/plot/_plotwidget.py
Eric89GXL: That would also be nice because you could eventually make it so the step and such could be adjusted, which would be nice. I'll turn it into a visual like I did for Histogram earlier.

examples/basics/plotting/quick.py
campagnola: I am -1 on inheriting baggage from MATLAB. This is a fairly simple API issue that will take most users only a couple of minutes to adjust to. I also don't think having a separate namespace in any way addresses that issue. These users will be equally confused whether they write `vp.plot()` or `vpq.plot()`. They'll look at the examples, and then they'll say "AH! That makes so much more sense than Matplotlib".

Eric89GXL: Well, if users have to adjust to an API, then I think it makes sense to just adopt the other one in this PR, and scrap the `quick` one entirely. In other words, just have this one:
```
fig = Fig()
fig[0, 0].plot(...)
```
For now at least I'll remove the `quick` stuff, and we can figure out later if we really need an additional API or not.

Eric89GXL:
I pushed a couple of commits to address the comments, but GitHub hasn't updated the PR (?). Hopefully it updates soon, otherwise I'll have to close and reopen.

PR #805, opened 2015-03-27
Made scroll wheel zoom work when fov > 0

inclement:
This copies the way right click zooming is handled.

### DIFF
--- a/vispy/scene/cameras/cameras.py
+++ b/vispy/scene/cameras/cameras.py
@@ -735,6 +735,8 @@ class PerspectiveCamera(BaseCamera):
         if event.type == 'mouse_wheel':
             s = 1.1 ** - event.delta[1]
             self._scale_factor *= s
+            if self._distance is not None:
+                self._distance *= s
             self.view_changed()
     
     @property


inclement:
It looks like the checks all failed for reasons unrelated to the PR.

Eric89GXL:
What is the problem that you're having that this fixes?

inclement:
@Eric89GXL If the fov of a PerspectiveCamera is zero, the scroll wheel zooms in and out (by modifying self._scale_factor). However, if the fov is non-zero, the scroll wheel does nothing.

Using right click and dragging zooms in and out regardless of the fov, because it modifies both self._scale_factor and self._distance, so I copied the latter modification.

almarklein:
> if the fov is non-zero, the scroll wheel does nothing.

It works for me. Did you try latest master?

almarklein:
Oh wait, what camera are you using exactly?

almarklein:
Ah now I see. It's when a distance is specified. I overlooked that, sorry.

almarklein:
Merging. Travis is half-down, so we don't want to wait for that. Locally checked that this worked. Thanks!

PR #808, opened 2015-03-27
All cameras now handle the "up" property correctly

almarklein:
PanZoom can thus operate in any plane, Turntable can be tilted in any plane, and FlyCamera will attempt to re-orient to make up point upward.

### DIFF
--- a/vispy/scene/cameras/cameras.py
+++ b/vispy/scene/cameras/cameras.py
@@ -645,8 +645,8 @@ class PanZoomCamera(BaseCamera):
                 # Translate
                 p1 = np.array(event.last_event.pos)[:2]
                 p2 = np.array(event.pos)[:2]
-                p1s = self._scene_transform.imap(p1)
-                p2s = self._scene_transform.imap(p2)
+                p1s = self._transform.imap(p1)
+                p2s = self._transform.imap(p2)
                 self.pan(p1s-p2s)
             
             elif 2 in event.buttons and not modifiers:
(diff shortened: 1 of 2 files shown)


vispy/scene/cameras/cameras.py
```diff
@@ -688,10 +689,42 @@ def _update_transform(self):
...
+        # transform When up is +z this is all. In other cases,
+        # self.transform is now set up correctly to allow pan/zoom, but
+        # for the scene we need a different (3D) mapping. When there
+        # is a minus in up, we simply look at the scene from the other
+        # side (as if z was flipped).
+        
+        if self.up == '+z':
+            # Keep it simple, but we give the scene an affinetransform
+            # to be consistent
+            thetransform = self.transform * AffineTransform()
```
campagnola: I'd rather it left as-is because affine should be a bit slower.

Eric89GXL: Doesn't it end up as a dot product on the GPU either way? Or do you mean in the scenegraph before it gets to the GPU?

campagnola: Most transforms are just scale+translate.

almarklein: I had it unmodified at first, but then when I cycled through the up vectors, the z+ would not work, maybe a problem with transform caching? Anyway, I cannot imagine that the performance difference is noticeable.

Eric89GXL: Yeah, but I thought these were done on GPU with matrix multiplies either way...

PR #817, opened 2015-03-27
Glumpy examples

rossant:
* [x] Fixed mandelbrot colors
* [x] Added a few functions in geometry subpackage (preparing for #819 which will require some improvements in `vispy.plot`)
* [x] Added quiver example


rossant:
Ready for review/merge.

Eric89GXL:
Failing because examples should not throw warnings. Can you clean up the example shader not to have this problem?

vispy/geometry/normals.py
```diff
@@ -0,0 +1,82 @@
...
+    L = np.sqrt(np.sum(N * N, axis=1))
+    L[L == 0] = 1.0  # prevent divide-by-zero
+    N /= L[:, np.newaxis]
+    normals = np.zeros_like(vertices)
+    normals[indices[:, 0]] += N
+    normals[indices[:, 1]] += N
+    normals[indices[:, 2]] += N
+    L = np.sqrt(np.sum(normals*normals, axis=1))
+    L[L == 0] = 1.0
+    normals /= L[:, np.newaxis]
```
Eric89GXL: We already have an optimized `_calculate_normals` function, this should probably make use of it

rossant: done

rossant:
> Failing because examples should not throw warnings. Can you clean up the example shader not to have this problem?

done

rossant:
travis is happy, this is ready to merge

Eric89GXL:
I will review and try examples shortly

vispy/geometry/parametric.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014, Nicolas P. Rougier
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+
+import numpy as np
+from .normals import normals
```
Eric89GXL: Did you delete this file? It looks like you did (since it was in the last review but not this one), but the ref is still here...?

Eric89GXL:
Other than my last comment, LGTM. I'll merge once we deal with that one.

vispy/geometry/parametric.py
rossant: Whoops. The surface function uses normals(), and I haven't tested that it works with the existing normal implementation. It's not possible to put back normals.py and merge the two implementations later once we properly test, compare, and benchmark them?

Eric89GXL: It's generally not a good idea to have the same thing done in multiple places in different ways (DRY), but given that the collections already need quite a bit of de-duplication, I can live with it.

Eric89GXL:
There is an error in `galaxy.py` that shouldn't be due to your changes, and is probably fixed in my PR, so this should be good.

PR #822, opened 2015-03-27
WIP: Visual filters and clipping

campagnola:
* Adds a basic API for attaching a filter to a visual:  `Visual.attach`
* Adds a few simple filters including clipping
* Adds fragment shader-based clipping to ViewBox

Still has a few bugs to work out.


campagnola:
Ready for review. The travis error is a failed download.

vispy/visuals/shaders/function.py
```diff
@@ -679,306 +546,6 @@ def remove_callback(self, hook, func):
...
-    def dtype(self):
-        return self._function.rtype
-    
-    def expression(self, names):
-        str_args = [arg.expression(names) for arg in self._args]
-        args = ', '.join(str_args)
-        fname = self.function.expression(names)
-        return '%s(%s)' % (fname, args)
-
-
```
Eric89GXL: This is all just copied to the new `variable.py`, right?

Eric89GXL:
I should be able to look more closely and try this out in the next couple of days if someone else (@almarklein?) doesn't beat me to it. FYI you can restart failed Travis runs on the travis site.

vispy/visuals/shaders/function.py
campagnola: Yes, I added a few new files because this was getting too long.

rossant:
@campagnola just wondering, how does this relate to #800 ?

examples/basics/visuals/visual_filters.py
```diff
@@ -0,0 +1,94 @@
...
+        for line in self.lines:
+            tr_sys = visuals.transforms.TransformSystem(self)
+            tr_sys.visual_to_document = line.transform
+            line.tr_sys = tr_sys
+
+        self.show(True)
+
+    def on_draw(self, ev):
+        gloo.clear('black', depth=True)
+        gloo.set_viewport(0, 0, *self.size)
```
Eric89GXL: should be `self.physical_size`

examples/basics/visuals/visual_filters.py
```diff
@@ -0,0 +1,94 @@
...
+        self.lines[4].transform = STTransform(translate=(0, 250))
+        self.lines[4].attach(Hatching())
+        
+        # mixing filters
+        self.lines[5].transform = STTransform(translate=(400, 250))
+        self.lines[5].attach(ColorFilter([1, 0, 0, 1]))
+        self.lines[5].attach(Hatching())
+        
+        app.Canvas.__init__(self, keys='interactive')
+        self.size = (800, 800)
```
Eric89GXL: better to do this as `__init__(self, ..., size=(800, 800))` to avoid unnecessary size event

examples/basics/scene/viewbox.py
```diff
@@ -24,7 +24,7 @@
 # First ViewBox uses a 2D pan/zoom camera
 vb1 = scene.widgets.ViewBox(name='vb1', border_color='yellow',
                             parent=canvas.scene)
-vb1.clip_method = 'fbo'
+vb1.clip_method = 'fragment'
```
Eric89GXL: Do we need this line here, or is `fragment` now the default?

vispy/scene/widgets/viewbox.py
```diff
@@ -291,14 +293,38 @@ def draw(self, event):
...
+        elif prefer == 'fragment':
+            self._clipper.bounds = event.visual_to_framebuffer.map(self.rect)
+            event.push_node(self.scene)
+            try:
+                self.scene.draw(event)
+            finally:
+                event.pop_node()
         else:
             # Just draw
             # todo: invoke fragment shader clipping
```
Eric89GXL: does this `todo` still make sense?

vispy/visuals/components/clipper.py
```diff
@@ -0,0 +1,48 @@
+from ..shaders import Function
```
Eric89GXL: Don't forget our boilerplate encoding and copyright

vispy/visuals/shaders/compiler.py
```diff
@@ -188,7 +188,7 @@ def _is_global(self, obj):
         # todo: right now we assume all Variables are global, and all
         # Functions are local. Is this actually correct? Are there any
         # global functions? Are there any local variables?
-        from .function import Variable
+        from .variable import Variable
```
Eric89GXL: This might not need to be nested anymore

Eric89GXL:
Other than my minor gripes, LGTM. Examples run well over here (OSX). @almarklein do you have time to take a look?

campagnola:
@rossant, #800 only brings the `ModularProgram` and `Transform` classes into collections. This is about providing a general interface for appending/prepending code to the shaders. I think the collections also address this issue to some extent (for example by providing a hook for clipping), but this PR implements a more general approach.

campagnola:
@Eric89GXL thanks for the review. I should be able to work on it this evening.

examples/basics/scene/viewbox.py
almarklein: What *should* be the default? (also in relation to the stereo.py example). Would 'viewport' not be the most efficient?

examples/basics/visuals/visual_filters.py
```diff
@@ -0,0 +1,94 @@
...
+            def _attach(self, visual):
+                visual._get_hook('frag', 'post').add(self.shader())
+
+        self.lines[4].transform = STTransform(translate=(0, 250))
+        self.lines[4].attach(Hatching())
+        
+        # mixing filters
+        self.lines[5].transform = STTransform(translate=(400, 250))
+        self.lines[5].attach(ColorFilter([1, 0, 0, 1]))
+        self.lines[5].attach(Hatching())
```
almarklein: I like the API

vispy/visuals/visual.py
```diff
@@ -101,3 +105,40 @@ def update(self):
...
+        self._hooks[key] = hook
+        return hook
+        
+    def attach(self, filter):
+        """Attach a Filter to this visual. 
+        
+        Each filter modifies the appearance or behavior of the visual.
+        """
+        filter._attach(self)
+        self._filters.add(filter)
```
almarklein: Circular reference. The Clipper filter stores the visual object, and the filters are also stored on the Visual. Either of these should probably be a weakref.

vispy/visuals/shaders/expression.py
```diff
@@ -0,0 +1,100 @@
...
+    
+    def __hash__(self):
+        return self._text.__hash__()
+
+
+class FunctionCall(Expression):
+    """ Representation of a call to a function
+    
+    Essentially this is container for a Function along with its signature. 
+    """
```
almarklein: Did we not already have this somewhere?

almarklein: Ah, it's right there. Carry on.

almarklein:
Running visual_filter.py gives me this:

screenshot from 2015-03-30 21 24 14

examples/basics/visuals/visual_filters.py
campagnola: Why is that better? Ideally it shouldn't make any difference..

examples/basics/scene/viewbox.py
campagnola: I think fragment should be the default to eventually allow collections across viewboxes. As always, profiling will be needed..

examples/basics/visuals/visual_filters.py
Eric89GXL: This sets the size twice, once during constriction and them immediately
thereafter. In some backends this will cause two resize events instead of
one. So it is less clean. It's also more consistent to have it in the
constructor since the majority of examples do it that way now.

campagnola:
I think all comments are addressed. @almarklein, please check that examples are working again.

almarklein:
nope, still the same, sorry

campagnola:
Both examples are the same? Can you try resizing the filters example window?

almarklein:
Weird, the content stays the same size (with respect to my real screen), resizing the window does not resize the viewboxes. The black bands rotate as I change the height of the window.

screenshot from 2015-03-31 14 46 35

almarklein:
viewbox.py is now more broken than before. At first I only saw the left line. Now I see none. If I resize, sometimes the left figure is shown briefly.

campagnola:
> the content stays the same size

That is the expected output.

> viewbox.py is now more broken than before.

Confirmed. I'm not sure why it was working for me before but not now..

campagnola:
Ok, should be fixed now.

almarklein:
Viewbox.py works again, but `visual_filters.py` still looks the same.

Eric89GXL:
Confirmed, `viewbox.py` works well but `visual_filters` still has bands moving based on window size. It otherwise looks like this (after resizing), which I assume is correct:

screen shot 2015-04-01 at 9 42 04 pm

campagnola:
As I said, that is the expected output. To make the bands stationary would complicate the filter and I want to keep it relatively simple..

almarklein:
Aha! All is well then, I think.

vispy/visuals/visual.py
almarklein: Should we do something with this?

campagnola: I think it is done--none of the components have strong references to their attached visual.

vispy/visuals/components/clipper.py
```diff
@@ -0,0 +1,53 @@
...
+        return self._bounds
+    
+    @bounds.setter
+    def bounds(self, b):
+        self._bounds = Rect(b).normalized()
+        b = self._bounds
+        self.clip_shader['view'] = (b.left, b.right, b.bottom, b.top)
+        
+    def _attach(self, visual):
+        self._visual = weakref.ref(visual)
```
almarklein: ah I suppose you added weakref here ...

almarklein:
in that case, LGTM

Eric89GXL:
Thanks @campagnola this does seem like a nice way of making the visuals more modular

PR #826, opened 2015-03-27
ENH: Text and console using doc coord sys

Eric89GXL:
@nippoo see if this works.

Closes #511.

PR #828, opened 2015-03-27
FIX: Minor example fixes

### LINKED ISSUE #801: examples/demo/gloo/shadertoy.py throws up a load of warnings
```
WARNING: Variable iSampleRate is not an active uniform
WARNING: Variable iChannelTime[0] is not an active uniform
WARNING: Variable iChannelTime[1] is not an active uniform
WARNING: Variable iChannelTime[2] is not an active uniform
WARNING: Variable iChannelTime[3] is not an active uniform
WARNING: Variable iChannelResolution[0] is not an active uniform
WARNING: Variable iDate is not an active uniform
WARNING: Program has unset variables: {'iMouse'}
```

This also isn't tested by Travis for some reason or another. Probably because of the above warnings.

Eric89GXL:
@nippoo can you see if this fixes #801 for you? It also moves us toward fixing #826, but isn't quite there yet. (I need to do more work on GLFW on OSX to really fix that probably.)

### DIFF
--- a/examples/benchmark/scene_test_1.py
+++ b/examples/benchmark/scene_test_1.py
@@ -96,14 +96,11 @@ class PanZoomCanvas(app.Canvas):
 
         self.width, self.height = self.size
         gloo.set_viewport(0, 0, self.physical_size[0], self.physical_size[1])
-
-        self._tr = TransformSystem(self)
-
-        self.show()
-
-    def on_initialize(self, event):
         gloo.set_state(clear_color='black', blend=True,
                        blend_func=('src_alpha', 'one_minus_src_alpha'))
+
+        self._tr = TransformSystem(self)
+        self.show()
 
     def on_resize(self, event):
         self.width, self.height = event.size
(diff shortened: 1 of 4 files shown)


rossant:
I would prefer to keep examples/demo/gloo/shadertoy.py as it is: basically there is a set of uniforms that may or may not be used by a given shadertoy example. Currently it is quite convenient because I can just copy and paste a shader example from shadertoy and it just works. But no example uses all possible variables. Can we make it such that these warnings are silent instead?

Eric89GXL:
Ahh okay, I'll add a comment at the top explaining the warnings.

rossant:
Thanks. But should we make these warnings more discrete? I would vote for downgrading them to DEBUG messages. These messages are useful to the developer of the visual, not to the end-user IMO.

Eric89GXL:
I think it's good to have them on by default, because in most applications if they show up it means that something is being done incorrectly, no? It would be nice if we could disable them for e.g. end-user applications, but at this point I think most users are developers. And if users of the high-level systems (scene, plot) see such warnings, we definitely want to know and fix it.

rossant:
> in most applications if they show up it means that something is being done incorrectly, no?

Not necessarily, I think. What about using INFO? This way, users will see the messages, but they won't be worried.

Eric89GXL:
I could live with that -- @campagnola @rougier @almarklein WDYT about making missing variable messages INFO instead of WARNING level?

almarklein:
Fine with me.

Eric89GXL:
Fixed, @nippoo seems satisfied enough, and Travis is happy. Ready for merge form my end @almarklein (we can change WARNING to INFO in a separate PR if we decide to).

campagnola:
Yes, INFO or even DEBUG would be fine, and we should consider a config option to make it more visible. In general we should optimize vispy's behavior for users and expect developers to have their own config (I think this same point was discussed wrt. CLI flags). Maybe we can provide a recommended config file for developers..

PR #829, opened 2015-03-28
[WIP] Axis Visual

nippoo:
A framework axis / ticks visual, with example.

### DIFF
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -36,27 +36,32 @@ class AxisVisual(Visual):
         self.tick_direction = tick_direction or self._get_tick_direction()
 
     def draw(self, transforms):
-        line = LineVisual(pos=self.extents, color=self.axis_color,
+        v_line = LineVisual(pos=self.extents, color=self.axis_color,
                           mode='gl', width=3.0)
 
         tick_fractions, tick_labels = self._get_tick_frac_labels()
 
-        tick_pos = self._get_tick_positions(tick_fractions)
+        tick_pos, tick_label_pos = self._get_tick_positions(tick_fractions)
 
-        ticks = LineVisual(pos=tick_pos, color=self.tick_color, mode='gl',
+        v_ticks = LineVisual(pos=tick_pos, color=self.tick_color, mode='gl',
                            width=2.0, connect='segments')
 
-        ticks.draw(transforms)
-        line.draw(transforms)
+        # TODO: make this a more efficient iterator
+        for x in range(0, len(tick_labels)):
+            TextVisual(str(tick_labels[x]), font_size=8, color='w',
+                       pos=tick_label_pos[x]).draw(transforms)
+
+        v_ticks.draw(transforms)
+        v_line.draw(transforms)
 
     def _get_tick_direction(self):
         """Determines the tick direction if not specified."""
         v = self.vec
 
-        if abs(v[0]) >= abs(v[1]):
-            v = np.dot(np.array([[0, -1], [1, 0]]), v) # right axis, down ticks
-        else:
-            v = np.dot(np.array([[0, 1], [-1, 0]]), v) # up axis, left ticks
+        if abs(v[0]) >= abs(v[1]): # rightward axis, rotate ticks clockwise
+            v = np.dot(np.array([[0, -1], [1, 0]]), v)
+        else: # upwards axis, rotate ticks counter-clockwise
+            v = np.dot(np.array([[0, 1], [-1, 0]]), v)
 
         # now return a unit vector
         return v / np.linalg.norm(v)
@@ -71,16 +76,18 @@ class AxisVisual(Visual):
 
         tick_endpoints = tick_vector + tick_origins
 
+        tick_label_pos = tick_origins + self.tick_direction*30
+
         c = np.empty([len(tick_fractions) * 2, 2])
         c[0::2] = tick_origins
         c[1::2] = tick_endpoints
 
-        return c
+        return c, tick_label_pos
 
     def _get_tick_frac_labels(self):
-        tick_num = 10 # number of ticks
+        tick_num = 11 # number of ticks
 
         tick_fractions = np.linspace(0, 1, num=tick_num)
-        tick_labels = tick_fractions.astype("str")
+        tick_labels = tick_fractions
 
         return tick_fractions, tick_labels
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -8,16 +8,32 @@ class AxisVisual(LineVisual):
     """
     Single axis.
     """
-    def __init__(self, extents, domain=(0., 1.), range=(0., 1.)):
-        color = np.array([[1, 1, 1, 1],
-                          [1, 1, 1, 1]])
+    def __init__(self, extents, domain=(0., 1.)):
+        color = (1, 1, 1, 1)
+
+        self.extents = extents
 
         LineVisual.__init__(self, pos=extents, color=color, mode='gl',
-                            width=4.0)
+                            width=3.0)
+
+    def draw_ticks(self, transforms):
+        t = TicksVisual(extents=self.extents)
+        t.draw(transforms)
 
 class TicksVisual(LineVisual):
-    def __init__(self, **kwargs):
-        pass
+    def __init__(self, extents, domain=(0., 1.)):
+        color = (0.7, 0.7, 0.7, 1)
+
+        # Generate 10 evenly-spaced ticks
+        tick_spaces = np.linspace(extents[0][0], extents[1][0], num=10)
+
+        x = np.repeat(tick_spaces, 2)
+        y = np.tile([extents[0][1], extents[0][1] + 10], len(tick_spaces))
+
+        tickpos=np.c_[x,y]
+
+        LineVisual.__init__(self, pos=tickpos, color=color, mode='gl',
+                            width=2.0, connect='segments')
 
 class TickTextVisual(TextVisual):
     def __init__(self, **kwargs):
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -28,6 +28,7 @@ class AxisVisual(LineVisual):
         color = (1, 1, 1, 1)
 
         self.extents = extents
+        self.vec = np.subtract(self.extents[1],self.extents[0])
 
         self.tick_direction = tick_direction or self._get_tick_direction()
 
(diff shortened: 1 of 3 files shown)


rossant:
see also https://github.com/glumpy/glumpy/blob/master/examples/spiral.py based on https://github.com/vispy/vispy/blob/master/vispy/visuals/collections/agg_segment_collection.py (now in VisPy)

campagnola:
Does anyone else get this?
```
>>> Fatal Python error: Segmentation fault

Current thread 0x00007ffbd3d00740:
  File "/home/luke/work/vispy/vispy/vispy/gloo/gl/_gl2.py", line 415 in glDrawArrays
  File "/home/luke/work/vispy/vispy/vispy/gloo/glir.py", line 791 in draw
  File "/home/luke/work/vispy/vispy/vispy/gloo/glir.py", line 352 in parse
  File "/home/luke/work/vispy/vispy/vispy/gloo/glir.py", line 157 in flush
  File "/home/luke/work/vispy/vispy/vispy/gloo/context.py", line 144 in flush_commands
  File "/home/luke/work/vispy/vispy/vispy/gloo/program.py", line 463 in draw
  File "/home/luke/work/vispy/vispy/vispy/visuals/shaders/program.py", line 50 in draw
  File "/home/luke/work/vispy/vispy/vispy/visuals/line/line.py", line 347 in draw
  File "/home/luke/work/vispy/vispy/vispy/visuals/line/line.py", line 246 in draw
  File "/home/luke/work/vispy/vispy/vispy/visuals/axis.py", line 54 in draw
  File "examples/basics/visuals/lineplot_axes.py", line 60 in on_draw
  File "/home/luke/work/vispy/vispy/vispy/util/event.py", line 448 in _invoke_callback
  File "/home/luke/work/vispy/vispy/vispy/util/event.py", line 436 in __call__
  File "/home/luke/work/vispy/vispy/vispy/app/backends/_qt.py", line 617 in paintGL
```

Eric89GXL:
I have never seen that before, you getting it only on this branch?

campagnola:
Only on the lineplot_axes example.

nippoo:
Heh, that's weird. I don't get it at all. Anyone else except @campagnola?

It does take up 100% CPU for a surprisingly long time (maybe 10 seconds?) but there's no error for me, much less a segfault.

vispy/visuals/axis.py
```diff
@@ -0,0 +1,93 @@
...
+
+        tick_pos, tick_label_pos = self._get_tick_positions(tick_fractions)
+
+        v_ticks = LineVisual(pos=tick_pos, color=self.tick_color, mode='gl',
+                           width=2.0, connect='segments')
+
+        # TODO: make this a more efficient iterator
+        for x in range(0, len(tick_labels)):
+            TextVisual(str(tick_labels[x]), font_size=8, color='w',
+                       pos=tick_label_pos[x]).draw(transforms)
```
Eric89GXL: It could be the creation of all of these `TextVisual` instances, you should create them once at the top in a list and then call `.draw` on them here instead of creating them again and again

Eric89GXL:
No segfaults here (OSX currently)

rossant:
There should be one GL cal for all labels and one for all segments,
otherwise perf will be terrible. Look at collections and how glumpy does it.

Eric89GXL:
@rossant yeah that's the plan eventually, starting simple for now

vispy/visuals/axis.py
nippoo: You mean creating them in the `__init__`? The only issue with this is that the number of ticks might change as you zoom in/out, so I might have to create some more. I'm happy to factor that out (in fact it's trivial at least for the LineVisuals) but is the Visual creation overhead large enough that it's worth hacking around with creating a fixed number of ticks and hiding unused ones?

In any case, they're only drawn once for the moment (well, for some reason, `draw` is called three times on OS X, but that shouldn't cause a segfault, right?!)

Eric89GXL: The creation overhead might be substantial for the `Text` visual. It's probably best to create as many as you need at the beginning, and append more as you need them. I wouldn't bother removing ones you no longer need, at least for a first pass. For example, if you go from 10->5->10 ticks, you should have only ever created 10 text visuals for that axis. If you go 10->5->15->10 you should end up with 15 sitting around. My guess is that the overhead for keeping any additional unused ones around will be less of an evil than having to create a ton of these things, at least for now. Plus we'll probably heavily refactor after the collections settle in anyway.

rossant: Here is how glumpy does it: https://github.com/glumpy/glumpy/blob/master/examples/lorenz.py



We have all of these collections in vispy/visuals/collections. Note that there are **two** visuals: one for labels, one for the grid and the ticks, and these are only created once. Changing them dynamically will be a bit tricky but you won't have to create new collections/visuals: you can just change the existing ones. cc @rougier 

Personally I don't think it's worth working on a slow implementation (such as one where dozens of visuals are created every frame) because it will be far too slow to be usable

nippoo: Okay. Shall I put this on hold until we've figured out what to do with
TextCollections / TextCollectionVisuals / GlyphCollections or whatever it's
being called? Or is it stable enough for me to go ahead and refactor that
all now to use the Collections?

... or should I carry on and create a fixed number of visuals and append as
needed (as @Eric89GXL suggested)?

rossant: I would suggest to implement the visual *without text* for now, so just the frame, grid, and ticks, using the SegmentCollection in VisPy. You'll find some collection examples in vispy/examples/collections. For the transform @campagnola will tell you how to do it (should be quite simple). Finally once the text collection is stable you can add it.

Eric89GXL: I completely agree that is what we should do eventually. But there is some
utility in continuing down the current path. Right now the collections
cannot be used directly as visuals, so that would need to be done before
any progress could be made here. In the meantime, creating the visuals once
should give at least decent enough performance that we could at least
address some high-level questions. For instance, if it at least works okay,
then I could start integrating it in my plotting PR within a week, instead
of having to wait what I assume would be a month or more before the
collections-as-visuals work well, and then get implemented here...

@nippoo if you want to try to take over or work with @campagnola on making
the collections proper visuals that integrate properly with the scenegraph
that would be great, but in my mind it really depends on how much time you
have to devote to this stuff.

rossant: > instead
> of having to wait what I assume would be a month or more before the
> collections-as-visuals work well

I sincerely hope this will happen this week rather than in a month or more.

campagnola: There is a lot of work that needs to be done to compute tick geometry and text placement (the latter is not yet solved in glumpy) so I think it makes sense to go ahead with this development, understanding that the actual text drawing will have to be replaced with a collection. Then we can work on getting a text collection visual implemented in parallel.

rossant: > tick geometry

wdym?

> I think it makes sense to go ahead with this development, understanding that the actual text drawing will have to be replaced with a collection

are you okay with @nippoo using segment collection for everything except text, and @Eric89GXL's solution for text for now?

> Then we can work on getting a text collection visual implemented in parallel.

+1

campagnola: By tick geometry I mean determining the placement and length of ticks. This is a nontrivial problem and the solution will look the same whether it is draw with a LineVisual or SegmentCollection.

> are you okay with @nippoo using segment collection for everything except text, and @Eric89GXL's solution for text for now?

That would be fine, except the segment collection still hasn't been turned into a visual.

Eric89GXL:
@nippoo I think your axes are starting to feel neglected...

nippoo:
@Eric89GXL you're right! ... what's the progress with the segment-collection-visual? Or at least have we reached a decision as to how we're going to implement collections vs simple visuals?

campagnola:
No progress on segment collection. I would just go ahead with the existing LineVisual with `method='gl', connect='segments'`; the code will look about the same anyway.

Eric89GXL:
+1 for just using lines and text objects for now. It will look similar, and will allow us to make progress incorporating it in `plot` the meantime

Eric89GXL:
@nippoo will you have time for this soon, or would it be better if someone else took over?

nippoo:
I reckon I'll have time in the next couple of weeks to get 2D axes working with arbitrary labels and text and zooming/panning. Is there anything in particular you wanted?

Eric89GXL:
Nope, just want to make sure we keep making progress. Axes are critical for plotting -- once we have something that works, we can figure out how to make them work well with the `vispy.plot` interface, and then make them fast. And then `vispy.plot` will be really useful :)

rossant:
And then we can make a release :)

Eric89GXL:
@nippoo can't you get @rossant to do some of your actual work so you can come back to your favorite project?

Eric89GXL:
I think this might be a blocker for release. @nippoo I take it you're too busy to work on this in the near term?

Eric89GXL:
@nippoo it looks like @campagnola is going to be doing some major refactoring of visuals. Do you think you can get this "working enough" to get merged in the next couple of days? @campagnola do you think it makes sense to get this into master so you can rework it with the rest of the visuals in your PR?

nippoo:
I had a look at it a couple of days ago. It seems like there might not be much point including any text in the axes until the TextCollectionVisual gets implemented (or performance of creating ~50 TextVisuals gets better!). 

What exactly is @campagnola refactoring?

We've got a big release coming up today, and working tomorrow/Thursday on a coding sprint. I don't think it should take too long to get axes and ticks working well enough (aka before the end of the week). I'll see if using a fixed number of TextVisuals gives acceptable performance - will report back before the end of the week. Sorry it's been so slow!

campagnola:
I didn't realize you were looking at creating many TextVisuals; we should definitely cut this down (1 text + 1 line per axis). I think it should be straightforward to modify the existing TextVisual to support multiple independent blocks of text; this does not require collections. @Eric89GXL is this something you have time to work on?

campagnola:
To answer the previous question: I don't think it makes much difference whether the visual is implemented before or after my refactor, as long as the changes are in the visual subclass files. If you find that you need  to modify anything deeper in the visual/scenegraph, then it might make more sense to wait.

Eric89GXL:
I might have time, yeah. What should the API be like? Maybe to start, if you pass a list of strings, you must also pass a list of positions for each string? That should be pretty straightforward.

campagnola:
Yes, that sounds good.

Eric89GXL:
@nippoo see #931

Eric89GXL:
#931 has been merged, you should be able to rebase and have better text speed using lists of strings and lists of positions

Eric89GXL:
ping @nippoo should I take this one over?

rossant:
@Eric89GXL I think it's safe to assume that @nippoo won't have much time to do it (sorry it's my fault..!)
you can take this one over unless @nippoo disagrees

Eric89GXL:
@nippoo I'm on this one. It is very expensive to create the objects on every draw call, so I just create them once and reuse them... more to come.

PR #833, opened 2015-03-28
Fixed a TubeVisual bug and rearranged doc

inclement:
Previously the tube tangents were calculated assuming the tube was closed, which messed up the first and last tangent values unless they happened to match anyway. The problem is shown in the images below. I also rearranged the doc (and added a missing parameter) to emphasise the main arguments.

Example tubes before this PR:
tubes_problem_orig

And after:
tubes_problem_fixed

### DIFF
--- a/vispy/visuals/tube.py
+++ b/vispy/visuals/tube.py
@@ -21,6 +21,12 @@ class TubeVisual(MeshVisual):
         tube will be extruded.
     radius : float
         The radius of the tube. Defaults to 1.0.
+    closed : bool
+        Whether the tube should be closed, joining the last point to the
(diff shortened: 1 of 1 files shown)


Eric89GXL:
@inclement are you on latest master? We did some refactoring of the transformation system, and I just wanted to make sure that the normals were still calculated correctly. If they aren't, could you fix as part of this PR?

inclement:
@Eric89GXL I think I was on master yesterday, but I also just merged the fixes since then and everything still seems fine.

Eric89GXL:
Good, glad I didn't mess it up :) Merging, thanks for the fix.

PR #835, opened 2015-03-29
FIX: Fix for Windows XP

### LINKED ISSUE #715: Vispy Windows XP Compatibility
When trying to run a Vispy Application in windows xp project will not start due to fault "Attribute Error: function 'SetProcessDPIAware' not found".  First instance traced back to "vispy\ext\gdi32plus.py line 165"  this line was commented out and program ran again and produced the same fault but pointed at "vispy\util\dpi_win32.py line 13" this line was also commented out and the program finally loaded.  Exhaustive testing is not yet complete but program is basically functional.

It appears the "SetProcessDPIAware" API is not supported in windows XP and only partially supported in windows vista and later, see....  https://msdn.microsoft.com/en-us/library/dn469266(v=vs.85).aspx and .... https://msdn.microsoft.com/en-us/library/dn302122(v=vs.85).aspx.

Further testing needs to be done to see whether Vispy is obtaining proper DPI settings.  
reference: http://www.experts-exchange.com/Programming/Languages/Scripting/Python/Q_21794611.html

Eric89GXL:
Closes #715.

Ready for review/merge, @almarklein might be quickest for you.

### DIFF
--- a/vispy/ext/gdi32plus.py
+++ b/vispy/ext/gdi32plus.py
@@ -162,7 +162,10 @@ user32.GetDC.argtypes = [UINT32]  # HWND
 
 user32.ReleaseDC.argtypes = [c_void_p, HDC]
 
-user32.SetProcessDPIAware.argtypes = []
+try:
+    user32.SetProcessDPIAware.argtypes = []
+except AttributeError:
+    pass  # not present on XP
 
 
 # gdiplus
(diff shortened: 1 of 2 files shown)


almarklein:
Does this imply we cannot measure DPI on XP? If so, maybe should document this somewhere?

Eric89GXL:
Honestly, I'm not sure what the result will be on XP other than the code will now run properly. I'm also not too worried about it because this must constitute a tiny portion of our user base.

PR #839, opened 2015-03-30
Optional install requirements

QuLogic:
This adds some Extra requirements, so that one can do something like `pip install vispy[ipython-vnc]` (or some other backend) and the additional dependencies would be installed.

Unfortunately, it seems like PyQt4/5 don't install eggs and don't have proper information on PyPI, so I wasn't able to add the correct dependencies there. I also added some minimum requirements when I could find checks for them, but couldn't see anything describing requirements for all backends.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -87,6 +87,17 @@ setup(
     platforms='any',
     provides=['vispy'],
     install_requires=['numpy'],
+    extras_requires={
+        'ipython-static': ['ipython'],
+        'ipython-vnc': ['ipython>=2'],
+        'ipython-webgl': ['ipython>=2', 'tornado'],
+        'pyglet': ['pyglet>=1.2'],
+        # 'pyqt4': [],  # Why is this on PyPI, but without downloads?
+        # 'pyqt5': [],  # Ditto.
+        'pyside': ['PySide'],
+        'sdl2': ['PySDL2'],
+        'wx': ['wxPython'],
+    },
     packages=package_tree('vispy'),
     package_dir={
         'vispy': 'vispy'},
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,7 @@ setup(
         'Programming Language :: Python',
         'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3.2',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -87,6 +87,17 @@ def package_tree(pkgroot):
     platforms='any',
     provides=['vispy'],
     install_requires=['numpy'],
+    extras_requires={
+        'ipython-static': ['ipython'],
+        'ipython-vnc': ['ipython>=2'],
+        'ipython-webgl': ['ipython>=2', 'tornado'],
+        'pyglet': ['pyglet>=1.2'],
+        'pyqt4': [],  # Why is this on PyPI, but without downloads?
+        'pyqt5': [],  # Ditto.
```
almarklein: Would it not be better to comment these out then? We might give the impression that `pyqt4` deps will be installed, but they will not.

almarklein:
Other than that, this LGTM. Travis errors seem due to data fetching issues, I restarted the offending builds.

setup.py
```diff
@@ -87,6 +87,17 @@ def package_tree(pkgroot):
     platforms='any',
     provides=['vispy'],
     install_requires=['numpy'],
+    extras_requires={
+        'ipython-static': ['ipython'],
+        'ipython-vnc': ['ipython>=2'],
+        'ipython-webgl': ['ipython>=2', 'tornado'],
+        'pyglet': ['pyglet>=1.2'],
+        'pyqt4': [],  # Why is this on PyPI, but without downloads?
```
Eric89GXL: Maybe it's so that people can have it in their required lists... not sure

QuLogic: But it provides no benefit since `pip` fails to download anything. Unless it's some kind of indicator that you should install it by hand, though I don't see how that would be clear.

setup.py
QuLogic: Yea, I was hoping to find some way to patch it so it'd print a warning at least, but I haven't had a chance to look yet.

Eric89GXL:
LGTM, ready for merge from your end @QuLogic?

QuLogic:
Yep, ready from my end.

Eric89GXL:
Thanks for the fixes!

Issue #838, opened 2015-03-29
Bunch of tests fail on Mac OS X


nippoo:
Mac OS X 10.10.1

Retina MacBook Pro 15" (mid-2012)

nVidia GeForce GT 650M

```
______________________________________________________________________ test_functionality_pyopengl ______________________________________________________________________

args = (), kwargs = {}

    def skipper_func(*args, **kwargs):
        """Skipper for normal test functions."""
        if skip_val():
            raise nose.SkipTest(get_msg(f, msg))
        else:
>           return f(*args, **kwargs)

../../../miniconda/envs/phy/lib/python3.4/site-packages/numpy/testing/decorators.py:146: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../miniconda/envs/phy/lib/python3.4/site-packages/numpy/testing/decorators.py:146: in skipper_func
    return f(*args, **kwargs)
vispy/gloo/gl/tests/test_functionality.py:53: in test_functionality_pyopengl
    _test_functonality('pyopengl2')
vispy/gloo/gl/tests/test_functionality.py:87: in _test_functonality
    objects = _prepare_vis()
vispy/gloo/gl/tests/test_functionality.py:226: in _prepare_vis
    assert_equal(gl.glGetShaderInfoLog(hvert), '')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

shader = 2

    def glGetShaderInfoLog(shader):
        res = GL.glGetShaderInfoLog(shader)
>       return res.decode('utf-8')
E       AttributeError: 'str' object has no attribute 'decode'

vispy/gloo/gl/_pyopengl2.py:132: AttributeError
_____________________________________________________
```

```
________________________________________________________________________ test_multiple_canvases _________________________________________________________________________

[...]

PR #834, opened 2015-03-29
FIX: Minor fixes to GL tutorial examples

Eric89GXL:
@nippoo can you see if the examples you mention in #806 work with this PR?

Closes #838.
Closes #837.
Closes #806.
Closes #667.
Closes #630.
Closes #576.

PR #840, opened 2015-03-30
small fix in pyopengl2 gl backend

almarklein:
This should fix some of the problems in #838

@nippoo Can you check if this helps?


Eric89GXL:
These are already more or less in #834, can you comment there if I should change them to something else?

Eric89GXL:
Closing since these changes should be covered already

PR #842, opened 2015-03-30
FIX: HiDPI testing

### LINKED ISSUE #551: Visuals should make use of document coordinate system

### LINKED ISSUE #789: All visuals need to use transform or scaling factor to support HiDPI
You can't just plot a fixed number of pixels to screen when writing your own visuals: everything has to be scaled up or down. i.e. lines, which should never be left as 1px, and instead should be scaled with the transform.

Top image is glfw which is hi-dpi aware (in this case, `canvas.physical_size == canvas.size * 2`)
Bottom is qt4 which is not hi-dpi aware and hides this from the program (`canvas.physical_size == canvas.size`)

Both on Mac OS X 10.10 on a 15" Retina MBP.

screenshot 2015-03-26 10 46 49

Eric89GXL:
Closes #789.
Closes #551.

@nippoo can you see if this makes tests pass for you?

### DIFF
--- a/vispy/geometry/calculations.py
+++ b/vispy/geometry/calculations.py
@@ -83,7 +83,7 @@ def _calculate_normals(rr, tris):
 
 
 def resize(image, shape, kind='linear'):
-    """Resize image using bilinear interpolation
+    """Resize an image
 
     Parameters
     ----------
--- a/vispy/geometry/calculations.py
+++ b/vispy/geometry/calculations.py
@@ -7,6 +7,8 @@
 
 import numpy as np
 
+from ..ext.six import string_types
+
 
 ###############################################################################
 # These fast normal calculation routines are adapted from mne-python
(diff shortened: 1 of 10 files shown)


Eric89GXL:
@bollu where you at? This should make your life less painful!

Eric89GXL:
sorry @bollu, I meant to ping @nippoo with that one... @nippoo this should make your life less painful.

Although @bollu if you happen to be using a HiDPI screen it might also make your life less painful, too :)

nippoo:
@Eric89GXL all tests pass :)

Eric89GXL:
Excellent, @almarklein @rossant feel free to merge then

vispy/geometry/calculations.py
```diff
@@ -78,3 +80,57 @@ def _calculate_normals(rr, tris):
...
+    shape : tuple
+        2-element shape.
+    kind : str
+        Interpolation, either "linear" or "nearest".
+
+    Returns
+    -------
+    scaled_image : ndarray
+        New image, will have dtype np.float64.
+    """
```
almarklein: Is there a reason to add this? I see it can be useful, but it does not seem like we're using it?

almarklein:
Other than that, LGTM

vispy/geometry/calculations.py
Eric89GXL: We are, it's used to rescale images to the testing image size

Eric89GXL: Right now we only use the `nearest` method, but the `linear` could come in handy in the future (I've wanted it before)

PR #844, opened 2015-03-30
Fix WebGL notebook examples

rossant:
The notebook WebGL examples are broken on master, because of the transform changes. This PR fixes them.

@cgestes can you try it?


cgestes:
both are working.

rossant:
thanks, merging

PR #845, opened 2015-03-30
Glir logger

izaid:
This PR implements a Glir logger that logs all Glir commands to a specified file. It is run via "python vispy --vispy-glir-file=somefile.txt".


vispy/__init__.py
```diff
@@ -27,6 +27,13 @@
 # Nice string for the version (mimic how IPython composes its version str)
 __version__ = '-'.join(map(str, version_info)).replace('-', '.', 2).strip('-')
 
+from argparse import ArgumentParser
+
+parser = ArgumentParser()
+parser.add_argument('--vispy-glir-file', help='Export glir commands to specified file.')
+args = parser.parse_args()
+glir_file = args.vispy_glir_file
+
```
rossant: Should we use the system in https://github.com/vispy/vispy/blob/5b01479e829e8f4b2dd348dff51ce864c1f834a5/vispy/util/config.py#L74 instead?

izaid: Yes, absolutely! Didn't know about that.

vispy/util/logs.py
```diff
@@ -327,3 +330,14 @@ def _handle_exception(ignore_callback_errors, print_callback_errors, obj,
             else:  # == 'node':
                 logger.error("Drawing node %s repeat %s"
                              % (node, this_print))
+
+class NumPyJSONEncoder(json.JSONEncoder):
+    def default(self, obj):
+        from ..app.backends._ipynb_util import _serialize_buffer
```
rossant: move this function to vispy util?

vispy/util/logs.py
```diff
@@ -327,3 +330,14 @@ def _handle_exception(ignore_callback_errors, print_callback_errors, obj,
...
+
+        if isinstance(obj, np.ndarray):
+            return _serialize_buffer(obj, array_serialization='base64')
+        elif isinstance(obj, np.generic):
+            return obj.item()
+
+        return json.JSONEncoder.default(self, obj)
```
rossant: new line at end of file

rossant:
The approach LGTM. Can you add some tests?

vispy/util/logs.py
izaid: Where in vispy.util do you want it? It is currently in vispy.util.logs.

rossant: sorry I was talking about _serialize_buffer

izaid: Yes, I'll move that. What about _serialize_command and _serialize_item? Should I be using those also?

rossant: if _serialize_buffer depends on them yes

izaid: _serialize_buffer doesn't depend on them. Was your intention with those to use them for glir commands?

rossant: OK you can just move _serialize_buffer and leave the others where they are

vispy/gloo/glir.py
```diff
@@ -294,7 +294,75 @@ def convert_shaders(self):
...
+                ob.set_framebuffer(*args)
+            elif cmd == 'SHADERS':  # Program
+                ob.set_shaders(*args)
+            elif cmd == 'WRAPPING':  # Texture1D, Texture2D, Texture3D
+                ob.set_wrapping(*args)
+            elif cmd == 'INTERPOLATION':  # Texture1D, Texture2D, Texture3D
+                ob.set_interpolation(*args)
+            else:
+                logger.warning('Invalid GLIR command %r' % cmd)
+
```
Eric89GXL: This is all copy-pasted without modification, right?

Eric89GXL:
Other than my clarification, LGTM!

vispy/gloo/glir.py
izaid: Yep.

izaid:
@rossant Except for tests, this is ready to go. How exactly should I do tests for this?

vispy/gloo/context.py
```diff
@@ -160,7 +161,13 @@ class GLShared(object):
...
     
     def __init__(self):
-        self._parser = GlirParser()
+        glir_file = config['glir_file']
+
+        parser_cls = GlirParser
+        if glir_file:
+            parser_cls = glir_logger(parser_cls, glir_file)
+
+        self._parser = parser_cls()
```
almarklein: Note the `parser.setter` a few lines below. I think these additions should go in there.

almarklein:
@rossant this seems very much related to the glir export mechanism that you want. Or is that the idea and am I simply put off by the name logger?

izaid:
@almarklein That's the idea, Cyrille and I discussed this. I called it logger.

almarklein:
Ok :+1:

izaid:
@almarklein With this technique, it can't go into parser.setter because we apply a decorator to the class. The parser.setter method takes an already constructed class.

almarklein:
Does this mean that the logger is disabled if a parser is set at the context?

izaid:
Yes. If you change the parser, the logger is disabled. The right way to handle this would be to change which parser is being created. (I think this makes sense.)

almarklein:
> How exactly should I do tests for this?

You can setup a simple visualization (e.g. just a few glClear commands), log that to a file (`import tempfile`), and check that the output is correct.

izaid:
@almarklein For testing, it's not clear to me how I can send Glir commands to the parser without setting up a canvas and everything. If I just send Glir commands to the parser without setting anything else up, I get a segfault.

What is the simplest infrastructure I need to create to get this to work? Is there an example somewhere?

Eric89GXL:
Look at the existing gl tests, gloo tests, app tests, or visual tests, we
do this in a number of places. Most will involve @requires_application and
"with Canvas", others monkey patch gloo (I think? ) to avoid actually
sending calls. Probably better to use the requires_application way.

vispy/gloo/tests/test_glir.py
```diff
@@ -1,8 +1,12 @@
...
+import tempfile
 
-from vispy.testing import run_tests_if_main
+from vispy import config
+from vispy.app import Canvas
+from vispy.gloo import glir
+from vispy.testing import requires_application, run_tests_if_main
 
 
 def test__queue():
```
rossant: @Eric89GXL @almarklein  is the double `_` intended?

vispy/gloo/tests/test_glir.py
```diff
@@ -45,6 +49,34 @@ def test__queue():
...
+    assert lines[i] == json.dumps(['FUNC', 'glClearColor', 1.0, 1.0, 1.0, 1.0])
+    i += 1
+    assert lines[i] == json.dumps(['FUNC', 'glClear', 17664])
+    i += 1
+    assert lines[i] == json.dumps(['FUNC', 'glFinish'])
+    i += 1
+
+    config.update(glir_file='')
+
+
```
rossant: delete the file at the end?

izaid: It's auto-deleted by tempfile.

vispy/gloo/tests/test_glir.py
almarklein: probably not

izaid:
This is *nearly* ready to go. I've updated this to include a test. It also is passing on every build configuration except one of Travis.

I can't figure out exactly what the problem is for that Travis configuration. It looks like Python 3.4 is unhappy for some stupid, silly reason, but I can't determine why. (Python 3.4 works fine on my Mac.)

Can someone else give this a shot? Maybe someone that uses Linux?

Eric89GXL:
The errors look related to encoding/decoding strings, and/or how files are opened. When I activate Python 3.4 environment on OSX I get simliar errors, e.g.:
```
  File "/Users/larsoner/custombuilds/vispy/vispy/gloo/glir.py", line 421, in _parse
    json.dump(command, self._file, cls=NumPyJSONEncoder)
  File "/Users/larsoner/anaconda/envs/py34/lib/python3.4/json/__init__.py", line 179, in dump
    fp.write(chunk)
TypeError: 'str' does not support the buffer interface
```
It's likely that you either need to use encode/decode, or open your file with the proper mode as binary or text.

Eric89GXL:
...in this case at least one offending line can be changed to this:
```
glir_file = tempfile.TemporaryFile(mode='w')
```
Are you sure you're running Python 3.4? These errors shouldn't really depend on the platform...

izaid:
@Eric89GXL Thanks! The mode='w' will fail the test, as it needs read access. I'm just trying with mode='r+' to see if that works (the default for tempfile is mode='r+b').

Yes, I was trying with Python 3.4. I have no idea why it works here.

Eric89GXL:
`r+` should work, it was the binary `b` that should have broken it

Eric89GXL:
`test_glir.py` now passes over here on 3.4 and 2.7, which bodes well for Travis :)

Eric89GXL:
Good to go from your end once Travis is happy? Looks like you've addressed previous comments

rossant:
I have a comment actually: would it be possible for the output file to be valid JSON? So to have a list of lists?

izaid:
Yep, good to go when green.

izaid:
@rossant Is it cool if I change that in my next PR? I'll try to implement the load and execute in WebGL, which requires exactly that.

rossant:
Also: can you make UNIFORM lists instead of base64 (or switch to lists if there are less than 20 values or something)? It might easier to read and to write the JS side.

rossant:
@izaid sure!

Eric89GXL:
sounds good, I'll merge when green then

izaid:
@rossant @Eric89GXL :+1:

Eric89GXL:
Thanks @izaid!

PR #847, opened 2015-03-30
FIX: Use info for unset vars

Eric89GXL:
@rossant happier now? :) This changes the "unset variables" and "not an active uniform/attribute" to be INFO instead of WARNING level. Note that I did *not* change the "invalid command" warnings to info, since those seem like they really shouldn't occur and would be indicative of deeper problems.


rossant:
Thanks! LGTM :)

Eric89GXL:
Excellent, Travis is happy so feel free to merge @rossant

PR #850, opened 2015-04-01
ENH: Arcball

### LINKED ISSUE #157: Implement mouse arcball for 3D interaction
Would make 3D examples much more convenient.

Renaud Blanch has a nice implementation using quaternions:

https://bitbucket.org/rndblnch/opengl-programmable/src/tip/linalg/
https://bitbucket.org/rndblnch/opengl-programmable/src/tip/09-gles2.py#cl-333

Eric89GXL:
Closes #157.

### DIFF
--- a/examples/basics/scene/viewbox.py
+++ b/examples/basics/scene/viewbox.py
@@ -32,7 +32,7 @@ vb2 = scene.widgets.ViewBox(name='vb2', border_color='blue',
                             parent=canvas.scene)
 vb2.parent = canvas.scene
 vb2.clip_method = 'viewport'
-vb2.camera = scene.TurntableCamera(elevation=30, azimuth=30, up='+y')
+vb2.camera = scene.ArcballCamera()
 
 
 # Move these when the canvas changes size
--- a/examples/basics/scene/viewbox.py
+++ b/examples/basics/scene/viewbox.py
@@ -32,7 +32,7 @@ vb2 = scene.widgets.ViewBox(name='vb2', border_color='blue',
                             parent=canvas.scene)
 vb2.parent = canvas.scene
 vb2.clip_method = 'viewport'
-vb2.camera = scene.ArcballCamera()
+vb2.camera = scene.TurntableCamera(elevation=30, azimuth=30, up='+y')
 
 
 # Move these when the canvas changes size
--- a/vispy/scene/cameras/cameras.py
+++ b/vispy/scene/cameras/cameras.py
@@ -934,21 +934,13 @@ class Base3DRotationCamera(PerspectiveCamera):
 
             elif 1 in event.buttons and keys.SHIFT in modifiers:
                 # Translate
-                w = self._viewbox.size[0]
+                norm = np.mean(self._viewbox.size)
                 if self._event_value is None:
                     self._event_value = self.center
-                dist = (p1 - p2) / w * self._scale_factor
+                dist = (p1 - p2) / norm * self._scale_factor
                 dist[1] *= -1
                 # Black magic part 1: turn 2D into 3D translations
-                sro, saz, sel = list(map(sind, (self._roll, self._azimuth,
-                                                self._elevation)))
-                cro, caz, cel = list(map(cosd, (self._roll, self._azimuth,
-                                                self._elevation)))
-                dx = (+ dist[0] * (cro * caz + sro * sel * saz)
-                      + dist[1] * (sro * caz - cro * sel * saz))
-                dy = (+ dist[0] * (cro * saz - sro * sel * caz)
-                      + dist[1] * (sro * saz + cro * sel * caz))
-                dz = (- dist[0] * sro * cel + dist[1] * cro * cel)
+                dx, dy, dz = self._dist_to_trans(dist)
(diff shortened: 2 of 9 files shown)


examples/basics/scene/viewbox.py
```diff
@@ -32,7 +32,7 @@
                             parent=canvas.scene)
 vb2.parent = canvas.scene
 vb2.clip_method = 'viewport'
-vb2.camera = scene.TurntableCamera(elevation=30, azimuth=30, up='+y')
+vb2.camera = scene.ArcballCamera()
```
Eric89GXL: Eventually we can modify a different example, this was just for testing

vispy/scene/cameras/cameras.py
```diff
@@ -1128,6 +1019,208 @@ def _update_projection_transform(self, fx, fy):
...
+        wh = self._viewbox.size
+        self._quaternion = (self._quaternion *
+                            Quaternion(*_arcball(self._event_value, wh)) *
+                            Quaternion(*_arcball(p2, wh)))
+        self._event_value = p2
+        self.view_changed()
+
+    def _rotate_tr(self, tr):
+        """Rotate the transformation matrix based on camera parameters"""
+        tr.matrix = np.dot(self._quaternion.get_matrix().T, tr.matrix)
```
almarklein: I've applied some fixes to the Quaternion class in my DefaultTransform branch. I think one of them involved `get_matrix()`. I should either backport these fixes into master, or I'll need to fix this camera in that PR.

Eric89GXL:
This PR now works, so maybe fix it in your PR?

Eric89GXL:
In other words, ready for review/merge from my end :)

almarklein:
In the volume example, if I change fov, the volume disappears

Eric89GXL:
Only for the arcball, or also for the turntable?

almarklein:
only the arcbal

Eric89GXL:
Sheesh, that took me a lot longer than I'd anticipated to work out. @almarklein the issue was actually shown with changes to `distance` (which was modified by changing `fov`). The problem went away when I used the `get_axis_angle` method of `Quaternion` instead of `get_matrix`, along with a few other adjustments. I'm not 100% sure why the `get_matrix` method wasn't working, but `get_axis_angle` should actually be more efficient anyway.

Then I realized I also needed to add the "move by shift-click" capability as well, which is now done. I better understand why the "black magic" comments are in there for the TurntableCamera now.

I think that [whoever suggested] (https://github.com/vispy/vispy/issues/157#issuecomment-87619039) that implementing the Arcball camera would be "easy" given the existing Quaternion code / capabilities owes me a beer or two...

The CIs are happy, ready for review/merge again @almarklein.

almarklein:
> I'm not 100% sure why the get_matrix method wasn't working

I think I know why. Its because the current `get_matrix` is incorrect. At least I think it is because I needed to update it for my transforms :P

almarklein:
There's one "easy" thing left. The Arcball camera does not work well with all up values. For -x and -y the rotation is wrong. See `flip_axis.py`. One solution (this one really is simple :)) is to ignore up-vectors for this camera, which makes sense, I think, since there is not really a notion of "up" (which is why this camera sucks so much :P).

Thanks for this, I'll buy you a beer next time we meet up.

Eric89GXL:
Done. Now the Arcball camera does nothing when the "up" value is changed. If someone is really interested in modifying it to do something they think is sensible regarding the "up" value, they can make it work properly :)

Feel free to merge if you're satisfied with this round @almarklein.

vispy/scene/cameras/cameras.py
```diff
@@ -1128,6 +1009,242 @@ def _update_projection_transform(self, fx, fy):
...
+              + dist[1] * (sro * caz - cro * sel * saz))
+        dy = (+ dist[0] * (cro * saz - sro * sel * caz)
+              + dist[1] * (sro * saz + cro * sel * caz))
+        dz = (- dist[0] * sro * cel + dist[1] * cro * cel)
+        return dx, dy, dz
+
+
+class ArcballCamera(Base3DRotationCamera):
+    """ 3D camera class that orbits around a center point while
+    maintaining a view on a center point.
```
rossant: noob question: what is the difference between turntable camera and arcball camera?

almarklein: With the turntable there is a notion of up, and by interaction the azimuth and elevantion are modified. With an arcball camera you go all over the place :)

flutefreak7:
Would it be accurate to say this is the kind of camera common in many CAD applications? (Pro-e, Solid Edge, NX, AutoCAD, SpaceClaim, etc.) The turntable camera common in simpler 3D plotting libraries has always frustrated me I think because I keep wishing it was more like a CAD application (or like a 6-DOF space flight sim for a gaming metaphor). I think I'm really looking forward to the arcball camera!

Eric89GXL:
@almarklein I'm going to assume you just forgot to merge this one :)

Eric89GXL:
@flutefreak7 I don't actually know, I haven't really used CAD applications before.

almarklein:
I, sorry. I somehow thought it was merged already :P

PR #852, opened 2015-04-02
Allow setting GL state variables for all visuals

### LINKED ISSUE #689: Allow custom blend functions in visuals
Can you allow custom blend functions in visuals?
For example, in the MarkersVisual class defined in vispy.visuals.markers.py, in the last function named "draw", the following code restricts ('src_alpha', 'one_minus_src_alpha') to be used as the blend function:

         set_state(depth_test=False, blend=True,

However, in my case, I would like the blend function to be ('src_alpha', 'one_minus_src_alpha', 'one', 'one'). I guess this is not possible in the current version, unless I write my own MarkersVisual class. Is that possible to allow other blend functions by providing a "set_state" function for this MarkersVisual class? Also for other visual classes?

campagnola:
This adds the methods `Visual.set_gl_state()` and `Visual.update_gl_state()` to let the user change the way GL state is configured by the visual before drawing. This is used to set depth testing, face culling, blend modes, etc. 

Before this PR, visuals completely hide the OpenGL API, so one question is whether we should implement a slightly higher-level API that does not refer to GL at all (but that could be left for a later PR).

Edit by Eric89GXL:
Closes #689.


Eric89GXL:
LGTM, +1 for leaving the more user-friendly API to a subsequent PR. Maybe we won't even need it. Travis error is unrelated (freetype problem again).

@campagnola do you think it's worth modifying one of the examples in some simple way to show what it could do (even if it's something weird)?

@almarklein did you want to have a look? If not, I'll run some examples, test it a bit, and merge if it all works as expected.

vispy/visuals/visual.py
```diff
@@ -37,6 +38,24 @@ def __init__(self):
...
+    
+    def update_gl_state(self, *args, **kwargs):
+        """Modify the set of GL state parameters to use when drawing
+        this visual.
+        """
+        if len(args) == 1:
+            self._gl_state['preset'] = args[0]
+        elif len(args) != 0:
+            raise TypeError("Only one positional argument allowed.")
+        self._gl_state.update(kwargs)
```
almarklein: `def update_gl_state(self, preset=None, **kwargs)` ?

almarklein:
If no state is set, `self._gl_state == {'preset': None}`, is the state reset? E.g. turn off depth testing etc?

What if I have two visuals A and B, where A specifies a state, and B does not. First A is drawn, and then B. We want B to appear the same as when A was not drawn before.

Also, what's the overhead? Is it worth remembering the state (in the drawing system?) so that we only need to update the parts of the state that were changed?

vispy/visuals/visual.py
campagnola: No; that would cause `preset` to be changed even if the user does not request it.

campagnola:
That's right--the items in `_gl_state` are the only ones applied and all others are left unchanged. Each visual class should implement a more complete set of state variables, though. I have no idea what the cost is..

campagnola:
I think this is good to merge. No need to add it to an example for now; hopefully we will pick suitable defaults and the user will very rarely need to use this API (and if they do, it's unlikely they'll find it buried in an example anyway).

Eric89GXL:
Ahh, true. I'll go ahead and take one last look and merge.

Eric89GXL:
I assume you ran examples to make sure things didn't break?

campagnola:
I did; everything looked ok to me (aside from some issues that are in master as well).

PR #853, opened 2015-04-02
Consistently use "method" instead of "style" and "mode"

campagnola:
Changed mode=>method for line visual and style=>method for volume visual

### DIFF
--- a/examples/basics/scene/colored_line.py
+++ b/examples/basics/scene/colored_line.py
@@ -29,7 +29,7 @@ canvas = scene.SceneCanvas(keys='interactive', size=(400, 200), show=True)
 
 # Create a visual that updates the line with different colormaps
 color = next(colormaps)
-line = scene.Line(pos=pos, color=color, mode='gl')
+line = scene.Line(pos=pos, color=color, method='gl')
 line.transform = STTransform(translate=[0, 140])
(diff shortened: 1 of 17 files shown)


almarklein:
You missed some.

campagnola:
@Eric89GXL, any idea if this error is ignorable? I can't reproduce it locally.

Eric89GXL:
I started seeing this early this week, with no real changes to how we do FreeType. I'll probably need to add some sort of exception. I'll try to reproduce it when I'm back on Linux on Monday, and if I can't succeed, I'll find a way to skip it.

I'll review this PR shortly, too.

Eric89GXL:
BTW sometimes restarting the test "fixes" the problem, which I've done here.

Eric89GXL:
Oh good, I've actually changed "style" -> "symbol" for the `MarkerVisual` in my plotting PR, so I think we'll be complementary instead of overlapping. Hooray, fewer merge conflicts!

Looking at the `git grep` it looks like you got the rest for `LineVisual` and `VolumeVisual`, so I'm happy to go ahead and merge this and clean up anything else if necessary in the plotting PR. It looks like you need a rebase, though, since this can't automatically be merged.

PR #860, opened 2015-04-04
Scenegraph profiling

campagnola:
Adds a Profiler class and instruments some visual/scenegraph code. Should add negligible overhead when profiling is not in use. Works like this:

```
# profile a specific code location (comma-separated list)
python example.py --vispy-profile=SceneCanvas.draw_visual

# enable all profilers
python example.py --vispy-profile=all
```

### DIFF
--- a/examples/benchmark/scene_test_1.py
+++ b/examples/benchmark/scene_test_1.py
@@ -7,6 +7,10 @@
 """
 Compare an optimal pan/zoom implementation to the same functionality
 provided by scenegraph.
+
+Use --vispy-cprofile to see an overview of time spent in all functions.
+Use util.profiler and --vispy-profile=ClassName.method_name for more directed
+profiling measurements.
 """
 import numpy as np
 import math
@@ -344,6 +348,7 @@ if __name__ == '__main__':
     svisual = Signals(data)
     view = scanvas.central_widget.add_view()
     view.add(svisual)
+    view.camera = 'panzoom'
 
     import sys
     if sys.flags.interactive != 1:
--- a/examples/benchmark/scene_test_1.py
+++ b/examples/benchmark/scene_test_1.py
@@ -15,6 +15,7 @@ from vispy import gloo, app, scene
 from vispy.visuals import Visual
 from vispy.visuals.shaders import ModularProgram, Function, Variable
 from vispy.visuals.transforms import TransformSystem, BaseTransform
+from vispy.util.profiler import Profiler
 
 
 class PanZoomTransform(BaseTransform):
@@ -143,6 +144,7 @@ class PanZoomCanvas(app.Canvas):
             self.update()
 
     def on_mouse_wheel(self, event):
+        prof = Profiler()
         if not event.modifiers:
             dx = np.sign(event.delta[1])*.05
             x0, y0 = self._normalize(event.pos)
@@ -176,9 +178,11 @@ class PanZoomCanvas(app.Canvas):
         return self._visuals
 
     def on_draw(self, event):
+        prof = Profiler()
         gloo.clear()
         for visual in self.visuals:
             visual.draw(self._tr)
+            prof('draw visual')
 
 
 X_TRANSFORM = """
--- a/examples/benchmark/scene_test_1.py
+++ b/examples/benchmark/scene_test_1.py
@@ -144,7 +144,7 @@ class PanZoomCanvas(app.Canvas):
             self.update()
 
     def on_mouse_wheel(self, event):
-        prof = Profiler()
+        prof = Profiler()  # noqa
         if not event.modifiers:
             dx = np.sign(event.delta[1])*.05
             x0, y0 = self._normalize(event.pos)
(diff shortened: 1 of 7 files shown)


examples/benchmark/scene_test_1.py
```diff
@@ -143,6 +144,7 @@ def on_mouse_move(self, event):
             self.update()
 
     def on_mouse_wheel(self, event):
+        prof = Profiler()  # noqa
```
Eric89GXL: how about we make this a decorator? That way this could just be:
```
@vispy_profile
on_mouse_wheel(self, event):
```
WDYT? That way anything we want to be profiled we can just add a decorator to.

Eric89GXL: ... although then you can't later have `prof(...)` calls like you do. So never mind I guess.

campagnola: I call `prof('message')` from inside the function to subdivide the measurement. Better to have just one way of doing it?

Eric89GXL:
@campagnola the CIs are happy. Should I review and test?

vispy/scene/systems.py
```diff
@@ -35,6 +38,7 @@ def process(self, event, node):
                 self.process(event, sub_node)
             finally:
                 event.pop_node()
+            prof('process child %s', sub_node)
```
Eric89GXL: should this be `%s' % sub_node`?

campagnola: No--just like with logging, we avoid string formatting until the message is displayed.

campagnola:
Yeah, this is ready for review/merge.

Eric89GXL:
A few ideas:

1. `--vispy-profile` could do the same thing as `--vispy-profile=all`, i.e. default would be to profile everything. I assumed/hoped it would when I first tried it.

2. Right now with the "all" mode, it's tough to see what's causing slowdowns because there is so much output. Could we have a mode where, instead of printing things out when they happened, it only gave an output e.g. every >= 10 seconds, listing the events that occurred in each interval? In this model, you could even ask it only to print the top `N` offenders during the last span of time. This could make it much easier to read. It also deals with the issue that having a call that takes 10 ms but only occurs rarely might be better or worse in some cases than a call that takes 1 ms but occurs very frequently -- but it's difficult to get that information from the current implementation. The current implementation seems to be good for targeted profiling, but less so for getting a sense of what should be targeted.

3. Could we make it so that the timing of each line is right-justified to e.g. 80 chars, even if optionally? This would also make it much easier to find the worst offenders.

These things could be done in subsequent PRs if necessary, too.

campagnola:
1. I don't think `getopt` supports this; we'd have to switch to argparse?
2. That's out of the scope of this PR, and maybe out of the scope of the Profiler class altogether. For general hunting we use `--vispy-cprofile`.
3. Not really worth the reorganization that would be required in the message-handling pipeline..   I generally only enable a single profiler at a time for very targeted profiling, and it's quite simple to trace the CPU effort to the biggest offenders. If we really want a more complex interface, it would make more sense to display the results in a QTreeWidget.

Sorry ;)

Eric89GXL:
Your explanations make sense for why those particular solutions won't work given the current implementation. My underlying concern is that this doesn't seem as easy to use as it could potentially be. I can envision outputs that would be simpler, and easier to relevant information out of. Am I just using this wrong? Or maybe the cprofile is what I should actually be using to get a general sense for where time is spent? It seems like this implementation requires knowing where to look for time consumption (the function or subset of functions), whereas I was hoping for a solution that would tell me where to look, at least as a first pass.

Eric89GXL:
I played with `--vispy-cprofile` a bit and I think it does something like what I want. I guess the workflow should be to use that to get an idea of what's going on, add `Profile` classes in appropriate areas, then use more targeted commands?

campagnola:
That's right--Profiler is meant for very targeted measurement. Or at least, that's the way I have used it in the past. You're welcome to try making it more useful..

Eric89GXL:
Can you rebase? Then later today I can try the examples and merge if all is well.

examples/benchmark/scene_test_2.py
```diff
@@ -0,0 +1,189 @@
+# -*- coding: utf-8 -*-
+# vispy: testskip
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014, Vispy Development Team.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
+Compare an optimal plot grid implementation to the same functionality
+provided by scenegraph.
```
Eric89GXL: Can you add comments to these files with helpers suggesting how to profile?

Also, I get some errors on this example, does it work for you?

```
  File "examples/benchmark/scene_test_2.py", line 49, in on_draw
    M = len(self.cells)
AttributeError: 'GridCanvas' object has no attribute 'cells'
```

campagnola: It works for me. I guess this is a backend issue--you are getting a call to `on_draw` before `__init__` has completed.

Eric89GXL:
Looks like this should make it easier for us to speed things up.

PR #861, opened 2015-04-04
Camera cleanup

campagnola:
We rushed the volume PR through and let a few things slip. This addresses some of them..

### DIFF
--- a/vispy/scene/cameras/cameras.py
+++ b/vispy/scene/cameras/cameras.py
@@ -647,7 +647,8 @@ class PanZoomCamera(BaseCamera):
                 # Zoom
                 p1c = np.array(event.last_event.pos)[:2]
                 p2c = np.array(event.pos)[:2]
-                scale = (1 + self.zoom_factor) ** ((p1c-p2c) * np.array([1, -1]))
+                scale = ((1 + self.zoom_factor) ** 
+                         ((p1c-p2c) * np.array([1, -1])))
                 center = self._transform.imap(event.press_event.pos[:2])
                 self.zoom(scale, center)
 
--- a/vispy/scene/cameras/cameras.py
+++ b/vispy/scene/cameras/cameras.py
@@ -527,7 +527,7 @@ class PanZoomCamera(BaseCamera):
                 raise TypeError("factor must be scalar or length-2 sequence.")
             scale = list(factor)
         if self.aspect is not None:
-            factor[0] = factor[1]
+            scale[0] = scale[1]
         
         # Init some variables
         center = center if (center is not None) else self.center
--- a/vispy/scene/cameras/cameras.py
+++ b/vispy/scene/cameras/cameras.py
@@ -18,15 +18,6 @@ from ...visuals.transforms import (STTransform, PerspectiveTransform,
                                    TransformCache)
 
 
-def sind(q):
-    return math.sin(q*math.pi/180)
-
-
-def cosd(q):
-    return math.cos(q*math.pi/180)
-
-# todo: allow panzoom camera to operate in other planes than Z (i.e. use up)
-
 # todo: Make 3D cameras use same internal state: less code, smooth transitions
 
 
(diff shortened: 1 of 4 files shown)


campagnola:
Ping @almarklein

vispy/scene/cameras/cameras.py
```diff
@@ -1148,10 +1138,9 @@ def _rotate_tr(self):
 
     def _dist_to_trans(self, dist):
         """Convert mouse x, y movement into x, y, z translations"""
-        sro, saz, sel = list(map(sind, (self.roll, self.azimuth,
-                                        self.elevation)))
-        cro, caz, cel = list(map(cosd, (self.roll, self.azimuth,
-                                        self.elevation)))
+        rae = np.array([self.roll, self.azimuth, self.elevation]) * np.pi / 180
+        sro, saz, sel = np.sin(rae)
+        cro, caz, cel = np.cos(rae)
```
almarklein: nice

vispy/visuals/text/text.py
```diff
@@ -463,7 +464,8 @@ def draw(self, transforms):
               transforms.framebuffer_to_render)
         px_scale = (tr.map((1, 0)) - tr.map((0, 1)))[:2]
         self._program.vert['transform'] = transforms.get_full_transform()
-        self._program.vert['text_scale'] = STTransform(scale=px_scale * n_pix)
+        self._text_scale.scale = px_scale * n_pix
+        self._program.vert['text_scale'] = self._text_scale
```
almarklein: Should we expose this as a property? Don't we already have a font-size?

vispy/scene/cameras/cameras.py
```diff
@@ -699,11 +693,7 @@ def _update_transform(self):
         # side (as if z was flipped).
 
         if self.up == '+z':
-            # Keep it simple, but we need to pass AffineTransform,
-            # because if we toggle to -z and then back to +z, things
-            # work incorrectly (test with flipped_axis.py)
-            # todo: fix this so we can just set to a STTransform
-            thetransform = self.transform * AffineTransform()
+            thetransform = self.transform
```
almarklein: I'd prefer to leave as-is until we actually fix the problem, which will probably be automatically fixed as we move forward with the transform system.

vispy/scene/cameras/cameras.py
```diff
@@ -85,7 +76,8 @@ class BaseCamera(Node):
     # These define the state of the camera
     _state_props = ()
 
-    ZOOM_FACTOR = 0.005
+    # The fractional zoom to apply for a single pixel of mouse motion 
+    zoom_factor = 0.007
```
almarklein: As it is now, the zooming feels too sensitive IMO.

vispy/visuals/text/text.py
campagnola: This change is only to avoid creating a new transform on every frame (and thus triggering recompiles). Unrelated to the PR, but I came across it while profiling.

vispy/scene/cameras/cameras.py
campagnola: I couldn't reproduce the bug, and this line triggers recompiling on every frame.

vispy/scene/cameras/cameras.py
campagnola: I forgot to mention: the example at `basics/scene/sensitivity.py` was added for everyone to try and report their findings. Use the +/- keys to adjust the mouse sensitivity while exploring the data. Report the value you like best as well as your OS, backend, and the input device you were using.
@rossant @Eric89GXL @rougier

vispy/scene/cameras/cameras.py
almarklein: oh ... in that case I guess we do it like this; the bug should be rare. Does this mean that if the if-factor is not '+z',  there is a re-compile each frame?

campagnola: Yes, that appears to be the case.

vispy/visuals/text/text.py
Eric89GXL: we do expose the font size in points as a property, this his to do with the scaling done in the GLSL code (related to the font size in points, but not the same thing)

campagnola:
Any outstanding issues? I think this is ready for merge.

vispy/scene/cameras/cameras.py
Eric89GXL: I'm fine with either one. I'm also okay with tweaking it in a subsequent PR where we take a mini poll :)

Eric89GXL:
Thanks for the fixes @campagnola

Issue #648, opened 2014-11-27
Implement more colormaps


rossant:
- [ ] [seaborn](https://github.com/mwaskom/seaborn/blob/master/seaborn/palettes.py)
- [ ] [colorbrewer][(http://colorbrewer2.org/)
- [x] [cubehelix](https://github.com/jradavenport/cubehelix)
- [x] implement routines to generate special colormaps (diverging, husl, hls...)
- [x] viridis, set as default
- [ ] inferno

sh4wn:
Do we want our own implementation for cubehelix our should we depend on the package? Or just copy the module with proper attribution?

almarklein:
No dependencies please :)  I think we should port the colormap to vispy's system, copying as much code as we can (and attributing it, of course).

sh4wn:
Ok, so I'm currently busy implementing this, but another question: What's the relation between the glsl_map defined in the class of a colormap and the GLSL programs in vispy/glsl/colormaps/[name].glsl? Are they both used?

rossant:
They are redundant, we'll have to merge them at some point. For now I think you can just use those in vispy.colors.

Eric89GXL:
BTW I have the `cubehelix` colormap in my open `vispy.plot` PR, so don't bother with that one...

almarklein:
cc @izaid because he did most of the works on colormaps.

ericdill:
Is this issue complete? There is a referenced PR that has been merged

Eric89GXL:
I think we have some but not all of those colormaps at this point

Eric89GXL:
We should probably switch to the same new default `viridis` as matplotlib:

https://www.youtube.com/watch?v=xAoljeRJ3lU

almarklein:
+1 for viridis

rougier:
Yes, veridis would be a great addition (and the default one maybe)

larsoner:
Closing for more recent discussion in #1397

PR #864, opened 2015-04-05
[WIP] Add more colormaps

sh4wn:
Hi all,

In this branch I will be working on adding some extra colormaps. This should address most of issue #648. Colormaps I would like to add:

- [x] Single Hue
- [x] HSL colormap (sea [seaborn] palettes)
- [x] [HUSL colormap][husl]
- [x] Diverging colormaps

[seaborn]: https://web.stanford.edu/~mwaskom/software/seaborn/tutorial/color_palettes.html
[husl]: http://www.husl-colors.org/


sh4wn:
Question: colormaps are retrieved by using the get_colormap function. How should we let the user configure the colormap (for example changing the hue)? 

Things I thought of:

- Just let the user import the class from the vispy.color.colormap module, and create their own instance
- Add a __call__ method to the class, which returns a new object with the new given values.

sh4wn:
SingleHue examples:

screenshot from 2015-04-05 11 56 55

rossant:
> Question: colormaps are retrieved by using the get_colormap function. How should we let the user configure the colormap (for example changing the hue)?

For example:

```python
get_colormap('single_hue', hue=x, value=y)
get_colormap(name, **kwargs)  # instanciate the given color map with the given keyword arguments
```

sh4wn:
That would be indeed a nice API. I've implemented it in that way. There is now also a class for a HSL colormap.

screenshot from 2015-04-05 14 14 31

vispy/color/colormap.py
```diff
@@ -438,6 +439,75 @@ def map(self, t):
...
+    of a color. The start color is almost white but still contains a hint of
+    the given color, and at the end the color is fully saturated.
+
+    Parameters
+    ----------
+    hue : scalar, optional
+        The number indicating the hue value. Must be in the range [0, 360].
+        Defaults to 200 (blue).
+    value : scalar, optional
+        The number represting the value component of a color. Must be in the
```
QuLogic: representing?

vispy/color/colormap.py
```diff
@@ -438,6 +439,75 @@ def map(self, t):
...
+class _HSL(Colormap):
+    """A colormap which is defined by n evenly spaced points in
+    a circular color space.
+
+    This means that we change the hue value while keeping the
+    saturation and value constant.
+
+    Parameters
+    ---------
+    n_colors : int (optional)
```
QuLogic: You used ` (optional)` here, but `, optional` in `_SingleHue`.

vispy/color/colormap.py
```diff
@@ -438,6 +439,75 @@ def map(self, t):
...
+        The number of colors to generate.
+    hue_start : int (optional)
+        The hue start value. Must be in the range [0, 360], the default is 0.
+    saturation : float (optional)
+        The saturation component of the colors generated. The default is fully
```
QuLogic: The tense is inconsistent. For `n_colors`, you used "to generate", but here it's "generated".

vispy/color/colormap.py
```diff
@@ -438,6 +439,75 @@ def map(self, t):
...
+        saturated (1.0). Must be in the range [0, 1.0].
+    value : float (optional)
+        The value component of the generated components or "brightness". Must
```
QuLogic: Leaving out some adjectives, "The component of the components"?

vispy/color/colormap.py
```diff
@@ -438,6 +439,75 @@ def map(self, t):
...
+        be in the range [0, 1.0], and the default is 1.0
+    controls : array-like
```
QuLogic: Aren't basically all arguments optional?

rossant:
Do you plan to add qualitative, sequential, and diverging colormaps here (see seaborn/color brewer) ?

sh4wn:
Welp, the comments were a bit rushed indeed. I'll fix it as soon as possible!

Diverging colormaps is definitely something I can do. Sequential is currently possible with the SingleHue colormap, but I can also add multi-hue sequential maps. I think some qualitative colormaps comes down to adding a few colormaps with hardcoded colors?

By the way, the python husl module is a bit larger than the cubehelix module, is it ok to depend on it? And gracefully fail if it is not available?

I'll be working on it this friday!

Eric89GXL:
How many lines is HUSL? Is it bigger than the other modules we've ported into `vispy.ext`?

Also, can you double-check that any code you take has BSD-compatible license (generally BSD, Apache, MIT)? We cannot use any code that uses GPL.

Eric89GXL:
Also, you need a rebase to get this mergeable

Eric89GXL:
Looks like you have a remaining style error:

https://travis-ci.org/vispy/vispy/jobs/58063588

The errors with GLFW segfaulting are not related to your PR, so ignore those for now. Let me know when this is ready for review.

vispy/color/colormap.py
```diff
@@ -3,12 +3,14 @@
...
 
 from __future__ import division  # just to be safe...
+import inspect
 
 import numpy as np
 
 from .color_array import ColorArray
 from ..ext.six import string_types
 from ..ext.cubehelix import cubehelix
+from ..ext import husl
```
Eric89GXL: looks like you only ever use `husl.husl_to_rgb` so might be worth just importing that directly as `from ..ext.husl import husl_to_rgb`

sh4wn:
Ok, I've added diverging colormaps, including some presets. 

screenshot from 2015-04-13 22-19-16

Any suggestions or comments?

vispy/color/colormap.py
```diff
@@ -508,7 +510,158 @@ def map(self, t):
...
+    ----------
+    hue : scalar, optional
+        The number indicating the hue value, which refers to a "true" color,
+        without any shading or tinting. Must be in the range [0, 360]. Defaults
+        to 200 (blue).
+    saturation_range : array-like, optional
+        The saturation represents how "pure" a color is. Less saturation means
+        more white light mixed in the color. A fully saturated color means
+        the pure color defined by the hue. No saturation means completely
+         hite. This colormap changes the saturation, and with this parameter
```
QuLogic: This should say `white`, no?

vispy/color/colormap.py
```diff
@@ -508,7 +510,158 @@ def map(self, t):
...
+        you can specify the lower and upper bound. Default is [0.2, 0.8].
+    value : scalar, optional
+        The number represting the value component of a color. This defines the
```
QuLogic: representing?

vispy/color/colormap.py
```diff
@@ -508,7 +510,158 @@ def map(self, t):
...
+    the given color, and at the end the color is fully saturated.
+
+    Parameters
+    ----------
+    hue : scalar, optional
+        The number indicating the hue value, which refers to a "true" color,
+        without any shading or tinting. Must be in the range [0, 360]. Defaults
+        to 200 (blue).
+    saturation_range : array-like, optional
+        The saturation represents how "pure" a color is. Less saturation means
```
QuLogic: It's inconsistent that this one is "The thing" while the other two are "The number indicating/representing the thing".

vispy/color/colormap.py
```diff
@@ -508,7 +510,158 @@ def map(self, t):
...
+    interpolation : str, optional
+        The interpolation mode of the colormap. Default: 'linear'. Can also
+        be 'zero'.
+        If 'linear', ncontrols = ncolors (one color per control point).
+        If 'zero', ncontrols = ncolors+1 (one color per bin).
+    """
+
+    def __init__(self, ncolors=6, hue_start=0, saturation=1.0, value=1.0,
+                 controls=None, interpolation='linear'):
+        hues = np.linspace(0, 360, ncolors + 1, dtype='int')[:-1]
```
QuLogic: Does this actually work? `np.linspace` does not accept the `dtype` keyword argument for me.

vispy/color/colormap.py
```diff
@@ -508,7 +510,158 @@ def map(self, t):
...
+        hues += hue_start
+        hues %= 359
```
QuLogic: Are you sure about this? 359 % 359 == 0. Why wouldn't it wrap at 360?

vispy/color/colormap.py
Eric89GXL: Things like this usually depend on version, looking at the docs you can see it was introduced in 1.9:

http://docs.scipy.org/doc/numpy/reference/generated/numpy.linspace.html

So safer to just use `.astype(int)`

QuLogic: Yea, sorry, I should have asked, "will this work for everyone?"

vispy/color/colormap.py
```diff
@@ -508,7 +510,158 @@ def map(self, t):
...
+    ---------
+    n_colors : int, optional
+        The number of colors to generate.
+    hue_start : int, optional
+        The hue start value. Must be in the range [0, 360], the default is 0.
+    saturation : float, optional
+        The saturation component of the colors to generate. The default is
+        fully saturated (1.0). Must be in the range [0, 1.0].
+    value : float, optional
+        The value component of the colors to generate or "brightness". Must
```
QuLogic: This means "(The value component or brightness) of the colours to generate.", right? Not '(The value component) or (the string "brightness")'? Because it could be interpreted the second way.

vispy/color/colormap.py
sh4wn: Ah, will change that!

vispy/color/colormap.py
sh4wn: Yeah I'm indeed not so sure about that value. Seaborn uses this value, and I couldn't find a quick answer why, so I kinda assumed there was a reason for it.

Eric89GXL: Might be worth opening an issue over there to ask

Eric89GXL:
ping @sh4wn do you have time to come back to this? If not someone else might be able to take it over. We don't want to leave PRs open too long otherwise we hit too many merge conflicts.

sh4wn:
Yes, I'll continue with this tommorow!

Eric89GXL:
I saw some coveralls comments -- let me know when the code is ready for another round of review.

sh4wn:
So I changed the hue wrap around value to 360. I [asked][tweet] of seaborn, and he said it would render invalid RGB values, but I can't seem to reproduce it. 

I also added a IPython notebook, because Github can currently render those inline, which is quite awesome.

[tweet]: https://twitter.com/michaelwaskom/status/597080579301974016

sh4wn:
Hmm too bad our colourmap rendering doesn't go too well with Github...

QuLogic:
> So I changed the hue wrap around value to 360. I asked the author of seaborn, and he argued that hues go from 0-359, but Wikipedia says otherwise. A hue of 360 doesn't produce an invalid RGB value.

It seems like one or both of you are confused about the modulus operator. `x % 360` _never returns_ 360.

sh4wn:
Oh yeah that's indeed true, makes the value 359 even weirder.

Eric89GXL:
Can you add an example that doesn't require a notebook?

Eric89GXL:
Also, why does the example need to be in the root folder? Do you not have `vispy` installed as `python setup.py develop`?

sh4wn:
> Also, why does the example need to be in the root folder? Do you not have vispy installed as python setup.py develop?

All these years of Python, and wow this does make a lot of things easier! But no I did not have vispy installed in that way.

As for the example, should I extend the docstring or create an example script in the examples folder? I also came across [this][colormaps] IPython notebook, so mine can be deleted.

[colormaps]: https://github.com/vispy/vispy/blob/master/examples/ipynb/colormaps.ipynb

Eric89GXL:
Please create a simple example script in the examples folder. It will help us make sure everything works well.

sh4wn:
There's an example (basics/scene/colored_line.py) which cycles through all colormaps. It is a very simple example, but it shows the colormaps are working. I could make a more advanced example (maybe loading an example shapefile with some data to visualise), but I'd rather spent my time today on porting several arrows from Glumpy. 

Anyone an idea why the Travis CI build with Python 3.4 minimal fails?

Eric89GXL:
Looks like you added and removed a file with thousands of lines. It would be good if we could squash your commits to get rid of it from the history, but I'm not sure it's possible since there was a merge in the middle. @campagnola is it possible? If not, that would be one reason to prefer rebasing to merging.

You also appear to have sorted out the Travis issue. I don't think a more advanced example is necessary. I'll go ahead and merge this assuming everything looks good in the code.

Eric89GXL:
It looks like [it is possible] (http://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit) but Travis isn't complaining about size so it's probably okay.

Eric89GXL:
Looks good, thanks @sh4wn!

PR #865, opened 2015-04-06
Add Colormap support for plot

brickgao:
Hi,

This PR is for adding Colormap support for plot.

```
cm = Colormap(['r', 'g', 'b'])
data = np.arange(15)
fig = vp.Fig()
fig[0:4, 0:4].plot(data, width=1, color=cm, face_color=cm, edge_color=cm)
```

The code above will show:

### DIFF
--- a/vispy/visuals/line/line.py
+++ b/vispy/visuals/line/line.py
@@ -10,7 +10,7 @@ from __future__ import division
 import numpy as np
 
 from ... import gloo
-from ...color import Color, ColorArray, get_colormap
+from ...color import Color, ColorArray, Colormap, get_colormap
 from ...ext.six import string_types
 from ..shaders import ModularProgram, Function
 from ..visual import Visual
@@ -87,7 +87,7 @@ class LineVisual(Visual):
 
     antialias : bool
         Enables or disables antialiasing.
-        For method='gl', this specifies whether to use GL's line smoothing, 
+        For method='gl', this specifies whether to use GL's line smoothing,
         which may be unavailable or inconsistent on some platforms.
     """
     def __init__(self, pos=None, color=(0.5, 0.5, 0.5, 1), width=1,
@@ -96,13 +96,13 @@ class LineVisual(Visual):
 
         self._changed = {'pos': False, 'color': False, 'width': False,
                          'connect': False}
-        
+
         self._pos = None
         self._color = None
         self._width = None
         self._connect = None
         self._bounds = None
-        
+
         # don't call subclass set_data; these often have different
         # signatures.
         LineVisual.set_data(self, pos=pos, color=color, width=width,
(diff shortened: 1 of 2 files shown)


Eric89GXL:
I think that eventually we want to be able to do this sort of thing in the fragment shader instead of Python, but I'm fine with this solution for now. @campagnola WDYT?

inclement:
This PR automatically generates the colours based on the number of points, do you (that is, the vispy devs) have a plan for how the api would allow more arbitrary settings, e.g. an optional `colormap_scalars` argument? I ask because I'd like to PR some other plotting commands with this functionality.

Eric89GXL:
I think that eventually we want to be able to just say that e.g. each point on a line has some "value", and that in GLSL does a colormap lookup (as opposed to doing the lookup in Python as is done here). @rossant @campagnola WDYT?

rossant:
@Eric89GXL +1. I imagine there are times where you want to assign RGBA values explicitely to each point, but in most cases, you'd probably use a GLSL colormap. There are a few colormap GLSL snippets in the repo already.

Eric89GXL:
@inclement I guess the answer to your question is yes, we'd like to be able to support that. Do you feel like implementing it? :)

In the meantime, I'm fine with merging this as-is -- we can go back and change the underlying implementation later if we want. @inclement does this PR work with the ideal API you're thinking of wanting to use eventually? We should make sure to get the API right so we break as little code as possible.

almarklein:
> I think that eventually we want to be able to just say that e.g. each point on a line has some "value"

+1 

If that value is a scalar, it is looked up in a colormap, if it has 2 elements, it is looked up in a texture, if it is a 3 elements it is interpreted as RGB, and with 4 RGBA. This does not have to be implemented all at once :)

inclement:
@almarklein Presumably that would also have a separate 'colormap' argument to actually pick what colormap is used in the case that a scalar is passed? That would sound great to me, and would also solve the problem of letting colormaps be specified by name or by the actual colormap object - I think that's a minor problem with this PR currently, because the color argument wants to interpret a string as a Color directly.

brickgao:
> If that value is a scalar, it is looked up in a colormap, if it has 2 elements, it is looked up in a texture, if it is a 3 elements it is interpreted as RGB, and with 4 RGBA.

@almarklein I think if `Colormap` is scaleable, it would be more common than changing the format of `Colormap` each time. I haven't find a way to distinguish the type of function parameters in Python. Is this viable in GLSL ;) ?

almarklein:
What I mean is that some Visuals, most notably the Mesh class, can have a way to specify a "value" for each vertex/point, and the Visual will decide whether to use colormap/texture/RGB based on the shape of this value. A colormap/texture can probably be associated via another method/property. The visual will have a GLSL snippet for each situation.

PR #868, opened 2015-04-08
Perspective correction

campagnola:
This fixes perspective transforms. Summary:

* OpenGL uses homogeneous coordinates; our transforms need to do the same.
* Corrected STTransform to work correctly when w != 1
* PerspectiveTransform is now just a thin wrapper around AffineTransform. Might just combine these both into MatrixTransform in a future PR.
* Fixed ImageVisual to work correctly in impostor mode when its transforms affect z
* Added a script that I used to test perspective. This will be converted into a unit test with image comparison in a future PR.
* There is a bizarre bug on my machine that I suspect is an OpenGL bug. The console example gives me a warning:

```
INFO: Variable u_scale is not an active uniform
```

This is due to the STTransform map function, which now looks like:

```
        vec4 st_transform_map(vec4 pos) {
            return vec4(pos.xyz * $scale.xyz + $translate.xyz * pos.w, pos.w);
        }
```

If I switch back to the old implementation, the warning goes away:

```
        vec4 st_transform_map(vec4 pos) {
            return (pos * $scale) + $translate * pos.w;
        }
```

I can't see why the $scale uniform would be optimized out of the shader in either case. I'm betting Travis won't have the same error, though..


vispy/scene/cameras/tests/perspective_test.py
```diff
@@ -0,0 +1,64 @@
...
+        image.transform = scene.STTransform(translate=(-12.8, -12.8),
+                                            scale=(0.1, 0.1))
+        v.add(image)
+        images.append(image)
+        
+
+@canvas.connect
+def on_key_press(ev):
+    if ev.key.name == '1':
+        print "Image method: impostor"
```
Eric89GXL: all of these print statements are non-Py3k compatible

vispy/scene/cameras/tests/perspective_test.py
```diff
@@ -0,0 +1,64 @@
...
+        v.camera.distance = 30
+        #v.camera = 'panzoom'
+        #v.camera.aspect = 1
+        
+        views.append(v)
+        image = scene.visuals.Image(imdata, method='impostor', grid=(4, 4))
+        image.transform = scene.STTransform(translate=(-12.8, -12.8),
+                                            scale=(0.1, 0.1))
+        v.add(image)
+        images.append(image)
```
Eric89GXL: Maybe a stupid question, but why have four of these?

vispy/scene/cameras/tests/perspective_test.py
```diff
@@ -0,0 +1,64 @@
...
+        for vb in views:
+            vb.clip_method = 'fragment'
+    elif ev.key.name == '4':
+        print "Viewbox method: viewport"
+        for vb in views:
+            vb.clip_method = 'viewport'
+    elif ev.key.name == '5':
+        print "Viewbox method: fbo"
+        for vb in views:
+            vb.clip_method = 'fbo'
```
Eric89GXL: Should add `.update()` calls after each of these, right? Or maybe just one at the end (updating is okay even if a non-meaningful key was pressed).

Eric89GXL: ...or does it auto-update because of the changed param?

Eric89GXL: FYI 1/2 seems to look the same, 3/4 look the same and 5 (FBO clipping) is messed up on my system. Is this what you get?

vispy/scene/cameras/tests/perspective_test.py
campagnola: Some of the bugs I was dealing with were dependent on the position of the viewbox within the canvas.

Eric89GXL:
You mean you have that warning in `examples/basics/scene/console.py`? I do not on OSX (NVIDIA mobile GPU), I can test Linux tomorrow.

vispy/scene/cameras/tests/perspective_test.py
campagnola: Yes, setting vb.clip_method triggers an update.
Yes, that is the output I get. The FBO issue is for another day.

Eric89GXL:
Looks like Travis is angry because `py.test` tried to run your `perspective_test.py` as a test. Renaming the file (not to have `test` in the name) should fix it, I think.

campagnola:
Ah! Travis has the same error. It does make sense, though. The shader looks like:
```
uniform vec4 u_scale;
uniform vec4 u_translate;
vec4 st_transform_map(vec4 pos) {
    return vec4(pos.xyz * u_scale.xyz + u_translate.xyz * pos.w, pos.w);
}

attribute vec2 a_position;
. . .
void main (void) {
    vec4 pos = st_transform_map(vec4(0., 0., 0., 1.));
    gl_Position = pos + vec4(a_position * u_logical_scale, 0., 0.);
    . . . 
}

```
Since the input to `st_transform_map` has xyz=0, `u_scale` can be optimized out. Will post a fix soon.

campagnola:
Green light! @almarklein would you like a peek?

almarklein:
Changes look good to me. However, for volume rendering this new perspective does not work as it should.

campagnola:
Check out the last commit--this fixes perspective in volume and also no longer requires face subdivision. The downside is that it moves the ray calculation to the fragment shader, which may slow the rendering down (but its probably not significant compared to the for-loop).

There is an issue with clipping that I wasn't able to fix. I don't quite have a handle on how the shaders work. Do you want to work on this in a sub-PR?

vispy/visuals/volume.py
```diff
@@ -51,37 +51,15 @@
...
-    
-    // Calculate a clip plane (in texture coordinates) for the camera
-    // position, in case that the camera is inside the volume.
-    pos_in_cam1.zw = vec2(u_cameraclip, 1.0);
-    vec3 cameraposinvol = $viewtransformi(pos_in_cam1).xyz;
-    cameraposinvol /= u_shape;  // express in texture coords
-    v_clipplane.xyz = v_ray;
-    v_clipplane.w = dot(v_clipplane.xyz, cameraposinvol);
+    v_position = a_position;
+    gl_Position = $transform(vec4(v_position, 1));
```
almarklein: 1.0

almarklein:
Wow, I truly believed it was not possible to get correct perspective without partitioning. You blew my mind :)

As for clipping, I suspect its related to the `w` factor which is now different. Also note the artifacts when fov=0.

vispy/visuals/volume.py
campagnola: Is a picky requirement of some GL drivers?

almarklein: Yes, we've had problems with some drivers because of this. Also, it will *always* cause compilation errors on WebGL.

Eric89GXL:
@campagnola looks like you addressed @almarklein's comments, except for clipping / fov=0 artifacts? Or are those fixed now / in a subsequent PR?

campagnola:
Last commit fixes clipping and moves the transformations back to the vertex shader where they belong.

Eric89GXL:
Good to go from your end, then @campagnola?

campagnola:
Yep. There is still an edge artifact, but that is present in master as well.

Eric89GXL:
Looks like Travis is failing on GLFW, almost certainly unrelated to your changes. Looks like we've probably found another bug for them :(

almarklein:
> but that is present in master as well.

mmm, you're right. I hadn't seen it yet (except for before I fixed it early on. I guess I did not fix it enough)

vispy/visuals/volume.py
```diff
@@ -51,37 +50,29 @@
...
 attribute vec3 a_texcoord;
 uniform vec3 u_shape;
-uniform float u_cameraclip;
 
 varying vec3 v_texcoord;
-varying vec3 v_ray;
-varying vec4 v_clipplane;
+varying vec3 v_position;
+varying vec4 v_position2;
+varying vec4 clipplane;
```
almarklein: Not used

vispy/visuals/volume.py
```diff
@@ -105,39 +96,44 @@
...
-    // Discart front facing
-    //if (!gl_FrontFacing)
-    //    discard;
-    
-    // Uncomment this to show a grid of the backfaces
-    //vec3 pcd = v_texcoord * u_shape;
-    //for (int d=0; d<3; d++)
-    //    if (pcd[d] > 1 && pcd[d] < (u_shape[d]-1) && sin(pcd[d]) > 0.9)
-    //            discard;
+    // Calculate ray. In the fragment shader we do another normalization
```
almarklein: This comment needs an update

vispy/visuals/volume.py
```diff
@@ -628,7 +614,7 @@ def _create_vertex_data(self):
         
         # Get indices and vertices for triangles. 
         indices, tex_coord, ver_coord = self._calc_coords(tex_coord, ver_coord,
-                                                          div)
+                                                          2)
```
almarklein: I suppose you can remove lots of lines above in this function. Which is awesome, it should even increase performance a wee bit.

vispy/visuals/volume.py
```diff
@@ -8,14 +8,13 @@
 
 In Python, we define the six faces of a cuboid to draw, as well as
 texture cooridnates corresponding with the vertices of the cuboid. In
-the vertex shader, we establish the ray direction and pass it to the
-fragment shader as a varying. In the fragment shader, we have the
-texture coordinate as the starting point of the ray, and we know the
-ray direction. We then calculate the number of steps and use that number
+the vertex shader, we compute the near clipping position along the view
+ray and pass it to the fragment shader as a varying. In the fragment shader, 
+we compute the ray direction and calculate the number of steps to use
```
almarklein: It seems this needs an update too.

almarklein:
Apart for some small comments, this LGTM. Nice work!

campagnola:
> mmm, you're right. I hadn't seen it yet (except for before I fixed it early on. I guess I did not fix it enough)

I think it is caused by the emulated 3D texture interpolating between z-slices. I have another PR coming up that touches some of that code, though, so let's hold off fixing it until then.

Eric89GXL:
@campagnola is this good to go from your end, then?

Eric89GXL:
(I restarted your test now that GLFW is fixed and the CIs are happy)

campagnola:
Cool, I'm happy with this one.

Eric89GXL:
Great, merged.

campagnola:
Huzzah!

PR #869, opened 2015-04-08
Fixed Cube mesh generation

inclement:
Currently the cube mesh generation function returns a 1D array of face indices, but the MeshData expects it to be 2D. This doesn't seem to affect anything right now as functions like `cube._meshdata.get_vertex_normals()` are not accessed, but exceptions are raised if that code is actually called. This PR just fixes that, everything seems to work fine afterwards.


Eric89GXL:
Nice, thanks for the fix!

PR #870, opened 2015-04-09
Image testing

campagnola:
This adds a basic UI for image comparison testing. It allows the user to compare the output of a unit test with a previously-generated standard image and decide (if there are any differences) whether to fail the test or to pass the test and save a new standard image.

Standards are stored and retrieved from a local clone of the test-data repository, so after running a batch of tests you can optionally commit the new standards and push upstream.

Give it a try:

```
$ python vispy/scene/cameras/tests/perspective_test.py
```

This causes the test-data repo to be cloned, but the test will fail because the standard has not
been added to the repo yet. To invoke the testing ui, use the `--vispy-audit` flag:

```
$ python vispy/scene/cameras/tests/perspective_test.py --vispy-audit
```

The window shows the test output, the standard (if any), and the difference between the two. Press (p)ass to store the test image as the new standard. Run again; this time the test passes.

Finally, try modifying perspective_test.py and run the test again.

@Eric89GXL, any thoughts about how to set the `audit_tests` config flag when running tests with `python make tests ...` ?


campagnola:
(this PR is built on #868; we'll need to merge that one first)

Eric89GXL:
Do you mean you want to be able to set the switch? The way the test runners execute currently, it spawn subprocesses for each test in case of segfaults. So you should be able to write a function that executes a script or set of scripts with the necessary command line option. That might work okay for now. Not sure if we want them enabled by default (in `make test`), since interactive tests can be a real pain.

campagnola:
Right, I want to be able to run tests locally, and then if any of the image comparison tests fail, I can run them again with the ``--audit`` flag (or some mechanism) to decide whether the new test results should replace the old ones.

Eric89GXL:
Ahh, so it's a two-stage thing -- one that is automated, and then one that you can run after your tests fail that allows you to potentially replace the result on the data repo. Sounds good.

What I'd do is make a new decorator `@audit` or so for tests that "flag" them as having an interactive option. Then make a new testing mode `make audit` / `python make test audit` that only runs the subset of tests that have that flag. For any test that fails, it launches the GUI for you to make your decision and replace. WDYT?

vispy/testing/_testing.py
```diff
@@ -408,9 +409,13 @@ def run_tests_if_main():
     except Exception:
         pass
     import __main__
-    print('==== Running tests in script\n==== %s' % fname)
-    run_tests_in_object(__main__)
-    print('==== Tests pass')
```
campagnola: @Eric89GXL, is this ok with you?

Eric89GXL: This seems like it should work okay. I assume the advantage is more informative output on error?

campagnola:
Green light! 
For now, we can just run the tests manually with `--vispy-audit` as in the example above. I think this will be sufficient as we work on adding individual tests or fixing bugs. If someone really needs the "audit all auditable tests" feature, it can be added at that time.

Eric89GXL:
So that will command-line option magically allows you choose whether or not to update a test image if a given test is failing? Does it work with all of our existing visual tests?

Eric89GXL:
Nice +1% coverage increase BTW

Eric89GXL:
Can you rebase or merge? This has some of the changes from #868 in it, which makes it harder to review

vispy/scene/widgets/grid.py
```diff
@@ -163,6 +163,12 @@ def _update_child_widgets(self):
         cols = np.linspace(rect.left, rect.right, n_cols+1)
         colstart = cols[:-1] + s2
         colend = cols[1:] - s2
+        
+        # snap to pixel boundaries to avoid drawing artifacts
+        colstart = np.round(colstart)
+        colend = np.round(colend)
+        rowstart = np.round(rowstart)
+        rowend = np.round(rowend)
```
Eric89GXL: Do these need to be `int` or `float`? `round` returns a `float`, so just wanted to check.

vispy/testing/_testing.py
```diff
@@ -341,6 +357,7 @@ def assert_image_equal(image, reference, limit=0.9):
                 corr = np.corrcoef(a.ravel(), b.ravel())[0, 1]
             if corr > max_corr:
                 max_corr = corr
+    print(max_corr, limit)
```
Eric89GXL: Why print every time, or is this left over from debug?

Eric89GXL:
Other than my minor comments, LGTM

vispy/scene/widgets/grid.py
campagnola: They ultimately become float vertices, so no need to convert..

campagnola:
@Eric89GXL and @almarklein, please check the changes in the last commit #7b99faa -- this fixes the bug that led to #880 for me. If it looks ok, then I will apply the change to all backends.

vispy/scene/cameras/tests/perspective_test.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
```
Eric89GXL: all of our other tests are named `tests/test_*`, so this should be renamed

vispy/visuals/tests/test_ellipse.py
```diff
@@ -19,20 +19,20 @@ def test_circle_draw():
         ellipse = visuals.Ellipse(pos=(75, 35, 0), radius=20,
                                   color=(1, 0, 0, 1))
         c.draw_visual(ellipse)
-        assert_image_equal("screenshot", 'visuals/circle1.png')
+        assert_image_approved("screenshot", 'visuals/circle1.png')
```
Eric89GXL: Do we need all three of `assert_image_equal`, `assert_image_approved` and `assert_image_match`? Can we eliminate one of them? I think match uses pixel-by-pixel matching whereas equal uses maximal cross-correlation value, and it looks like approved just maps to match. I can see why we'd want cross-corr in some cases and pixel differences in another, but can we combine approved and match? Or make it so approved can have a mode where you choose which comparison to use?

vispy/app/backends/_qt.py
```diff
@@ -590,7 +590,6 @@ def _vispy_set_current(self):
         if self._vispy_canvas is None:
             return  # todo: can we get rid of this now?
         if self.isValid():
-            self._vispy_canvas.set_current()  # Mark as current
```
Eric89GXL: This seems weird to me -- inside the backend's `_vispy_set_current`, this seems to make it so that it no longer makes the Canvas's GLIR queue current? (Right?)

Eric89GXL: Ahh, I see the related change below that does it instead. I think the problem is that there might be places in these backends where we call `self._vispy_set_current()`, those will need to be changed to `self._vispy_canvas.set_current()` instead.

vispy/scene/canvas.py
```diff
@@ -207,6 +207,7 @@ def draw_visual(self, visual, event=None, viewport=None):
             Optionally specifies the viewport to use. If None, the entire
             physical size is used.
         """
+        self.set_current()
```
Eric89GXL: I seem to remember reading at one point that context switches can be slow. Can you check to make sure this doesn't hurt performance (see how non-vsync'ed framerate drops with and without this line even if nothing is drawn)?

vispy/visuals/tests/test_ellipse.py
Eric89GXL: ...also, once all comments are addressed I'll need to run tests in HiDPI mode to make sure these are all still okay, given that the matching algorithm has changed from xcorr to pixel counting.

campagnola: Yes, the plan is to fold `assert_image_equal` into `assert_image_match` using an `image_corr` argument or similar.

Eric89GXL:
...and Travis is pretty broken by the changes currently.

campagnola:
New test images have not been uploaded yet. More to come.

vispy/scene/canvas.py
campagnola: Context switches are slow, but requesting a context that is already current is free (at least on my machine). I checked examples/basics/scene/line_update; ~630fps either way. We need a standard set of benchmarks...

vispy/app/backends/_qt.py
campagnola: Done for qt backend.

vispy/scene/cameras/tests/perspective_test.py
campagnola: done.

campagnola:
Ok, all visual tests have been updated to use the new system. Everything passes locally for me, but there is a snag: If I push my changes to the test-data repository, then all other PRs (and master) will have test failures until the new tests are merged. 

We could fix this by having a variable in `testing/image_tester.py` specify the exact commit ID that should be used for running tests. Then every time the test_data repo is changed, the commit ID must change as well (but older commits should always pass their tests since they know which version of the test data to compare against). The change would have to be introduced in a separate PR that goes in before this one..   thoughts?

campagnola:
Ok, ready for final review.

vispy/util/wrappers.py
```diff
@@ -132,3 +132,24 @@ def run_subprocess(command, return_code=False, **kwargs):
...
+        output = proc.stdout.read()
+        proc.wait()
+        if proc.returncode != 0:
+            msg = ("Command '%s' returned non-zero exit status %d" % 
+                   (args[0], proc.returncode))
+            ex = subprocess.CalledProcessError(msg)
+            ex.returncode = proc.returncode
+            ex.output = output
+            raise ex
+        return output
```
Eric89GXL: I think `run_subprocess` already does this. Can you try it instead? Or get rid of where `run_subprocess` is used? DRY and all that.

vispy/visuals/tests/test_markers.py
```diff
@@ -13,10 +13,10 @@ def test_markers():
     # work for people
     with TestingCanvas() as c:
         marker = Markers()
-        marker.set_data(np.array([[50, 50]], np.float32))
+        np.random.seed(57983)
+        marker.set_data(np.random.normal(size=(30, 2), loc=50, scale=10))
```
Eric89GXL: FYI you can use `rng = np.random.RandomState(57983)` then `rng.normal(...)` to avoid needing to use globals at all

vispy/util/wrappers.py
campagnola: The main difference is that check_output leaves stderr unchanged. I guess I can add a flag to run_subprocess..

Eric89GXL:
Tests pass over here on Linux, I'll give it a shot on OSX to make sure HiDPI doesn't break everything.

vispy/util/wrappers.py
Eric89GXL: might be easier just to get rid of `run_subprocess` instead -- more standard anyway.

campagnola: Well there are some places that do use the stderr return value.. and it turns out all you need is `stderr=None` to make `run_subprocess` behave like `check_output`, so I removed it :)

Eric89GXL:
You appear to have some bug:

https://travis-ci.org/vispy/vispy/jobs/59756318

Eric89GXL:
I have an additional failing test when running OSX in HiDPI mode that I do not have in current `master`:
```
vispy/visuals/tests/test_markers.py:19: in test_markers
    assert_image_approved("screenshot", "visuals/markers.png")
vispy/testing/image_tester.py:114: in assert_image_approved
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

im1 = array([[[   0.,    0.,    0.,  255.],
        [   0.,    0.,    0.,  255.],
  ...,
        [   0.,    0.,    0.,  255.],
        [   0.,    0.,    0.,  255.]]])
im2 = array([[[  0,   0,   0, 255],
        [  0,   0,   0, 255],
        [  0,   0,...55],
        [  0,   0,   0, 255],
        [  0,   0,   0, 255]]], dtype=uint8)
min_corr = 0.9, px_threshold = 50.0, px_count = None, max_px_diff = None, avg_px_diff = None, img_diff = None

...

        if min_corr is not None:
            with np.errstate(invalid='ignore'):
                corr = np.corrcoef(im1.ravel(), im2.ravel())[0, 1]
>           assert corr >= min_corr
E           assert 0.8635152548646563 >= 0.9

vispy/testing/image_tester.py:196: AssertionError
``` 
It's probably fine to relax that one to 0.8 or so. Fix that and make Travis happy and we should be good here.

campagnola:
Travis is happy, not sure what's up with appveyor..

Eric89GXL:
On HiDPI I still have the same error as before, see comment above

campagnola:
One better: I added the ability to use a simulated pixel scale factor. After that, I was able to make a correction to MarkerVisual that improves the correlation (I get 0.96 now; please test on your end).

Eric89GXL:
What, you don't like me and @nippoo having to do all of the HiDPI testing? :) Seriously though, thanks for making it so I shouldn't need to change my display to 720x540 to run HiDPI tests. It's fixed at my end and the changes look good, so I'll merge once Travis comes back happy.

Eric89GXL:
Thanks @campagnola a 1% coverage increase is great, and it looks like the new testing features will be very helpful. What's next?

campagnola:
Next is to clean up the volume PR, and then a scenegraph overhaul.

Issue #871, opened 2015-04-11
Depth testing broken for some mesh Nodes


inclement:
This can be seen in master in the basics/visuals/tube.py example.

I bisected the problem to the introduction of set_gl_state for the Visual class, and I think the actual cause is that MeshVisual and Node both call `Visual.__init__(...)`, but Node does so second. This overwrites the MeshVisual setting (enabling depth testing) with the default.

In testing, I found that e.g. replacing the relevant `Visual.__init__` line with the following made things seem to work as normal, to demonstrate the issue.

    if not hasattr(self, '_gl_state'):
        self._gl_state = {'preset': None}

However, I don't understand why this issue doesn't seem to affect the other examples using scene.visual classes, such as basics/scene/isosurface.py

Eric89GXL:
Could it be fixed by switching the `Node` super-class `__init__` to occur before the `Mesh` `set_gl_state` calls instead of after? @campagnola if you have an idea of the correct answer here feel free to jump in.

inclement:
@Eric89GXL I tried this by switching the order in the constructed `__init__` method of vispy/scene/visuals.py, but it doesn't work - running the tube example gives

    Traceback (most recent call last):
      File "tube.py", line 59, in <module>
        canvas.view.add(l1)
      File "/home/asandy/devel/vispy/vispy/scene/widgets/viewbox.py", line 182, in add
        node.add_parent(self.scene)
      File "/home/asandy/devel/vispy/vispy/scene/node.py", line 148, in add_parent
        self.events.parents_change(added=parent)
    AttributeError: 'EmitterGroup' object has no attribute 'parents_change'

I think this is a similar problem, the Node adds these events to itself but then self.events is overridden by the MeshVisual's own `__init__`.

Eric89GXL:
Ahh, so it's a more general problem than just `gl_state` then. Somehow we need to make it so that either `Node` respects what is done by its subclasses and have `Node.__init__` run at the end of the subclass `__init__`, or the subclasses need to respect what `Node` does and do `Node.__init__` at the beginning of each subclass `__init__`. I'm fine either way. @campagnola WDYT?

campagnola:
I have already fixed this in a branch I am working on.

PR #873, opened 2015-04-13
Volume rendering methods

campagnola:
The main purpose of this PR is to add a couple of new volume rendering methods, but many ancillary changes are included.

* Add "translucent" rendering method (better quality compositing but requires more effort in selecting colormaps)
* Add "additive" rendering method (nice for visualizing 3D functions like electron orbitals)
* Reorganized a "bit" of the volume fragment shader core--improved performance, simpler code, somewhat easier to understand.
* Tweaked isosurface rendering; no more ray randomization.
* Added Z-interpolation to emulated 3D texture
* Disabled comment removal from shader preprocessing (@rougier, was this necessary?)
* Fixed `VisualNode` double-init bug (see #871) 

@almarklein, please be critical--I imagine I have not tested the volume rendering under all of the cases that you have.

Depends on #870.

### DIFF
--- a/examples/basics/scene/volume.py
+++ b/examples/basics/scene/volume.py
@@ -109,7 +109,7 @@ def on_key_press(event):
     elif event.text == '0':
         cam1.set_range()
         cam3.set_range()
-    elif event.text in '[]':
+    elif event.text != '' and event.text in '[]':
         s = -0.025 if event.text == '[' else 0.025
         volume1.threshold += s
         volume2.threshold += s
--- a/examples/basics/scene/volume.py
+++ b/examples/basics/scene/volume.py
@@ -10,11 +10,12 @@ Example volume rendering
 
 Controls:
 
-* 1 - toggle camera between first person (fly) and regular 3D (turntable)
-* 2 - toggle between volume rendering methods
-* 3 - toggle between stent-CT / brain-MRI image
-* 4 - toggle between colormaps
-* 0 - reset cameras
+* 1  - toggle camera between first person (fly) and regular 3D (turntable)
+* 2  - toggle between volume rendering methods
+* 3  - toggle between stent-CT / brain-MRI image
+* 4  - toggle between colormaps
+* 0  - reset cameras
+* [] - decrease/increase isosurface threshold
 
 With fly camera:
 
@@ -50,7 +51,7 @@ view = canvas.central_widget.add_view()
 emulate_texture = True
 
 # Create the volume visuals, only one is visible
-volume1 = scene.visuals.Volume(vol1, parent=view.scene, threshold=0.5,
+volume1 = scene.visuals.Volume(vol1, parent=view.scene, threshold=0.1,
                                emulate_texture=emulate_texture)
 volume1.transform = scene.STTransform(translate=(64, 64, 0))
 volume2 = scene.visuals.Volume(vol2, parent=view.scene, threshold=0.5,
@@ -73,10 +74,13 @@ def on_key_press(event):
         view.camera = cam_toggle.get(view.camera, 'fly')
     elif event.text == '2':
         methods = ['mip', 'translucent', 'iso', 'additive']
+        cmaps = {'mip': 'grays', 'translucent': 'tgrays', 'iso': 'grays', 'additive': 'tgrays'}
         method = methods[(methods.index(volume1.method) + 1) % 4]
         print("Volume render method: %s" % method)
         volume1.method = method
+        volume1.cmap = cmaps[method]
         volume2.method = method
+        volume2.cmap = cmaps[method]
     elif event.text == '3':
         volume1.visible = not volume1.visible
         volume2.visible = not volume1.visible
@@ -88,16 +92,16 @@ def on_key_press(event):
         cam1.set_range()
         cam3.set_range()
     elif event.text == '[':
-        volume1.threshold -= 0.1
+        volume1.threshold -= 0.025
         volume2.threshold = volume1.threshold
     elif event.text == ']':
-        volume1.threshold += 0.1
+        volume1.threshold += 0.025
         volume2.threshold = volume1.threshold
 
 # for testing performance
-#@canvas.connect
-#def on_draw(ev):
-    #canvas.update()
+@canvas.connect
+def on_draw(ev):
+    canvas.update()
 
 if __name__ == '__main__':
     print(__doc__)
--- a/examples/basics/scene/volume.py
+++ b/examples/basics/scene/volume.py
@@ -45,7 +45,7 @@ canvas.measure_fps()
 view = canvas.central_widget.add_view()
 
 # Set whether we are emulating a 3D texture
-emulate_texture = True
+emulate_texture = False
 
 # Create the volume visuals, only one is visible
 volume1 = scene.visuals.Volume(vol1, parent=view.scene, threshold=0.225,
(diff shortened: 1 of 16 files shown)


rougier:
I removed comments because else, parsing do not work in some cases:
```
uniform vec4 /* */ color;
```

inclement:
I'm finding meshes to not work, e.g. examples/basics/scene/surface_plot.py only shows the axes (not the mesh), examples/basics/visuals/mesh.py shows nothing, and examples/basics/visuals/tube.py shows only the tube with shading='lines'.

vispy/scene/cameras/cameras.py
```diff
@@ -436,13 +436,14 @@ def _set_scene_transform(self, tr):
...
-                finally:
-                    cam._linked_cameras_no_update = False
+        for cam in self._linked_cameras:
+            if cam is self._linked_cameras_no_update:
+                continue
+            try:
+                cam._linked_cameras_no_update = self
+                cam.set_state(self.get_state())
+            finally:
+                cam._linked_cameras_no_update = None
```
almarklein: Ah, linking A to B and then C to B! Good catch.

almarklein:
The translucent method is sweet! 

However, the iso method looks crap on my machine. It ha all sorts of stuff in the bg. Also toggling volume fails as well as toggling colormaps.

There are quite a few subtle details in the code that you touched. I'm happy to go with your changes as long as it keeps working as expected :)

While you're in here, can you look at the spine that shines through at the front?

campagnola:
Ready for review. I think I have addressed all comments above.

Eric89GXL:
@campagnola I assume you don't want to deal with #888 here?

vispy/gloo/preprocessor.py
```diff
@@ -45,7 +45,7 @@ def replace(match):
                 logger.critical('"%s" not found' % filename)
                 raise RuntimeError("File not found", filename)
             text = '\n// --- start of "%s" ---\n' % filename
-            text += remove_comments(open(path).read())
+            text += open(path).read()
```
Eric89GXL: This will leak file descriptors, can you change to use `with`?

vispy/visuals/mesh.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, vertices=None, faces=None, vertex_colors=None,
         Visual.__init__(self, **kwargs)
         
         self.set_gl_state('translucent', depth_test=True,
-                          cull_face='front_and_back')
+                          cull_face=False)
```
Eric89GXL: At first I thought this would cause a change in behavior, but now I'm thinking that it won't because on `master` just the property type of culling gets set but the culling isn't enabled...?

Eric89GXL:
Other than my minor comments LGTM. @almarklein can you have a look?

vispy/visuals/mesh.py
campagnola: Originally these were set to 'front_and_back', which should cause both faces to be _culled_. However, a separate bug prevented that being set correctly. After I fixed that bug, the meshes mysteriously stopped drawing :)

campagnola:
I'll deal with #888 separately.

vispy/gloo/preprocessor.py
campagnola: The descriptors should be freed along with the File objects. Is that not the case?

Eric89GXL: On py3k you get warnings about descriptors that are not explicitly closed

campagnola: Is that an OSX thing? I've never seen anything like that. At least on my system I can confirm using `os.fdopen` that descriptors are not leaked.

Eric89GXL: Yeah I guess they're not leaked, you just get annoying warnings in some circumstances. I've gotten them on Linux before, and in the past I've had problems with Travis logs being full of them on some repos. You can leave it if you want and we can change it later if need be. Personally I prefer the explicitness of `with` in any case.

Eric89GXL: Thinking back, it's possible I was seeing `ResourceWarnings` from opening other types of objects -- and that might have prompted me to try to use `with` whenever possible.

QuLogic: You should definitely put this in a `with`. On Py3k you'd get the `ResourceWarning` on exit, usually.

campagnola: Ok, I found it:  https://docs.python.org/3.5/whatsnew/3.2.html

> A new warning category, ResourceWarning, has been added. It is emitted when potential issues with resource consumption or cleanup are detected. It is silenced by default in normal release builds but can be enabled through the means provided by the warnings module, or on the command line.

> A ResourceWarning is also issued when a file object is destroyed without having been explicitly closed. While the deallocator for such object ensures it closes the underlying operating system resource (usually, a file descriptor), the delay in deallocating the object could produce various issues, especially under Windows.

almarklein:
Mostly looking good. Do I see it correctly that you fixed the spine shining through at the front?

The rendering is wrong when flipping axis though, try the flipped_axis.py example. Should not be too hard to fix (once you know where to fix it).

(I'm away next week, but don't let me hold this up!)

campagnola:
Yes, the texture coordinates are now clipped in emulated 3d.

The volume stops working when axes are flipped because that inverts the face normals, causing the front surfaces to be drawn (I draw only the back faces to cut down on the fragment shader cost). I can't immediately think of a good fix for that. We should expect all kinds of trouble in that case--meshes drawn inverted, broken lighting calculations, etc.

almarklein:
All this can be taken into account, and I think we *should* take it into account. In the previous volume implementation there was no culling, but the fragments of the front faces were discarded asap.

campagnola:
*grumble*
You are correct. We definitely need to be able to flip the volume..

campagnola:
Ok @almarklein, axis flipping is fixed. Anything else?

examples/basics/scene/volume.py
```diff
@@ -10,11 +10,12 @@
...
-* 2 - toggle between mip and iso render styles
-* 3 - toggle between stent-CT / brain-MRI image
-* 4 - toggle between colormaps
-* 0 - reset cameras
+* 1  - toggle camera between first person (fly) and regular 3D (turntable)
+* 2  - toggle between volume rendering methods
+* 3  - toggle between stent-CT / brain-MRI image
+* 4  - toggle between colormaps
+* 0  - reset cameras
+* [] - decrease/increase isosurface threshold
```
Eric89GXL: Minor issues:

* camera reset doesn't seem to do anything for me for TurntableCamera, and for Arcball it only resets the distance (not rotation)
* isosurface threshold can be set negative (should probably be limited here >= 0)
* scroll wheel with FlyCamera doesn't seem to work:
```
  File "/Users/larsoner/custombuilds/vispy/vispy/scene/cameras/cameras.py", line 1567, in viewbox_mouse_event
    PerspectiveCamera.viewbox_mouse_event(self, event)
  File "/Users/larsoner/custombuilds/vispy/vispy/scene/cameras/cameras.py", line 764, in viewbox_mouse_event
    if self._distance is not None:
```
These issues seem to also be in `master`, so I can open a new issue for those if you don't want to fix them here.

Eric89GXL:
Other than my minor issues with the example which can be addressed in a subsequent PR, LGTM +1 for merge. @almarklein do you have time to look?

campagnola:
Yeah, let's leave unrelated issues for later. This PR is starting to smell bad.

Eric89GXL:
I'll go ahead and merge this, @almarklein if you have time to look feel free to make comments and @campagnola can fix them in #895

Eric89GXL:
Thanks @campagnola

campagnola:
Hooray!

almarklein:
Thanks Luke!

PR #881, opened 2015-04-21
Text visual: changed shader and position to make it 3d compatible

gabyx:


vispy/visuals/text/text.py
```diff
@@ -438,7 +438,10 @@ def pos(self):
     @pos.setter
     def pos(self, pos):
         pos = [float(p) for p in pos]
-        assert len(pos) == 2
+        l = len(pos);
+        assert(l == 3 or l == 2)
```
Eric89GXL: `assert` statements are skipped when optimization is used, do a proper check and throw an error (see other instances in code base)

vispy/visuals/text/text.py
```diff
@@ -438,7 +438,10 @@ def pos(self):
...
+        if(l == 2):
+            pos.append(float(0.0));
```
Eric89GXL: This assumes `pos` is a list, which you don't know. It could be tuple, it could be an ndarray, etc. Here is how I would structure it:
```
pos = np.array(pos, float)
if pos.ndim != 1 or pos.size not in (2, 3):
    raise ValueError('pos must be array-like with 2 or 3 elements')
if pos.size == 2:
    pos = np.concatenate((pos, [0.]))
```

Eric89GXL:
Other than my minor comments, LGTM! The example is neat :)

Eric89GXL:
Travis is cranky about style:

https://travis-ci.org/vispy/vispy/jobs/59689509#L1489

Then we should be good to go. Thanks for working on this!

Eric89GXL:
Thanks for all the iterations @gabyx !

gabyx:
ah no worries :-)
its a pleasure to see how an opensource project on such a high level works :-)

just crazy =), all these webtools, checks, commitments from people like you and so on :-)


BR 


Am 24.04.2015 um 18:03 schrieb Eric Larson <notifications@github.com>:

PR #882, opened 2015-04-22
Explicit commit ID for image testing

campagnola:
Each commit explicitly references the test-data commit ID it should be tested against.
This should allow #870 to be merged without breaking other branches.

### DIFF
--- a/vispy/util/fetching.py
+++ b/vispy/util/fetching.py
@@ -17,6 +17,12 @@ from ..ext.six import string_types
 from ..util.config import config
 
 
+# This is the commit ID in the vispy/test-data repository that should be used
+# when doing image comparison tests. This ID must be updated as needed to 
+# follow changes to the test-data repository.
+test_data_commit_id = '27f2d5ab1cc20342d255694a5a21f35d2d7e376b'
+
+
 ###############################################################################
 # Vispy data directory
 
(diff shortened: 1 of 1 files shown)


vispy/util/fetching.py
```diff
@@ -90,7 +96,10 @@ def get_testing_file(fname, directory=None, force_download=False):
     fname : str
         The path to the file on the local system.
     """
-    _url_root = 'https://github.com/vispy/test-data/raw/master/'
+    global test_data_commit_id
```
Eric89GXL: Why does this need to be global? I don't see it get modified in here.

Eric89GXL:
What do you want this to do? Once a user has a testing file, it will only get updated if they force it to be updated. I think that is still the case here. I don't mind having to update images every once in a while, though -- is that what we'll have to do now?

vispy/util/fetching.py
campagnola: I just want it to be easy to find since it needs to be updated whenever we change the test-data repo.

campagnola:
All I want this to do is ensure that travis won't complain in unrelated PRs when I push new commits into the test-data repository. #870 will implement a slightly different strategy where it manages a git clone of test-data and checks out the necessary commit before testing.

vispy/util/fetching.py
Eric89GXL: I don't follow -- how does it make it easier to find if it's here?

Eric89GXL: It's already at the top of the file, and used two lines below...?

Eric89GXL:
Ahh, okay, cool

vispy/util/fetching.py
campagnola: Actually now that I think about it, this code will only be around for a short while in order to bootstrap the new image tests in place. #870 will remove it and replace with a git-based system.

Eric89GXL:
Sounds good, onto #870 then!

PR #889, opened 2015-05-03
Defer string formatting in log messages

### LINKED ISSUE #888: A recent commit caused a big framerate drop in 3d
The commit (found by bisection) is 6e82de1b3408555f6ce23b45b4568b66e2e86eb4. I'm not sure if it only affects 3d scenes, but that's the only place I've directly experienced it. After asking on gitter about it, I checked if it's fixed by #873, but it seems it isn't.

I first noticed this in one of my own scenes, with a small number of visuals, that previously could be updated at 60fps with no issues but following this commit takes well over a second to update under mouse interaction.

It's also detectable (but less obvious) in e.g. the examples/basics/visuals/tube.py example, though the framerate drop is not too large.

This can be demonstrated by directly reverting 6e82de1b3408555f6ce23b45b4568b66e2e86eb4 , after which the problem is resolved (others are reintroduced, including the breaking of depth-testing, but it demonstrates that this commit is where the issue directly appears).

campagnola:
Improves performance in many examples, hopefully fixes #888.
@inclement can you verify this helps?

### DIFF
--- a/vispy/visuals/shaders/function.py
+++ b/vispy/visuals/shaders/function.py
@@ -465,7 +465,7 @@ class Function(ShaderObject):
         if '$' in code:
             v = parsing.find_template_variables(code)
             logger.warning('Unsubstituted placeholders in code: %s\n'
-                           '  replacements made: %s' % 
+                           '  replacements made: %s', 
                            (v, list(self._expressions.keys())))
         
         return code + '\n'
--- a/vispy/visuals/shaders/function.py
+++ b/vispy/visuals/shaders/function.py
@@ -19,6 +19,7 @@ trigger a recompile.
 
 import re
 import logging
+import numpy as np
 
 from ...util.eq import eq
 from ...util import logger
@@ -238,14 +239,10 @@ class Function(ShaderObject):
             # try just updating its value.
             variable = storage.get(key, None)
             if isinstance(variable, Variable):
-                try:
+                if np.any(variable.value != val):
                     variable.value = val
                     self.changed(value_changed=True)
-                    return
-                except Exception:
-                    # Setting value on existing Variable failed for some
-                    # reason; will need to create a new Variable instead. 
-                    pass
+                return
             
             # Could not set variable.value directly; instead we will need
             # to create a new ShaderObject
--- a/vispy/visuals/shaders/program.py
+++ b/vispy/visuals/shaders/program.py
@@ -38,7 +38,7 @@ class ModularProgram(Program):
     @vert.setter
     def vert(self, vcode):
         if hasattr(self, '_vert') and self._vert is not None:
-            self._vert._dependents.remove(self)
+            self._vert._dependents.pop(self)
 
         self._vert = vcode
         if self._vert is None:
@@ -46,7 +46,7 @@ class ModularProgram(Program):
 
         vcode = preprocess(vcode)
         self._vert = MainFunction(vcode)
-        self._vert._dependents.add(self)
+        self._vert._dependents[self] = None
 
         self._need_build = True
         self.changed(code_changed=True, value_changed=False)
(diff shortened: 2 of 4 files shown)


inclement:
It's better, but still much slower than before - I guess an improvement from less than 1fps to more like 5-10fps.

campagnola:
Ok, I'll keep looking. It would be helpful if you can post a script that demonstrates the problem, since the performance hit is smaller in other examples..

inclement:
After pulling my mesh out into a standalone example, the performance is actually fine following your fix. My original code still gives a very slow scene, but I'll assume this is my fault somehow, I can't immediately work out what's causing it.

Eric89GXL:
Good to go then @campagnola? Thanks for the quick fix for this one.

inclement:
With further testing, I think the remaining issue I had is that the performance penalty for multiple Visuals in a scene has also increased significantly from what it was at e.g. 0f5eb3b48568ff129e9f4a24584a10abb6000269

inclement:
Here's an example that does demonstrate this performance change with multiple visuals. Performance is fine until the commit in my original issue, but poor afterwards. It is improved by this PR, but still nowhere near what it used to be.

    import sys
    from vispy import scene
    from vispy.geometry.torusknot import TorusKnot

    from colorsys import hsv_to_rgb
    import numpy as np

    points = TorusKnot(5, 3).first_component[:-1]
    points[:, 0] -= 20.
    points[:, 2] -= 15.

    colors = np.linspace(0, 1, len(points))
    colors = np.array([hsv_to_rgb(c, 1, 1) for c in colors])


    for _ in range(15):
        tube = scene.visuals.Tube(points,
                                shading='flat',
                                color=colors,
                                tube_points=8)
        points = points.copy() + 8.
        canvas.view.add(tube)

    canvas.view.camera = 'arcball'
    canvas.bgcolor = 'white'
    canvas.show()

    if __name__ == '__main__':
            canvas.app.run()

campagnola:
Thanks @inclement! I just added a couple more commits that bring the performance for your example back to previous levels (maybe a little faster). Can you verify?

Eric89GXL:
https://code.google.com/p/weakrefset/source/browse/weakrefset.py

You can add that to externals to get `WeakSet` in 2.6

inclement:
@campagnola It's a lot better, but actually I'm still finding it well below 60fps both for my example above and the other examples where I originally noticed it.

Eric89GXL:
@campagnola okay to merge this one for now? Maybe after this and the volume PR are merged @inclement you can take a look again?

inclement:
Sounds good to me - presumably it can still be improved, since it was faster before (so it must be vispy's end and not too many opengl contexts), but it's still a big improvement over master.

Eric89GXL:
@inclement do you have time to try out `vispy`'s profiling capabilities?

inclement:
@Eric89GXL At some point I will (but maybe not in the next few days). Actually I'm not sure what special profiling capabilities vispy has, is there something that can help with this?

Eric89GXL:
IIRC you can run a file with the command-line `--vispy-cprofile` to get a sense of what functions are eating time, and then you can use a more targeted approach after that. See discussion [here] (https://github.com/vispy/vispy/pull/860).

campagnola:
Let's go ahead and merge this. Here's what I did to test: 

1. Set your test script to generate 100 tube visuals to add a bit more stress
2. Made test script redraw as fast as possible by adding

```
t = app.Timer(interval=0.01, connect=canvas.update, start=True)
```

3. Ran with `--vispy-fps` to measure frame rate.
4. Got 11 fps after `git checkout 6e82de1~1`  (the commit immediately prior to the bug)
5. Got 0.1 fps after `git checkout 6e82de1`  (the commit that introduced the bug)
6. Got 11 fps from the tip of this PR.

So everything looks good on my end. @inclement if you get different results, then we might also have some platform-dependent issues going on.

Eric89GXL:
@inclement feel free to open (a) new issue(s) as performance problems crop up. We're gradually working to squash all those bugs, so your feedback (along with @SimonDanisch) should help us do that.

Issue #894, opened 2015-05-09
Optimize and complete the scenegraph


campagnola:
Thinking out loud: here is my plan for completing the scenegraph and taking care of performance issues.

The major issues right now are (1) multi-parenting in the scenegraph suggests an architecture where visuals do not know their transformations until draw time. Consequently, a lot of unnecessary work is done on every redraw. (2) The scenegraph needs picking and (3) some concept of a "scene" coordinate frame (mainly for lighting). 

- [x] **Visual views:** (#895) similar to @rougier's collection views, this would allow each visual to operate with multiple copies of its shader program, and each copy may have different transforms. This is meant to solve (1) while still allowing a visual to be drawn multiple times per frame. This will also allow creating extra views for exporting at different resolutions, views with picking filters applied, etc. 

    - [x] MultiProgram - Behaves like ModularProgram, but internally wraps multiple copies of the program. This will make it easier for visuals to handle multiple shader programs.
    - [x] New visual class organization: `BaseVisual` defines the basic API (draw, bounds, attach, etc.), `Visual` is anything that can be drawn with a single shader program (or MultiProgram), and `CompoundVisual` consists only of sub-Visuals with no program of its own. This organization will be necessary in order to implement the `BaseVisual` API in a generic way (otherwise each visual would have to reimplement the complete API).
    - [x] Remove multi-parenting from scenegraph and replace with the ability to automatically create a new view on an entire branch of the graph. This allows us to statically define the transform system used by each node in the graph, and only apply changes as they are needed. 
- [x] Scenegraph draw should only involve calling a cached list of functions:

    ```
    def draw_scene():
        if self.draw_order is None:
            self.generate_draw_order()
        for visual in self.draw_order:
            visual.draw()
    ```

- [x] Transforms need a `simplified` property that contains a collapsed (and cached) version of the transform (#832). To avoid re-simplifying chains every time a view changes, we can mark specific transforms as 'dynamic', which will prevent them being joined during simplification. For example, here is a typical transform chain:

    ``` 
    framebuffer -> canvas -> view_transform -> visual_transform1 -> visual_transform2
    ```

    If we collapse this chain to a single matrix transform, then the matrix needs to be recomputed from its constituent pieces every time the view transform is updated. Instead, we can mark the view transform as dynamic, which forces the simplified transform to look like:

    ```
    simplified_transform1 -> view_transform -> simplified_transform2
    ```

- [x] **Picking.** This requires drawing the scene with a constant, flat ID color for each visual.
[...]

PR #895, opened 2015-05-09
WIP: Visual views

campagnola:
This PR will implement part of #894:

- [x] Visual views
- [x] MultiProgram
- [x] Visual class organization: BaseVisual, Visual, VisualView, CompoundVisual
- [x] TransformSystem that responds to canvas resize / viewport / fbo
- [x] Preliminary VisualCollection class that lacks support for collectionizing transforms

@almarklein, @Eric89GXL have a look at examples/basics/visuals/visual_view.py and comment on the API.


Eric89GXL:
Should be good to rebase. I'll take a look later today or tomorrow.

campagnola:
Done.

Eric89GXL:
I like it. I would prefer something like this:
```
self.image.add_view()
views = self.image.views
```
so `self.image.views` is just a simple list, and it starts out with a single element (what is currently `image.default_view`, so you'd just use `image.views[0]` for that). It seems simpler than using multple attribute names.

campagnola:
Will add `visual.views` but I would rather the visual only keeps weak references to the views that are created so we don't have to worry about removing them.

Eric89GXL:
So it will keep a strong reference to `visual.default_view` but a weak one to all the others? That doesn't seem very good. What do you think will be the problem with strong refs?

campagnola:
The default view allows you to treat the visual as its own view (for example, call `visual.draw()`). That default view needs to last as long as the visual does, hence the strong ref. All others should be weak to avoid having to do manual garbage collection:

```
# create a temporary view for exporting at high res
view = visual.view()
view.draw()
. . .
visual.remove_view(...)
```

Eric89GXL:
The manual garbage collection doesn't sound so bad to me. But if we're going to organize it that way, then don't bother with `visual.views` since it doesn't really make sense to treat all views the same way.

campagnola:
This is ready for a preliminary review. All visuals are currently broken and there is a new example `examples/basics/visuals/line_prototype.py` that shows how I think visuals should be written and used. There are currently a few things not implemented: collection support, compound visuals, and some issues related to transform systems + canvas.

Previously the base definition of a `Visual` was left very thin--any class with `draw()` and `bounds()` methods. This PR is a major departure in that Visuals now have a much more rigid definition. The downside to this is that developers of new Visuals are forced to adhere to this framework, which might prove awkward or constraining in some cases (however I think all current visuals will fit into the framework nicely). The major benefit to having a rigid structure is that it allows us to automate features that previously had to be implemented individually for each visual (views and collections being two of the major targets here). It should also be the case that the current library of visuals can be reimplemented with many fewer lines of code, due to the use of shared machinery. 

I am requesting a review of the new system, especially focusing on the `line_prototype.py` example and `visuals/visual.py`. 

@almarklein @Eric89GXL @rossant @rougier

Eric89GXL:
So just to make sure I have a handle on this -- the idea is that we're considering trading off on some level of flexibility in the `Visuals` in order to gain:

1. simpler / more DRY code
2. easier integration of collections
3. all current visuals are expected to be adapted fairly straightforwardly under the new system
4. more efficiency beyond `collection` integration (perhaps? maybe transforms will work better?) 
5. easier to write future visuals (perhaps?)

+100 for me based on that, even if. I think if 5% of use cases become awkward or difficult the benefits are worth it. And even if points (4) and (5) that I hope for don't pan out, I'm still +25 :)

Now to actually look at the changes...

examples/basics/visuals/line_prototype.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(self, pos=None, color=(1, 1, 1, 1)):
+        LineVisual.__init__(self, pos, color)
+        self._draw_mode = 'points'
+    
+
+class PlotLineVisual(visuals.CompoundVisual):
+    def __init__(self, pos=None, line_color=(1, 1, 1, 1), point_color=(1, 1, 1, 1)):
+        self._line = LineVisual(pos, color=line_color)
+        self._point = PointVisual(pos, color=point_color)
+        visuals.CompoundVisual.__init__(self, [self._line, self._point])
```
Eric89GXL: So a `CompoundVisual` allows creating a new visual class by combining other existing `Visual` classes. And adding this additional class allows us to be smarter about dealing with transforms, buying efficiency?

Eric89GXL:
I'll find the time this weekend to digest the code more closely. In the meantime if you have a few spare minutes, a few sentences about the purpose of the new classes would be good (can be in the PR as comments to save some legwork eventually). Otherwise I'll try to infer it from the code directly.

campagnola:
Comments added.

Eric89GXL:
Comments LGTM. Is the list of advantages I made above correct, including the more speculative points 4 and 5?

campagnola:
(1) True

I hope (2) is true, but that remains to be seen. There is a lot of overlap between the visual and collection classes; some of that will need to be disentangled.

(3) True

(4) Very true. One of the major issues this solves is the ability to have multiple programs (one per view) for a visual rather than modifying (and sometimes recompiling) the same program in between each draw. A deeper consequence of this is that each view knows it has exactly one program, so we can remove from `draw()` a lot of checks that are no longer necessary. We can also remove the multiple-parent capability from the scenegraph, which will greatly simplify a lot of the code and effort required when drawing.

(5) True

Eric89GXL:
Sounds great to me. @almarklein @rossant @rougier do you have time to take a look and comment?

Eric89GXL:
@campagnola would it be good to review and merge this since it appears to be done? If so, give it a rebase and I'll take a closer look.

campagnola:
No, this PR can't be merged because it breaks pretty much everything  :)
#928 supersedes it and will include all of the necessary visual/scenegraph updates to get everything working again. This PR is kept separate (and open) in order to allow others to review the basic changes to the visual system more easily.

Eric89GXL:
Ahh, right. Makes sense.

vispy/visuals/visual.py
```diff
@@ -3,52 +3,164 @@
...
+    * Minimal framework for creating views of the visual.
+    * A data structure that is shared between all views of the visual.
+    * Abstract `draw`, `bounds`, `attach`, and `detach` methods.
+    
     """
-
-    def __init__(self):
-        self._visible = True
+    def __init__(self, vshare=None, key=None):
+        self._view_class = getattr(self, '_view_class', VisualView)
```
almarklein: key and vshare arguments must probably described in the docstring

vispy/visuals/visual.py
```diff
@@ -3,52 +3,164 @@
...
+        return self.transforms.visual_to_document
+    
+    @transform.setter
+    def transform(self, tr):
+        self.transforms.visual_transform = tr
+
+    def view(self, key=None):
+        """Return a new view of this visual.
+        """
+        if key is None:
```
almarklein: Again, describe purpose of `key`, that while loop below scares me a bit :)

almarklein:
Can you maybe provide a class diagram? E.g. I was somewhat surprised to see VisualView inherting from Visual.

almarklein:
Though in general I think this seems a very good alternative to have multiple inheritance (and other features) in what seems to become a simpler system.

vispy/visuals/visual.py
campagnola: These are private arguments used when creating views. I have renamed them to have underscores.

vispy/visuals/visual.py
campagnola: The keys were just intended as unique names for debugging purposes. They ended up being useless so I have removed them :)

campagnola:
@almarklein, I added more documentation to the top of `visual.py`. 

>  I was somewhat surprised to see VisualView inherting from Visual.

This is because a `Visual` is defined as having a single shader program, and implements methods for dealing with that program. A `VisualView` is the same--it must have its own program, and generally operates on that program in exactly the same way that the `Visual` does. It should also make sense that, from the user's perspective, a `VisualView` can be used in all the same ways that a `Visual` can be used.

```
# works both for View and VisualView
if isinstance(obj, Visual):
    obj.draw()
```

almarklein:
Aha, thanks for the clarification. Would it make sense to instead have one `Visual` class that has its own transform and filters, and each such `Visual` having an `InternalVisual` that contains the actual data, etc. Multiple visuals can then share one `InternalVisual`. This would be somewhat as if each Visual is a `VisualView`, and the actual `Visual` instance is more hidden.

campagnola:
That's close to how it works--the `VisualShare` class is meant to contain all of the data that is shared between the views of a visual. This is really just a shared data structure, though, to avoid having to create  a new subclass of VisualShare for every subclass of Visual. I went through several different class structures before finally settling on this one. The important constraints were:

1. User can instantiate a Visual and use it directly without ever thinking about views. 
1. Relatively simple to make a new Visual class.
1. No extra effort needed for creating views on new Visual classes.
1. Views can be used just like other visuals wrt transforms and filters. 

If you'd like to take a stab at a class structure that makes more sense, then I'd be happy to see what you come up with. I'll bet there is a better solution out there..

almarklein:
I think what I mean is that (sticking with your terminology), the user always instantiates a `VisualView` (and the real underlying `Visual` is automatically created the first time):

    # Your current approach
    line1 = LineVisual()
    line2 = line1.view()
    # line1 and line2 are not of the same class
    # Question: is line2 a VisualView or a LineVisualView of some sort?
  
    # What I am proposing
    line1 = LineVisualView()  # although if we go this way, we should obviously rename it
    line2 = line1.view()
    # Both objects are of the same class and are a view on the same underlying Visual

My (minor) issues with the current approach is that I find the class hierarchy a bit complex, I don't like mixins so much, and I slightly dislike that line1 and line2 are not equals. Although with what I am proposing you'd need two classes (`SomethingVisual` and `SomethingVisualView`) for each type of visual (thought the view can probably be defined in 2 lines or perhaps autogenerated).

I'm not nearly as much into this code as you are, so I leave it to your judgement what would work best :)

campagnola:
I agree that it would be nice to have the class symmetry. One of the versions I tried writing looked more like what you describe.

> Although with what I am proposing you'd need two classes (SomethingVisual and SomethingVisualView) for each type of visual

That's the main reason I ended up abandoning it.

> thought the view can probably be defined in 2 lines or perhaps autogenerated

That's a possibility. I'll give it another try.

Eric89GXL:
Closing this one in favor of new PR

Issue #900, opened 2015-05-12
"edge_color" parameter of "CubeVisual" and array_like input support.


KelSolaar:
I'll describe first the context of this issue: In the long term I would like to provide [Colour](https://github.com/colour-science/colour) with alternative 3d plotting capabilities and be able to produce plots such as:

![image](https://cloud.githubusercontent.com/assets/99779/7582162/41dddc8a-f88b-11e4-9c56-4d6a2268639e.png)

The main reason for this is that [Matplotlib](http://matplotlib.org) is not really suited for heavy 3d plotting, there are massive z ordering issues and it is also quite slow.

Back to the issue itself, in order to have varying coloured wireframes I would like the `vispy.visuals.cube.CubeVisual` `edge_color` parameter to support array_like input. I have a [patch](https://github.com/colour-science/vispy/commit/8148e4c70b763cf5cba9f07e42bb7a8e32206fe0) to allow this behaviour in our [fork](https://github.com/colour-science/vispy/commit/8148e4c70b763cf5cba9f07e42bb7a8e32206fe0).

Given the following code...

```python
import sys

import numpy as np
from vispy import scene
from vispy.geometry import create_cube
from vispy.scene.visuals import Cube


canvas = scene.SceneCanvas(keys='interactive',
                           size=(800, 600),
                           show=True,
                           bgcolor=(0.5, 0.5, 0.5))

view = canvas.central_widget.add_view()

vertices, _, _ = create_cube()

RGB_f = np.copy(vertices['color'])
RGB_f[..., 3] *= 0.1
RGB_e = vertices['color']

cube1 = Cube(vertex_colors=RGB_f,
             edge_color=RGB_e,  # Relevant statement for the issue.
             parent=view.scene)
vertices = cube1.mesh_data.get_vertices()
vertices[..., 0] -= 1.5
cube1.mesh_data.set_vertices(vertices)

cube2 = Cube(vertex_colors=RGB_f,
             edge_color='k',  # Relevant statement for the issue.
             parent=view.scene)
vertices = cube2.mesh_data.get_vertices()
vertices[..., 0] += 1.5
cube2.mesh_data.set_vertices(vertices)

view.camera = 'turntable'
view.camera.fov = 45

if __name__ == '__main__' and sys.flags.interactive == 0:
    canvas.app.run()
```

... I can render this:

![image](https://cloud.githubusercontent.com/assets/99779/7582550/0ac5e3f6-f88f-11e4-830b-842a5ea454cb.png)

[...]

PR #901, opened 2015-05-13
PR: Support for parametric plane and box geometries.

KelSolaar:
This PR implements support for a parametric plane and a parametric box that builds on it through the following objects:

- `vispy.geometry.generation.create_plane`

image

- `vispy.geometry.generation.create_box`

image

Some things to consider:

- It would be neat to actually return a `MeshData` instance instead of the current tuple for `create_cube`, `create_plane` and `create_box` definitions, however the class may not have all the needed fields, especially one for the outline. Upon really quick inspection it doesn't seem to support arbitrary data.
- As I didn't wanted to break anything, I did not replaced `make_cube`, it is maybe worth for it to be an alias to `make_box` or whatever you feel the best.
- I haven't created matching visuals for the new geometries, but I have some that fit my usage accordingly to #900 there: https://github.com/colour-science/colour-vispy/blob/master/colour_vispy/visuals.py

### DIFF
--- a/examples/basics/visuals/box.py
+++ b/examples/basics/visuals/box.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (c) 2014, Vispy Development Team.
+# Copyright (c) 2015, Vispy Development Team.
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 
 """
@@ -15,9 +15,6 @@ canvas = scene.SceneCanvas(keys='interactive', size=(800, 600), show=True)
 
 view = canvas.central_widget.add_view()
 
-camera = scene.cameras.TurntableCamera(fov=45, parent=view.scene)
-view.camera = camera
-
 vertices, faces, outline = geometry.create_box(width=2, height=4, depth=8,
                                                width_segments=4,
                                                height_segments=8,
@@ -28,6 +25,9 @@ box = scene.visuals.Box(width=4, height=4, depth=8, width_segments=4,
                         vertex_colors=vertices['color'],
                         edge_color='k',
                         parent=view.scene)
+
+camera = scene.cameras.TurntableCamera(fov=45, azimuth=60, parent=view.scene)
+view.camera = camera
 
 if __name__ == '__main__' and sys.flags.interactive == 0:
     canvas.app.run()
--- a/examples/basics/visuals/box.py
+++ b/examples/basics/visuals/box.py
@@ -9,8 +9,7 @@ Simple demonstration of the box geometry.
 import sys
 
 from vispy import scene
-from vispy import visuals
-from vispy.geometry import create_box
+from vispy import geometry
 
 canvas = scene.SceneCanvas(keys='interactive', size=(800, 600), show=True)
 
@@ -19,45 +18,16 @@ view = canvas.central_widget.add_view()
 camera = scene.cameras.TurntableCamera(fov=45, parent=view.scene)
 view.camera = camera
 
-vertices, faces, outline = create_box(width=1,
-                                      height=2,
-                                      depth=3,
-                                      width_segments=4,
-                                      height_segments=8,
-                                      depth_segments=12)
+vertices, faces, outline = geometry.create_box(width=2, height=4, depth=8,
+                                               width_segments=4,
+                                               height_segments=8,
+                                               depth_segments=16)
 
-box_f = scene.Mesh(vertices=vertices['position'],
-                   faces=faces,
-                   vertex_colors=vertices['color'],
-                   parent=view.scene)
-
-box_w1 = scene.Mesh(vertices=vertices['position'],
-                    faces=outline,
-                    vertex_colors=vertices['color'],
-                    mode='lines',
-                    parent=view.scene)
-
-transform = visuals.transforms.AffineTransform()
-transform.translate((0, 0, 2.5))
-box_w1.transform = transform
-
-vertices, faces, outline = create_box(width=1,
-                                      height=2,
-                                      depth=3,
-                                      width_segments=4,
-                                      height_segments=8,
-                                      depth_segments=12,
-                                      planes=('-x', '-y', '-z'))
-
-box_w2 = scene.Mesh(vertices=vertices['position'],
-                    faces=outline,
-                    vertex_colors=vertices['color'],
-                    mode='lines',
-                    parent=view.scene)
-
-transform = visuals.transforms.AffineTransform()
-transform.translate((0, 3.5, 0))
-box_w2.transform = transform
+box = scene.visuals.Box(width=4, height=4, depth=8, width_segments=4,
+                        height_segments=8, depth_segments=16,
+                        vertex_colors=vertices['color'],
+                        edge_color='k',
+                        parent=view.scene)
 
 if __name__ == '__main__' and sys.flags.interactive == 0:
     canvas.app.run()
--- a/examples/basics/visuals/box.py
+++ b/examples/basics/visuals/box.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2014, Vispy Development Team.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+
+"""
+Simple demonstration of the box geometry.
+"""
+
+import sys
+
+from vispy import scene
+from vispy import visuals
+from vispy.geometry import create_box
+
+canvas = scene.SceneCanvas(keys='interactive', size=(800, 600), show=True)
+
+view = canvas.central_widget.add_view()
+
+camera = scene.cameras.TurntableCamera(fov=45, parent=view.scene)
+view.camera = camera
+
+vertices, faces, outline = create_box(width=1,
+                                      height=2,
+                                      depth=3,
+                                      width_segments=4,
+                                      height_segments=8,
+                                      depth_segments=12)
+
+box_f = scene.Mesh(vertices=vertices['position'],
+                   faces=faces,
+                   vertex_colors=vertices['color'],
+                   parent=view.scene)
+
+box_w1 = scene.Mesh(vertices=vertices['position'],
+                    faces=outline,
+                    vertex_colors=vertices['color'],
+                    mode='lines',
+                    parent=view.scene)
+
+transform = visuals.transforms.AffineTransform()
+transform.translate((0, 0, 2.5))
+box_w1.transform = transform
+
+vertices, faces, outline = create_box(width=1,
+                                      height=2,
+                                      depth=3,
+                                      width_segments=4,
+                                      height_segments=8,
+                                      depth_segments=12,
+                                      planes=('-x', '-y', '-z'))
+
+box_w2 = scene.Mesh(vertices=vertices['position'],
+                    faces=outline,
+                    vertex_colors=vertices['color'],
+                    mode='lines',
+                    parent=view.scene)
+
+transform = visuals.transforms.AffineTransform()
+transform.translate((0, 3.5, 0))
+box_w2.transform = transform
+
+if __name__ == '__main__' and sys.flags.interactive == 0:
+    canvas.app.run()
--- a/examples/basics/visuals/plane.py
+++ b/examples/basics/visuals/plane.py
@@ -9,8 +9,7 @@ Simple demonstration of the plane geometry.
 import sys
 
 from vispy import scene
-from vispy import visuals
-from vispy.geometry import create_plane
+from vispy import geometry
 
 canvas = scene.SceneCanvas(keys='interactive', size=(800, 600), show=True)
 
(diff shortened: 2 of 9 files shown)


vispy/geometry/generation.py
```diff
@@ -89,6 +89,250 @@ def create_cube():
...
+        Array of vertices suitable for use as a VertexBuffer.
+    faces : array
+        Indices to use to produce a filled plane.
+    outline : array
+        Indices to use to produce an outline of the plane.
+
+    References
+    ----------
+    [1] Cabello, R. (n.d.). PlaneBufferGeometry.js. Retrieved May 12, 2015,
+        from https://github.com/mrdoob/three.js/blob/master/src/extras/geometries/PlaneBufferGeometry.js  #noqa
```
Eric89GXL: I think the `# noqa` actually needs to go after the end `"""`

vispy/geometry/generation.py
```diff
@@ -89,6 +89,250 @@ def create_cube():
     return vertices, filled, outline
 
 
+def create_plane(width=1,
+                 height=1,
+                 width_segments=1,
+                 height_segments=1,
+                 direction='+z'):
```
Eric89GXL: we try to fit as many of these on the same line as possible while still being PEP8

vispy/geometry/generation.py
```diff
@@ -89,6 +89,250 @@ def create_cube():
...
+    vertices = np.zeros(positions.shape[0],
+                        [('position', np.float32, 3),
+                         ('texcoord', np.float32, 2),
+                         ('normal', np.float32, 3),
+                         ('color', np.float32, 4)])
+
+    vertices['position'] = positions
+    vertices['texcoord'] = texcoords
+    vertices['normal'] = normals
+    vertices['color'] = colours
```
Eric89GXL: we also use the American spelling of colors

vispy/geometry/generation.py
```diff
@@ -89,6 +89,250 @@ def create_cube():
...
+               depth=1,
+               width_segments=1,
+               height_segments=1,
+               depth_segments=1,
+               planes=None):
+    """ Generate vertices & indices for a filled and outlined box.
+
+    Parameters
+    ----------
+    width : numeric
```
Eric89GXL: s/numeric/float/g

vispy/geometry/generation.py
```diff
@@ -89,6 +89,250 @@ def create_cube():
...
+    outline = np.zeros((0, 2), dtype=np.uint32)
+
+    offset = 0
+    for vertices_p, faces_p, outline_p in planes_m:
+        positions = np.vstack((positions, vertices_p['position']))
+        texcoords = np.vstack((texcoords, vertices_p['texcoord']))
+        normals = np.vstack((normals, vertices_p['normal']))
+
+        faces = np.vstack((faces, faces_p + offset))
+        outline = np.vstack((outline, outline_p + offset))
```
Eric89GXL: using `vstack` in this way will cause a bunch of unnecessary memory reallocations, it would be better to just make a list of everything and call `np.concatenate` once -- or better, preallocate the proper array size and populate (but this is probably overkill)

vispy/geometry/generation.py
```diff
@@ -89,6 +89,250 @@ def create_cube():
...
+    if '-x' in planes:
+        planes_m.append(create_plane(depth, height, d_s, h_s, '-x'))
+        planes_m[-1][0]['position'][..., 0] -= width / 2
+    if '+x' in planes:
+        planes_m.append(create_plane(depth, height, d_s, h_s, '+x'))
+        planes_m[-1][0]['position'][..., 0] += width / 2
+
+    positions = np.zeros((0, 3))
+    texcoords = np.zeros((0, 2))
+    normals = np.zeros((0, 3))
```
Eric89GXL: it's also better to use `np.float32` for these

examples/basics/visuals/plane.py
```diff
@@ -0,0 +1,60 @@
...
+                      vertex_colors=vertices['color'],
+                      mode='lines',
+                      parent=view.scene)
+
+transform = visuals.transforms.AffineTransform()
+transform.translate((4.5, 0, 0))
+plane_w2.transform = transform
+
+if __name__ == '__main__' and sys.flags.interactive == 0:
+    canvas.app.run()
```
Eric89GXL: these examples should probably only go in `basics/visuals/` if they actually demonstrate a new visual. It's probably worth actually adding your visuals in this PR.

Eric89GXL:
Looks good so far. WDYT about adding your visuals to this PR?

vispy/geometry/generation.py
QuLogic: In fact, `planes_m` is already a list of arrays, so why not just `np.concatenate` the entire thing?

vispy/geometry/generation.py
```diff
@@ -89,6 +89,250 @@ def create_cube():
...
+    x_grid1 = x_grid + 1
+    y_grid1 = y_grid + 1
+
+    # Positions, normals and texcoords.
+    positions = np.zeros(x_grid1 * y_grid1 * 3)
+    normals = np.zeros(x_grid1 * y_grid1 * 3)
+    texcoords = np.zeros(x_grid1 * y_grid1 * 2)
+
+    offset = offset1 = 0
+    for i_y in range(y_grid1):
```
QuLogic: I think these loops can be cut down a bit:
```
y = np.arange(y_grid1) * height / y_grid - height / 2
x = np.arange(x_grid1) * width / x_grid - width / 2

positions[::3] = np.tile(x, y_grid1)
positions[1::3] = - np.repeat(y, x_grid1)

normals[2::3] = 1

texcoords[::2] = np.tile(np.arange(x_grid1) / x_grid, y_grid1)
texcoords[1::2] = np.repeat(1 - np.arange(y_grid1) / y_grid, x_grid1)
```
(I may have messed up a few of the limits.)

vispy/geometry/generation.py
```diff
@@ -89,6 +89,250 @@ def create_cube():
...
+            outline[offset1 + 4] = c
+            outline[offset1 + 5] = d
+
+            outline[offset1 + 6] = d
+            outline[offset1 + 7] = a
+
+            offset += 6
+            offset1 += 8
+
+    positions = np.reshape(positions, (-1, 3))
```
QuLogic: Wait, if you were going to do this anyway, why not just allocate it with that shape in the first place instead of messing around with offsets?

vispy/geometry/generation.py
KelSolaar: I need to double check but I think it is fine this way, just on the offending line.

vispy/geometry/generation.py
KelSolaar: Sure sounds good!

vispy/geometry/generation.py
KelSolaar: Ok with that!

vispy/geometry/generation.py
KelSolaar: I'm not sure to understand this one.

vispy/geometry/generation.py
KelSolaar: I need to check but I think the original reason I did it this way was because I had to offset the faces / outline indexes for each plane: `faces = np.vstack((faces, faces_p + offset))`

I can probably offset the indexes first and do a concatenation for everything at once after.

vispy/geometry/generation.py
KelSolaar: You mean specifying the `dtype` here: `positions = np.zeros((0, 3))`

examples/basics/visuals/plane.py
KelSolaar: Sounds good!

vispy/geometry/generation.py
Eric89GXL: substitute `float` for `numeric`

vispy/geometry/generation.py
KelSolaar: It is almost a straight port of [three.js](https://github.com/mrdoob/three.js/blob/master/src/extras/geometries/PlaneBufferGeometry.js) code, as it was working fine, I didn't really looked at vectorising it. If you think it is worth it, we can probably give it a go.

vispy/geometry/generation.py
KelSolaar: Oh right that was a regex above :)

vispy/geometry/generation.py
KelSolaar: Gave it a whirl but I didn't really managed to get anything elegant.

Eric89GXL: fair enough, just put in a comment like `# XXX this could probably be done without reallocations`

KelSolaar:
I didn't managed to find an elegant way to replace the loop stacking the planes data together in `vispy.vispy.geometry.generation.create_box` definition.

I added the `vispy.vispy.visuals.plane.PlaneVisual` and `vispy.vispy.visuals.plane.BoxVisual` visuals based on the `CubeVisual` although I'm not happy with them because of the fact I'm unable to use colours on the edges. I would rather prefer having a visual where I can just toggle the `triangle` or `line` mode with full control, than one like the `CubeVisual` which embeds 2 visuals.

The code is now also almost the same for the `PlaneVisual`, `BoxVisual` and `CubeVisual` which just begs for refactoring.

I actually ended up doing something like this in my code, I have generic `PrimitiveVisual` that inherits from `MeshVisual` and use the data from the various `create_*` definitions: https://github.com/colour-science/colour-analysis/blob/master/colour_analysis/visuals.py#L28

It would probably more elegant that those `create_*` definitions return a `MeshData` instance but in order to we need to have support for the outline, maybe in a kind of user attribute (`MeshData.user_data`).

Eric89GXL:
Hmmm... it does seem like it would be nice to be able to set whether you want to draw triangles, lines, or both. I wonder if it makes sense to add to `MeshVisual` the ability to draw the edge as well as the mesh itself. Would that work for your use case?

KelSolaar:
@Eric89GXL Provided you have full control of the colour of triangles / lines it would be perfect for my usage. It is essentially what I'm doing.

Eric89GXL:
@campagnola WDYT about modifying `MeshVisual` so that the edges can be drawn?

Eric89GXL:
(as well as the triangles themselves)

campagnola:
I think that was originally part of the plan, but it's also helpful if we limit visuals to either drawing from a single shader program, or drawing entirely from sub-visuals (like PolygonVisual does). So perhaps we should keep MeshVisual triangles-only and make a new visual for combining triangles+lines?  (or maybe rename MeshVisual to TrianglesVisual and create a new MeshVisual?)

Eric89GXL:
+1 for renaming `MeshVisual` to `TrianglesVisual`... but we don't want to break people's code. Maybe the new `MeshVisual` could act like `TrianglesVisual` by default and thus be backward-compatible?

campagnola:
That sounds good to me.

Eric89GXL:
@KelSolaar do you want to give this a shot? It sounds like you might already have the code to do this. If not, again I can do it, but you might have to wait a few days :)

KelSolaar:
I don't mind waiting a few days, I'm so swamped that I barely manage to code my own stuff ;)

Eric89GXL:
@KelSolaar for now, should we move forward with this PR so we can get your Visuals into `master`? Then later I / you can refactor with `TrianglesVisual` / `MeshVisual`.

KelSolaar:
Sounds good!

KelSolaar:
I fixed the latest *flake8* issues. I think the `TrianglesVisual` / `MeshVisual` stuff should go with another issue / PR.

examples/basics/visuals/box.py
```diff
@@ -0,0 +1,33 @@
...
+                                               depth_segments=16)
+
+box = scene.visuals.Box(width=4, height=4, depth=8, width_segments=4,
+                        height_segments=8, depth_segments=16,
+                        vertex_colors=vertices['color'],
+                        edge_color='k',
+                        parent=view.scene)
+
+if __name__ == '__main__' and sys.flags.interactive == 0:
+    canvas.app.run()
```
Eric89GXL: When I run this example I get a box that is zoomed in too far. Can you fix it? You should be able to tell the camera to set the view to fit all objects.

vispy/geometry/__init__.py
```diff
@@ -19,5 +19,6 @@
 from .torusknot import TorusKnot  # noqa
 from .calculations import (_calculate_normals, _fast_cross_3d,  # noqa
                            resize)  # noqa
-from .generation import create_arrow, create_cone, create_cube, \
-           create_cylinder, create_sphere  # noqa
+from .generation import (create_arrow, create_box, create_cone,  # noqa
+                         create_cube, create_cylinder, create_plane,  # noqa
+                         create_sphere)  # noqa
```
Eric89GXL: when adding things to `__init__` you need to also update `__all__` above, as it's used to set the order when creating docs and to clean up imports

vispy/geometry/generation.py
```diff
@@ -89,6 +89,214 @@ def create_cube():
...
+        Array of vertices suitable for use as a VertexBuffer.
+    faces : array
+        Indices to use to produce a filled plane.
+    outline : array
+        Indices to use to produce an outline of the plane.
+
+    References
+    ----------
+    [1] Cabello, R. (n.d.). PlaneBufferGeometry.js. Retrieved May 12, 2015,
+        from http://git.io/vU1Fh
```
Eric89GXL: Can you follow the `numpy` way, "references" section number 11 here:

https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections

Basically you'd add a note above, proabably after the 1-line function description, saying it is derived from MIT-licensed code?

vispy/visuals/box.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014, Vispy Development Team. All Rights Reserved.
```
Eric89GXL: I think I've changed all of these to 2015s in the rest of the code base now, please change in any files you've added in this PR

Eric89GXL:
Other than my minor gripes, LGTM! Thanks for iterating on this a couple of times.

How is the actual use case coming along? Hopefully once @campagnola is done with his scenegraph overhaul #928 you can take a look and see if you see some performance benefits.

vispy/geometry/__init__.py
KelSolaar: Sure, I completely forgot.

vispy/geometry/generation.py
KelSolaar: Well, I'm actually following it (I omitted the `..` by mistake, my bad): https://github.com/numpy/numpy/blob/master/numpy/core/fromnumeric.py#L2719

That being said Numpy doesn't seem to have a standard / common citation style. I used APA here (like in our API,  https://github.com/colour-science/colour/issues/135) which is quite common in the scientific community.

We do use a citation manager (Mendeley) and export everything from it for each release: https://github.com/colour-science/colour/blob/develop/BIBLIOGRAPHY.bib along a scrapping from all the API references: https://github.com/colour-science/colour/blob/develop/BIBLIOGRAPHY.rst

Now for the MIT thing, I don't remember it is something you find in citations but it can definitely be included.

Just let me know what do you want.

KelSolaar:
You are welcome! :)

As for the use case, it is slowly turning into something quite cool :) There are a lot of small annoying issues here and there but I'm finding a way through them *most* of the time.

So far it looks like that:

image

Really annoying things for me are the transparency handling and the lack of AA. The transparency issue forces me to layer the visuals in a precise order so that I mitigate the problems but it is far from being perfect.

vispy/geometry/generation.py
Eric89GXL: I don't actually care about the actual citation style, just the docstring
style (the ..) so Sphinx sets up links properly. :)

Eric89GXL:
I'll give one more quick run through the examples tomorrow morning then merge, thanks for the quick turnaround @KelSolaar. Hopefully we can fix the transparency issues this summer.

Eric89GXL:
Awesome, thanks @KelSolaar!

PR #909, opened 2015-05-17
Windows end-user doesn't have rights to symlink

stonebig:


vispy/app/backends/_ipynb_webgl.py
```diff
@@ -61,7 +62,7 @@ def _prepare_js(force=False):
     else:
         kwargs = {}
     install_nbextension(jsdir, overwrite=force, destination='vispy',
-                        symlink=True, **kwargs)
+                        symlink=(False if os.name == 'nt' else True), **kwargs)
```
QuLogic: `symlink=(os.name != 'nt')`?

Eric89GXL: +1, simpler solution

Eric89GXL:
Thanks @stonebig!

Issue #271, opened 2014-06-14
add `on_click` and `on_dlbclick` in canvas events


rossant:

Eric89GXL:
@nippoo can you add this now that you've added the `doubleclick` event?

PR #910, opened 2015-05-18
Double-click

nippoo:
- [x] Natively pass-through double-click on backends which support it...
- [x] and 'fake it' on those that don't.

### DIFF
--- a/vispy/app/backends/_template.py
+++ b/vispy/app/backends/_template.py
@@ -129,7 +129,7 @@ class CanvasBackend(BaseCanvasBackend):
         self._vispy_canvas.events.mouse_release(pos=(x, y), button=1,
                                                 modifiers=())
         self._vispy_canvas.events.mouse_double_click(pos=(x, y), button=1,
-                                                modifiers=())
+                                                     modifiers=())
         self._vispy_canvas.events.mouse_move(pos=(x, y), modifiers=())
         self._vispy_canvas.events.mouse_wheel(pos=(x, y), delta=(0, 0),
                                               modifiers=())
--- a/vispy/app/backends/_template.py
+++ b/vispy/app/backends/_template.py
@@ -128,6 +128,8 @@ class CanvasBackend(BaseCanvasBackend):
                                               modifiers=())
         self._vispy_canvas.events.mouse_release(pos=(x, y), button=1,
                                                 modifiers=())
+        self._vispy_canvas.events.mouse_double_click(pos=(x, y), button=1,
+                                                modifiers=())
         self._vispy_canvas.events.mouse_move(pos=(x, y), modifiers=())
         self._vispy_canvas.events.mouse_wheel(pos=(x, y), delta=(0, 0),
                                               modifiers=())
--- a/vispy/app/canvas.py
+++ b/vispy/app/canvas.py
@@ -494,7 +494,8 @@ class MouseEvent(Event):
     """Mouse event class
 
     Note that each event object has an attribute for each of the input
-    arguments listed below.
+    arguments listed below, as well as a "time" attribute with the event's
+    precision start time.
 
     Parameters
     ----------
@@ -539,6 +540,7 @@ class MouseEvent(Event):
         self._delta = np.zeros(2) if (delta is None) else np.array(delta)
         self._last_event = last_event
         self._press_event = press_event
+        self._time = time()
 
     @property
     def pos(self):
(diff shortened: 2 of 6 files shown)


Eric89GXL:
Related PR, at least WX should support the event natively:

https://github.com/vispy/vispy/pull/777

Any chance you can also make a `on_click` event as part of this? See:

https://github.com/vispy/vispy/issues/271

nippoo:
I actually made an `on_click` event in https://github.com/nippoo/vispy/commit/da0b90a38ccea63bf42007d8ba9bb14c6922f5b7 but got rid of it when I realised Qt didn't support it! I'll add that if it's something people are interested in, but it's important to realise its limitations where backends don't support - a `click` on a button, for example, should be triggered when the `mouse_press` and `mouse_release` are both on the button.

This is, AFAICT, pretty much impossible if the backend isn't aware of where your button (or square/shape/polygon) is on the canvas, if that makes any sense... so it won't be perfect - and it might be better to use `mouse_release` or `mouse_press` in most circumstances...

rossant:
AFAIC I'd need dbl click only. For click, using `mouse_press` is sufficient, especially when you have to press a key on the keyboard at the same time (like `ctrl+click`)

campagnola:
Agreed with @nippoo, the app system should stick to press/release/double and we can implement a click event in the scenegraph later on if needed.

nippoo:
I think I'll just implement double-click here then - click will be great for certain things but I'll (or someone else can!) do that later...

Eric89GXL:
Fine by me, thanks for tackling this @nippoo

nippoo:
Okay - so, assuming Travis-Veyor passes, this is good to merge from my end. I've tested it on Qt and GLFW, which are the only two I've got working here... and happy to entertain any suggestions about a nicer way than hard-coding a dt of 0.3 seconds in the code (though it seems a sensible default)

Eric89GXL:
[Stackoverflow suggests] (http://ux.stackexchange.com/questions/40364/what-is-the-expected-timeframe-of-a-double-click) that closer to 500 ms is a safer option, but I'm fine with 0.3 for now

vispy/app/backends/_template.py
```diff
@@ -128,6 +128,8 @@ class CanvasBackend(BaseCanvasBackend):
                                               modifiers=())
         self._vispy_canvas.events.mouse_release(pos=(x, y), button=1,
                                                 modifiers=())
+        self._vispy_canvas.events.mouse_double_click(pos=(x, y), button=1,
+                                                modifiers=())
```
Eric89GXL: boooo docstring has bad PEP8 :)

vispy/app/canvas.py
```diff
@@ -98,6 +99,13 @@ class Canvas(object):
         * touch
         * close
 
+    The ordering of the mouse_double_click, mouse_press, and mouse_release
+    events are not guaranteed to be consistent between backends. Only certain
+    backends natively support double-clicking (currently Qt and WX); on other
+    backends, they are detected manually with a fixed time delay.
+    This can cause problems with accessibility, as increasing the OS detection
+    time or using a dedicated double-click button will not be respected.
```
Eric89GXL: At some point we might be able to query this info from osx/windows/linux... it looks doable for [Windows] (https://msdn.microsoft.com/en-us/library/windows/desktop/ms646258%28v=vs.85%29.aspx) and [OSX] (https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSEvent_Class/#//apple_ref/occ/clm/NSEvent/doubleClickInterval), Linux is probably harder. I'm going to assume YAGNI for now :)

nippoo: Your assumption is correct! I wonder how many people use all the other random backends on Windows/Mac anyway?! And my mouse has a fourth button which is mapped directly to a double-click action, which I presume won't work.

nippoo:
[Pyglet](https://www.assembla.com/code/pywiic/subversion/nodes/trunk/lib/pyglet/text/caret.py?html_id=85&load_comments=true&rev=31#ln527) has 250ms, which is where I got my figure from. I've no strong views - I suppose there's no disadvantage in a longer time!

vispy/app/backends/_template.py
nippoo: Fixed to prevent your sleepless nights. ;)

Eric89GXL:
Thanks @nippoo!

Issue #905, opened 2015-05-15
Implement support for visual visibility toggle.


KelSolaar:
It would be great to have a `visible` attribute on the various visuals so that we can toggle their drawing at will without removing them from the scene graph (maybe skipping the `draw` call). I have seen a maye related attribute in the [`vispy.visuals.visual.Visual`](https://github.com/vispy/vispy/blob/master/vispy/visuals/visual.py#L36) class but it doesn't seem to be used.

Eric89GXL:
+1 for `.visible` property that toggles drawing by just returning if it is `False`, okay with you @campagnola?

campagnola:
Sounds good to me. The scenegraph should also handle this by skipping drawing the children of a hidden node.

KelSolaar:
I have just found a `visible` attribute in the [vispy.scene.node.Node](https://github.com/vispy/vispy/blob/master/vispy/scene/node.py#L69) class that seems to be doing the job just fine: https://drive.google.com/file/d/0B_IQZQdc4Vy8YVBTdExYUVJMSUk/view?usp=sharing

I have some unrelated redrawing issues though when I toggle the `XYZAxisVisual`.

Eric89GXL:
Ahh, so visibility can be toggled that way iff you're using scenegraph @KelSolaar? If so, can you think of a place we can improve the docs so that this will be clearer to users?

For the redrawing issue, can you check to see if it's due to the removal of that item, or if it's just due to causing a draw update (like add a keyboard command that just calls update and see if that also causes the issue)?

KelSolaar:
Yes it does work and I'm using it in the screencast above. The attribute is actually documented in the section for the [vispy.scene.node.Node](http://api.vispy.org/en/latest/scene.html#module-vispy.scene.node).

I definitely didn't think about looking at the `vispy.scene.node.Node` class for that since it is specifically a display related attribute. The _looks like it does nothing_ `self._visible = True` statement in [vispy.visuals.visual.Visual](https://github.com/vispy/vispy/blob/master/vispy/visuals/visual.py#L36) class confused me. It is also not really obvious that this class gets mixed with `Node` when looking at the code since it inherits from `object`. Definitely worth a mention in the docstrings about that, especially for people spending more time in the codebase than the documentation.

It is probably a good idea to put a reminder in the [vispy.scene.visuals](http://api.vispy.org/en/latest/scene.html#module-vispy.scene.visuals) introductory paragraph. Something like this maybe:

[...]

PR #911, opened 2015-05-19
Improve "vispy.scene.visuals" module and "vispy.visuals.visual.Visual" class docstrings.

KelSolaar:
References: https://github.com/vispy/vispy/issues/905#issuecomment-102419889


vispy/visuals/visual.py
```diff
@@ -30,6 +30,11 @@ class Visual(object):
     bounds_change : Event
         Emitted when the bounds of the visual have changed.
 
+    Notes
```
campagnola: This is not accurate, as it is possible to use the Visual classes without the scenegraph.

vispy/scene/visuals.py
```diff
@@ -3,7 +3,9 @@
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 """
 The classes in scene.visuals are visuals that may be added to a scenegraph
-using the methods and properties defined by vispy.scene.Node. 
+using the methods and properties defined in vispy.scene.Node such as name,
+visible, parent, children, etc...
```
campagnola: It would be great if we can figure out how to make these show up in the documentation for each individual VisualNode class.

vispy/visuals/visual.py
KelSolaar: The sentence doesn't say it can only be used within the scenegraph, (on the contrary), it states that in order to be used within it, it gets mixed with `vispy.scene.Node`.

campagnola: How about something like "When used in the scenegraph, all Visual classes are mixed with vispy.scene.Node in order to implement . . ."

KelSolaar: Sure thing!

vispy/scene/visuals.py
Eric89GXL: Yeah, it would. We can tackle it later. We should at least be able to do it for the `Scene` versions of visuals because we manually construct their docstrings anyway.

vispy/visuals/visual.py
```diff
@@ -30,6 +30,11 @@ class Visual(object):
...
+    -----
+    -   When used in the scenegraph, all Visual classes are mixed with
+    vispy.scene.Node in order to implement the methods, attributes and
```
Eric89GXL: single-backticks around `vispy.scene.Node` here. If you're really curious see:

https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

But the short version is, it will make it get properly linked in our docs.

vispy/visuals/visual.py
```diff
@@ -30,6 +30,11 @@ class Visual(object):
     bounds_change : Event
         Emitted when the bounds of the visual have changed.
 
+    Notes
+    -----
+    -   When used in the scenegraph, all Visual classes are mixed with
```
Eric89GXL: I think just one space after the `-` is standard for restructuredtext bullet lists

vispy/visuals/visual.py
KelSolaar: As I noticed they weren't used here: http://api.vispy.org/en/latest/scene.html#module-vispy.scene.visuals I actually skipped them.

vispy/visuals/visual.py
KelSolaar: Sure!

vispy/visuals/visual.py
Eric89GXL: Yeah, we should be using them but haven't been...

KelSolaar: Should I open a separate general issue for objects links in documentation?

Eric89GXL: Please open an issue, yeah, and please change this in your PR so we at least start making things better incrementally :)

Eric89GXL:
Travis is angry about style:

https://travis-ci.org/vispy/vispy/jobs/63196396#L1601

FYI you can test locally with `make flake`

KelSolaar:
It looks like I did an unwanted code formatting in PyCharm. It trimmed a few empty lines but although moved the bracket incorrectly.

Eric89GXL:
Thanks again @KelSolaar!

PR #913, opened 2015-05-19
FIX: Fix background colors

### LINKED ISSUE #908: "bgcolor" parameter has no effect.
I'm maybe overlooking something but I don't manage to have the `bgcolor` parameter to work in this example:

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

from vispy.app import run


class Canvas(SceneCanvas):
    def __init__(self):
        SceneCanvas.__init__(self, keys='interactive', bgcolor=(0.2, 0.2, 0.2))

        self._grid = self.central_widget.add_grid()

        self._view1 = self._grid.add_view(row=0, col=0)
        self._view1.border_color = 'red'
        self._view1.bgcolor = 'blue'

        self._view2 = self._grid.add_view(row=1, col=0, border_color='green',
                                          bgcolor='yellow')

        self.show()


if __name__ == '__main__':
    canvas = Canvas()
    run()
```

Passing `bgcolor='yellow'` here for example is fine: https://github.com/vispy/vispy/blob/master/examples/basics/scene/nested_viewbox.py#L115

Eric89GXL:
Closes #908.

@KelSolaar can you see if this fixes your use case?

@campagnola feel free to have a look and see if you agree with the changes.

### DIFF
--- a/examples/basics/scene/background_borders.py
+++ b/examples/basics/scene/background_borders.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# vispy: gallery 30
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014, Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
+Demonstration of various features of Line visual.
+"""
+
+from vispy.scene import SceneCanvas
+
+canvas = SceneCanvas(keys='interactive', bgcolor='w', show=True)
+grid = canvas.central_widget.add_grid(spacing=0, bgcolor='gray',
+                                      border_color='k')
+view1 = grid.add_view(row=0, col=0, margin=10, bgcolor=(1, 0, 0, 0.5),
+                      border_color=(1, 0, 0))
+view2 = grid.add_view(row=0, col=1, margin=10, bgcolor=(0, 1, 0, 0.5),
+                      border_color=(0, 1, 0))
+
+if __name__ == '__main__':
+    canvas.app.run()
--- a/examples/basics/scene/background_borders.py
+++ b/examples/basics/scene/background_borders.py
@@ -5,7 +5,7 @@
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 # -----------------------------------------------------------------------------
 """
-Demonstration of various features of Line visual.
+Demonstration of borders and background colors.
 """
 
 from vispy.scene import SceneCanvas
--- a/vispy/scene/node.py
+++ b/vispy/scene/node.py
@@ -36,7 +36,7 @@ class Node(object):
     # Needed to allow subclasses to repr() themselves before Node.__init__()
     _name = None
 
-    def __init__(self, parent=None, name=None, **kwargs):
+    def __init__(self, parent=None, name=None):
         self.name = name
         self._visible = True
 
--- a/vispy/scene/widgets/viewbox.py
+++ b/vispy/scene/widgets/viewbox.py
@@ -296,13 +296,6 @@ class ViewBox(Widget):
                 self.scene.draw(event)
             finally:
                 event.pop_node()
-        # else:
-        #     # Just draw
-        #     event.push_node(self.scene)
-        #     try:
-        #         self.scene.draw(event)
-        #     finally:
-        #         event.pop_node()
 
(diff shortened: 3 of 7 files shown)


vispy/scene/node.py
```diff
@@ -36,7 +36,7 @@ class Node(object):
     # Needed to allow subclasses to repr() themselves before Node.__init__()
     _name = None
 
-    def __init__(self, parent=None, name=None, **kwargs):
+    def __init__(self, parent=None, name=None):
```
Eric89GXL: This is the "endpoint" for the `**kwarg` passing, so it was dangerous for us to still have it here. If the kwargs haven't been consumed by this point, we should throw an error.

vispy/scene/widgets/viewbox.py
```diff
@@ -255,34 +253,33 @@ def draw(self, event):
             viewport = self._prepare_viewport(event)
         elif method == 'fbo':
             fbo = self._prepare_fbo(event)
+        else:
+            raise ValueError('Unknown clipping method %s' % method)
```
Eric89GXL: Added this as a safeguard against errant param passing

vispy/scene/widgets/widget.py
```diff
@@ -224,21 +244,21 @@ def add_grid(self, *args, **kwargs):
         """
         Create a new Grid and add it as a child widget.
 
-        All arguments are given to add_widget().
+        All arguments are given to Grid().
```
Eric89GXL: This is a change because `add_widget` didn't take any `args` or `kwargs`, so I think they should probably go to `Grid` (and below, `ViewBox`)

KelSolaar:
Yes it does work for me, thanks!

image

examples/basics/scene/background_borders.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# vispy: gallery 30
+# -----------------------------------------------------------------------------
+# Copyright (c) 2014, Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
+Demonstration of various features of Line visual.
```
campagnola: update me.

vispy/scene/node.py
campagnola: Good catch.

vispy/scene/widgets/widget.py
```diff
@@ -182,21 +186,37 @@ def _update_line(self):
...
+        faces = np.array([
+            [0, 2, 1],
+            [1, 2, 3],
+            [2, 4, 3],
+            [3, 5, 4],
+            [4, 5, 6],
+            [5, 7, 6],
+            [6, 0, 7],
+            [7, 0, 1],
+            [5, 3, 1],
```
campagnola: Last two triangles should be left out if the background color is clear. I have seen big performance hits for filling large triangles like these.

vispy/scene/widgets/widget.py
campagnola: I think that was because `Grid.add_widget()` does accept arguments (row, col, etc.)

campagnola:
Looks good!

vispy/scene/widgets/widget.py
Eric89GXL: From what I saw, the signature ~15 lines up is `add_widget(self, widget)` unless it is subclassed with more options elsewhere...

campagnola: Right; `Grid` being the subclass  ;)

Eric89GXL:
@campagnola thanks for the quick review, comments addressed

campagnola:
Two comments still pending above..

Eric89GXL:
The triangles thing has been addressed, GitHub just hasn't updated. Did you want me to revert the changes for `add_widget`?

vispy/scene/widgets/widget.py
Eric89GXL: I'm not sure this is a problem, `Grid` has its own signature for both `add_grid` and `add_widget`, and they don't call the super `Widget` versions...?

Eric89GXL: ...and actually `Grid.add_grid(..., **kwargs)` ends up passing the `kwargs` to the `Grid` call, too.

campagnola:
Nah, I retract my comment.
Thanks!

PR #915, opened 2015-05-20
Fixed out-of-bounds mouse errors on example resize

nippoo:
On window managers where the requested window size is not always respected (I'm on AwesomeWM but I assume other tiling WMs like XMonad etc do the same thing), the tests will fail since they assume the window has not been resized. We should be resilient to this.

### DIFF
--- a/examples/demo/gloo/imshow_cuts.py
+++ b/examples/demo/gloo/imshow_cuts.py
@@ -159,6 +159,10 @@ class Canvas(app.Canvas):
 
         yf = 1 - y/(h/2.)
         xf = x/(w/2.) - 1
+
+        x_norm = (x*512)/w
+        y_norm = (y*512)/h
+
         P = np.zeros((4+4+514+514, 2), np.float32)
 
         x_baseline = P[:4]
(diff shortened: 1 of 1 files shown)


examples/demo/gloo/imshow_cuts.py
```diff
@@ -159,6 +159,10 @@ def on_mouse_move(self, event):
 
         yf = 1 - y/(h/2.)
         xf = x/(w/2.) - 1
+
+        x_norm = (x*512)/w
+        y_norm = (y*512)/h
```
Eric89GXL: This isn't integer division on Python2.7, so you should do it more explicitly:
```
x_norm = (x*512)//w
```

examples/demo/gloo/imshow_cuts.py
```diff
@@ -170,11 +174,11 @@ def on_mouse_move(self, event):
         y_baseline[...] = (xf, -1), (xf, -1), (xf, 1), (xf, 1)
 
         x_profile[1:-1, 0] = np.linspace(-1, 1, 512)
-        x_profile[1:-1, 1] = yf+0.15*I[y, :]
+        x_profile[1:-1, 1] = yf+0.15*I[y_norm, :]
```
Eric89GXL: ...otherwise later you get warnings here about non-integer indices that they say will become errors eventually

nippoo: Oops. Yes, that should definitely be integer division!

Eric89GXL:
Travis is struggling today, but I checked locally and it's good. Thanks @nippoo

PR #920, opened 2015-05-22
Switch to SVG version of badges.

QuLogic:
Current badges are a bit inconsistent. Switching to the SVG versions makes them all the same height and shading.


Eric89GXL:
Thanks for the first 0.5 PR :)

QuLogic:
> first 0.5 PR

Oh, hah, lucky me! :confetti_ball:

PR #922, opened 2015-05-24
FIX: Minor fixes

### LINKED ISSUE #867: GL Error when running Offscreen.py
Hey all,

Thanks for the library! I'm trying to use gloo for offscreen rendering, but am getting an error when I run the offscreen.py example script. Paste from the error message is at:

https://gist.github.com/ardot/f300779d24102f2d15d4

Is this a KP? Am I configuring something wrong?
Thanks,
-Tim

### LINKED ISSUE #921: 'Isoline' object has no attribute 'set_levels', in "isocurve_for_trisurface_qt.py" example
trying isocurve_for_trisurface_qt.py example on windows/PyQt4/vispy0.4, it seems there is a basic compile error

````
  File "D:\result_tests\winpython-3.4.3.3\python-3.4.3\Scripts\isocurve_for_trisurface_qt.py", line 127, in <module>
    win = MainWindow()
  File "D:\result_tests\winpython-3.4.3.3\python-3.4.3\Scripts\isocurve_for_trisurface_qt.py", line 88, in __init__
    self.update_view()
  File "D:\result_tests\winpython-3.4.3.3\python-3.4.3\Scripts\isocurve_for_trisurface_qt.py", line 93, in update_view
    self.props.combo.currentText())
  File "D:\result_tests\winpython-3.4.3.3\python-3.4.3\Scripts\isocurve_for_trisurface_qt.py", line 121, in set_data
    self.iso.set_levels(cl)
AttributeError: 'Isoline' object has no attribute 'set_levels'
````

### PR DESCRIPTION
**Pull request by Eric89GXL**
Closes #921 
Closes #867

WIP because I'm hoping to tackle a few other little bugs.

### DIFF
--- a/examples/demo/gloo/offscreen.py
+++ b/examples/demo/gloo/offscreen.py
@@ -103,7 +103,7 @@ class Canvas(app.Canvas):
         app.Canvas.__init__(self, show=False, size=size)
         self._t0 = time()
         # Texture where we render the scene.
-        self._rendertex = gloo.Texture2D(self.size)
+        self._rendertex = gloo.Texture2D(shape=self.size + (4,))
         # FBO.
         self._fbo = gloo.FrameBuffer(self._rendertex,
                                      gloo.RenderBuffer(self.size))
--- a/examples/demo/scene/isocurve_for_trisurface_qt.py
+++ b/examples/demo/scene/isocurve_for_trisurface_qt.py
@@ -96,7 +96,7 @@ class MainWindow(QtGui.QMainWindow):
 class Canvas(scene.SceneCanvas):
 
     def __init__(self):
-        scene.SceneCanvas.__init__(self, keys='interactive')
+        scene.SceneCanvas.__init__(self, keys=None)
         self.size = 800, 600
         self.view = self.central_widget.add_view()
         self.view.camera = scene.TurntableCamera()
@@ -118,7 +118,7 @@ class Canvas(scene.SceneCanvas):
     def set_data(self, n_levels, cmap):
         self.iso.set_color(cmap)
         cl = np.linspace(-self.radius, self.radius, n_levels + 2)[1:-1]
-        self.iso.set_levels(cl)
+        self.iso.levels = cl
 
 
 # -----------------------------------------------------------------------------
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -1193,10 +1193,10 @@ class GlirFrameBuffer(GlirObject):
 
     def set_framebuffer(self, yes):
         if yes:
+            self.activate()
             if not self._validated:
                 self._validated = True
                 self._validate()
-            self.activate()
         else:
             self.deactivate()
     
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: vispy/gloo/glir.py [hunk 1]
```diff
@@ -233,7 +233,7 @@ def convert_shaders(convert, shaders):
                 lines.append(line.rstrip())
             # Write
             if not has_version:
-                lines.insert(0, '#version 120\n#line 2\n')
```
Eric89GXL: @izaid `git blame` says this line is yours -- why add the `#line 2` comment? Was that a leftover from debugging?

examples/demo/gloo/offscreen.py
```diff
@@ -103,7 +103,7 @@ def __init__(self, size=(600, 600)):
         app.Canvas.__init__(self, show=False, size=size)
         self._t0 = time()
         # Texture where we render the scene.
-        self._rendertex = gloo.Texture2D(self.size)
+        self._rendertex = gloo.Texture2D(shape=self.size + (4,))
```
Eric89GXL: So it turns out the error on intel platforms is that the renderbuffer we were creating wasn't complete -- I guess it was creating a luminance instead of RGB(A) texture here.

In any case, the GL spec says that it's strongly recommended to check that framebuffers are complete before rendering to them:

https://www.khronos.org/opengles/sdk/docs/man/xhtml/glCheckFramebufferStatus.xml

We currently don't have that implemented. I plan to put it in `gloo` to hopefully save us some future pain.

vispy/gloo/glir.py
campagnola: I seem to recall having fixed this once as well. I think the various OpenGL implementations represent their line numbers differently.

examples/demo/gloo/offscreen.py
campagnola: Cool. Thanks for debugging this one.

examples/demo/scene/isocurve_for_trisurface_qt.py
```diff
@@ -96,7 +96,7 @@ def update_view(self):
 class Canvas(scene.SceneCanvas):
 
     def __init__(self):
-        scene.SceneCanvas.__init__(self, keys='interactive')
+        scene.SceneCanvas.__init__(self, keys=None)
```
campagnola: Why?

Eric89GXL: This is an example of embedding the canvas within a Qt window, so the interactive keys like "escape" and "fullscreen" aren't useful. Hitting escape just makes the plot go away and the window stays open with controls, etc.

vispy/gloo/glir.py
```diff
@@ -1193,10 +1193,10 @@ def delete(self):
 
     def set_framebuffer(self, yes):
         if yes:
+            self.activate()
             if not self._validated:
                 self._validated = True
                 self._validate()
-            self.activate()
```
Eric89GXL: Turns out we did have a check for the framebuffer status, we were just doing the check *before* actually activating the framebuffer, which seems to be the wrong order of operations. (Now on Intel if I have a bad attachment, I get an appropriate error as below.)

vispy/gloo/glir.py
campagnola: Got it. I should have read the code :/

Eric89GXL:
@campagnola added one last fix, comments addressed, CIs are happy. Ready for merge from my end.

campagnola:
Looks good to me.

vispy/gloo/glir.py
izaid: @Eric89GXL I actually don't remember doing this...

Eric89GXL: It was in a block of lines that were yours so I bet it was part of a big copy-paste, don't worry about it

PR #924, opened 2015-05-24
[WIP] Bezier lines and arrows

sh4wn:
Hi all!

I've created a branch where I will develop the first part of my GSoC project: Bezier lines and arrows. Currently I've just copied the geometry.curves module from Glumpy, but this will change of course in the coming weeks.

So, the idea for arrows:
- Separate the head from the body/tail. 
- Create a subclass of LineVisual called ArrowVisual
- Add options to specify which vertices get an arrow head

Questions:
The line visual has many vertices which may or may not be connected. To extend on this, it would be only natural to specify which vertices will get an arrow head. But what would be the best way to determine the arrow head orientation? 

Orientation could be determined by getting the two last vertices of a line, determine its direction vector, and calculate the corner with the unit vector pointing upwards. But, what would be the best way to determine those last two vertices? As far as I can tell the given vertices are not necessarily in order, plus you're not sure if they're connected.

I'll probably have some questions about the OpenGL shaders too, they look quite advanced, but I'm also still busy with [this OpenGL tutorial][opengl].

Anyway, please let me hear your suggestions!

[opengl]: http://antongerdelan.net/opengl/


Eric89GXL:
@campagnola do we need the connected / unconnected capability moving forward? It seems like this problem is nicely solved by collections. In other words, can we make the simplification that every "LineVisual" used at some point will be a single connected line? I think it would simplify the API and coding, assuming we can gain back the speed by using collections.

Eric89GXL:
@sh4wn can you infer from @rougier's code how this direction is computed? It seems like he must use that information to get the direction somehow, unless it's computed on the CPU (unlikely).

campagnola:
> do we need the connected / unconnected capability moving forward?

It's hard to say. It seems likely that some thing can be accomplished more efficiently by simply specifying an array of connections (for example, there should be an easy way to set up a vector field of 100k arrows). If there was a way to do the same with collections that is both natural and efficient, then we might be able to do without it..

sh4wn:
Orientation is part of the vertex buffer data I believe:

https://github.com/glumpy/glumpy/blob/master/glumpy/library/arrows/arrow.vert#L45

sh4wn:
So, yesterday I investigated the code a lot, and the current plan is:

- [x] Create ArrowVisual class
- [ ] Write/Modify OpenGL shaders which draw the arrow heads
    - To determine the orientation and position, the VBO will consists of two vertices specifying the arrow body. The head center position will be the second vertex. 
    - The orientation can be determined by calculating the direction vector between the two vertices. By specifying the two vertices as VBO we can calculate the orientation on the GPU, which is I guess a better idea than on the CPU. 
- [ ] Create some extra arrow heads.

There's already some GLSL code for arrows in vispy/glsl/arrows/ (copied from Glumpy). I did some grep searches on some of the filenames and functions, but I couldn't find any code which uses these files. 

So, what shall we do? Create new shaders without an arrow body which I will use in the ArrowVisual class, or modify the currently unused shaders?

rougier:
Also, have a look at http://jcgt.org/published/0003/04/01/, it might be of some help.

Eric89GXL:
@sh4wn you probably didn't find any use for those files because the GLSL code was pulled from `glumpy`, but corresponding examples / use cases were not. So take a look and see if they are used in glumpy somewhere. It's possible that they aren't used there, either. I think the best approach is to make use of the existing code if it seems feasible, since in general it's easier to reuse code than to start from scratch if the existing code is decent, and @rougier generally knows what he's doing when it comes to GLSL :)

sh4wn:
Yeah in Glumpy they're used in one or more examples. 

I've currently copied these shaders to vispy/glsl/arrowheads, and in this folder I will apply any changes for this new ArrowVisual. 

Also: the current approach consists of a separate OpenGL program for drawing the arrowheads. It would be probably better to integrate them with the current line shaders. But I think I'm going to focus now on getting these arrow heads to work in a separate program, and see about performance and integration later.

Eric89GXL:
+1, start easy -- get something that works, then make it work well, then make it fast

Eric89GXL:
(although in this case @rougier's code is probably already fast enough)

sh4wn:
So, I'm getting a segfault right now when trying to start my arrows example:
https://github.com/sh4wn/vispy/blob/bezier-arrows/examples/basics/visuals/arrows.py

What are the most probable causes? Is my VertexBuffer not properly initialized? Something in the shader code?

sh4wn:
Any tips for debugging those OpenGL shaders?

sh4wn:
Segfault fixed (commenting out include directives wasn't enough)

sh4wn:
Some nice progress today! Ported all Glumpy arrow heads to the "new" system. Some last code cleaning will be done tomorrow, and I'll also investigate the orientation issue. And then implement some of my own arrow heads this sunday. 

I'll probably need to rebase this before it's in any mergeable state.

Eric89GXL:
Sounds good! Once that's done, you'll get to (have to) port it to @campagnola's slightly new Visuals system from #928, too :)

sh4wn:
Hmm, sort of fixed the automatic orientation feature. But the orientation is always slightly off:

screenshot from 2015-06-17 15-36-18

Anyone ideas for calculating the orientation properly?

Eric89GXL:
It looks like the orientation information might be calculated from the vertex positions (current - last), which will not be the same as the bezier curve tangent at the coordinate that the arrow is drawn (unless the curve is a straight line). Does this problem go away / do things look correct if you just use straight lines? Does it look worse if you make the bend more prominent?

sh4wn:
What would be the best way to apply my changes to the branch of @campagnola?

Eric89GXL:
You could open a PR into his branch, but it depends how much longer his PR
will likely be open (unmerged). @campagnola should new visual PRs go into
your branch, master, or wait a little bit until your branch is merged?

sh4wn:
Oh haha, saw your comment too late. I've just rebased this branch based on that of @campagnola.

Eric89GXL:
That should work :) if his branch gets merged in the meantime (or not) it
should be fine. Let us know when we should have a look. @bollu has been
translating a number of visuals so he can help answer questions as you go.

campagnola:
Yeah--it doesn't matter how your code gets merged eventually, it'll just be a lot less work if you begin writing from my branch.

sh4wn:
Ok, so I'm busy with moving the ArrowHead visual to the new system, but I see a few possibilities, and I'd like to hear which one you guys would prefer:

We'll have an ArrowVisual class which derives from CompoundVisual (other possibility: derive from LineVisual which is already a compound visual). This visual needs to combine a line and an arrow head. 

For the arrow heads: should I integrate them with the MarkerVisual? A drawback from that would be that the orientation for each arrowhead should be computed on the CPU. 

I'm currently leaning towards:
- ArrowVisual derived from LineVisual. Combines the arrow head visual with _GLLineVisual or _AggLineVisual.
- Separate ArrowHeadVisual with orientation calculation on the GPU.

Eric89GXL:
I think separate makes more sense. I don't think we want visuals deriving
from other visuals in general, instead they should derive from
CompoundVisual if possible. I think that's especially true here where we
get gpu calculation instead of cpu calculation by deriving from
CompoundVisual.

Eric89GXL:
Oh, I misread your proposal. I think the ArrowHeadVisual would be good, but
maybe still have the ArrowVisual be a CompoundVisual that combines the two
underlying line classes and one arrow head class. Not sure. @campagnola
what fits best with your class hierarchy / design?

campagnola:
> I think the ArrowHeadVisual would be good, but maybe still have the ArrowVisual be a CompoundVisual that combines the two underlying line classes and one arrow head class.

I think that's the way I would do it.

> For the arrow heads: should I integrate them with the MarkerVisual? A drawback from that would be that the orientation for each arrowhead should be computed on the CPU.

It would be great if MarkerVisual could be used for drawing arrow heads. How do you propose to calculate orientations on the GPU?I assume this would come from the angle of the line ends, but you won't have access to the line's vertex buffer from the ArrowHeadVisual. 

A different approach would be to embed the arrow head code in the AGG line visual, since it has code specifically for handling line endings (I think we currently just have round and square ends).

Let's keep all of these possibilities in mind because ideally, the arrow head shader code you write could be reused in many different contexts..

sh4wn:
> It would be great if MarkerVisual could be used for drawing arrow heads. How do you propose to calculate orientations on the GPU?I assume this would come from the angle of the line ends, but you won't have access to the line's vertex buffer from the ArrowHeadVisual. 

Currently one needs to specify two vertices which represent the arrow body. With two vertices we can easily calculate a direction vector, and the second vertex is used as position for the arrow head. However, it appears that this calculated direction vector is often slightly off the real tangent of the line. So an other method is probably better.

> I think the ArrowHeadVisual would be good, but maybe still have the ArrowVisual be a CompoundVisual that combines the two underlying line classes and one arrow head class.

Yes, possibly replacing ArrowHeadVisual by MarkerVisual, and integrating the arrow heads with the MarkerVisual.

bollu:
I've read through the AGG papers briefly, and I think one can control line
endings? I remember seeing different arrow heads laid out in a spiral.

I'm not sure if my memory of it is accurate.

sh4wn:
@bollu Do you mean the paper with the techniques for drawing markers?

bollu:
Yep, that's the one.

sh4wn:
@bollu I'm not sure what you mean by controlling the line endings?

Furthermore: I'm still in doubt wether to integrate the arrow heads with the marker visuals. When I think of a possible application (visualising graphs), and thinking how they can animate when you use an automatic layout, makes me question whether it's a good idea to calculate the arrow head orientation on the CPU each frame. 

I think it would be much less CPU intensive when you just pass the last two vertices of a line to the GPU buffer and calculate the orientation on the GPU, but only if we can accurately calculate the orientation from the last two line vertices (but I have some ideas for that).

Eric89GXL:
It does sound better to get orientation calculation working on the GPU. However, I would propose that you do whatever is the path of least resistance first. We can always optimize it later. In other words, if it is easy to get the ArrowHeadVisual working while calculating the orientation on the CPU, go ahead and do that first, and then later you or someone else can go back and make it more efficient using GPU-based orientation calculation. Having some working code in place (in `master`) is an important first step because it means that other people can start to make enhancements and fixes instead of you having to do everything. Make sense?

Have we converged on an API? If so, I think the top priority should be to get some working code with simple tests and a nice example in `master`. Luke's #928 should be merged ASAP, so you can for now start working off of that branch to make your visual. If you've already started making a visual using the old API, that is okay -- the differences are not great between the old and new way.

rougier:
For "regular" agg lines, you can control the line ending between butt/cap/round/triangle up to the limit where the line ending is fully contained in the line enveloppe. Else, you need to have separate markers as you did. The mis-alignement makes me think the orientation might have been computed between the n-3 and n-2 instead of n-2 and n-1. Can you make some test with a very simple 2-lines segments and check orientation is ok in such a case ?

sh4wn:
Ok, cool I've got the arrow visual working with the new visual system. Next up is a quiver plot example using this new visual. 

@rogier: I'm pretty sure I use n-1 and n-2 as vertices (example here: https://github.com/sh4wn/vispy/blob/de25918048ecf471ad0e0ae77f781962f036b4fa/examples/basics/visuals/arrows.py#L17), but I will create a more simple example, and see what happens.

Personally I think it's because the arrow head partly overlaps the line (the second vertex is used as *center* point of the arrow head), so maybe we should translate it a bit in the direction of the line.

Eric89GXL:
@sh4wn I think you meant @rougier :)

Yes please let us know when the example is done. It looks like this is still a PR into `master`, you probably need to rebase off Luke's branch to get things working.

rougier:
One other explanation could the insufficient tesselation of the curve (geomtry.curves). In this file, you will find the following:
```
curve_distance_epsilon = 1e-30
curve_collinearity_epsilon = 1e-30
curve_angle_tolerance_epsilon = 0.01
curve_recursion_limit = 32
m_cusp_limit = 0.0
m_angle_tolerance = 10 * math.pi / 180.0
m_approximation_scale = 1.0
m_distance_tolerance_square = (0.5 / m_approximation_scale)**2
```

Those are directly taken from http://antigrain.com/research/adaptive_bezier/ but the idea is that there is an approximation scale that make you tesselate the curve more or less depending on the size. You can also try to play with these parameters to see if it changes things (for example, number of points returned).

sh4wn:
Ok, the quiver plot is finished. But I'd argue if this is a good example for the ArrowVisual. The Arrow Visual was designed with graphs in mind where you often have moving nodes and don't care much for the specific angle an arrow has with a certain axis. 

In the case of a quiver plot you want them at a specific angle, and therefore have to calculate quite a lot on the CPU (rotate the vertices).

Anyway, here's a screenshot:
screenshot from 2015-07-16 15-10-12

And for the straight lines the arrow head looks oriented in the right way, so that's a plus.

Eric89GXL:
How fast is the example? Is the on-CPU orientation calculation slow? I guess it could be for that many arrows.

sh4wn:
At full screen you'll start noticing a slight delay, but at the initial window size you won't notice it (Haswell i5).

Eric89GXL:
I'm surprised the screen size matters. Does that change the number of arrows or something?

sh4wn:
Haha yes I should've mentioned it, the size of the arrows will stay the same, it will increase the number of rows and columns with arrows as the screen size increases.

Eric89GXL:
Can you try just updating the vertices? It doesn't seem like that should be that expensive.

I think it might be the fact that you're not on #928. I think the shaders might be rebuilt with every upload, instead of just updating the coords attribute.

Eric89GXL:
Oh, you've been merging with Luke's branch instead of rebasing. That will make reviewing a bit difficult until his branch gets merged, unfortunately. I didn't realize you were already on his branch -- in that case, you probably won't see huge performance gains.

sh4wn:
I'm quite sure I'm on #928, and I believe I don't recompile the program everytime. 

Anyway: the code is here: https://github.com/sh4wn/vispy/blob/bezier-arrows/examples/basics/visuals/arrows_quiver.py

But each arrow has two vertices which need to be rotated, and each arrows has a different rotation matrix, so we can't use any efficient numpy function. So I find it not that strange you'll notice a slight delay at some point.

Hmm, when #928 gets merged into master, will this PR not detect that those commits are already there, and does not belong here anymore?

Eric89GXL:
Try to vectorize the computation. Don't use a matrix, that is one way to think about applying rotation but it is an inefficient way here.

Eric89GXL:
If I have a set of points (x, y) and a mouse at (mouse_x, mouse_y), then I can easily in a vectorized way compute the unit vector in that direction for each point at the same time, and then shift that unit vector back to the origin of each point. Make sense?

Eric89GXL:
In other words, those ~20 lines should become 3 or 4, I think...

sh4wn:
Do we have a minimum Numpy version we support? 

NumPy 1.9 has a nice feature to quickly normalize a matrix: https://stackoverflow.com/questions/7741878/how-to-apply-numpy-linalg-norm-to-each-row-of-a-matrix

Eric89GXL:
Probably 1.7, so don't use norm with axis argument. It is only sometimes
faster anyway.

Eric89GXL:
Let me know if you can't figure out the math and I'll have a look

rougier:
For quiver plot with regular placement of arrow, you don't need to compute anything on the CPU. Have a look at https://github.com/glumpy/glumpy/blob/master/examples/quiver.py. I only draw a quad (two triangles) and the shader takes care of drawing arrows. The only information needed is the size of the viewport in pixels and mouse location (in pixels). If this was a real quiver, you would only need a textire giving orientation and strength for each arrow.

Eric89GXL:
Yeah, those methods would be more efficient. We are trying to start simple for this first pass, though. Then maybe some more efficient coder (maybe named @rougier) could come through and optimize it :)

sh4wn:
Ok I fixed it :) Performance is definitely a bit better. I can't see any delay anymore at fullscreen. A tailored GLSL program is of course better, but it's a nice example for demonstrating the usage of ArrowVisual.

Eric89GXL:
I would comment in-line, but the gigantic diff makes it brutal, so here:

* `np.abs(direction_vectors)` you don't need the `abs` because they are real numbers and you then immediately square them

* `class Canvas(app.Canvas):` two lines before class def, Travis will complain about this

Can you use `make flake` and e.g. `make pyqt4` and `make nobackend` to get tests passing for the bits of code you've written?

Actually, can you open your branch as a PR against Luke's branch, as well? We'll only merge one of them eventually, but in the meantime it would make reviewing much easier.

bollu:
`make docs` as well, it checks for parameters in your docstrings.

Eric89GXL:
Yep, although I think that actually gets run by `make nobackend` as well

sh4wn:
There's a new PR into @campagnola's branch here: https://github.com/campagnola/vispy/pull/22

PR #926, opened 2015-05-25
Added example for mouse editing of line plot

fschill:
Hi, 

I'm looking into a simple vector graphics sketching framework, possibly as an add-on to vispy. As a first test I extended the line plot visual to add mouse events and simple mouse editing. The demo is based on various line-plot and pan-zoom demos from existing examples. In principle the demo works with multiple lines, but as a click into empty space creates a new point for all lines in this demo, it is somewhat unintuitive. 

It seems to me that other visuals could be extended in a similar way, maybe even with a generic wrapper for basic point editing. Comments are welcome how to best integrate or add this feature.

A potential issue I discovered is that the left-click pan interferes with left-click dragging line points. It would be great if the mouse buttons for camera actions could be configurable (e.g. right button for pan). 

cheers, 
Felix

### DIFF
--- a/examples/basics/visuals/line_draw.py
+++ b/examples/basics/visuals/line_draw.py
@@ -15,6 +15,7 @@ import numpy as np
 
 from vispy import app, scene
 
+
 class EditLineVisual(scene.visuals.Line):
     """
     Mouse editing extension to the Line visual.
@@ -29,10 +30,11 @@ class EditLineVisual(scene.visuals.Line):
         # initialize point markers
         self.markers = scene.visuals.Markers()
         self.marker_colors = np.ones((len(self.pos), 4), dtype=np.float32)
-        self.markers.set_data(pos=self.pos, symbol="s",edge_color='red', size=6)
+        self.markers.set_data(pos=self.pos, symbol = "s", edge_color = "red",
+                              size=6)
 
         # snap grid size
-        self.gridsize=10
+        self.gridsize = 10
 
     def draw(self, transforms):
         # draw line and markers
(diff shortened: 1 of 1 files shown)


Eric89GXL:
Cool example! FYI we are working on picking of visuals using GLSL (should scale nicely to thousands and millions of points), but this is a good example of picking using the CPU.

Travis is unhappy because of style issues -- could you fix those? I'll point out a few as well.

In terms of camera configuration -- that could be a useful feature. We'd have to dig into the API a bit to make that work. How would you ideally change the parameter? I imagine something like `camera.pan_button = 'right'` or something, but I'd have to think about it.

examples/basics/visuals/line_draw.py
```diff
@@ -0,0 +1,150 @@
...
+    def on_mouse_release(self, event):
+        self.print_mouse_event(event, 'Mouse release')
+        self.selected_point = None
+        self.update_markers()
+
+    def on_mouse_move(self, event):
+        # left mouse button
+        if event.button == 1:
+            # self.print_mouse_event(event, 'Mouse drag')
+            if self.selected_point != None:
```
Eric89GXL: you should generally use `is` for comparisions to `None`, like `if self.selected_point is not None:` here

fschill:
Thanks for the feedback - I briefly went through it again and fixed the style issues I could find. I also changed my method comments to docstring format - not very familiar with it but I hope I got it right. As it's just a demo at this stage I didn't add full docstring comments everywhere as they can make the code a bit long for an example (at least the other demos I looked at were similar).

Regarding the camera - I was thinking the same as you, I imagine that would be the simplest way. The Pan-Zoom camera has a few hard-coded checks, e.g. line 698:
```            if 1 in event.buttons and not modifiers: ```
which could be changed to:
```            if self.drag_button in event.buttons and not modifiers: ```
or similar, which would be initialised in the constructor to show the same behaviour as currently if nothing else is specified. Some methods for setting the buttons would have to be added too.

I read the sections regarding event handling in Vispy, but it seems that the camera gets the events before my EditLineVisual handlers, so I wasn't able to block the events. Ideally it should be possible to handle some events (e.g. only if an point-dragging operation is in progress) and let them through otherwise. Up to now I couldn't figure out the best way to do this, other than turning off the event handling completely for mouse_move. It would maybe be possible to disconnect and re-connect depending on state, but I wasn't sure if this is the way it's supposed to be done.

fschill:
>FYI we are working on picking of visuals using GLSL (should scale nicely to thousands and millions of points), but this is a good example of picking using the CPU.

Great to hear, sounds interesting! I assume that the new GLSL mechanism will essentially do what my "select_point" method currently does, i.e. return primitives that are close to the mouse pointer? Does it also mean that visuals will only get mouse events that are spatially close (e.g. for a line only if the mouse pointer is on/near the line)?

fschill:
ok, took a few attempts until I figured out how to configure code style checks in my IDE, but it seems Travis is happy now... :)

Eric89GXL:
LGTM. @campagnola okay to merge this for now, and you (or someone else) can update the picking part with the GPU-based method later?

fschill:
Sounds good. I'm planning to build a small and simple sketching library on top now, but I think the higher levels are fairly independent of the exact method of picking. In any case, I'll keep in mind that the GPU picking is coming up and will try to update once it's there. Do you already have some rough design of what it will look like that I could have a look at?

Eric89GXL:
#928 has the new picking, it might even have an example (haven't checked)

campagnola:
Thanks! 
The new picking system is currently able to determine which visual occupies each pixel. We would need a slightly different approach for picking individual vertices in a visual (but this should be simple).

PR #931, opened 2015-05-26
FIX: Allow text lists

Eric89GXL:
Ready for review/merge @campagnola.

@nippoo with this PR you can do:
```
text.text = ['foo', 'bar']
text.pos = [[100, 200], [100, 300]]
canvas.update()
```
and it should show both bits of text.

### DIFF
--- a/vispy/visuals/text/text.py
+++ b/vispy/visuals/text/text.py
@@ -208,7 +208,7 @@ class TextVisual(Visual):
     ----------
     text : str | list of str
         Text to display. Can also be a list of strings.
-        Note: support for list of str will likely be removed soon
+        Note: support for list of str might be removed soon
         in favor of text collections.
     color : instance of Color
         Color to use.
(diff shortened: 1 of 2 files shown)


vispy/visuals/text/text.py
```diff
@@ -206,8 +206,10 @@ class TextVisual(Visual):
 
     Parameters
     ----------
-    text : str
-        Text to display.
+    text : str | list of str
+        Text to display. Can also be a list of strings.
+        Note: support for list of str will likely be removed soon
+        in favor of text collections.
```
campagnola: Why? This seems like a reasonable API to keep. 
In the long run, it is not at all clear how the collection and visual APIs will look when they are combined. I could imagine something like:

```
t1 = TextVisual("...", color1, font_size1)
t2 = TextVisual("...", color2, font_size2)
collection = TextCollection(t1 + t2)
```

.. but this API is quite cumbersome if you wanted to have, for example, 100k letters assigned to points in a scatter plot. That should look more like the API you have proposed here.

campagnola:
Looks good! I'm happy to merge this after we sort out the single docstring issue.

vispy/visuals/text/text.py
Eric89GXL: Yeah that's true. I don't want to lock us into an API until we've seen how all the collections stuff will work, though. I'll change the comment here to say that it might be removed in favor of collections, instead of probably will.

Eric89GXL:
Thanks for the merge @campagnola, can't ask for better response time than that :)

PR #932, opened 2015-05-27
rewrite dictionary comprehension as a loop (Python 2.6 support)

groutr:
Filling the dictionary now takes about half the time too!

### DIFF
--- a/vispy/util/svg/color.py
+++ b/vispy/util/svg/color.py
@@ -154,13 +154,14 @@ _keyword_colors = {
     "yellow":               (255, 255,   0),
     "yellowgreen":          (154, 205,  50)}
 
+from itertools import combinations
+from string import hexdigits
 
-_NUMERALS = '0123456789abcdefABCDEF'
-
-
-_HEXDEC = {v: int(v, 16)
-           for v in (x + y for x in _NUMERALS for y in _NUMERALS)}
-
+_HEXDEC = {}
+for x, y in combinations(hexdigits, 2):
+    v = x + y
+    _HEXDEC[v] = int(v, 16)
+    
 
 def _rgb(triplet):
     return _HEXDEC[triplet[0:2]], _HEXDEC[triplet[2:4]], _HEXDEC[triplet[4:6]]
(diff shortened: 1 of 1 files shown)


vispy/util/svg/color.py
```diff
@@ -154,13 +154,14 @@
     "yellow":               (255, 255,   0),
     "yellowgreen":          (154, 205,  50)}
 
+from itertools import combinations
+from string import hexdigits
```
Eric89GXL: I didn't know about this, neat

Eric89GXL: Minor gripe, though -- please put imports at the top (PEP8)

Eric89GXL:
Other than my minor comment LGTM. Thanks for the fix @groutr. Let me know when the imports are moved and I'll merge.

groutr:
I moved the imports.

Eric89GXL:
Thanks for that @groutr -- I'll merge once the CIs are happy

Eric89GXL:
Thanks @groutr!

MatthieuDartiailh:
This could have been done using a list comprehension building a list of pair which can then be passed to the dict constructor. I believe this would be more efficient than an explicit loop.

groutr:
The list comprehension will be slower.  And after doing the timings, I confirmed that it is indeed slightly slower (it's constructing a lot of tuples that we don't need to construct).
I also point out that the explicit loop is much easier/faster to read than the list comprehension.

Eric89GXL:
Good to know! Also, probably not a huge deal either way because of the size of the operation :)

groutr:
@Eric89GXL: You're right, given the size _and_ the fact that the size stays constant.

MatthieuDartiailh:
Ok nice to know. Apologies for the noise...

vispy/util/svg/color.py
```diff
@@ -155,12 +159,11 @@
...
-_NUMERALS = '0123456789abcdefABCDEF'
-
-
-_HEXDEC = {v: int(v, 16)
-           for v in (x + y for x in _NUMERALS for y in _NUMERALS)}
-
+_HEXDEC = {}
+for x, y in combinations(hexdigits, 2):
+    v = x + y
+    _HEXDEC[v] = int(v, 16)
```
dengemann: ```dict([(v, int(v, 16)) for v in (x + y for x in _NUMERALS for y in _NUMERALS)])```

@Eric89GXL et al this would have been possible btw. in Python 2.6

PR #938, opened 2015-05-28
FIX #937 left Click and shif key 

### LINKED ISSUE #937: Problem with turn ans shift key event
In a Canvas with a turnable Camera, if you try to make a left click to turn the mesh and if you maintain the left click and after you press shift key you are an error message.

sylm21:
Closes #937

### DIFF
--- a/vispy/scene/cameras/cameras.py
+++ b/vispy/scene/cameras/cameras.py
@@ -1005,6 +1005,8 @@ class Base3DRotationCamera(PerspectiveCamera):
                 norm = np.mean(self._viewbox.size)
                 if self._event_value is None:
                     self._event_value = self.center
+                elif len(self._event_value) == 2:
+                    self._event_value = self.center
                 dist = (p1 - p2) / norm * self._scale_factor
                 dist[1] *= -1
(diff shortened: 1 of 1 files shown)


vispy/scene/cameras/cameras.py
```diff
@@ -1005,6 +1005,8 @@ def viewbox_mouse_event(self, event):
                 norm = np.mean(self._viewbox.size)
                 if self._event_value is None:
                     self._event_value = self.center
+                elif len(self._event_value) == 2:
+                    self._event_value = self.center
```
Eric89GXL: a nitpick on a fix of this few lines, but it is more DRY to do:
```
if self._event_value is None or len(self._event_value) == 2:
```
that way the operation is only called once

Eric89GXL:
FYI I added "Closes #937" to your description, which causes GitHub to auto-close that issue once your PR is merged

Eric89GXL:
This fix does remove the error for me. Interaction isn't quite smooth with it, but it's better than an error. Thanks for yet another fix @sylm21!

PR #940, opened 2015-05-29
Unbundle-able libraries

QuLogic:
This moves bundled libraries into a single directory, and adds shims so that the system copies _may_ be used instead. `six` was a bit of a pain, so I pulled the work that was in astropy/astropy#2623. This should make it easier for packagers, while still leaving things self-contained for you.


vispy/ext/decorator.py
```diff
@@ -1,253 +1,14 @@
...
-    ContextManager = type(
-        'ContextManager', (GeneratorContextManager,),
-        dict(__call__=__call__, __init__=__init__))
+from __future__ import absolute_import
 
-contextmanager = decorator(ContextManager)
+try:
+    from vispy.ext._bundled.decorator import *
+except ImportError:
+    from decorator import *
```
Eric89GXL: but the above should never fail, right?

vispy/ext/decorator.py
```diff
@@ -1,253 +1,14 @@
...
-    def __init__(self, f, *a, **k):
-        return GeneratorContextManager.__init__(self, f(*a, **k))
-    ContextManager = type(
-        'ContextManager', (GeneratorContextManager,),
-        dict(__call__=__call__, __init__=__init__))
+from __future__ import absolute_import
 
-contextmanager = decorator(ContextManager)
+try:
+    from vispy.ext._bundled.decorator import *
```
Eric89GXL: ...and we prefer to use relative imports, why not do from .decorator import * here?

Eric89GXL:
What problem does this fix? I'm still not clear on that.

QuLogic:
Packages don't like bundled libraries. Before, you'd have to change all references to `vispy.ext.*`, but now, you can just delete the `_bundled` directory, and add package dependencies as usual.

Eric89GXL:
By packagers you mean like Anaconda? I'm missing who this will help

QuLogic:
[Fedora](https://fedoraproject.org/wiki/Packaging:No_Bundled_Libraries), [openSUSE](https://en.opensuse.org/openSUSE:Packaging_guidelines#Bundling_of_multiple_projects), Debian (can't find a link, but there are [cases](https://wiki.debian.org/Teams/RustPackaging/Unbundling)), etc.

vispy/ext/decorator.py
QuLogic: If you wish, but I think you mean `from ._bundled.decorator import *`?

campagnola:
I ran into the same issues with pyqtgraph. Even had a hard time packaging the documentation because of a single .js file that sphinx dragged in.

Eric89GXL:
Ahh makes sense. To properly test support, we should modify one of our Travis runs (probably 3.4 full) so that the external dependencies are installed as well, and then also `vispy/ext/_bundled` gets deleted before any tests are run. @QuLogic can you add that to this PR? It will test that the unbundling actually works as expected.

Eric89GXL:
Also @QuLogic do you have particular bundling targets in mind that you (or others) plan on working on?

QuLogic:
Well, technically speaking, all bundled libraries should be removed, but practically speaking, I think that's all I'm up for at the moment. Of the items in `ext`:
 * `cubehelix` hasn't made any releases,
 * `egl` appears home-grown,
 * `fontconfig` could maybe be unbundled, but it would mean a hard dependency on pyglet, which I'd rather avoid if possible (haven't looked into it yet),
 * `gdi32plus` is not a big deal here,
 * `glfw` does not seem to have an upstream,
 * `ipy_inputhook` maybe, but IPython would be a heavy dep, and looking at other tickets, it looks like it might get modified/simplified even more.

Maybe could look at this one:
 * `freetype` says "adapted from freetype-py", so I'd have to look into the changes first, but there may already be packages for it.

I assume everything bundled is in `ext`.

Eric89GXL:
We have tried to keep bundled things in there, yes. That sounds reasonable to me, thanks for taking a stab at this @QuLogic. I think there is a `pyglw` or `glfwpy`, but I'm not sure what types of releases they've had. Probably safer to leave that one.

Do you have time to modify Travis appropriately as I commented above?

Eric89GXL:
I think it would mean a conditional in the `.travis.yml` checking if the run is Python 3.4 full, and if so, delete the `ext/_bundle` directory, and use `conda` and if not available then `pip` to install the packages.

QuLogic:
OK, I'll look into it in a little bit, then.

Eric89GXL:
Thanks -- it should ensure that the code all works as expected, and ensure that we keep pace with any changes in those packages as we go.

QuLogic:
Freetype seems to be causing trouble somehow. Maybe I shouldn't have added that one...

QuLogic:
Looks like [that was commented out](https://github.com/vispy/vispy/blob/master/vispy/ext/freetype.py#L192-201). I think a better fix is upstream rougier/freetype-py#21, but it's not in a released version yet.

rougier:
I'll try to make a new release. Can you open an issue on freetype-py (just to remind me of making a release).

QuLogic:
OK, I left out freetype-py for now, so that this will work. I'll hold on to that change until freetype-py gets a new release.

Eric89GXL:
Thanks a lot @QuLogic for getting Travis on board

QuLogic:
That... seems a little late there...

Eric89GXL:
Coveralls does that sometimes, guess he gets lazy

PR #948, opened 2015-05-30
Different method of associating glir queues (for better performance)

campagnola:
Ready for review/merge.

This resolves a performance issue where scenes with a large number of visuals were visiting `GlirQueue.clear()` thousands of times per draw.

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -69,7 +69,23 @@ def as_enum(enum):
 
 
 class _GlirQueueShare(object):
-    """
+    """This class contains the actual queues of GLIR commands that are
+    collected until a context becomes available to execute the commands.
+    
+    Instances of this class are further wrapped by GlirQueue to allow the
+    underlying queues to be transparently merged when GL objects become
+    associated.
+    
+    The motivation for this design is that it allows most glir commands to be
+    added directly to their final queue (the same one used by the context),
+    which reduces the effort required at draw time to determine the complete
+    set of GL commands to be issued.
+    
+    At the same time, all GLObjects begin with their own local queue to allow
+    commands to be queued at any time, even if the GLObject has
+    not been associated yet. This works as expected even for complex topologies
+    of GL objects, when some queues may only be joined at the last possible
+    moment.
     """
     def __init__(self, queue):
         self._commands = []  # local commands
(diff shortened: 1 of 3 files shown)


Eric89GXL:
Can you comment briefly on what it does? Maybe by commenting on / docstringing the `_GlirQueueShare` class? Thousands of time per draw is pretty rough indeed.

Have you tried this out with the IPython WebGL backend to see if it works? If not I can verify it before merging.

campagnola:
Working on docstring updates. Please do check ipython, thanks!  (it always seems to take me several minutes to get it up and running again)

campagnola:
Basically: in the existing implementation, a graph of dependencies is set up that links objects like `context -> program -> texture`. Each object has its own queue of glir commands, which means that to flush the entire set of commands for a context, you must recursively search the entire tree for stray commands (and this flush occurs once per program draw). 

This PR changes the flow such that, once objects are connected in the graph, they all share the same command queue and it is no longer necessary to search the tree.

Eric89GXL:
So with this new approach is it possible that you might now get multiple threads putting calls into the same queue out of order, whereas previously the order would have been guaranteed (since it the commands would have been extracted from each object in the tree by the drawing thread)? I'm not sure if it matters if this is the case, but I want to know for future reference if we should keep this in mind.

campagnola:
Yes, that could change the order of commands (and we would also need to add a lot of locking to make everything thread safe). I had actually added some code that marked each command based on the order it was generated and sorted before parsing, but ultimately it turned out to be unnecessary.

Eric89GXL:
WebGL appears to work just as well, and tests pass. I trust you there is a performance benefit :)

Eric89GXL:
Thanks. Did you need this for your big PR, or did you just notice this could be split off into a bite-sized PR?

campagnola:
It was something I fixed in the PR but broke it off because I knew it would need a lot of testing and didn't depend on the visual/scenegraph changes (and the tests don't run so well in the PR).

almarklein:
Does this also hold up when *changing* the topology of a scene graph? E.g. when a visual is moved to another canvas? Do changes that have been applied to the visual propagate to the correct parser in this case?

campagnola:
No, this would end in catastrophe by merging together the queues of two different contexts. That will need to be dealt with at some point, but I think it will take a lot of work (beyond the effects of this PR) to make visuals usable on more than one canvas.

almarklein:
Ok, and what about changing the topology where just one context is involved?

campagnola:
If there is only one context, then the topology is irrelevant--all GlirQueues should (eventually) become associated and will share a single underlying queue of glir commands.

Issue #832, opened 2015-03-28
Refactoring of node transformations and caching system 


almarklein:
During the 2015 sprint in Paris, Luke and I discussed a plan for the transformation caching. It allows allow for all the flexibility that we currently have, while realizing a performance that is (we think) close to using bare visuals.

Steps:

* [ ] Implementation of a DefaultTransform class that can do scaling, translation and rotation, which is aware of the complexity of the transform and is efficient with composition and inversion when the described transformation does not include shearing. 
* [ ] Make the DefaultTrasform class the default transform for all nodes.
* [ ] Transform chains cache a simplified (collapsed) version of itself. Transforms can be marked as "dynamic", which will prevent them from being collapsed with neighboring transforms. In that way, updating these will not require any re-simplification and or program rebuilding.
* [ ] Each node keeps its own cache of transformation chains. When a parent is changed, all child nodes will be made to update their transform chains. In this way the transform chains won't have to be rebuild on each draw.  This will replace the global node-chain-to-transform caching mechanism currently in use.

I'm working on the DefaultTransform now.

Eric89GXL:
Sounds great to me!

rougier:
Some questions:
* Why not using identity as the default transform ? (scaling(1) + translate(0) + rotation(0) has a cost)
* Does not collapsing transforms impact performances significantly ?
* Is there an option to not have transform for a node (just inheriting parent's transform) ?

almarklein:
> Why not using identity as the default transform ? (scaling(1) + translate(0) + rotation(0) has a cost)

You mean like an IdentityTransform class? I think it would be good if we can allow the user to simply do `node.transform.translate(x, y, z)`, without needing to set the type of transform. The class I'm now working on keeps track of its own complexity, so that if its still the identity, it will be fast.

> Does not collapsing transforms impact performances significantly ?

I think we can attain the best performance if we can collapse all static transform, but keep the dynamic ones intact. That way we can update the parameters of the transform without needing any re-collapsing or re-compilation.

> Is there an option to not have transform for a node (just inheriting parent's transform) ?

If the transform is not changed, it represents the identity transform, which is what you describe.

campagnola:
> Is there an option to not have transform for a node (just inheriting parent's transform) ?

To answer more directly, yes--currently all visuals use a `NullTransform` by default (should be renamed to `IdentityTransform` as you suggested) and even if we change the default, it will always be possible to reassign it to any other type.

PR #950, opened 2015-05-30
WIP on a singular efficient Transform class

almarklein:
(I had this branch laying around for a while and make it "public" so it might be useful for Luke's epic quest for a better scene graph).

issues:  implements part of #832, #894

This PR implements a `DefaultTransform` class that can deal with translation, scaling, rotation, and anything else than can be described with a 4x4 matrix. Its smart enough to keep the internal description as simple as possible, which means it's also fast.

Having one transform class simplifies things, especially for the user.

cc @campagnola note that I made some fixes/changes in the quaternion class, which needed some changes in the cameras as well.

### DIFF
--- a/vispy/visuals/transforms/default_transform.py
+++ b/vispy/visuals/transforms/default_transform.py
@@ -1,0 +1,394 @@
+import numpy as np
+
+from vispy.visuals.transforms import BaseTransform, AffineTransform
+from vispy.util.quaternion import Quaternion
+from vispy.util import transforms
+
+# todo: rename to simply Transform
+
+class DefaultTransform(BaseTransform):
+    """ A transform that supports translation, rotation and scaling
+
+    Instances of this class keep track of the complexity of the
+    transformation, and internally describe it in a manner that allows
+    for the most effective composition and inversion.
+    
+    Parameters
+    ----------
+    scale : None | scalar | tuple of 3 scalars
+        The initial scaling for this transform. See ``scale()`` for
+        details,
+    rotate : None | quaternion
+        The initial rotation for this transform. See ``rotate()`` for
+        details.
+    translate : None | tuple of 3 scalars
+        The initial translation for this transform. See ``translate()``
+        for details.
+    matrix : None | 4x4 numpy array
+        The initial transform represented with a matrix. See
+        ``set_matrix()`` for details.
+    
+    Efficiency
+    ----------
+    If possible, the transformation is internally described using a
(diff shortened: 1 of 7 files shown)


almarklein:
ping @campagnola I don't think any of this made it into #1021? I (or someone else) should probably revive this after merging #1021.

Eric89GXL:
I don't think any of it did, either

vispy/visuals/transforms/default_transform.py
```diff
@@ -0,0 +1,394 @@
...
+    """
+    
+    Linear = True
+    
+    @property
+    def IsIdentity(self):
+        return self._is_identity
+    
+    @property
+    def IsShearless(self):
```
campagnola: Isn't this the same as `Orthogonal` ?

vispy/visuals/transforms/default_transform.py
```diff
@@ -0,0 +1,394 @@
...
+            qq = self._rotation
+            self._has_rotation = qq.w != 1.0 or any(tuple(qq)[1:])
+        else:
+            rotmat = q.get_matrix()
+            self._matrix = np.dot(self._matrix, rotmat)
+        
+        self.update()
+    
+    def scale(self, scale):
+        """ Scale the transform with a scalar value or a 3-element tuple.
```
campagnola: It would be nice to have the option of changing the `_scaling` and `_rotation` parameters directly, rather than having to rebuild the entire transform.

vispy/visuals/transforms/default_transform.py
almarklein: Yes, I think so. That sounds like a better name.

vispy/visuals/transforms/default_transform.py
almarklein: That won't work if the transform is stored as a matrix. I wonder if we should even allow it, because it can lead to unexpected results. If you want to scale afterwards, maybe you should two transform objects instead.

vispy/visuals/transforms/default_transform.py
campagnola: (I mean that Orthogonal already exists in the base transform class)

vispy/visuals/transforms/default_transform.py
campagnola: I think you're right.

djhoese:
So @almarklein I guess this is no longer worth it?

PR #928, opened 2015-05-25
Scenegraph overhaul

campagnola:
This PR is a continuation of #895 (and supersedes it). I am keeping both PRs open for now so that the prior changes may be reviewed in isolation. Major tasks for this PR are:

* [x] Remove multi-parenting in scenegraph
* [x] Nodes have a single transform (and related simplifications)
* [x] SceneCanvas.draw_visual should be simple and efficient
* [x] Implement picking (finally)
* [ ] ViewNode and a way to automatically create a view on a scenegraph subtree
* [x] Other simplifications:
    * [x] Remove all ViewBox clip methods except fragment. 
    * [x] Remove render/framebuffer/canvas scenegraph nodes (might restore these later if there turns out to be a need)

Visuals to migrate:

* [x] Line
* [x] Mesh
* [x] GridLines
* [x] Markers
* [x] Volume
* [x] Image
* [x] Text
* [x] Axis
* [x] Polygon
* [x] Box
* [x] Cube
* [x] Ellipse
* [x] Isocurve
* [x] Isosurface
* [x] Plane
* [x] Rectangle
* [x] RegularPolygon
* [x] Spectrogram
* [x] Tube
* [x] XYZAxis
* [x] Histogram
* [x] LinePlot
* [x] Isoline
* [x] SurfacePlot
* [x] ColorBar

EDIT BY Eric89GXL:

Related issues:
* [x] Toggling a node toggles children #986
* [x] Text visual changing viewport #981
* [x] Type of Canvas size #976
* [x] Centering of text visuals #975: working now?
* [x] Scene benchmarks broken #965: need to fix?
* [x] Depth drawing precision #953: need to add?
* [x] Get SceneCanvas from node #927: exists? 
* [x] Plane and box PR #901: need to update visuals
* [x] Original PR #895
* [x] Reuse FBO in viewbox #791
* [x] Picking example #693
* [x] Axes, ticks, grid #677
* [x] Grid plot with individual zooming #434: does it work?
* [x] Picking #140
* [ ] Line editing example PR #926: modify to use picking?
* [ ] Camera translate speed #914: need to add?
* [ ] Performance of the SceneGraph #628

closes #986
closes #981
closes #976
closes #975
closes #965
closes #953
closes #927
closes #791
closes #693
closes #677
closes #434
closes #140

PR #959, opened 2015-06-02
Added an implicitly connected grid mesh visual

inclement:
This PR replaces https://github.com/vispy/vispy/pull/756, which I accidentally closed with a bad push.

This isn't yet ready to merge, I need to fix the color parameter to be consistent with the surface plot.


Eric89GXL:
Travis is cranky because you have an un-caught warning:
```
Example basics/visuals/grid_mesh.py failed (0):
/home/travis/build/vispy/vispy/vispy/geometry/meshdata.py:336: RuntimeWarning: invalid value encountered in divide
  norm /= (norm**2).sum()**0.5  # and re-normalize
```

Eric89GXL:
...actually that points at a slightly deeper problem. Could you change that line to only divide the elements whose norm > 0?

inclement:
Ah, I was wondering how to placate travis, thanks.

Eric89GXL:
He is picky, but it's for our own good :)

inclement:
Okay, I think this is ready for review. The colors argument currently only takes ndarrays of rgb or rgba colors, but this is consistent with the surface_plot and maybe other visuals. I'd like to fix this to accept the range of arguments previously discussed, but I think it would be better in a separate PR so that I can demonstrate whatever I propose in several visuals at once, without mixing it with this new one.

Eric89GXL:
+1 for separate PR that goes through all the visuals to unify color handling. but that should probably wait until #928 is merged since it will change how the visuals work a bit under the hood

examples/basics/visuals/grid_mesh.py
```diff
@@ -0,0 +1,53 @@
+"""
+Demonstration of a mesh constructed in a grid about x,y,z coordinates.
+"""
```
Eric89GXL: I think in our other examples that require `scipy` we make a note of it at the top, can you `git grep scipy` to find them and do whatever we did in those here?

vispy/geometry/__init__.py
```diff
@@ -20,5 +20,5 @@
 from .calculations import (_calculate_normals, _fast_cross_3d,  # noqa
                            resize)  # noqa
 from .generation import (create_arrow, create_box, create_cone,  # noqa
-                         create_cube, create_cylinder, create_plane,  # noqa
-                         create_sphere)  # noqa
+                         create_cube, create_cylinder, create_grid_mesh,
+                         create_plane, create_sphere)  # noqa
```
Eric89GXL: please add to `__all__`

vispy/visuals/gridmesh.py
```diff
@@ -0,0 +1,95 @@
...
+        A 2d array of y coordinates for the vertices of the mesh. Must
+        have the same dimensions as xs and zs.
+    zs : ndarray
+        A 2d array of z coordinates for the vertices of the mesh. Must
+        have the same dimensions as xs and ys.
+    colors : ndarray
+        The colors of the points of the mesh. Should be either a
+        (width, height, 4) array of rgba colors at each grid point or
+        a (width, height, 3) array of rgb colors at each grid point.
+        Defaults to None, in which case
```
Eric89GXL: `None` should be in the `colors : ` line and this description should be finished

vispy/visuals/gridmesh.py
```diff
@@ -0,0 +1,95 @@
...
+    zs : ndarray
+        A 2d array of z coordinates for the vertices of the mesh. Must
+        have the same dimensions as xs and ys.
+    colors : ndarray
+        The colors of the points of the mesh. Should be either a
+        (width, height, 4) array of rgba colors at each grid point or
+        a (width, height, 3) array of rgb colors at each grid point.
+        Defaults to None, in which case 
+    shading : str | None
+        Same as for the `MeshVisual` class. Defaults to 'smooth'.
```
Eric89GXL: I think we usually list the `**kwargs` as well here

vispy/visuals/gridmesh.py
```diff
@@ -0,0 +1,95 @@
...
+                self.__vertices = None
+            self._zs = zs
+
+        if self.__vertices is None:
+            vertices, indices = create_grid_mesh(self._xs, self._ys, self._zs)
+            self.__meshdata.set_vertices(vertices)
+            self.__meshdata.set_faces(indices)
+
+        if colors is not None:
+            self.__colors = colors
```
Eric89GXL: I don't think this is ever used...?

Eric89GXL: If you copied it from another visual, I think we recently figured out it was there in error

Eric89GXL:
example works well for me, so other than my comments LGTM

vispy/visuals/gridmesh.py
```diff
@@ -0,0 +1,95 @@
...
+                self.__vertices = None
+            self._ys = ys
+
+        if zs is not None:
+            if self._zs is not None:
+                self.__vertices = None
+            self._zs = zs
+
+        if self.__vertices is None:
+            vertices, indices = create_grid_mesh(self._xs, self._ys, self._zs)
```
Eric89GXL: I think this way of operating is a bit unsafe. What happens if `xs` or `ys` is the wrong shape? now we have overwritten our `self._xs` and `self._ys` with potentially bad values. It seems like it would be cleaner to do creation in one place (that does input validation), and then set all `self` attributes. For example:
```
xs = xs or self._xs
ys = ys or self._ys
zs = zs or self._zs
vertices, indices = create_grid_mesh(xs, ys, zs)
# now that this has passed, we are safe and can set all attributes
self._xs, self._ys, self._zs = xs, ys, zs
self.__meshdata...
```

inclement: Ah, I see, that's a better idea. I actually followed the lead of SurfacePlot here, but noticed it has the same issue - and actually in both cases I'm not sure it's even possible to set_data if you want to change the mesh size because the colors array is already in the MeshData with a certain shape. Maybe it would be better to explicitly check the sizes of everything...I'll think about it and implement at least your fix.

examples/basics/visuals/grid_mesh.py
inclement: I could only find that the tests have a 'requires_scipy' decorator, and that one example in demo/gloo/unstructured_2d.py has this information in a comment at the top (but other examples don't seem to follow this pattern). Should I do something like the latter, or am I missing something?

Eric89GXL: argh, I thought we had better documentation than that. Don't worry about it for now, then. Sorry for the noise.

Eric89GXL:
@inclement I think the only outstanding thing is the "unification" / checking issue, feel free to fix `SurfacePlot` while you're at it if you want as well. Ping me whenever it's ready for review again.

PR #960, opened 2015-06-02
[WIP] ENH: Use new QOpenGL API with Qt 5.4+.

QuLogic:
This is not really complete yet; it's really more for documenting the change since it doesn't work just yet. Most of the visual tests fail, not all the others do somehow.

Apparently, the new [QOpenGLWidget embeds better and is more compatible than the old QGLWidget](https://blog.qt.io/blog/2014/09/10/qt-weekly-19-qopenglwidget/). Unfortunately, there are some differences, so it may not integrate into vispy quite so easily.
- [QGLWidget](http://doc.qt.io/qt-4.8/qglwidget.html) -> [QOpenGLWidget](https://doc-snapshots.qt.io/qt5-5.4/qopenglwidget.html)
- [QGLContext](http://doc.qt.io/qt-4.8/qglcontext.html) -> [QOpenGLContext](http://doc.qt.io/qt-5/qopenglcontext.html)
- [QGLFormat](http://doc.qt.io/qt-4.8/qglformat.html) -> [QSurfaceFormat](http://doc.qt.io/qt-5/qsurfaceformat.html)

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -304,7 +304,11 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
         # Qt supports OS double-click events, so we set this here to
         # avoid double events
         self._double_click_supported = True
-        self._physical_size = p.size
+
+        # We call this here to set the physical size. We do this rather than set
+        # the physical size directly, because resizeGL takes into account any
+        # non-unity device pixel ratio.
+        self.resizeGL(*p.size)
 
         # Activate touch and gesture.
         # NOTE: we only activate touch on OS X because there seems to be
@@ -727,6 +731,12 @@ class CanvasBackendDesktop(QtBaseCanvasBackend, QGLWidget):
     def resizeGL(self, w, h):
         if self._vispy_canvas is None:
             return
+        if qt_lib == 'pyqt5':
+            # We take into account devicePixelRatio, which is non-unity on
+            # e.g HiDPI displays.
+            ratio = self.devicePixelRatio()
+            w = w * ratio
+            h = h * ratio
         self._vispy_set_physical_size(w, h)
         self._vispy_canvas.events.resize(size=(self.width(), self.height()),
                                          physical_size=(w, h))
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -713,6 +713,12 @@ class CanvasBackendDesktop(QtBaseCanvasBackend, QGLWidget):
         else:
             self.swapBuffers()
 
+    def _vispy_get_fb_bind_location(self):
+        if QT5_NEW_API:
+            return self.defaultFramebufferObject()
+        else:
+            return QtBaseCanvasBackend._vispy_get_fb_bind_location(self)
+
     def initializeGL(self):
         if self._vispy_canvas is None:
             return
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -26,6 +26,7 @@ import os
 import sys
 import atexit
 import ctypes
+from distutils.version import LooseVersion
 
 from ...util import logger
 from ..base import (BaseApplicationBackend, BaseCanvasBackend,
@@ -66,6 +67,7 @@ def _check_imports(lib):
 # Get what qt lib to try. This tells us wheter this module is imported
 # via _pyside or _pyqt4 or _pyqt5
 QGLWidget = object
+QT5_NEW_API = False
 if qt_lib == 'pyqt4':
     _check_imports('PyQt4')
     if not USE_EGL:
@@ -75,7 +77,13 @@ if qt_lib == 'pyqt4':
 elif qt_lib == 'pyqt5':
     _check_imports('PyQt5')
     if not USE_EGL:
-        from PyQt5.QtOpenGL import QGLWidget, QGLFormat
+        from PyQt5.QtCore import PYQT_VERSION_STR
+        if LooseVersion(PYQT_VERSION_STR) >= '5.4.0':
+            from PyQt5.QtWidgets import QOpenGLWidget as QGLWidget
+            from PyQt5.QtGui import QSurfaceFormat as QGLFormat
+            QT5_NEW_API = True
+        else:
+            from PyQt5.QtOpenGL import QGLWidget, QGLFormat
     from PyQt5 import QtGui, QtCore, QtWidgets, QtTest
     QWidget, QApplication = QtWidgets.QWidget, QtWidgets.QApplication  # Compat
 elif qt_lib == 'pyside':
@@ -191,15 +199,21 @@ def _set_config(c):
     glformat.setGreenBufferSize(c['green_size'])
     glformat.setBlueBufferSize(c['blue_size'])
     glformat.setAlphaBufferSize(c['alpha_size'])
-    glformat.setAccum(False)
-    glformat.setRgba(True)
-    glformat.setDoubleBuffer(True if c['double_buffer'] else False)
-    glformat.setDepth(True if c['depth_size'] else False)
+    if QT5_NEW_API:
+        # Qt5 >= 5.4.0 - below options automatically enabled if nonzero.
+        glformat.setSwapBehavior(glformat.DoubleBuffer if c['double_buffer']
+                                 else glformat.SingleBuffer)
+    else:
+        # Qt4 and Qt5 < 5.4.0 - buffers must be explicitly requested.
+        glformat.setAccum(False)
+        glformat.setRgba(True)
+        glformat.setDoubleBuffer(True if c['double_buffer'] else False)
+        glformat.setDepth(True if c['depth_size'] else False)
+        glformat.setStencil(True if c['stencil_size'] else False)
+        glformat.setSampleBuffers(True if c['samples'] else False)
     glformat.setDepthBufferSize(c['depth_size'] if c['depth_size'] else 0)
-    glformat.setStencil(True if c['stencil_size'] else False)
     glformat.setStencilBufferSize(c['stencil_size'] if c['stencil_size']
                                   else 0)
-    glformat.setSampleBuffers(True if c['samples'] else False)
     glformat.setSamples(c['samples'] if c['samples'] else 0)
     glformat.setStereo(c['stereo'])
     return glformat
@@ -211,6 +225,9 @@ class ApplicationBackend(BaseApplicationBackend):
 
     def __init__(self):
         BaseApplicationBackend.__init__(self)
+        if QT5_NEW_API:
+            # For Qt5 >= 5.4.0 - Enable sharing of context between windows.
+            QApplication.setAttribute(QtCore.Qt.AA_ShareOpenGLContexts, True)
 
     def _vispy_get_backend_name(self):
         name = QtCore.__name__.split('.')[0]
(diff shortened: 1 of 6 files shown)


vispy/app/backends/_qt.py
```diff
@@ -179,15 +185,21 @@ def _set_config(c):
     glformat.setGreenBufferSize(c['green_size'])
     glformat.setBlueBufferSize(c['blue_size'])
     glformat.setAlphaBufferSize(c['alpha_size'])
-    glformat.setAccum(False)
-    glformat.setRgba(True)
-    glformat.setDoubleBuffer(True if c['double_buffer'] else False)
-    glformat.setDepth(True if c['depth_size'] else False)
+    if hasattr(glformat, 'setAccum'):
```
QuLogic: These toggles don't seem to exist. I think they are just automatically enabled when you set the size > 0.

vispy/app/backends/_qt.py
```diff
@@ -581,25 +596,33 @@ def _init_specific(self, p, kwargs):
...
         else:
             hint = QtCore.Qt.Widget  # can also be a window type
-        QGLWidget.__init__(self, glformat, p.parent, widget, hint)
+        if hasattr(QGLWidget, 'isSharing'):
+            # Qt4 and Qt5 < 5.4.0
+            QGLWidget.__init__(self, p.parent, widget, hint)
+        else:
+            # Qt5 >= 5.4.0
+            QGLWidget.__init__(self, p.parent, hint)
+        self.setFormat(glformat)
```
QuLogic: `QOpenGLWidget` doesn't accept this directly in the constructor.

vispy/app/backends/_qt.py
```diff
@@ -581,25 +596,33 @@ def _init_specific(self, p, kwargs):
...
         self._initialized = True
-        if not self.isValid():
+        if hasattr(QGLWidget, 'isSharing') and not self.isValid():
```
QuLogic: `QOpenGLWidget` context does not become valid until being shown.

vispy/app/backends/_qt.py
```diff
@@ -581,25 +596,33 @@ def _init_specific(self, p, kwargs):
...
             raise RuntimeError('context could not be created')
-        self.setAutoBufferSwap(False)  # to make consistent with other backends
+        if hasattr(self, 'setAutoBufferSwap'):
+            self.setAutoBufferSwap(False)  # to make consistent with other backends
```
QuLogic: I can't find a way to disable automatic swapping; it's probably causing some of the troubles.

vispy/app/backends/_qt.py
```diff
@@ -581,25 +596,33 @@ def _init_specific(self, p, kwargs):
...
 
-        # first arg can be glformat, or a gl context
         if p.always_on_top or not p.decorate:
             hint = 0
             hint |= 0 if p.decorate else QtCore.Qt.FramelessWindowHint
             hint |= QtCore.Qt.WindowStaysOnTopHint if p.always_on_top else 0
         else:
             hint = QtCore.Qt.Widget  # can also be a window type
-        QGLWidget.__init__(self, glformat, p.parent, widget, hint)
+        if hasattr(QGLWidget, 'isSharing'):
```
QuLogic: Sharing is automatic for all widgets in the same window.

vispy/app/backends/_qt.py
```diff
@@ -199,6 +211,9 @@ class ApplicationBackend(BaseApplicationBackend):
 
     def __init__(self):
         BaseApplicationBackend.__init__(self)
+        if hasattr(QtCore.Qt, 'AA_ShareOpenGLContexts'):
+            # For Qt5 >= 5.4.0
+            QApplication.setAttribute(QtCore.Qt.AA_ShareOpenGLContexts, True)
```
QuLogic: This enables context sharing for widgets not on the same window.

QuLogic:
Also, the default framebuffer may not be 0 for a context, so that may also contribute to the problem if we are binding the framebuffer to 0.

vispy/app/backends/_qt.py
Eric89GXL: that would make sense

Eric89GXL:
Thanks for all the notes. I think they should actually go in the code as comments so that e.g. the next person doesn't errantly undo your changes

QuLogic:
OK, added the comments. I still need to figure out if I can use more manual buffer swapping. Also, is there a way for the backend to override `glBindFramebuffer` in some way? Or maybe I should add some internal method to query the framebuffer for a backend?

Eric89GXL:
No way to override currently, no. Do we need to do something special for Qt 5.4+?

QuLogic:
Yea, sorry, I forgot to link it, but it says to never use 0:
> All rendering happens into an OpenGL framebuffer object. makeCurrent() ensure that it is bound in the context. Keep this in mind when creating and binding additional framebuffer objects in the rendering code in paintGL(). Never re-bind the framebuffer with ID 0. Instead, call defaultFramebufferObject() to get the ID that should be bound.

Eric89GXL:
Well that's a bummer :( I think that somewhere within GLIR it knows which context it's a part of. I suppose we could add a private property `._fb_bind_location` to the `GLContext` class (I think?) that would be `0` for all other backends. @almarklein does that make sense to you?

almarklein:
Yes, that makes sense.

QuLogic:
OK, but I don't see how `GLContext` is linked back to a `Canvas`, since that is the level at which Qt is?

Eric89GXL:
Canvas has a .context property (or something like it), so whenever that is
instantiated is where I'd look to pass information.

QuLogic:
OK, before going off on a tangent, I did a quick check. The `defaultFrameBuffer` call returns 0 in the constructor, but once you get to the paint callback, it returns 1. Probably it gets changed when the window is shown as that is when the context is created. In any case, this is likely the reason I get a completely black window when using this new backend setup, and I'll need to implement a property like the above.

QuLogic:
Success, somewhat. I ~~hacked~~ upgraded the `CURRENT` GLIR command to take an argument for the framebuffer  (though maybe I should have just changed `id_`?). Now instead of pretty much every thing failing, I just get some sporadic failures. I _think_ that's because I can't disable automatic buffer swapping on this widget.

I also changed the original commit to better indicate which conditions are based on the Qt version.

QuLogic:
Strangely, the remaining failures now are not on image diffs, but shaders that are randomly not compiling. And the program log is empty, too. Must be some other thing Qt5 is modifying somewhere...

vispy/app/backends/_qt.py
```diff
@@ -626,7 +652,17 @@ def _vispy_swap_buffers(self):
         # Swap front and back buffer
         if self._vispy_canvas is None:
             return
-        self.swapBuffers()
+        if QT5_NEW_API:
+            ctx = self.context()
+            ctx.swapBuffers(ctx.surface())
```
Eric89GXL: I wonder if somehow repeated calls to these is problematic, you could try storing them in a private attribute of `self`

vispy/gloo/context.py
```diff
@@ -166,7 +166,12 @@ def flush_commands(self, event=None):
         """
         if self._do_CURRENT_command:
             self._do_CURRENT_command = False
-            self.shared.parser.parse([('CURRENT', 0)])
+            canvas = get_current_canvas()
+            if canvas and hasattr(canvas, '_backend'):
+                fbo = canvas._backend._vispy_get_fb_bind_location()
+            else:
+                fbo = 0
+            self.shared.parser.parse([('CURRENT', 0, fbo)])
```
Eric89GXL: so I guess before the implied/effective call was `[('CURRENT', 0, 0)]`, even though it was written as `[('CURRENT', 0)]`?

QuLogic: Well, before there was no argument (no last `0`), but effectively yes. That being said, I could replace the first 0 with `fbo` instead. I just wasn't certain if the `id_` had some special meaning for GLIR.

Eric89GXL: yeah I'm not 100% sure how this stuff works, we might have to wait until @almarklein pops up again to get some insight if you can't get it from the code

almarklein: This looks good to me, but we need to update the GLIR spec in the wiki. I suspect the vispyjs implementation will just continue to work ok. cc @rossant

QuLogic: Ah, I didn't know about that wiki page. It says:
> `<ID>` is an integer unique within the current GL context that is used as a reference to a GL object.

If I read that correctly, I think that means I can just replace the hard-coded `0` with the fbo?

almarklein: Ah, yes, I think we can just do that :)

astrofrog:
@QuLogic - just for info, I tried this branch and I have an issue when trying to run the volume example on a HiDPI display:

screen shot 2016-09-19 at 14 11 18

I guess that somewhere, we need to check what the device resolution is?

hadim:
Any news here ? Is it possible to use vispy with PyQt 5.7 ?

QuLogic:
This PR only makes sure vispy uses non-deprecated functionality; unless PyQt 5.7 deleted stuff, I don't see why you can't use vispy with it already.

I rebased this today, but there are still issues with it. I _think_ the problem is that with the new widget, Qt manages the framebuffer but vispy assumes it is always attached in the same way it last left things. Every once in a while, Qt will swap out something and vispy's assumptions are no longer true, so we end up with randomly failing stuff.

As for HiDPI displays, I think vispy already has scaling support, but I guess the Qt backend doesn't implement it properly. I don't have such a display, so I couldn't say.

Eric89GXL:
@QuLogic do you plan to continue working on this, or should someone else take it over if they have time?

QuLogic:
I think it's most of the way there; the problem, as I laid out before, is that it seems like Qt is swapping out some buffer that vispy assumes is still in use. Unfortunately, trying to figure out where those assumptions are being made is not an easy task, so any help with that would not be unwelcome.

Eric89GXL:
One quick thing I notice in the Qt 5 docs is that it says:

> The initialization of OpenGL rendering state, etc. should be done by overriding the initializeGL() function, rather than in the constructor of your QGLWidget subclass.

So maybe we are setting some properties in the wrong place

Eric89GXL:
Also maybe related:

http://www.qtcentre.org/threads/61820-Using-QOpenGLWidget-swapBuffers

Eric89GXL:
And also good reading:

http://stackoverflow.com/questions/21722852/difference-in-opengl-speed-between-qt-4-5-and-opengl-api
http://stackoverflow.com/questions/15671816/opengl-vs-qopengl-qtopengl-in-qt-5-differences-and-limitations

But I assume you have seen and explored most of these @QuLogic...?

QuLogic:
> So maybe we are setting some properties in the wrong place

This might be possible; I think somethings are set in some common method that is called by `__init__`. I'm not sure if that would cause the flaky errors.

The other links are interesting reading, but I think nothing that's there is being done wrong on our side, though I could be wrong.

astrofrog:
@QuLogic - I opened a PR to your branch to fix things on HiDPI displays. With this, I can successfully run the ``volume.py`` example on my Retina mac.

QuLogic:
@astrofrog [This build](https://travis-ci.org/vispy/vispy/jobs/167166001) seems to show that maybe calling `resizeGL` at that point is too early?

astrofrog:
@QuLogic - ok, I need to investigate. We do need to make sure the setting of the physical size is done as in resizeGL basically, but maybe the last call in resizeGL would need to be skipped that early on.

astrofrog:
@QuLogic - can you try replacing the call to ``resizeGL`` I added by simply:

```
        if qt_lib == 'pyqt5':
            # We take into account devicePixelRatio, which is non-unity on
            # e.g HiDPI displays.
            ratio = self.devicePixelRatio()
            w = w * ratio
            h = h * ratio
        self._vispy_set_physical_size(w, h)
```

? (it's the last line in resizeGL that is causing the issue)

Eric89GXL:
On `master` I can run examples using PyQt5 fine, but on this PR I get this error (Ubuntu 64-bit):
```
larsoner@bunk:~/custombuilds/vispy$ python -i examples/basics/plotting/spectrogram.py
>>> WARNING: QOpenGLShaderProgram: could not create shader program
WARNING: QOpenGLShader: could not create shader
WARNING: bool QOpenGLTextureBlitter::create() Could not link shader program:
 ""
```
I get a different set of errors on Anaconda Python, but I think it's a problem with [their build of Qt] (http://stackoverflow.com/questions/40400045/anaconda-and-python-qt-5-6-with-opengl-support).

astrofrog:
Just to say that I'm using a version of this PR for glue (I bundle a fork of vispy in the package so I've included the changes here) and it works fine for markers and volume rendering - so at least it's on the right track, but indeed there are some failing examples still.

PR #961, opened 2015-06-03
FIX: Restore missing values in hex-to-decimal dict.

QuLogic:
`itertools.combinations` != double loop comprehension. `combinations` excludes doubled entries, so values like '00' did not exist. This fixes the broken tiger example (are these not tested on Travis?).

Possibly the reason why it seemed in #932 that the explicit loop was twice as fast may be due to the fact that this code produces only 231 entries, when there should be 484.


Eric89GXL:
@QuLogic no, the collections examples are not currently tested. That should change in the coming months as we make more progress integrating the scenegraph and collections systems. In the meantime, thanks for the fix!

PR #964, opened 2015-06-03
Include Eric Larson

outofculture:
Maybe a test should be added asserting that the listings of core contributors all match each other.


campagnola:
And we also need a unit test to make sure the contributor-checking unit test is correctly checking all of the contributor lists.

Eric89GXL:
Hah! I didn't even realize we had a list there...

Eric89GXL:
thanks for the tweak @outofculture

PR #968, opened 2015-06-03
Add new signature to `message_handler` for pyqt5

hadim:
A small bug fix in Qt backend.

In pyqt5 `QtCore.qInstallMessageHandler` takes a function with a different signature than pyqt4. The result was that all pyqt5 log message would display `TypeError: message_handler() takes 2 positional arguments but 3 were given`.

This PR fix the issue by checking wether `qt_lib` is pyqt4 or pyqt5.

PS: I also add a `.decode()` to pyqt4 logging message signature because it's return a byte and not an str. However it could raise issue with Python 2. This code has only been tested with Python 3.4.

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -155,7 +155,7 @@ def message_handler(*args):
                "event."):
         return
     else:
-        if qt_lib == "pyqt4":
+        if qt_lib in ("pyqt4", "pyside"):
             logger.warning(msg.decode())
         elif qt_lib == "pyqt5":
             logger.warning(msg)
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -140,7 +140,7 @@ BUTTONMAP = {0: 0, 1: 1, 2: 2, 4: 3, 8: 4, 16: 5}
 # Also, ignore spam about tablet input
 def message_handler(*args):
 
-    if qt_lib == "pyqt4":
+    if qt_lib in ("pyqt4", "pyside"):
         msg_type, msg = args
     elif qt_lib == "pyqt5":
         msg_type, context, msg = args
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -138,14 +138,26 @@ BUTTONMAP = {0: 0, 1: 1, 2: 2, 4: 3, 8: 4, 16: 5}
 
 # Properly log Qt messages
 # Also, ignore spam about tablet input
-def message_handler(msg_type, msg):
+def message_handler(*args):
+
+    if qt_lib == "pyqt4":
+        msg_type, msg = args
+    elif qt_lib == "pyqt5":
+        msg_type, context, msg = args
+    elif qt_lib:
+        raise RuntimeError("Invalid value for qt_lib %r." % qt_lib)
+    else:
+        raise RuntimeError("Module backends._qt should not be imported directly.")
+
     if msg == ("QCocoaView handleTabletEvent: This tablet device is "
                "unknown (received no proximity event for it). Discarding "
                "event."):
         return
     else:
-        logger.warning(msg)
-
+        if qt_lib == "pyqt4":
+            logger.warning(msg.decode())
+        elif qt_lib == "pyqt5":
+            logger.warning(msg)
 try:
     QtCore.qInstallMsgHandler(message_handler)
 except AttributeError:
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -32,6 +32,7 @@ from ..base import (BaseApplicationBackend, BaseCanvasBackend,
                     BaseTimerBackend)
 from ...util import keys
 from ...ext.six import text_type
+from ...ext.six import string_types
 from ... import config
 from . import qt_lib
 
@@ -155,10 +156,8 @@ def message_handler(*args):
                "event."):
         return
     else:
-        if qt_lib in ("pyqt4", "pyside"):
-            logger.warning(msg.decode())
-        elif qt_lib == "pyqt5":
-            logger.warning(msg)
+        msg = msg.decode() if not isinstance(msg, string_types) else msg
+        logger.warning(msg)
 try:
     QtCore.qInstallMsgHandler(message_handler)
 except AttributeError:
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -147,7 +147,8 @@ def message_handler(*args):
     elif qt_lib:
         raise RuntimeError("Invalid value for qt_lib %r." % qt_lib)
     else:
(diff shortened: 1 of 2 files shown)


hadim:
A small script to test this : 

```python
import vispy
from vispy import app

qt_lib = "pyqt4"
qtapp = app.use_app(qt_lib)

QtCore = qtapp.backend_module.QtCore

if qt_lib == "pyqt4":
    QApplication = qtapp.backend_module.QtGui.QApplication
elif qt_lib == "pyqt5":
    QApplication = qtapp.backend_module.QtWidgets.QApplication

app = QApplication([''])

QtCore.qDebug("Debug message")
QtCore.qCritical("Critical message")
```

hadim:
Last thing, you may also want to catch `msg_type` and bind `QtMsgType` (http://doc.qt.io/qt-5/qtglobal.html#QtMsgType-enum) to python logging level (https://docs.python.org/3.4/library/logging.html#logging.Logger.info).

But that's another topic.

hadim:
The last commit fix some compatibility issue in `vispy/app/qt.py`.

It basically adds this :+1: 

```python
from .backends import qt_lib

if qt_lib == 'pyqt4':
    from PyQt4 import QtGui, QtCore
    QWidget, QGridLayout = QtGui.QWidget, QtGui.QGridLayout  # Compat
elif qt_lib == 'pyqt5':
    from PyQt5 import QtGui, QtCore, QtWidgets
    QWidget, QGridLayout = QtWidgets.QWidget, QtWidgets.QGridLayout  # Compat
elif qt_lib == 'pyside':
    from PySide import QtGui, QtCore
    QWidget, QGridLayout = QtGui.QWidget, QtGui.QGridLayout  # Compat
elif qt_lib:
    raise RuntimeError("Invalid value for qt_lib %r." % qt_lib)
else:
    raise RuntimeError("Module backends._qt should not be imported directly.")
```

It's maybe an heavy solution for later if you want to add more Qt related widgets ...

vispy/app/backends/_qt.py
```diff
@@ -138,14 +138,27 @@ def _check_imports(lib):
 
 # Properly log Qt messages
 # Also, ignore spam about tablet input
-def message_handler(msg_type, msg):
+def message_handler(*args):
+
+    if qt_lib == "pyqt4":
+        msg_type, msg = args
+    elif qt_lib == "pyqt5":
+        msg_type, context, msg = args
```
Eric89GXL: what about `pyside`? do we not need to set the message handler for it?

vispy/app/qt.py
```diff
@@ -14,13 +16,28 @@
...
-    
+if qt_lib == 'pyqt4':
+    from PyQt4 import QtGui
+    QWidget, QGridLayout = QtGui.QWidget, QtGui.QGridLayout  # Compat
+elif qt_lib == 'pyqt5':
+    from PyQt5 import QtGui, QtWidgets
+    QWidget, QGridLayout = QtWidgets.QWidget, QtWidgets.QGridLayout  # Compat
+elif qt_lib == 'pyside':
+    from PySide import QtGui
+    QWidget, QGridLayout = QtGui.QWidget, QtGui.QGridLayout  # Compat
```
Eric89GXL: this can be combined with `pyqt4` as `if qt_lib in ('pyqt4', 'pyside')` right? should be more DRY

Eric89GXL:
Other than my minor gripes LGTM -- thanks for working on these fixes!

vispy/app/backends/_qt.py
hadim: Is that ok for you if we assume the signature is the same than for `pyqt4` ?

vispy/app/qt.py
hadim: Ok

vispy/app/backends/_qt.py
```diff
@@ -138,14 +138,27 @@ def _check_imports(lib):
...
+
     if msg == ("QCocoaView handleTabletEvent: This tablet device is "
                "unknown (received no proximity event for it). Discarding "
                "event."):
         return
     else:
-        logger.warning(msg)
-
+        if qt_lib == "pyqt4":
+            logger.warning(msg.decode())
```
Eric89GXL: again don't forget pyside

Eric89GXL:
And yeah, I think pyside and pyqt4 should be the same, since we weren't triaging in `_qt` based on which one was loaded before

vispy/app/backends/_qt.py
```diff
@@ -138,14 +138,27 @@ def _check_imports(lib):
...
                "unknown (received no proximity event for it). Discarding "
                "event."):
         return
     else:
-        logger.warning(msg)
-
+        if qt_lib in ("pyqt4", "pyside"):
+            logger.warning(msg.decode())
+        elif qt_lib == "pyqt5":
+            logger.warning(msg)
```
almarklein: What I like in a case like this is `msg = msg.decode() if not isinstance(msg, string_types)` (assuming `msg` is never `None`)

almarklein:
Other than that minor comment, LGTM. Travis is complaining about `vispy/app/qt.py:23:5: F401 'QtGui' imported but unused`

vispy/app/backends/_qt.py
hadim: Could you please elaborate ? What should I put for `string_types` since `unicode` only exist in Python 2.

Eric89GXL: do a `git grep string_types` and you'll see the other uses, it's in `vispy/ext/six` so something like `from ...ext.six import string_types`

hadim:
Very nice !

If travis is happy, I let you merge it.

Thanks for the review guys.

Eric89GXL:
Thanks for the fixes @hadim!

dzhoshkun:
I'm still experiencing this issue when running `python -c "from vispy import app; c = app.Canvas(keys='interactive')"`.

I have the following versions:
* `python --version` => `3.5.2`
* PyQt5 version => `5.9.1`
* VisPy version => `0.4.0`

Any ideas?

davidh-ssec:
@dzhoshkun This fix was not made available in the 0.4.0 version of vispy (at least from what I can tell). You should try installing vispy from the github master branch. Until we complete all of the pending issues for a 0.5.0 release, installing from github is the only way to get the best working version of vispy.

dzhoshkun:
@davidh-ssec Thank you very much for the quick reply! This has solved the problem.

Miiike89:
@dzhoshkun Had the same problem. Worked for me too. Thanks a lot.

By the way, are there any rumors, when version 0.50 will be released?

davidh-ssec:
@Miiike89 Hopefully this month (if not the next week) if nothing else gets in the way. See https://github.com/vispy/vispy/milestone/5

Miiike89:
Thanks fior the quick reply. Looking forward to 0.5.0.

PR #971, opened 2015-06-03
FIX: More font tests

### LINKED ISSUE #970: Test failure in test_font_glyph (possible race condition)
Intermittently, I get:

```
====================================================== FAILURES ======================================================
__________________________________________________ test_font_glyph ___________________________________________________

    def test_font_glyph():
        """Test loading glyphs"""
        # try both a vispy and system font
        sys_fonts = set(list_fonts()) - set(_vispy_fonts)
        assert_true(len(sys_fonts) > 0)
        for face in ('OpenSans', list(sys_fonts)[0]):
            font_dict = dict(face=face, size=12, bold=False, italic=False)
            glyphs_dict = dict()
            chars = 'foobar^C&#'
            if face != 'OpenSans' and os.getenv('APPVEYOR', '').lower() == 'true':
                continue  # strange system font failure
            if 'true' in os.getenv('TRAVIS', '') and sys.version[0] == '3':
                continue  # as of April 2015 strange FontConfig error on Travis
            for char in chars:
                # Warning that Arial might not exist
>               _load_glyph(font_dict, char, glyphs_dict)

vispy/util/fonts/tests/test_font.py:40: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
vispy/util/fonts/_freetype.py:48: in _load_glyph
    face.set_char_size(f['size'] * 64)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <vispy.ext.freetype.Face object at 0x2af9dc656208>, width = 768, height = 0, hres = 72, vres = 72

    def set_char_size(self, width=0, height=0, hres=72, vres=72):
        error = FT_Set_Char_Size(self._FT_Face, width, height, hres, vres)
        if error:
>           raise RuntimeError('Could not set size: %s' % hex(error))
E           RuntimeError: Could not set size: 0x17

vispy/ext/freetype.py:397: RuntimeError
```

Automating this test 10,000 times passes each time but running it manually (or via the test suite) seems to fail occasionally with the same error - I've repeated it twice.

Ubuntu 14.04 LTS, AwesomeWM (tiling window manager), NVidia GTX 980, Anaconda Python 3.4.

Eric89GXL:
Closes #970

cc @nippoo

### DIFF
--- a/vispy/ext/fontconfig.py
+++ b/vispy/ext/fontconfig.py
@@ -113,6 +113,6 @@ def find_font(face, bold, italic):
 
 def _list_fonts():
     """List system fonts"""
-    stdout_, stderr = run_subprocess(['fc-list', ':', 'family'])
+    stdout_, stderr = run_subprocess(['fc-list', ':scalable=true', 'family'])
     vals = [v.split(',')[0] for v in stdout_.strip().splitlines(False)]
     return vals
--- a/vispy/ext/freetype.py
+++ b/vispy/ext/freetype.py
@@ -30,6 +30,7 @@ FT_LOAD_NO_AUTOHINT = 32768
 FT_LOAD_TARGET_LCD = 196608
 FT_LOAD_TARGET_LIGHT = 65536
 FT_LOAD_NO_SCALE = 1
+FT_FACE_FLAG_SCALABLE = 1
 
 _64_bit = (8 * struct.calcsize("P")) == 64
 ##############################################################################
--- a/vispy/util/fonts/_freetype.py
+++ b/vispy/util/fonts/_freetype.py
@@ -26,7 +26,7 @@ _font_dict = {}
 # and isn't using fonts (Windows)
 
 def _load_font(face, bold, italic):
-    from ...ext.freetype import Face
+    from ...ext.freetype import Face, FT_FACE_FLAG_SCALABLE
     key = '%s-%s-%s' % (face, bold, italic)
     if key in _font_dict:
         return _font_dict[key]
@@ -35,6 +35,9 @@ def _load_font(face, bold, italic):
     else:
         fname = find_font(face, bold, italic)
     font = Face(fname)
+    if (FT_FACE_FLAG_SCALABLE & font.face_flags) == 0:
+        raise RuntimeError('Font %s is not scalable, so cannot be loaded'
+                           % face)
     _font_dict[key] = font
     return font
 
--- a/vispy/util/fonts/_win32.py
+++ b/vispy/util/fonts/_win32.py
@@ -23,7 +23,7 @@ from ...ext.gdi32plus import (gdiplus, gdi32, user32, winreg, LOGFONT,
 
 # XXX This isn't perfect, but it should work for now...
 
-def find_font(face, bold, italic):
+def find_font(face, bold, italic, orig_face=None):
     style_dict = {'Regular': 0, 'Bold': 1, 'Italic': 2, 'Bold Italic': 3}
 
     # Figure out which font to actually use by trying to instantiate by name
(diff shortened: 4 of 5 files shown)


nippoo:
Ha, you got there before me!

Eric89GXL:
I'm not on Linux currently so I can't debug the scalable property, so I expect this PR to fail currently. Do you have time to look into it?

nippoo:
Yeah, this PR fails every time for me, consistently, with the same error. I'm not sure what you want to do with it, though?  Not all fonts are resizeable...?

Eric89GXL:
I think we should only enumerate fonts that are scalable

nippoo:
Right. Do we want to catch and handle the error gracefully in `_freetype.py`? What do we want to do if the user requests a non-scaleable font?

Eric89GXL:
I think we should throw an error if they request a non-scalable font, but that is of secondary importance I guess. So the full changes would require changes to the font-enumerating code, and changes to freetype to check the scalability during load.

Eric89GXL:
@nippoo you taking a stab at the font stuff, or should I go for it?

nippoo:
Hey, I'm happy to debug tomorrow (I'm home now - not at a Linux box) - what exactly d'you want me to do? Find some way of removing non-scalable fonts from `face` in the test? Or just `try/catch`ing and ignoring that specific error in the test?

Eric89GXL:
Yeah, the first order would be to eliminate non-scalable fonts from the font list enumeration code. That should fix the test that is now failing.

If you really want, you can also make it so that if a user requests a font, check if it's not-scaling and throw an error. We could add such a font to one of our repos so it can be tested, too.

nippoo:
Hm, okay. I'll have a look tomorrow to see if there's an easy way to distinguish non-scalable from scalable fonts without loading them and catching the error!

Eric89GXL:
There really should be a way to query the properties. You might need to actually load each font, but it should be very fast.

nippoo:
Okay, so you want to `_load_font()` and test whether the `num_fixed_sizes == 0` (this seems to be exposed in freetype: fonts which fail return 1; and fonts which pass are 0.

Does this just fail on Linux, or has it failed on any other platform (namely Windows with Freetype)? Should I just put in a conditional to check the correct OS, or will it fail similarly for certain fonts on Mac (that you know of)?

Eric89GXL:
IIRC we use freetype on all operating systems, so you shouldn't need to
triage by OS at that level. You might need to add something to the global
font enumeration function (which calls an os specific font finding
function) to cull the initial list of found fonts, though. Make sense?

Eric89GXL:
Nope, we definitely use Quartz on OSX. So I guess for now you just need to modify windows and Linux. I've never encountered the issue on Windows or OSX, though, so maybe just modifying the font-finding and font-loading functions on Linux is enough for now.

Eric89GXL:
If you want to be really complete you can even add a font that you know will fail to https://github.com/vispy/test-data, and add a test that if the sys is linux, it downloads it, tries to load it, and uses `assert_raises`. You know, if you want to go the extra mile.

Eric89GXL:
ping @nippoo do you think you'll have time for this soon? if not, I can do it

Eric89GXL:
Latest commit makes it so that only scalable fonts are enumerated, and then I manually verified that attempting to load a non-scalable font throws an error (too hard to get Travis to test this).

Ready for review/merge from my end.

nippoo:
Agh sorry - I completely forgot about this.

I'll test it tomorrow if you like, though the code looks like it should work just fine...

Eric89GXL:
You can if you want, but Travis being happy after my change suggests it should work.

PR #976, opened 2015-06-03
Fix type of canvas size.

QuLogic:
It must be a tuple to match other usage of dimensions.

For example, [here](../blob/master/vispy/scene/canvas.py#L266):
```python
         prof('create SceneDrawEvent')

         vp = (0, 0) + self.physical_size if viewport is None else viewport
         scene_event.push_viewport(vp)
         prof('push_viewport')
         try:
```


Eric89GXL:
@campagnola this shouldn't mess up #928 right?

Eric89GXL:
Thanks @QuLogic!

PR #978, opened 2015-06-04
fixed doc string for pyqt5

### LINKED ISSUE #972: docs error - should be PyQt5
was reading [config.py ](https://github.com/vispy/vispy/blob/75b683ed080440e881a2d0066b9ce182f4217569/vispy/util/config.py#L97), and I think it should be ```pyqt4|pyqt5|...``` rather that ```pyt5``` 

line no. 97

bollu:
fixes #972 - changed docs

### DIFF
--- a/vispy/util/config.py
+++ b/vispy/util/config.py
@@ -94,7 +94,7 @@ def _init():
 VISPY_HELP = """
 VisPy command line arguments:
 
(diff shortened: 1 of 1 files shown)


Eric89GXL:
thanks for the quick fix

Issue #975, opened 2015-06-03
Text in TextVisual example does not stay centred


QuLogic:
The `TextVisual` anchors to the centre by default, plus the resize handler moves the text position to be in the centre of the screen. However, as [you can see](http://youtu.be/FqJv5L8RPm4), the text does _not_ stay centred. Printing out `self.size` seems to show that the window size is determined correctly, but the text does not seem to be in the right place.

Eric89GXL:
I suspect this will be fixed by #928 because it changes a lot about the transform system

QuLogic:
Not just yet...
```
$ python3 text_visual.py 
Traceback (most recent call last):
  File "text_visual.py", line 41, in <module>
    c = Canvas()
  File "text_visual.py", line 15, in __init__
    self.text = visuals.TextVisual('', bold=True)
  File "/home/elliott/code/vispy/vispy/visuals/text/text.py", line 374, in __init__
    Visual.__init__(self, **kwargs)
  File "/home/elliott/code/vispy/vispy/visuals/visual.py", line 165, in __init__
    self._prepare_transforms(self)
  File "/home/elliott/code/vispy/vispy/visuals/visual.py", line 234, in _prepare_transforms
    raise NotImplementedError()
NotImplementedError
```

PR #979, opened 2015-06-04
[WIP] ColorBar visual

### LINKED ISSUE #682: Colorbar visual
* with/without border
* ticks
* tick labels

colorbar

bollu:
closes #682 

Started working on this, learnt a couple of things about vispy's model in the process. I've got the basic structure down.

- [x] add a ```orientation``` field for ```horizontal``` or ```vertical```
- [x] add a ```label``` field
- [x] ~~add a ```ticks``` field for the colorbar~~
- [ ] add an accompanying node for the SceneGraph
- [ ] draw a customizable border around the colorbar
- [ ] add test cases for the colorbar
- [x] actually respect the ```clim``` parameter (this should be easy, just change the ```texcoords```
- [x] document the example, so that there's proper explanation
- [x] write documentation (docstrings) for the files
- [x] add a license notice for the files

***Screenie of colorbar :***

screenshot from 2015-06-04 23 48 12


Eric89GXL:
I don't think you need to do anything special to respect the `clim` bounds -- `clim` sets what the start and end of the colorbar mean for the data, so you should be able to just plot the bar, and then label the start and end of it with `clim[0]` and `clim[1]` for now.

Eric89GXL:
And let's not bother with ticks, we should eventually do that using the `AxisVisual` somehow

bollu:
@Eric89GXL: okay, I'm confused, doesn't clim give you the start and end range for the colorbar interpolation? like, by default, you go from ```colormap([0, 1])```, but with clim, you go from ```colormap([clim_start, clim_end])``` correct?

Eric89GXL:
No, think about the common use case. For example, I have some data that ranges from -100 to 1000. But when I plot it I clip it a little bit, so I do:
```Python
>>> clim = [-5, 900]
>>> cmap = 'cubehelix'
>>> ax.image(..., colormap=cmap, clim=clim)
```
It would be nice to be able to then do:
```
ax.colorbar(cmap, clim)
```
and have it "do the right thing", which in this case is to display the full colorbar, with the left side marked as -5 and the right side marked as 900. Does that make sense?

Eric89GXL:
In other words, I don't think the most common use case will be to "cut into" part of the colormap for display, but rather to use the entire colormap, but have the start and end points mean something different for the data.

I suppose eventually we'll also have to figure out what happens if they do want to "cut into" the colormap, or only display a subsection of the colormap, or if the colormap is scaled in some nonlinear way... but we should be able to tackle that later once visuals actually support that sort of thing.

bollu:
So, in that case, the range ```[-100, 1000]``` gets linearly mapped to the range ```[-5, 100]``` ? or is the data clamped? ```[-100, -5] -> { -5 }```, ```[-5, 900]``` remains identical, ```[900, 1000] -> { 900 }```

(I think clamping is the right behaviour?)

So, in case of clamping, even the colors get clamped? like, ```[-100, -5] -> 0```, ```[-5, 900] -> [0, 1]```, ```[900, 1000] -> 1```?

Eric89GXL:
Yeah, colors outside the range get clamped.

bollu:
@Eric89GXL so, there's nothing "extra" to be done by the colorbar, right? like, it doesn't need to modify the color rendering, just rendering linearly would work?

Eric89GXL:
Yep, should be pretty basic for now

bollu:
@Eric89GXL how would I take the label for the colorbar? as a string? if so, how does one allow styling of the label?

Eric89GXL:
the label should be a string, yeah. We should probably eventually allow styling during construction, but this might blow up the constructor arguments which is not good. It's probably easier to expose the underlying `label` as an attribute, where `colorbar.label` is an instance of `TextVisual`, then people can change properties as they wish

bollu:
@Eric89GXL ah, yes. That does sound like a nice way to go about it. I'll do it that way.

bollu:
@Eric89GXL: I can't really use the ```clim``` field unless there's an axis to it, right? So do I simply leave it as a dummy field?

Eric89GXL:
For now just place labels at the start and end using a single TextVisual object, eventually we will replace it with an `AxisVisual` but we can do that later

bollu:
is there anyway to get the bounding box / size of a ```TextVisual```? I don't really see a straightforward way of doing this. 

When I place the label, I need to make sure that it doesn't intersect with the ```ColorBar``` whatever the dimensions of the ```ColorBar``` are, or whatever font size the ```TextVisual``` uses. I'd like to query the ``TextVisual``` for it's dimensions for this reason.

If this is not possible, what is a reasonable compromise when it comes to placing the labels?

QuLogic:
@bollu I think what you want is to set the `TextVisual`'s `anchor_x` and `anchor_y` settings to something other than `'center'` (though that may be a bit broken #975).

bollu:
@QuLogic: thank you. That worked.

bollu:
@Eric89GXL should the border around the ColorBar measured in pixels? I'm assuming it should be.

Eric89GXL:
Yep

bollu:
@Eric89GXL that will require doing the calculation in a coordinate system measured in pixels. I'm assuming that's the ```Visual``` coordinate system?

Eric89GXL:
You want the `document` coordinate system. Look through this example:

https://github.com/vispy/vispy/blob/master/examples/tutorial/visuals/T02_measurements.py

I strongly recommend going through each of those examples, they will save you some pain when developing visuals.

bollu:
current screenshot:

screenshot from 2015-06-07 23 51 30

It's not very flattering, but it works :)

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,223 @@
...
+            # https://github.com/vispy/vispy/issues/981
+            self._ticks = []
+            self._ticks.append(TextVisual(str(clim[0]),
+                               pos=begin_tick_pos,
+                               anchor_y="top"))
+            self._ticks.append(TextVisual(str(clim[1]),
+                               pos=end_tick_pos,
+                               anchor_y="top"))
+
+        elif orientation == "vertical":
```
bollu: I recognize that there is code duplication here, but I prefer the explicitness of it, since the two halves of the code setup the colorbar differently. (```rotation``` is setup, along with different positions for the markers).

I'd like your opinions.

bollu:
@Eric89GXL the test cases for visuals do a direct pixel-by-pixel comparison? ```assert_image_approved``` keeps popping up in the test cases. So, if I'm going to write tests, I'll need to use the same function to do it?

Eric89GXL:
Yeah, but I have a feeling the colorbar might change a little bit when using the AxisVisual, so for now just add a `XXX TODO`, and just test that it actually works and renders something non-zero.

Eric89GXL:
But yes, eventually you'll want to use the image comparison functions. There is a cross-correlation type mode, and a pixel-difference type mode. Either should eventually work for this visual.

vispy/visuals/colorbar.py
Eric89GXL: I think it's okay to have this level of redundancy. But I'd change it to `else:  # orientation == 'vertical':`, that way 1) we save one comparison, 2) it's clear that we know that to be the state at this point, and 3) it's easy to add the next `elif` mode by deleting a few chars.

bollu:
@Eric89GXL cool, so, add a TODO and check if it renders _anything_?

Eric89GXL:
Yeah for now that should be okay. Not ideal but it will remind us to check it for real later. If you really want, you can add the full test now, and modify it later.

bollu:
New screenshot (picked flattering colors :) )

screenshot from 2015-06-09 06 29 09

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,330 @@
...
+        tex_coords = np.array([[0, 0], [1, 0], [1, 1],
+                              [0, 0], [1, 1], [0, 1]],
+                              dtype=np.float32)
+
+        self._program.frag['color_transform'] = Function(self._cmap.glsl_map)
+        self._program['a_texcoord'] = tex_coords.astype(np.float32)
+
+        self.regenerate()
+
+    def regenerate(self):
```
bollu: I'm not a fan of this name, but I couldn't think of anything better than ```regenerate```. Do you have any suggestions?

Eric89GXL: when does it need to be called? it should probably be private for now

Eric89GXL:
Nice example! Can you e.g. show an Image with a colormap, and put colorbars on each side of it? That would make a potential real-life use case clearer :)

vispy/visuals/colorbar.py
bollu: Ah, yes. It should have been private in the first place.

Well, whenever you need to recreate shader data or recreate the label or
ticks or something, it's called.

So, for example, when you change the border width, the positions of the
labels need to be regenerated to not overlap with the new border.

bollu:
Sure :) that was mostly me messing around with it and changing a bunch of
properties to get something pretty.

vispy/visuals/colorbar.py
QuLogic: So, simple old `_update` then?

bollu: @QuLogic that works too :) But, I felt that ```_update``` was too generic for some reason.

Eric89GXL: as long as it doesn't collide with any super-class properties, sure -- but be sure to check if you use that name

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,330 @@
...
+        self._border_program['a_position'] = vertices.astype(np.float32)
+
+        self._border_program.vert['border_width'] = self._border_width
+        self._border_program.frag['border_color'] = self._border_color.rgba
+
+    @staticmethod
+    def _get_orientation_error(orientation):
+        return ValueError("orientation must"
+                          " be \'horizontal\' or \'vertical\'."
+                          " given: %s" % orientation)
```
QuLogic: I would say `... or 'vertical', not '%s'."` Also, the single quotes do not need to be escaped. Finally, it is safer to encapsulate `orientation` in a tuple as `"..." % (orientation, )`.

bollu: noted. Will change!

bollu:
@Eric89GXL are there any canonical colormap images I could use?

Eric89GXL:
Not sure. I'd probably just construct something artificial like:
```
y, x = np.meshgrid(np.linspace(-1, 1, 30), np.linspace(0, 2, 20))
data = np.exp(-(x*x + y*y))
```
And then do a colormap'ed image of `data`

bollu:
Decided to plot the mandlebrot set for more authenticity :)

screenshot from 2015-06-09 19 42 48


@Eric89GXL is this okay as an example?

rougier:
Nice !

Eric89GXL:
Yes, LGTM :)

Maybe we should have orientation = 'left' | 'right' | 'top' | 'bottom', which would set the tick directions accordingly to always be toward the outside -- it seems like the label and ticks in your example should be pointing the other direction to me

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,398 @@
...
+            self._label.pos = text_x, text_y
+
+            # TODO, HACK: This should ideally be a single TextVisual
+            # However, one TextVisual with multiple strings
+            # does not seem to be working as of now. (See #981)
+            # https://github.com/vispy/vispy/issues/981
+            self._ticks[0].pos = x - halfw, text_y
+            self._ticks[1].pos = x + halfw, text_y
+
+        elif self._orientation == "top":
```
bollu: Number of branches has gone from 2 to 4. This piece of code needs some love, but I'm not sure as to _how_ to rewrite it.

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,398 @@
...
+        self._clim = clim
+        # new TextVisuals need to be created
+        self._update()
+
+    @property
+    def label(self):
+        return self._label
+
+    @label.setter
+    def label(self, label):
```
bollu: As of now, I'm letting the user have direct access to the ```label``` ```TextVisual``` that's present inside the colorbar. This breaks encapsulation. However, if we want to allow styling, it'll require more getters / setters (```set_label_color```, ```set_label_font_size```,...) What's the right compromise?

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,398 @@
...
+        self._label = label
+        # position, styling has to be applied
+        self._update()
+
+    @property
+    def ticks(self):
```
bollu: Same problem as that of ```label```. I'm giving the user direct access, which I'm not too fond of

examples/basics/visuals/colorbar_visual_types.py
```diff
@@ -0,0 +1,108 @@
...
+colormap = get_colormap("ice")
+
+
+def style_colorbar(colorbar):
+    colorbar.label.font_size = 18
+    colorbar.label.color = "black"
+
+    colorbar.clim = (0, 10)
+
+    colorbar.ticks[0].font_size = 14
```
bollu: setting ticks needs you to access them individually. This should change once ```AxisVisual``` stabilizes, but until then, is there anyway to make this nicer? (overload the ```=``` operator by returning a custom ```Ticks``` class came to my mind, but it felt like over-engineering)

bollu:
@Eric89GXL  Do the new orientations look correct to you?
screenshot from 2015-06-10 16 27 51

Also, I reworked the example with the new orientation:

screenshot from 2015-06-10 16 28 48


I think this is stable enough to write a testcase against. What do you think?

rossant:
in the last example i'd rather see the text going from bottom to top

bollu:
@rossant hm, that's fair. Will change

bollu:
@rossant  pushed the change. Does this work?
screenshot from 2015-06-10 18 15 22
screenshot from 2015-06-10 18 15 04

rossant:
it seems more natural to me like this, but let's wait for others' opinions!

bollu:
@rossant, yeah, you're right. If I was drawing a graph on paper as well, that would be way I would label the y axis. It does feel more natural.

@Eric89GXL thoughts?

bollu:
Also, @Eric89GXL  shall I start writing test-cases for this?

vispy/visuals/colorbar.py
Eric89GXL: usually there is some way to set up parameters (scale factors, variable names) in the branches, and then set all the attributes after the branching. see if that's possible here

vispy/visuals/colorbar.py
Eric89GXL: don't bother. eventually we can just give them axis to the ticks somehow, but it will probably change once we change it to use the axis visual

examples/basics/visuals/colorbar_visual_types.py
Eric89GXL: I would say for now, don't allow setting these things to keep it simpler

Eric89GXL:
you can write tests, but they will probably need to change quickly. it's probably better to focus instead of translating visuals, and then once that's done finishing axes, once that's done then making this use that visual, then once that's done it's probably stable enough to write some tests :)

jdreaver:
@rossant @bollu +1 on the bottom-to-top text orientation :+1:

Eric89GXL:
@bollu perhaps just add some basic cases that make sure it renders for the various orientations and choices of limits, and we can improve the tests later. that way we can get this merged and you can move on to other things.

bollu:
please merge the [test data](https://github.com/vispy/test-data/pull/2) first so the CI passes

bollu:
@Eric89GXL. @rossant: I feel that this is okay to get reviewed now.

examples/basics/visuals/colorbar_visual.py
```diff
@@ -0,0 +1,154 @@
...
+# Copyright (c) 2015, Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+
+"""Using Colorbars with the Canvas"""
+
+from vispy import app
+from vispy import gloo
+from vispy.visuals.transforms import NullTransform, STTransform
+from vispy.visuals.transforms import TransformSystem
```
Eric89GXL: By convention we group imports like this if we need more lines
```
from vispy.visuals.transforms import (a, b,
                                      c)
```

vispy/testing/image_tester.py
```diff
@@ -74,6 +74,9 @@ def assert_image_approved(image, standard_file, message=None, **kwargs):
     This function will automatically clone the test-data repository into 
     ~/.vispy/test-data. However, it is up to the user to ensure this repository
     is kept up to date and to commit/push new images after they are saved.
+
+    Run the test with python <test-path> --vispy-audit-tests to bring up
+    the auditing GUI.
```
Eric89GXL: thanks for improving the docs

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,443 @@
...
+# visually "sticking" to the Colorbar
+_TEXT_PADDING_FACTOR = 1.2
+
+
+class ColorBarVisual(Visual):
+    """Visual subclass displaying a colorbar
+
+    Parameters
+    ----------
+
```
Eric89GXL: no empty newline here

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,443 @@
...
+    center_pos : tuple (x, y)
+        Position where the colorbar is to be placed with
+        respect to the center of the colorbar
+    halfdim : tuple (half_width, half_height)
+        Half the dimensions of the colorbar measured
+        from the center. That way, the total dimensions
+        of the colorbar is (x - half_width) to (x + half_width)
+        and (y - half_height) to (y + half_height)
+    cmap : str | vispy.color.ColorMap
+        either the name of the ColorMap to be used from the standard
```
Eric89GXL: start with capital letter

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,443 @@
...
+        set of names (refer to `vispy.color.get_colormap`),
+        or a custom ColorMap object.
+
```
Eric89GXL: no need for this extra empty newline

vispy/visuals/colorbar.py
```diff
@@ -0,0 +1,443 @@
...
+        The label that is to be drawn with the colorbar
+        that provides information about the colorbar.
+    clim : tuple (min, max)
+        the minimum and maximum values of the data that
+        is given to the colorbar. This is used to draw the scale
+        on the side of the colorbar.
+    border_width : float (in px)
+        The width of the border the colormap should have. This measurement
+        is given in pixels
+    border_color : string | vispy.color.Color
```
Eric89GXL: s/string/str/g

vispy/testing/image_tester.py
bollu: you're welcome :)

Eric89GXL:
@bollu you moving on to helping @campagnola with #928 by translating visuals next? That would probably be the highest impact move since a) you'll learn about a bunch of visuals that way and b) that PR should help improve performance quite a bit.

campagnola:
FYI I have text, axis, and plotline visuals in progress.

bollu:
@campagnola I was checking out the ```LineVisual``` example and was planning to port ```polygon``` as you suggested

Eric89GXL:
Thanks for this @bollu, this should be useful with the plotting interface. Hopefully the old visuals system makes sense, and the new one won't be too hard to learn.

bollu:
@Eric89GXL I'm reading through the new visuals code :) It seems pretty cool, especially with the shared shaders and everything.

PR #982, opened 2015-06-06
[WIP] IPython extension mechanism for Vispy

### LINKED ISSUE #977: automatically launch vispy in webgl mode if in notebook
This could be a nice usability improvement, to have vispy automatically pickup the webgl back end if its running in a notebook.

AFAIK, this is possible to detect since ipython provides a config object that can be used to query status.

If there's a consensus on this, I'll try and write a pull request for this

bollu:
closes #977  

So, I tried to take a stab at this, and here's my first attempt. I foresee this needing a _lot_ of iteration to get right.

***Things to decide:***
- location of the ```ipython``` extension file
- features of the extension 
- different ```magic``` operators that this could have

***TODO:***

- [ ] I'm not sure how to write test cases for this. I'm guessing ```iPython``` provides some way to do this, but I'll have to look into this

- [ ] Documentation for this, so that it's discover-able

- [ ] list the extension on iPython's [extension wiki](https://github.com/ipython/ipython/wiki/Extensions-Index)

- [ ] Publish the extension ( push it to ```PyPi``` with the tag ```Framework :: IPython```)


Eric89GXL:
maybe `vispy/app/ipython` for the IPython file?

Looking into tests would be great. If it's not easy don't worry too much about it, we can let use cases drive it. Hopefully the extension is fairly simple. Speaking of which, I think the first step should just be to make it easy to use the ipynb_webgl backend for full integration like we talked about. That would be good enough for this PR, especially if it is integrated in an example or two with tests.

bollu:
@Eric89GXL is the current path okay? The import is relatively painless right now - ```%load_ext vispy.ipython``` rather than ```%load_ext vispy.app.ipython```. This is subjective, but I think the first is more ergonomic than the second

Eric89GXL:
I think it's fine for the function to live in that namespace (I agree it's nicer), but I think the actual file that contains the relevant code should go in `vispy/app`

Eric89GXL:
agh... never mind that last comment. That will break some tests about orthogonality of modules. So you can leave it where it is.

bollu:
@Eric89GXL very well. I'll also add an example on using this under ```examples/ipynb```.

Eric89GXL:
Sure, or can you just update the existing examples to use this nicer syntax?

bollu:
@Eric89GXL will do.

bollu:
A question of UX:

As of now, a user uses ```%load_ext vispy.ipython``` to load the vispy extension, and __then does__
```%vispy_ipython``` to launch the webgl enabled app.

So, currently, the user needs 2 lines:
```
%load_ext vispy.ipython
%vispy_ipython
```

for the webgl backend to work


Should the first line (i.e, ```%load_ext vispy.ipython```)  also auto
matically launch the app with the right backend? IMO, just loading the extension should also load the backend, but @rougier  or someone else might have workflows that I have not considered.

**Comment by Eric89GXL** [comment]
What does the vispy extension do, if it doesn't set the `app.use_app`?

**Comment by bollu** [comment]
It sets up the other magic functions (for now, just one). It injects the ```vispy_ipython``` function into the ```iPython``` interpreter object

**Comment by Eric89GXL** [comment]
I think it should enable the `ipynb_webgl` backend by default. @rossant WDYT?

**Comment by rossant** [comment]
@Eric89GXL +1

**Comment by bollu** [comment]
@Eric89GXL, @rossant : cool, then :) I'll make it autoload the webgl backend

**Comment by rossant** [comment]
i think you can rename `%vispy_ipython` to `%vispy_webgl`

**Comment by bollu** [comment]
@rossant : all we'll have now is the ```%load_ext vispy.ipython```. Do you want to rename that as ```%load_ext vispy.webgl``` ?

**Comment by rossant** [comment]
ah, ok. no need then :)

**Comment by bollu** [comment]
@rossant could you do me a favor and check if the webgl ipython notebooks work for you? I just pushed the updated versions

**Comment by rossant** [comment]
didn't work at first, I tried again and it worked
but it might not be related to this PR (maybe due to the recent changes with async loading)

**Comment by bollu** [comment]
@rossant did you have the same behavior on all notebooks?

**Comment by bollu** [comment]
as for testing, I've been taking a cursory glance as to how other libraries do it, and they seem to have a god object-like thing that the notebook extension instantiates. 

So, for testing, they instantiate the same object and call it's methods and test that output. I don't think that would work for us, since we need to test the client itself. Although, I'm not sure

[example of beautifulsoup testcase](https://github.com/Psycojoker/ipython-beautifulsoup/blob/master/test.py)

[example of python-matlab-bridge](https://github.com/arokem/python-matlab-bridge/blob/master/pymatbridge/tests/test_run_code.py#L13)

**Comment by Eric89GXL** [comment]
It's not the end of the world if we can't test this, it's only a few lines that are a wrapper, and it should be heavily tested by users. If it starts to get more complex we can work harder to test. Of course if you want to try to set something up I'm not going to stop you :)

**Comment by bollu** [comment]
@Eric89GXL : If I wanted to add a test for this, where would it have to be located?

**Comment by Eric89GXL** [comment]
I would put it in vispy/app/tests/

**Comment by bollu** [comment]
@Eric89GXL  thanks!

**Comment by bollu** [comment]
@Eric89GXL, @rossant : time for code review? I added a simple test to check if the webgl backend is loaded using ```iPython```'s testing classes.

Big shout-out to [arokem](https://github.com/arokem/python-matlab-bridge) who answered my questions over gitter :)

**Comment by Eric89GXL** [comment]
You will need to 1) add a nice error message if the user doesn't have IPython of the proper version (3.0+), use `LooseVersion` for this, 2) create a `requires_ipython` decorator for the test that uses it, and 3) modify Travis (and maybe AppVeyor) to install IPython.

**Comment by bollu** [comment]
okay, this is interesting - it needs the ```ipython``` module for the testcase to run, but it doesn't exist on the Travis CI server since we didn't ask for it to install in the ```.travis.yml`` file. do I create a pull request and add that as a dependency?

**Comment by Eric89GXL** [comment]
It can all be part of this PR

**Comment by bollu** [comment]
@Eric89GXL: I had not known about ```LooseVersion```. Will check that out.
And will do

File: examples/ipynb/webgl_example_1.ipynb [hunk 1]
```diff
@@ -27,7 +27,9 @@
     "import vispy.gloo as gloo\n",
     "from vispy import app\n",
     "from vispy.util.transforms import perspective, translate, rotate\n",
-    "app.use_app('ipynb_webgl');"
+    "\n",
+    "# load the vispy bindings for the iPython notebook which enables webGL\n",
```
rossant: please change all instances of "iPython" to "IPython" -- we're not in the fruit business here :)

bollu:
@Eric89GXL the nice error message is in the testcase or in the notebook?

examples/ipynb/webgl_example_1.ipynb
bollu: whoa, my bad :)

Eric89GXL:
I just mean that a user running IPython < 3.0 should get a nice error message when doing `app.use_app('ipynb_webgl')` and thus also your IPython extension, just make sure it's the case that happens

bollu:
gotcha. but, looking around a bit, [LooseVersion is going to be deprecated](http://stackoverflow.com/questions/11887762/how-to-compare-version-style-strings)?


[Here is the actual PEP (```PEP 0386```)](https://www.python.org/dev/peps/pep-0386/)

Eric89GXL:
Hmm... we already use it in many places. So for now you can still use it. If you want, you could do whatever is "the right thing" (NormalizedVersion?) but make sure it works on old platforms, too. So we might need to have something like `vispy.util.check_version` that wraps to `NormalizedVersion` if available and `LooseVersion` if not. It can be part of this PR if you wan to do it now, since the number of lines changed is still low (should be easy to review both parts).

bollu:
@Eric89GXL It looks like ```NormalizedVersion``` is a part of ```distutils2```, and is 

1. not part of the standard library
2. not yet ready for developers

It looks like [```disutils2``` is dead](https://hg.python.org/distutils2), with the latest commits being 2012. So, I don't think we have to worry.


I'm just using ```LooseVersion```. Does that work?

Eric89GXL:
Yeah that should be fine

bollu:
@Eric89GXL Why do we need a ```requires_ipython``` decorator? Like, what's the point? what's supposed to happen if ```IPython``` doesn't exist? skip the test I assume?

Eric89GXL:
Yes exactly, look at e.g. requires_pyopengl and you will get the idea

bollu:
That technique is really cool, a decorator returning a decorator that is then applied to a function is something I haven't seen before :D

bollu:
@Eric89GXL I think I did it right - Travis, AppVeyor and the decorator update.

.travis.yml
```diff
@@ -127,6 +127,9 @@ install:
         travis_retry sudo apt-get -qq -y install libegl1-mesa;
       fi;
 
+    # Install IPython
+    - pip -q install "ipython[notebook]"
```
QuLogic: Pretty sure this should only be installed if `${DEPS} == full`, not always.

appveyor.yml
```diff
@@ -25,8 +25,8 @@ install:
 
   # Install the dependencies of the project.
   # "conda install --yes --quiet numpy scipy matplotlib setuptools flake8 pyqt nose pytest coverage"
-  - "conda install --yes --quiet numpy scipy matplotlib setuptools nose pytest coverage"
-  - "pip install -q pytest-cov"
+  - "conda install --yes --quiet numpy scipy matplotlib setuptools nose pytest coverage ipython ipython-notebook"
+  - "pip inst all -q pytest-cov"
```
QuLogic: `inst all`?

examples/ipynb/webgl_example_1.ipynb
```diff
@@ -189,21 +191,21 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 2",
```
QuLogic: This and the following changes are not really necessary.

vispy/ipython.py
```diff
@@ -0,0 +1,49 @@
...
+
+    IPython : IPython interpreter
+        An instance of the IPython interpreter that is handed
+        over to the extension
+    """
+
+    # don't continue if IPython version is < 3.0
+    ipy_version = LooseVersion(IPython.__version__)
+    if ipy_version < LooseVersion("3.0.0"):
+        ipython.write_err("You IPython version is less than "
```
QuLogic: s/You/Your/

vispy/ipython.py
```diff
@@ -0,0 +1,49 @@
...
+                          "version 3.0.0, the minimum for Vispy's"
+                          "IPython backend. Please upgrade you IPython"
```
QuLogic: s/you/your/

vispy/testing/__init__.py
```diff
@@ -40,7 +40,9 @@
 to determine what automatic tests are run.
 """
 
-from ._testing import (SkipTest, requires_application, requires_img_lib,  # noqa
+from ._testing import (SkipTest, requires_application,  # noqa
+                      requires_ipython,   # noqa
```
QuLogic: Huh, this `# noqa` is masking the fact that this line is misaligned.

.travis.yml
Eric89GXL: Indeed

vispy/testing/__init__.py
Eric89GXL: and actually this can fit on the same line as `requires_img_lib` so let's save a line and pop that one up

vispy/testing/_testing.py
```diff
@@ -242,6 +242,30 @@ def requires_img_lib():
...
+        import IPython  # noqa
+    except Exception:
+        return False, "IPython library not found"
+    else:
+        if LooseVersion(IPython.__version__) >= LooseVersion(version):
+            return True, "IPython present"
+        else:
+            message = (
+                "current IPython version: (%s) is "
+                "less than expected version: (%s)") % \
```
Eric89GXL: older than required version

.travis.yml
Eric89GXL: and I think Travis should be able to use `conda` as well, which should be faster

vispy/app/tests/test_ipython.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2015, Vispy Development Team.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+"""Tests for Vispy's IPython bindings"""
+
+# from vispy import IPython
+from numpy.testing import assert_equal
+from IPython.testing.globalipapp import get_ipython
```
Eric89GXL: nest this import, it will fail on systems without IPython, defeating the purpose of the decorator (Travis will also tell you this once we put the `ipython` install only on the `DEPS=full` case

vispy/app/tests/test_ipython.py
```diff
@@ -0,0 +1,22 @@
...
+from vispy.testing import requires_ipython
+
+
+@requires_ipython()
+def test_webgl_loading():
+    """Test if the vispy.ipython extension loads
+        the webGL backend"""
+    ipy = get_ipython()
+    ipy.run_cell("from vispy import app")
```
Eric89GXL: you shouldn't need this import, is it necessary?

vispy/app/tests/test_ipython.py
```diff
@@ -0,0 +1,22 @@
...
+    ipy.run_cell("%load_ext vispy.ipython")
+    ipy.run_cell("backend_name = app.use_app().backend_name")
+
+    # make sure that the webgl backend got loaded
+    assert_equal(ipy.user_ns["backend_name"], "ipynb_webgl")
```
Eric89GXL: nice

vispy/ipython.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2015, Vispy Development Team.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+"""Entry point for vispy's IPython bindings"""
+
+import IPython
```
Eric89GXL: same, nest this import, put inside `try/except`, and in fact try to make this code more DRY with the testing code. Maybe put IPython version checking function in `vispy/util` somewhere, and import that both from `testing` and from here.

Eric89GXL:
I think you also need to add `vispy/ipython` to the list of modules. In fact I think that part might be auto-generated, so make `vispy/ipython` a directory and maybe have `vispy/ipython/__init__.py` and put your real code in `vispy/ipython/ipython.py` -- silly I know but all our other main submodules work that way

almarklein:
With ipython becoming jupiter, might it make sense to call it vispy-jupyter, or vispy-notebook instead?

rossant:
since vispy is only in Python, it will always be IPython+Jupyter, so either way is fine IMO

vispy/app/tests/test_ipython.py
bollu: @Eric89GXL which import are you referring to?

vispy/ipython.py
bollu: @Eric89GXL but, the two pieces of code aren't exactly the same, right?

The test case decorator causes the test case to skip. Here, we need to throw a human-readable error for the user. I'm not sure what you want me to reuse

vispy/testing/_testing.py
bollu: @Eric89GXL: what is the right version?

examples/ipynb/webgl_example_1.ipynb
bollu: @QuLogic I opened it in the ipython notebook kernel for python2, which was why those changes took place. Should I revert that commit and redo it in the python3 kernel?

vispy/testing/_testing.py
Eric89GXL: I mean say "older than required version" instead of "less than expected version"

vispy/app/tests/test_ipython.py
```diff
@@ -0,0 +1,23 @@
...
+from numpy.testing import assert_equal
+from vispy.testing import requires_ipython
+
+
+@requires_ipython()
+def test_webgl_loading():
+    from IPython.testing.globalipapp import get_ipython
+
+    """Test if the vispy.ipython extension loads
+        the webGL backend"""
```
Eric89GXL: This should be at the top, your nested import should go below it

Eric89GXL: and this should be one line preferably

vispy/ipython/ipython.py
```diff
@@ -0,0 +1,49 @@
...
+    IPython : IPython interpreter
+        An instance of the IPython interpreter that is handed
+        over to the extension
+    """
+    import IPython
+
+    # don't continue if IPython version is < 3.0
+    ipy_version = LooseVersion(IPython.__version__)
+    if ipy_version < LooseVersion("3.0.0"):
+        ipython.write_err("Your IPython version is less than "
```
Eric89GXL: older than

vispy/ipython/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+"""
+This module provides the binding between Vispy and IPython in the form of
+an IPython extension
+"""
+
+# load the two functions that IPython uses to instantiate an extension
+# that way, the user only needs to run %load_ext vispy.ipython rather that
+# %load_ext vispy.ipython.ipython
+from .ipython import load_ipython_extension, unload_ipython_extension  # noqa
```
Eric89GXL: missing newline at end of file

.travis.yml
```diff
@@ -127,6 +127,11 @@ install:
         travis_retry sudo apt-get -qq -y install libegl1-mesa;
       fi;
 
+    # Install IPython
+    -if "${DEPS}" == "full"; then
+      conda install --yes --quiet ipython ipython-notebook ipython-qtconsole
```
Eric89GXL: we shouldn't need `ipython-qtconsole`

Eric89GXL:
Can you take a look at coveralls to confirm that the new code is covered? I think it is since coverage went up, but it's good to look at the reports to make sure.

bollu:
How exactly do I do that? (Check if all functions are covered) I'm sorry,
I'm not very familiar with coveralls as a tool :)

Eric89GXL:
poke around here

https://coveralls.io/builds/2753896

bollu:
@Eric89GXL: It didn't notice the new files at all? [in the list here](https://coveralls.io/builds/2753896), none of the files under the ipython directory are listed?

examples/ipynb/webgl_example_1.ipynb
QuLogic: Yes, I understood that, but it clutters up the commit unnecessarily. Reverting would just add another unnecessary commit. You'd need to rebase to fix it correctly. In the future, you should use `git add -p` and `git diff`/`git diff --cached` to verify what you are adding to a commit.

bollu: Understood. I'm sorry about this

QuLogic: It's good practice in general to review the diffs. Not everyone does it though...

Eric89GXL: @bollu don't sweat it too much. FYI I usually try to do it when opening a PR, and before asking for final review. I inevitably find things I meant to change or didn't mean to change. It generally saves some time.

.travis.yml
```diff
@@ -127,6 +127,11 @@ install:
         travis_retry sudo apt-get -qq -y install libegl1-mesa;
       fi;
 
+    # Install IPython
+    -if "${DEPS}" == "full"; then
+      conda install --yes --quiet ipython ipython-notebook
```
Eric89GXL: this line should probably have a semicolon

Eric89GXL:
I'm not sure why the latest build didn't do any Travis runs... hopefully on the next commit it will.

Regarding Coveralls, I think that's a bug with their display, if you go to the "ALL" list of files, `vispy/ipython.py` is there with 76% coverage.

bollu:
It looks like ```anaconda``` is not installed on the travis boxes that build Python 2.7. So, I've added a rule to switch to ```pip``` on 2.7. I'd like to check if the build succeeds with that.

Eric89GXL:
Ahh, I forgot that we use system Python there. Then yes, please use `pip` on those builds.

bollu:
```
The command "if [ "${DEPS}" == "full" ]; then if ["${PYTHON}" == "2.7" ]; then pip install -q \"ipython[notebook]\"; else conda install --yes --quiet ipython ipython-notebook; fi; fi;" failed and exited with 127 during .
```

Any idea why this fails, @Eric89GXL ?

QuLogic:
You are missing a space after the second `[`.

bollu:
@QuLogic whoa, thanks. And I find it interesting that [travis lint](https://lint.travis-ci.org/) did not pick that up.

bollu:
@QuLogic , build still fails :(

```
The command "if [ "${DEPS}" == "full" ]; then if [ "${PYTHON}" == "2.7" ]; then pip install -q \"ipython[notebook]\"; else conda install --yes --quiet ipython ipython-notebook; fi; fi;" failed and exited with 2 during .
```

Eric89GXL:
do you need those escape characters `\"`? Does `pip install -q ipython[notebook]` by itself work? Seems like it should

bollu:
@Eric89GXL , you *should* need the ``` " ``` around the ``` ipython[notebook] ``` or else ```pip``` fails. But, I'll give it a shot without the ```"``` if the current build fails as well

Eric89GXL:
Looks like `'` failed, so I killed that build. I just tried running it with quotes and with no quotes and `pip` said "ipython 3.1.0 does not provide the extra 'notebook'" in both cases

Eric89GXL:
(locally I tried I mean)

bollu:
```
$ if [ "${DEPS}" == "full" ]; then if [ "${PYTHON}" == "2.7" ]; then pip install -q ipython[notebook]; else conda install --yes --quiet ipython ipython-notebook; fi; fi;
You are using pip version 6.0.7, however version 7.0.3 is available.
```

Whoa, that's cool. No quoting at all works in the Travis build?

Eric89GXL:
Actually maybe on the `2.7` builds you should just use `sudo apt-get install python-ipython` or whatever the line would be to make sure old, incompatible builds are handled properly...

Eric89GXL:
You could try no quoting to see if it works, but on my local Anaconda it doesn't like the `[notebook]` option, so it might still fail

bollu:
@Eric89GXL: no quoting [seems to be working so far](https://travis-ci.org/vispy/vispy/jobs/65992995)

Eric89GXL:
Actually, it just occurred to me that we probably do want to test an old version of IPython. You could modify the 2.6 build to install an old version of IPython using conda, see:

http://www.continuum.io/blog/conda

You can add an env var IPYTHON="" by default but IPYTHON="=2.0" or something for the Python 2.6 build. Then you can modify the test that requires IPython to just require *any* version of IPython, and if version < 3.0, use `raises` or `assert_raises` to check that an appropriate assertion is raised, and if IPython >= 3.0 check that the tests run okay.

Whether you want to do this now, in a subsequent PR, or not at all is up to you. I get a kick out of setting up CI platforms, but if you'd rather move on to some actual coding you can :)

bollu:
@Eric89GXL might as well be thorough :) I'm on it. I'll do it now, and if it takes too long, I'll complete it once I crash and wake up.

Eric89GXL:
Cool. FYI the 2.6 build is where we have previously dumped old versions of Pyglet and GLFW. However, Pyglet has since updated their version so there's no reason to use `tip.zip`. While you're digging around in there, could you unify these lines to just use the standard `pip` version? We don't need to triage anymore.
```
          pip install -q pyglet;
          pip install -q https://bitbucket.org/pyglet/pyglet/get/tip.zip;
```

bollu:
sure thing :)

bollu:
Whee! looks like this builds ^_^

Eric89GXL:
Excellent :)

bollu:
@Eric89GXL Why do I need an environment variable for the IPython version when I can just query the actual module itself? ```Ipython.__version__```?

Eric89GXL:
You don't query the env var from Python, you query it from `.travis.yml` to decide which version to install. By default `IPYTHON=""` so doing `conda install ipython$IPYTHON` will do `conda install ipython` and install the default version, whereas `IPYTHON="=2.0"` will do `conda install ipython=2.0` effectively. See e.g.:

https://github.com/mne-tools/mne-python/blob/master/.travis.yml#L24

bollu:
that is really slick. will do!

bollu:
@Eric89GXL is there any way I can test out what happens? because I'm not sure what exactly IPython does when an extension fails to load (which is what will happen if IPython version < 3.0)

bollu:
I just made the changes that I wanted to and pushed it to travis for now. I'll need to find a better solution though

Eric89GXL:
You could set up your own virtual environments for testing. It's easy with Anaconda, slightly less easy but still easy with Linux system python. That way you can have whatever version of IPython you want installed.

.travis.yml
```diff
@@ -127,6 +127,15 @@ install:
         travis_retry sudo apt-get -qq -y install libegl1-mesa;
       fi;
 
+    # Install IPython
+    - if [ "${DEPS}" == "full" ]; then
+        if [ "${PYTHON}" == "2.7" ]; then
+          pip install -q ipython[notebook];
+        else
+          conda install --yes --quiet ipython$IPYTHON ipython-notebook;
```
Eric89GXL: you *might* also need the `$IPYTHON` on the `ipython-notebook` call, might as well add it to be safe. Also, make sure this is the last call to conda, otherwise it's possible some other package we install with `conda` later will recommend a higher version, and it will override our previous manual installation of this version. (Please add a note in Travis about this, too.)

bollu:
@Eric89GXL turns out, Python 2.6 anaconda [doesn't have ipython](http://docs.continuum.io/anaconda/pkg-docs.html) (check the package list). So, well, what now?

Eric89GXL:
You could make 2.7 or 3.4 run an old version. Probably 2.7 is the better choice.

bollu:
@Eric89GXL but, well, we would need to check both the success branch and the failure branch, right? But AFAIK there is only _one_ Python2.7 machine with full dependencies on the CI server. So, how do we test both branches?

Eric89GXL:
Py3k should have the newer working version, Python 2.7 the old one

bollu:
@Eric89GXL rationale for that? :)

Eric89GXL:
We want to test both cases ideally (having the right version, and not), so I just picked one to get the old version. Since Python 2.7 already runs some older stuff since it's using system Python, it seemed like the natural choice.

bollu:
@Eric89GXL travis passed. So, well, what now?

Eric89GXL:
I think you might be done. Just double-check that you've addressed all the comments, and I'll pull, test, take one last look, and merge if it's all good.

Eric89GXL:
How's the coverage?

bollu:
@Eric89GXL last I checked, it had gone up. I don't think that's changed since I didn't add any functions

Eric89GXL:
It should have gone up again since you improved Travis's tests, can you double-check?

bollu:
@Eric89GXL will do

bollu:
@Eric89GXL it says [test_ipython file has new coverage of 25](https://coveralls.io/files/835927821) for the latest commit

Eric89GXL:
Was it higher before?

Eric89GXL:
perhaps your requires decorator must not work... does it work locally?

Eric89GXL:
Oh wait -- you showed the result for just the first job, 2.6, which doesn't have any IPython...

bollu:
ahh, my bad.

Eric89GXL:
I think this is the report you want 

https://coveralls.io/builds/2778857

bollu:
@Eric89GXL - yep, [coverage increase](https://coveralls.io/files/836656486)

bollu:
@Eric89GXL could this be added to the PyPi index under, say, testing?

Eric89GXL:
Not sure, I haven't dealt with that before. Does it list the testing modules or something? Not sure if any of the repos I work on have used it, so it might not be a big deal.

Eric89GXL:
Or do you mean by the tag you mention in your top post (which should be updated BTW)?

bollu:
yes, that. Also, add it to the [IPython wiki](https://github.com/ipython/ipython/wiki/Extensions-Index)

Eric89GXL:
Sure, whatever you think makes sense, I don't know much about it

Eric89GXL:
For now add to `setup.py`, once it's merged you can put it on their Wiki. Or maybe we should wait until release to do that bit.

bollu:
@Eric89GXL  yeah, that's fair. Wait for version 0.5.

bollu:
I [created an issue for this](https://github.com/vispy/vispy/issues/new). Could you give the ```0.5``` milestone tag please?

Eric89GXL:
Done. Go ahead and update `setup.py` and hopefully I won't have any other comments so we can merge :)

Eric89GXL:
I think it's okay to add to `setup.py` now

bollu:
@Eric89GXL done. pushed the change.

bollu:
@Eric89GXL  this passed CI as well.

Eric89GXL:
Great, I'll take a look and merge if everything runs well shortly

.travis.yml
```diff
@@ -113,7 +116,7 @@ install:
           pip install -q pyglet;
           conda install --yes --quiet wxpython > ${REDIRECT_TO};
         else
-          pip install -q https://bitbucket.org/pyglet/pyglet/get/tip.zip;
+          pip install -q pyglet;
```
Eric89GXL: please combine this with the `pip install -q pyglet` above, they don't need to be separate commands anymore because they are the same

.travis.yml
```diff
@@ -127,6 +130,20 @@ install:
         travis_retry sudo apt-get -qq -y install libegl1-mesa;
       fi;
 
+    # Install IPython
+    # No IPython for 2.6
+    # Install using conda with old version for 2.7
+    # Install using pip with new version for > 2.7
```
Eric89GXL: we should use conda with new version for > 2.7, and system version `apt-get install python-ipython` or whatever for 2.7

Eric89GXL:
Other than my final comments LGTM, let me know when they're taken care of and I'll merge

.travis.yml
bollu: I don't think that will pull the latest versions of ```ipython``` which we need. IMO pip is the right way to pull packages where ```conda``` is not present

bollu:
@Eric89GXL done :)

bollu:
@Eric89GXL can this be merged?

Eric89GXL:
Yeah I think so, just keep my minor comments in mind for the next round. Just making sure things pass over here.

Eric89GXL:
Argh, I meant about those minor comments to refer to #979, but it looks like you already fixed them :) I'll also merge that one once the CIs come back happy.

PR #983, opened 2015-06-07
Fixed SurfacePlotVisual attribute name

inclement:
set_data later sets self.__colors (though nothing ever reads from either attribute), so I guess this is a small and currently irrelevant typo


Eric89GXL:
Yeah, nothing reads from that attribute, please delete both occurrences

inclement:
Okay, I replaced the commit.

Eric89GXL:
Thanks @inclement

PR #988, opened 2015-06-17
Minor fix to Volume docstring

astrofrog:
The Volume docstring originally mentioned that the data should be 2-dimensional, which I presume was a typo?


almarklein:
:) thanks!

PR #991, opened 2015-06-18
typo: dpi attribute shadows property

midgetfc:
noob learning : stumbled upon this while reading your great code.
cheers

### DIFF
--- a/vispy/app/canvas.py
+++ b/vispy/app/canvas.py
@@ -135,7 +135,7 @@ class Canvas(object):
             dpi = util_config['dpi']
         if dpi is None:
             dpi = get_dpi(raise_error=False)
-        self.dpi = dpi
+        self._dpi = dpi
 
         # Create events
         self.events = EmitterGroup(source=self,


Eric89GXL:
@midgetfc actually what happens here doesn't end up being shadowing. Check out where the property is defined, near to it there is a `@dpi.setter` function. When you do `canvas.dpi = 92`, the `setter` gets called with the `92` arg. It's a way of getting and setting properties while having control over some (preferably small) things that happen during those processes -- in this case, converting the input to float and redrawing the canvas. Does that make sense?

midgetfc:
Python is still quite new for me. I do understand that I have improperly stated things here. Your getter and setter functions are not shadowed.
I still have a question though : when you initialize dpi from global config or by a call to get_dpi in __init__, line 138, the attribute is called dpi, and the property getter/setter refer to an attribute called _dpi. Should'nt they refer to the same name, like you do for _title ?

Eric89GXL:
It depends. You can either do `self.dpi = X`, which will go through the setter. We know the setter sets the private variable `self._dpi`. So if we want to skip the processing that the setter does -- there are some cases where we want to do this, like with the title -- then we can just do `self._title`. So the choice of which statement to use is made based on whether we want to go through the setter, or skip that process.

midgetfc:
I understand. I had not realized that self.dpi = x in __init__ would go through the setter. I have too much c++ background, and haven't quite completely realized that __init__ **is not** a constructor.
All my apologies, and thanks for your patience. How do you close the PR ?

Eric89GXL:
There should be a close button for you since you opened the PR, but maybe you don't have it. I'll go ahead and close. Feel free to open other PRs if you find other potential issues.

PR #995, opened 2015-06-20
add test docs option to make.py

### LINKED ISSUE #993: Allow make.py test flake to run docstring tests as well
As of now, ```make.py test flake``` does _not_ test the docstrings, which means the docstring tests present at  ```vispy/util/tests/test_docstring_parameters.py``` need to be run separately.

Forgetting that step means waiting for the CI to pass the build, have that fail after 5 minutes, and then fix the docstrings again.

I propose to allow ``make.py test flake``` to run the docstring tests as well (all forms of linting we have, basically)

Thoughts?

bollu:
closes #993 

Also added an option to the makefile so you can write ```make docs```.

the syntax to invoke this test is ```python make.py test docs```.

### DIFF
--- a/vispy/testing/_runners.py
+++ b/vispy/testing/_runners.py
@@ -8,6 +8,7 @@ from __future__ import print_function
 
 import sys
 import os
+import warnings
 from os import path as op
 from copy import deepcopy
 from functools import partial
@@ -114,9 +115,14 @@ def _unit(mode, extra_arg_string, coverage=False):
 def _docs():
     """test docstring paramters
     using vispy/utils/tests/test_docstring_parameters.py"""
-    print("warning: this imports vispy from the vispy installation. it is"
-          " recommended to setup vispy using 'python setup.py develop'"
-          " so that the latest sources are used automatically")
+    dev = _get_import_dir()[1]
+
+    if not dev:
+        warnings.warn("Docstring test imports Vispy from"
+                      " Vispy's installation. It is"
+                      " recommended to setup Vispy using"
+                      " 'python setup.py develop'"
+                      " so that the latest sources are used automatically")
     try:
         # this should always be importable
         from vispy.util.tests import test_docstring_parameters
--- a/vispy/testing/_runners.py
+++ b/vispy/testing/_runners.py
@@ -114,9 +114,13 @@ def _unit(mode, extra_arg_string, coverage=False):
 def _docs():
     """test docstring paramters
     using vispy/utils/tests/test_docstring_parameters.py"""
+    print ("warning: this imports vispy from the vispy installation. it is"
+           " recommended to setup vispy using 'python setup.py develop'"
+           " so that the latest sources are used automatically")
     try:
         # this should always be importable
         from vispy.util.tests import test_docstring_parameters
+        print("Running docstring test...")
         test_docstring_parameters.test_docstring_parameters()
     except AssertionError as docstring_violations:
         # the test harness expects runtime errors,
--- a/vispy/testing/_runners.py
+++ b/vispy/testing/_runners.py
@@ -111,8 +111,23 @@ def _unit(mode, extra_arg_string, coverage=False):
         os.rename(op.join(cwd, '.coverage'), out_name)
 
 
+def _docs():
+    """test docstring paramters
+    using vispy/utils/tests/test_docstring_parameters.py"""
+    orig_dir = os.getcwd()
+    import_dir, dev = _get_import_dir()
+    os.chdir(op.join(import_dir, '..'))
+    try:
+        from vispy.util.tests import test_docstring_parameters
+        test_docstring_parameters.test_docstring_parameters()
+    except Exception as e:
+        print("unable to test docs: %s" % str(e))
+
+    os.chdir(orig_dir)
+
+
 def _flake():
-    """Test flake8 and docstrings"""
+    """Test flake8"""
     orig_dir = os.getcwd()
     import_dir, dev = _get_import_dir()
     os.chdir(op.join(import_dir, '..'))
@@ -304,7 +319,7 @@ def test(label='full', extra_arg_string='', coverage=False):
     ----------
     label : str
         Can be one of 'full', 'unit', 'nobackend', 'extra', 'lineendings',
-        'flake', or any backend name (e.g., 'qt').
+        'flake', 'docs', or any backend name (e.g., 'qt').
     extra_arg_string : str
         Extra arguments to sent to ``pytest``.
     coverage : bool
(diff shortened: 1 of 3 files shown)


vispy/testing/_runners.py
```diff
@@ -111,6 +111,25 @@ def _unit(mode, extra_arg_string, coverage=False):
         os.rename(op.join(cwd, '.coverage'), out_name)
 
 
+def _docs():
+    """test docstring paramters
+    using vispy/utils/tests/test_docstring_parameters.py"""
+    print ("warning: this imports vispy from the vispy installation. it is"
```
bollu: I'm not super happy with the solution. Is there a way to ensure that the newest versions are automatically tested? This was the simplest solution I could come up with (i.e, import the test from the vispy install), but I think there's a better solution that I'm missing

vispy/testing/_runners.py
```diff
@@ -344,6 +363,9 @@ def test(label='full', extra_arg_string='', coverage=False):
         runs.append([_check_line_endings, 'lineendings'])
     if label in ('full', 'extra', 'flake'):
         runs.append([_flake, 'flake'])
+    if label in ('full', 'extra', 'docs'):
+        runs.append([_docs, 'docs'])
```
Eric89GXL: not for full, full already runs it as part of nosetests

vispy/testing/_runners.py
```diff
@@ -111,6 +111,25 @@ def _unit(mode, extra_arg_string, coverage=False):
         os.rename(op.join(cwd, '.coverage'), out_name)
 
 
+def _docs():
+    """test docstring paramters
+    using vispy/utils/tests/test_docstring_parameters.py"""
+    print("warning: this imports vispy from the vispy installation. it is"
+          " recommended to setup vispy using 'python setup.py develop'"
+          " so that the latest sources are used automatically")
```
Eric89GXL: I don't think we want this warning, it will get old after a while. Better to document it somewhere else instead.

Eric89GXL: Like installation or development instructions somewhere.

vispy/testing/_runners.py
```diff
@@ -111,6 +111,25 @@ def _unit(mode, extra_arg_string, coverage=False):
...
+        # this should always be importable
+        from vispy.util.tests import test_docstring_parameters
+        print("Running docstring test...")
+        test_docstring_parameters.test_docstring_parameters()
+    except AssertionError as docstring_violations:
+        # the test harness expects runtime errors,
+        # not AssertionError. So wrap the AssertionError
+        # that is thrown by test_docstring_parameters()
+        # with a RuntimeError
+        raise RuntimeError(docstring_violations)
```
Eric89GXL: you should add an `else` clause that says "Ok" in some nice way

vispy/testing/_runners.py
bollu: @Eric89GXL does such a place exist on the wiki?

vispy/testing/_runners.py
bollu: The test runner does that already, doesn't it?
Here's my output on running it:

```
$ python make/make.py test docs
warning: this imports vispy from the vispy installation. it is recommended to setup vispy using 'python setup.py develop' so that the latest sources are used automatically
Running docstring test...
Passed

Testing succeeded (0 failed, 0 skipped) in 5.833 seconds
```

Eric89GXL: Oh right

vispy/testing/_runners.py
Eric89GXL: Not yet that I know of, we should add it

vispy/testing/_runners.py
bollu: I'd much rather be safe than sorry when it comes to warning like these. How about creating an issue to migrate it away, but keep the warning?

vispy/testing/_runners.py
bollu: In that case, I think I'll spend a few hours tonight documenting development for Vispy (the environment (some opengl backend, nosetests, flake), offline testing, CI.

Eric89GXL: Actually we have a heuristic to detect if someone is using a dev version or not, you should only throw a warning `if not _get_import_dir()[1]` (and use `warnings.warn` instead of print)

QuLogic: This Warning is missing Many Capital Letters...

vispy/testing/_runners.py
```diff
@@ -111,6 +112,30 @@ def _unit(mode, extra_arg_string, coverage=False):
...
+    """test docstring paramters
+    using vispy/utils/tests/test_docstring_parameters.py"""
+    dev = _get_import_dir()[1]
+
+    if not dev:
+        warnings.warn("Docstring test imports Vispy from"
+                      " Vispy's installation. It is"
+                      " recommended to setup Vispy using"
+                      " 'python setup.py develop'"
+                      " so that the latest sources are used automatically")
```
bollu: @QuLogic: fixed the capitalization :) Sorry for that!

vispy/testing/_runners.py
```diff
@@ -111,6 +112,30 @@ def _unit(mode, extra_arg_string, coverage=False):
         os.rename(op.join(cwd, '.coverage'), out_name)
 
 
+def _docs():
+    """test docstring paramters
+    using vispy/utils/tests/test_docstring_parameters.py"""
+    dev = _get_import_dir()[1]
```
bollu: the code now uses ```_get_import_dir()``` to check if Vispy has been setup correctly,

bollu:
@Eric89GXL: Is this merge-able?

vispy/testing/_runners.py
```diff
@@ -111,6 +112,30 @@ def _unit(mode, extra_arg_string, coverage=False):
...
+        warnings.warn("Docstring test imports Vispy from"
+                      " Vispy's installation. It is"
+                      " recommended to setup Vispy using"
+                      " 'python setup.py develop'"
+                      " so that the latest sources are used automatically")
+    try:
+        # this should always be importable
+        from vispy.util.tests import test_docstring_parameters
+        print("Running docstring test...")
+        test_docstring_parameters.test_docstring_parameters()
```
Eric89GXL: It is best/clearest to put as few lines as possible in the `try`. So that would be the actual function call. Also, you can use a relative import here like `from ..util.tests import test_docstring_parameters`.

Eric89GXL:
@bollu LGTM other than my last comment. We can make that minor tweak in some future PR.

Eric89GXL:
Thanks for the enhancement, hopefully it makes the other coding easier for you :)

vispy/testing/_runners.py
bollu: Oh, thanks for the tips on ```try/catch```. That makes a lot of sense! Will keep that in mind for the future.

PR #996, opened 2015-06-23
Add volume check

midgetfc:
Proposal to add a sanity check to the contrast limits range.
Avoids division by zero when clim is set from the volume data, and the data is a 'brick' (every voxel the same value)

### DIFF
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -456,11 +456,11 @@ class VolumeVisual(Visual):
         # Apply clim
         vol = np.array(vol, dtype='float32', copy=False)
         if self._clim[1] == self._clim[0]:
-            if not self._clim[0] == 0.:
+            if self._clim[0] != 0.:
                 vol *= 1.0 / self._clim[0]
         else:
             vol -= self._clim[0]
-            vol *= 1.0 / (self._clim[1] - self._clim[0])
+            vol /= self._clim[1] - self._clim[0]
         
         # Apply to texture
         self._tex.set_data(vol)  # will be efficient if vol is same shape
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -455,8 +455,12 @@ class VolumeVisual(Visual):
         
         # Apply clim
         vol = np.array(vol, dtype='float32', copy=False)
-        vol -= self._clim[0]
-        vol *= 1.0 / (self._clim[1] - self._clim[0])
+        if self._clim[1] == self._clim[0]:
+            if not self._clim[0] == 0.:
+                vol *= 1.0 / self._clim[0]
+        else:
+            vol -= self._clim[0]
+            vol *= 1.0 / (self._clim[1] - self._clim[0])
         
         # Apply to texture
         self._tex.set_data(vol)  # will be efficient if vol is same shape


QuLogic:
You should rebase and remove the 3 changes from the PR that was never applied.

Eric89GXL:
@midgetfc read this over first to help you understand what's going on:

https://github.com/LABSN/expyfun/blob/master/git_flow.rst

Then you can do something like this:
```
git checkout master
git fetch upstream  # add this as a repo if you don't have it already
git reset --hard upstream/master
git checkout -b add-volume-check2
git cherry-pick c1b5ddf
git push origin -u --force add-volume-check2:add-volume-check
```
Do those commands make sense to you after reading the information above?

midgetfc:
@Eric89GXL Yes, it makes sense. Thank you for the link.
This is what I think I've done :

   git checkout master
   git fetch upstream  # add this as a repo if you don't have it already
   git reset --hard upstream/master

restored my repository master branch history to the exact copy of upstream/master.
-  removed the two commits 7fb6733 and 9f541c. My mistake here was that I had not
   made a branch, but had commited to master. Correct ?
-  removed the commit e57937a. I don't see clearly here what I should've done, and
   what was wrong. Git asked for this commit when I wanted to make a new branch.
   Was it also a consequence of the two previous incorrect commits ?
 
   git checkout -b add-volume-check2
   git cherry-pick c1b5ddf

create a new branch from the head of the now clean master branch and pick-up the changes proposed originaly in c1b5ddf
 
   git push origin -u --force add-volume-check2:add-volume-check

copy the new repository configuration to github.

So now, if this PR gets closed, I will put my master branch up to date with
upstream/master and delete add-volume-check and add-volume-check2 branches as is
explained in the link you gave. Correct ?

And if I need to propose a fix again, must I make sure my local master branch has the exact same history as the upstream/master before creating my new branch and pushing it to github ?

I'm still learning...

Eric89GXL:
You don't need to close this PR, as now it's in the correct state.

In general, before starting a new branch to eventually open a PR, I always do:
```
git fetch upstream
git checkout master
git reset --hard upstream/master
git checkout -b new-branch
... hack hack hack ...
git commit -am 'FIX: Some fixes'
git push origin new-branch
```
This ensures that my commits go on top of whatever is up-to-date when I start working. It's not absolutely necessary to do that bit -- as long as your `master` only contains commits that are in the same order as `upstream/master` it should be fine, but the risk with branching from a not up-to-date `master` is that then you might get conflicts, where your new changes hit the same lines as changes that have been made in `upstream/master`.

vispy/visuals/volume.py
```diff
@@ -455,8 +455,12 @@ def set_data(self, vol, clim=None):
         
         # Apply clim
         vol = np.array(vol, dtype='float32', copy=False)
-        vol -= self._clim[0]
-        vol *= 1.0 / (self._clim[1] - self._clim[0])
+        if self._clim[1] == self._clim[0]:
+            if not self._clim[0] == 0.:
```
Eric89GXL: `if self._clim[0] != 0.:`

vispy/visuals/volume.py
```diff
@@ -455,8 +455,12 @@ def set_data(self, vol, clim=None):
...
+                vol *= 1.0 / self._clim[0]
+        else:
+            vol -= self._clim[0]
+            vol *= 1.0 / (self._clim[1] - self._clim[0])
```
Eric89GXL: `vol /= self._clim[1] - self._clim[0]`

midgetfc:
@Eric89GXL Thanks for your comments. I hope I've done all right with git this time. From the way it's described in the docs, I would'nt have tought about using git reset --hard routinely. But effectively, if the local master branch is solely used as a trunk to fork branches, it makes sense.

QuLogic:
If you never commit anything on `master`, you can just `git merge upstream/master`, of course. This should be a fast-forward merge, so you can use `--ff-only` to be extra careful.

Eric89GXL:
Sorry for the slow response @midgetfc, thanks for the fix!

PR #998, opened 2015-06-26
add a warning when setup.py is unable to load setuptools

bollu:
This is useful if someone is trying to use `python setup.py develop` but doesn't have `setuptools`.

There's a nested import (which loads `warnings.warn`). I will move that to the top level if it is  necessary, but I think the nested import is "neater" since it isolates usage

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,12 +34,12 @@ Announcing:
 
 import os
 from os import path as op
+from warnings import warn
 
 try:
     # use setuptools namespace, allows for "develop"
     import setuptools  # noqa, analysis:ignore
 except ImportError:
-    from warnings import warn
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -39,6 +39,8 @@
     # use setuptools namespace, allows for "develop"
     import setuptools  # noqa, analysis:ignore
 except ImportError:
+    from warnings import warn
```
Eric89GXL: warnings is a built-in package, you don't need to nest it

bollu:
this passed CI, and I changed the import to be top-level. Can this be merged?

Eric89GXL:
Thanks @bollu!

PR #999, opened 2015-07-05
FIX: Better skipping

Eric89GXL:
I noticed our testing suite was not efficiently skipping backends that aren't installed or don't work, this should fix it.


Eric89GXL:
Ready for review/merge from my end.

almarklein:
LGTM

PR #1003, opened 2015-07-08
FIX: Move docs

Eric89GXL:
The docs are now totally contained within the https://github.com/vispy/vispy-website repo. Ready for review/merge from my end.


campagnola:
What's the purpose of this? Isn't it standard practice to include documentation with the source?

Eric89GXL:
I think most packages do it that way, yeah. But for us it is much simpler for us to organize it so that the documentation is part of `vispy-website`. You can just do `make html` there, and it has all the right files to build properly. Otherwise you need some convoluted system to move files back and forth, since some of the website is in one place (vispy-website) and some of it is in a different repo (vispy). I can reconfigure it to do it that way again if you think it's worth it, though.

campagnola:
What if `make html` just clones (or pulls) the vispy source and builds the documentation from there?

Eric89GXL:
That would be okay I guess. It's a bit clunky because you have to build up extra machinery, have two copies of the repo floating around (or keep track of where each is), etc. But if keeping the `doc` folder in `vispy` is important then that might be the best solution.

almarklein:
I think having the docs in the website repo is ok. Most docs that we had in the source repo was auto-generated anyway. Putting it all (gallery, tutorials, examples, ref guide) in one place should simplify things somewhat...

Eric89GXL:
It definitely simplified the doc-building / uploading process, at least.

Eric89GXL:
And actually the docs that I removed contain very little actual documentation -- 95% of it is in the code itself. Our `.rst` files are really more like placeholders for docs. So we don't actually ship docs with our code at this point, we ship something that can potentially be built into docs...

campagnola:
I'm convinced. I don't think anyone builds their own docs anyway  :)
Ok to merge?

Eric89GXL:
Yeah good to go

Issue #555, opened 2014-09-10
Testing backends on Py 3.4 spams import warnings


almarklein:
When I run `python3 make test pyside` (or any backend) I get this:

```
OK (SKIP=14)
sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>
/usr/lib/python3.4/importlib/_bootstrap.py:2127: ImportWarning: sys.meta_path is empty
/usr/lib/python3.4/importlib/_bootstrap.py:2127: ImportWarning: sys.meta_path is empty
... and 200 or so more of those
```

If I run the full suite, I get this at the end of each backend test.

Apparently, this is a known problem in Python 3.4, see http://bugs.python.org/issue21049 and http://bugs.python.org/issue21052. It seems related to an unclosed open file or socket which is cleaned up very late because its stored at a class or is in a circular reference. The Python shutdown mechanism tries to close the resource, and this somehow invokes an import, which cannot work because `sys.meta_path` is empty.

I've tried to see if I can find the error but to no avail. Maybe @Eric89GXL has an idea? I already tried removing the `with open(os.devnull, 'w') as fid:` in test_app.py. Maybe something related to the subprocesses that we spamn during testing?

foxx:
@Eric89GXL You just saved me hours of debugging, it was driving me insane. Thank you.

Eric89GXL:
Heh -- we actually have the warnings again because we put the `import IPython` back in some of our tests, so I'm going to subscribe to that issue you opened.

jakirkham:
So, I think the issue from this `ResourceWarning` caused by the unclosed `devnull` file handle will be fixed by this ( https://github.com/ipython/ipython/pull/8929 ).

PR #1008, opened 2015-07-13
ENH: Use container-based infrastructure

Eric89GXL:
Let's see if we can speed up the build system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,48 @@
 language: python
 
-# Here we use anaconda for 2.6 and 3.4, since it provides the simplest
+# use container-based infrastructure
+sudo : false
+
+# install a newer cmake since at this time Travis only has version 2.8.7
+addons:
+  apt:
+    sources:
+    - kalakris-cmake
+    packages:
+    - cmake
+    - libgl1-mesa-dri
+    - libegl1-mesa
+    # old GLFW
+    - libglfw2
+    # GLFW build
+    - xorg-dev
+    - libglu1-mesa-dev
+    # SDL2 build
+    - mercurial
+    - libdbus-1-dev
+    - libgl1-mesa-dev
+    - libglu1-mesa-dev
+    - libpulse-dev
+    - libx11-dev
+    - libxcursor-dev
+    - libxext-dev
+    - libxi-dev
+    - libxinerama-dev
+    - libxrandr-dev
+    - libxss-dev
+    - libxt-dev
+    - libxv-dev
+    - libxxf86vm-dev
+    - libasound2-dev
+    - libts-dev
+    - libudev-dev
+
+# Here we use Anaconda for our builds, since it provides the simplest
 # interface for running different versions of Python. We could also use
 # it for 2.7, but the Ubuntu system has installable 2.7 Qt4-GL, which
 # allows for more complete testing.
 
 # Size testing can be skipped by adding "[size skip]" within a commit message.
-
-virtualenv:
-    system_site_packages: true
-
 
 env:
     # Enable python 2 and python 3 builds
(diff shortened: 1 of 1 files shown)


bollu:
@Eric89GXL what kind of containers are you talking about? do you want to use pre-baked Docker containers so we only need to pull the latest Vispy?

Eric89GXL:
We could perhaps do that eventually, but for now the idea is just to have the builds start and run faster:

http://docs.travis-ci.com/user/migrating-from-legacy/

QuLogic:
So many fixes, you ran out of x's...

FYI, you can also place `addons` in the `env` matrix so that you don't need to install the full deps on the minimal test.

.travis.yml
```diff
@@ -76,14 +96,9 @@ before_install:
...
     # Install numpy, flake
-    - if [ "${PYTHON}" != "2.7" ]; then
-        conda create -n testenv --yes --quiet pip python=$PYTHON > ${REDIRECT_TO};
-        source activate testenv > ${REDIRECT_TO};
-        conda install --yes --quiet numpy nose pytest > ${REDIRECT_TO};
-      else
-        travis_retry sudo apt-get -qq -y install python-numpy python-setuptools > ${REDIRECT_TO};
-        pip install -q pytest > ${REDIRECT_TO};
-      fi;
+    - conda create -n testenv --yes --quiet pip python=$PYTHON > ${REDIRECT_TO};
```
QuLogic: If it's `--quiet`, what's the use of redirecting?

Eric89GXL:
Yeah I am still tweaking :)

I didn't know that about the `env` matrix, that would be useful indeed

.travis.yml
Eric89GXL: it still has output

.travis.yml
```diff
@@ -19,36 +52,23 @@ env:
...
-    - PYTHON=2.7 DEPS=full TEST=examples  # test examples
-    - PYTHON=3.4 DEPS=full TEST=examples
-
-    - PYTHON=3.4 DEPS=minimal TEST=extra  # test file sizes, style, line endings
+    - PYTHON=3.4 DEPS=full TEST=examples  # test examples
+    - PYTHON=3.4 DEPS=minimal TEST=standard  # also tests file sizes, style, line endings
 
 
 before_install:
     - REDIRECT_TO=/dev/stdout  # change to /dev/null to silence Travis
```
QuLogic: With Travis' folding, I never really saw the point in this.

.travis.yml
```diff
@@ -150,38 +136,32 @@ install:
...
         if [ "${PYTHON}" != "2.6" ]; then
-          cd ~;
-          travis_retry sudo apt-get -qq -y install mercurial libdbus-1-dev libgl1-mesa-dev libglu1-mesa-dev libpulse-dev libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxrandr-dev libxss-dev libxt-dev libxv-dev libxxf86vm-dev libasound2-dev libts-dev libudev-dev > ${REDIRECT_TO};
           wget -q http://www.libsdl.org/release/SDL2-2.0.3.tar.gz;
           tar xzf SDL2-2.0.3.tar.gz &> ${REDIRECT_TO};
           cd SDL2-2.0.3;
           mkdir build;
           cd build;
-          ../configure --disable-audio --disable-render --disable-joystick --disable-haptic --disable-power --disable-cpuinfo &> ${REDIRECT_TO};
+          ../configure --prefix=$HOME --disable-audio --disable-render --disable-joystick --disable-haptic --disable-power --disable-cpuinfo &> ${REDIRECT_TO};
```
QuLogic: Does `$HOME/.local` work? (with possibly pre-existing search paths?)

.travis.yml
Eric89GXL: There was a span of time (a week or a month?) where the folding didn't work properly, so it was necessary back then. Now it's probably not necessary, but also isn't doing too much harm, so we'll probably remove it eventually.

.travis.yml
Eric89GXL: I've tried setting LD_LIBRARY_PATH, that doesn't work. We used to use standard install directories and use `sudo ldconfig`, but we can't do that anymore.

Eric89GXL:
@QuLogic the matrix does seem to speed up a few of the builds quite a bit, thanks for that tip. Feel free to have another look and see if it seems optimized to you.

Eric89GXL:
@campagnola @almarklein this one is ready for merge, this speeds up our builds a bit

almarklein:
Out of interest, what is the speed gain? Anyway, the `.travis` seems a bit cleaner. LGTM

Eric89GXL:
Examples take about 7 min instead of about 14, primary builds about 9 instead of about 16

Eric89GXL:
feel free to merge if you're happy, one set of eyes is probably enough on this one

almarklein:
You asked @QuLogic to have another look above. Is he happy? :)

almarklein:
(I find `.travis` files not the easiest to read)

Eric89GXL:
We can always do another revision if there are additional ideas :)

.travis.yml
```diff
@@ -1,54 +1,78 @@
...
-    - PYTHON=3.4 DEPS=full TEST=examples
-
-    - PYTHON=3.4 DEPS=minimal TEST=extra  # test file sizes, style, line endings
+    - env: PYTHON=2.6 DEPS=full TEST=standard   # functionality
+      addons:
+        apt:
+          sources:
+          packages:
+            - *mesa_apt
+    - env: PYTHON=2.7 DEPS=full TEST=standard IPYTHON="=2" # NUMPY="=1.7"
```
QuLogic: Any reason to keep the `NUMPY` bit here?

Eric89GXL: yeah, I want to put it back, but doing so breaks a file that I don't want to change until #928 is merged :)

QuLogic:
Yea, seems fine to me.

campagnola:
Looks good to me. Any idea what all the `_bootstrap.py:2150: ImportWarning: sys.meta_path is empty` is about?

Eric89GXL:
I've seen that warning before, I think it's some weird Py3k thing from
another package.

QuLogic:
#555 ?

Eric89GXL:
Yeah, that one had it. That issue suggests it's from IPython. I think I've seen it before in other contexts, too.

bollu:
Just for the record, [from the Travis blog](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/):

>Can I provide my own Docker containers?

>The build environments we're currently using on our Docker-based setup provide the same services, programming languages and tools offered by our legacy stack. ...

>We're not there yet in allowing you to bring your own, but it's something we have in mind for the future.

I think once it is allowed, we could roll our own `Docker` images?

Eric89GXL:
Yeah probably, but I wouldn't hold my breath. They actually currently allow caching of directories, too, but only for paid users. The custom images will probably also go to paid users first, too, if I had to guess.

Issue #914, opened 2015-05-19
Add support for "TurntableCamera" translate speed.


KelSolaar:
I have some issues when using a `vispy.scene.cameras.TurntableCamera` with user defined distance.

In my application the camera is sometime set quite close of the origin and I suspect the combination of that distance and the large scene bounding box is enough to generate values in `vispy.scene.cameras.TurntableCamera._dist_to_trans` that predate correct traveling as you can see in this video: https://drive.google.com/file/d/0B_IQZQdc4Vy8S0dTNEd5Q1hLWE0/view?usp=sharing

For now I have created a new camera that inherits from `TurntableCamera` and overrides `_dist_to_trans` as follows:

``` python
    def _dist_to_trans(self, distance):
        translate = np.asarray(TurntableCamera._dist_to_trans(self, distance))
        translate *= self.__translate_speed

        return translate
```

jdreaver:
I had a pull request (#714) that implemented a consistent panning speed for cameras, but the camera code changed so much before it was merged that it didn't work anymore. It worked like this: if you imagine a flat surface in front of the camera, as if there is a glass window in front of you, then panning the view is like grabbing a part of that window and moving yourself. That is, the point under your mouse when you start panning stays under your mouse. I kind of don't like the current panning system because it feels a little "slippery". What you have shown here is an extreme case.

In order to reimplement the PR, I would have to sit down and really understand the camera changes (the camera code is more complex so it could be more general). However, I do explain the concept inside the PR.

KelSolaar:
That looked great and seems to be perfect in my context.

Eric89GXL:
I don't see a problem with having some multiplicative parameter like you have @KelSolaar. @campagnola what did you and @almarklein have in mind for this sort of thing?

@jdreaver yeah we want picking, hopefully one day the "handle" thing will be possible. But for now we might have to go for a less ideal but more easily workable solution. It would be good if you could try to reimplement your changes in the new system at some point, but that will probably take some work.

jdreaver:
> @jdreaver yeah we want picking, hopefully one day the "handle" thing will be possible. But for now we might have to go for a less ideal but more easily workable solution. It would be good if you could try to reimplement your changes in the new system at some point, but that will probably take some work.

Agreed. Simple is definitely best for now.

almarklein:
I agree that the mouse should stay in the same spot. What's shown in the video is just wrong :)

KelSolaar:
[...]

PR #1021, opened 2015-07-27
ENH: Scenegraph update

### LINKED ISSUE #140: ENH: picking
Vispy needs a system for picking that can provide:

* The top-most visual drawn at (x,y) on a canvas
* A depth-ordered list of all visuals drawn at (x,y)
* The same for larger / arbitrary areas such as all visuals drawn within a rectangular area or a lasso
* Which primitives (triangle / line / sprite) of a visual are in the picking area
* The location in the local coordinate system of an entity that corresponds to a canvas pixel

### LINKED ISSUE #434: New example: grid plot with individual zooming
Porting [this example from glumpy](https://github.com/rougier/glumpy/blob/master/examples/collection-custom-lines.py). Requires @rougier's collections (#410).

Further examples that could be ported to Vispy:

https://github.com/rougier/glumpy/blob/master/examples/snippet-cubes.py
https://github.com/rougier/glumpy/blob/master/examples/snippet-lines.py

### LINKED ISSUE #677: Axes, ticks, labels, grid visuals

### LINKED ISSUE #693: Picking example
Showing how to implement picking or tooltip manually with `on_mouse_press` and `on_mouse_release` (might be better with `on_click` #271 but we don't have that yet), with a scatter plot or a mesh for example.

Useful until picking is properly implemented in the scene graph.

### LINKED ISSUE #791: Reuse FBO in viewbox
Right now a new FBO object is created each draw, because at the time of writing gloo was not mature enough.

### LINKED ISSUE #927: Add a way to retrieve the current "SceneCanvas" instance from the "SceneGraph".
It is maybe implemented but I didn't find it, I need a way to find the current `SceneCanvas` instance a `SceneGraph` node, specifically a `ViewBox`, at the moment I'm passing the current `SceneCanvas` instance to my views which is not particularly elegant. I tried to walkup the graph, there is a node named **canvas_cs** up there but I didn't see which attribute links it to the `SceneCanvas` instance, I'm not even sure it is related.

### LINKED ISSUE #953: Drawing Precision Issues with 3D Cameras
I have important drawing precision issues with the 3D cameras that I pinpointed to be coming from the `BaseCamera._get_depth_value ` private method: https://github.com/vispy/vispy/blob/master/vispy/scene/cameras/cameras.py#L128

Oddly the doc string specifies that:

```
For 24 bits and more, we're fine with 100.000,
```

but the value returned is **100000.0**, putting that back to **100** or even **1** fixes the issues.

### LINKED ISSUE #965: scene_test_* examples broken
Will hopefully be fixed by changes in #928 (cc @campagnola), thanks for finding it @QuLogic

### LINKED ISSUE #975: Text in TextVisual example does not stay centred

### LINKED ISSUE #981: using a TextVisual leads to unintended scaling of other Visual
I'm not sure if creating an issue is the right thing to do, but I haven't been able to debug this, so I decided to go ahead and just create one.

The problem is that when I use a ```TextVisual``` with multiple strings (as the constructor allows, it throws off the scaling _completely_ and causes the entire scene to become really tiny. I don't know why this happens.

Also, it __does  not__ happen if you have a ```TextVisual``` with just one string. This to me is even more perplexing. 

***Example of extreme scaling down:***

screenshot from 2015-06-07 01 02 36

[example in which this happens](https://github.com/bollu/vispy/blob/colorbar-visual/examples/basics/visuals/colorbar_visual.py)

[offending line of code](https://github.com/bollu/vispy/blob/colorbar-visual/vispy/visuals/colorbar.py#L195)

***Proper rendering with one ```TextVisual```***

screenshot from 2015-06-07 01 07 16

### LINKED ISSUE #986: Toggling visibility on a "Node" should "optionally" toggle visibility of its children.
This one is actually highly debatable: at the moment it is possible to make a `Node` invisible in the `SceneGraph` but its children will still remain visible.

Usually in DCC application, I'm especially thinking about Autodesk Maya as it is the one that is the most used in VFX and Game industries, when you hide a node all the children nodes are getting hidden.

Right now I have to manually walk the `Node` children down to have that behaviour. I will probably end up monkey patching `vispy.scene.visuals.Node` directly as I don't want to create classes for things that I can build with a simple definition (for example a visual that is built around `Markers` and `Lines` with everything parented under a `Node`).

Eric89GXL:
Opening as a PR in-repo so we can see if the last few fixes make Travis happy, and so that additional PRs can be opened against a branch that people get notified about. Thanks for the suggestion @QuLogic!

* [x] Remove multi-parenting in scenegraph
* [x] Nodes have a single transform (and related simplifications)
* [x] SceneCanvas.draw_visual should be simple and efficient
* [x] Implement picking (finally)
* [ ] ViewNode and a way to automatically create a view on a scenegraph subtree
* [x] Other simplifications:
    * [x] Remove all ViewBox clip methods except fragment. 
    * [x] Remove render/framebuffer/canvas scenegraph nodes (might restore these later if there turns out to be a need)

Visuals to migrate:

* [x] Line
* [x] Mesh
* [x] GridLines
* [x] Markers
* [x] Volume
* [x] Image
* [x] Text
* [x] Axis
* [x] Polygon
* [x] Box
* [x] Cube
* [x] Ellipse
* [x] Isocurve
* [x] Isosurface
* [x] Plane
* [x] Rectangle
* [x] RegularPolygon
* [x] Spectrogram
* [x] Tube
* [x] XYZAxis
* [x] Histogram
* [x] LinePlot
* [x] Isoline
* [x] SurfacePlot
* [x] ColorBar

EDIT BY Eric89GXL:

Related issues:
* [x] Toggling a node toggles children #986
* [x] Text visual changing viewport #981
* [x] Type of Canvas size #976
* [x] Centering of text visuals #975: working now?
* [x] Scene benchmarks broken #965: need to fix?
* [x] Depth drawing precision #953: need to add?
* [x] Get SceneCanvas from node #927: exists? 
* [x] Plane and box PR #901: need to update visuals
* [x] Original PR #895
* [x] Reuse FBO in viewbox #791
* [x] Picking example #693
* [x] Axes, ticks, grid #677
* [x] Grid plot with individual zooming #434: does it work?
* [x] Picking #140
* [ ] Line editing example PR #926: modify to use picking?
* [ ] Camera translate speed #914: need to add?
* [ ] Performance of the SceneGraph #628

closes #986
closes #981
closes #976
closes #975
closes #965
closes #953
closes #927
closes #791
closes #693
closes #677
closes #434
closes #140


bollu:
Should we rebase our code on top of this?

Eric89GXL:
Sure, that would work

Eric89GXL:
@campagnola green light :) Ready for review/merge from my end.

almarklein:
I'll try to review this today :)

vispy/scene/canvas.py
```diff
@@ -241,73 +262,205 @@ def draw_visual(self, visual, event=None, viewport=None):
         event : None or DrawEvent
             Optionally specifies the original canvas draw event that initiated
             this draw.
-        viewport : tuple | None
-            Optionally specifies the viewport to use. If None, the entire
-            physical size is used.
         """
-        self.set_current()
         prof = Profiler()
```
almarklein: Is this profiler stuff supposed to be here, or is it an artifact that should be removed?

Eric89GXL: It should be okay to stay, if profiling is disabled it should have almost no overhead

vispy/scene/canvas.py
```diff
@@ -241,73 +262,205 @@ def draw_visual(self, visual, event=None, viewport=None):
...
+            vis = VisualNode._visual_ids.get(id_[10, 10], None)
+        except RuntimeError:
+            # Don't have read_pixels() support for IPython. Fall back to
+            # bounds checking.
+            return self._visual_bounds_at(pos)
+        return vis
+
+    def _visual_bounds_at(self, pos, node=None):
+        """Find a visual whose bounding rect encompasses *pos*.
+        """
```
almarklein: This also works for 3D?

vispy/scene/canvas.py
```diff
@@ -241,73 +262,205 @@ def draw_visual(self, visual, event=None, viewport=None):
...
+        visual : instance of Visual | None
+            The visual at the position, if it exists.
+        """
+        tr = self.transforms.get_transform('canvas', 'framebuffer')
+        fbpos = tr.map(pos)[:2]
+
+        try:
+            id_ = self._render_picking(region=(fbpos[0]-10, fbpos[1]-10,
+                                               20, 20))
+            vis = VisualNode._visual_ids.get(id_[10, 10], None)
```
almarklein: Why render a 20x20 region if we only use one pixel?

vispy/scene/canvas.py
```diff
@@ -241,73 +262,205 @@ def draw_visual(self, visual, event=None, viewport=None):
...
+        ----------
+        pos : tuple
+            (x, y) position at which to find visuals.
+        radius : int
+            Distance away from *pos* to search for visuals.
+        """
+        tr = self.transforms.get_transform('canvas', 'framebuffer')
+        pos = tr.map(pos)[:2]
+
+        id = self._render_picking(region=(pos[0]-radius, pos[1]-radius,
```
almarklein: I would expect a fallback for IPython here too, or at least a try-except.

vispy/scene/visuals.py
```diff
@@ -13,11 +13,88 @@
...
+        if clipper is not None:
+            self.attach(clipper)
+            self._clippers[node] = clipper
+
+    @property
+    def picking(self):
+        """Boolean that determines whether this node (and its children) are
+        drawn in picking mode.
+        """
+        return self._picking
```
almarklein: This looks weird to me. Should this not be an attribute of the draw event? Also, what happens when the user sets picking to True (since its a public property)?

vispy/scene/node.py
```diff
@@ -86,6 +100,62 @@ def name(self, n):
...
+        return self._clipper
+        
+    @property
+    def order(self):
+        """A value used to determine the order in which nodes are drawn.
+        
+        Greater values are drawn later. Children are always drawn after their
+        parent.
+        """
+        return self._order
```
almarklein: This is nice.

vispy/scene/node.py
```diff
@@ -242,14 +384,16 @@ def set_transform(self, type_, *args, **kwargs):
         """
         self.transform = create_transform(type_, *args, **kwargs)
 
-    def _transform_changed(self, event):
+    def _update_trsys(self, event):
```
almarklein: What is this? maybe a more descriptive name, or a docstring to explain

vispy/scene/widgets/widget.py
```diff
@@ -130,6 +134,54 @@ def inner_rect(self):
...
+        
+        If either stretch factor is None, then it will be assigned when the
+        widget is added to a layout based on the number of columns or rows it
+        occupies.
+        """
+        return self._stretch
+
+    @stretch.setter
+    def stretch(self, s):
+        self._stretch = s
```
almarklein: `self._stretch = float(s[0]), float(s[1])`

vispy/scene/widgets/widget.py
```diff
@@ -130,6 +134,54 @@ def inner_rect(self):
...
+        self._update_layout()
+
+    @property
+    def fixed_size(self):
+        """Fixed size (w, h) of the widget.
+        
+        Specifying a fixed size for either axis forces the widget to have a 
+        specific size in a layout. Setting either axis to None allows the 
+        widget to be resized by the layout.
```
almarklein: There is a contradiction here. What happens if I provide `None, 100`? As I understand it, this allows a widget to override a layout? Will the layout simply ignore the widget?

vispy/scene/widgets/widget.py
```diff
@@ -130,6 +134,54 @@ def inner_rect(self):
...
+        """
+        return self._fixed_size
+    
+    @fixed_size.setter
+    def fixed_size(self, s):
+        self._fixed_size = s
```
almarklein: a sanity check would be nice.

campagnola:
Sorry everybody! I have been travelling and interviewing non-stop since SciPy ended. I will finally have more time again. Also, THIS IS AWESOME! Reviews forthcoming...

vispy/visuals/regular_polygon.py
```diff
@@ -20,56 +19,41 @@ class RegularPolygonVisual(EllipseVisual):
...
-        super(EllipseVisual, self).__init__()
-        self.mesh.mode = 'triangle_fan'
-        self._pos = pos
-        self._color = Color(color)
-        self._border_color = Color(border_color)
-        self._radius = radius
-        self._sides = sides
-        self._update()
+    def __init__(self, center=None, color='black', border_color=None,
+                 border_width=1, radius=0.1, sides=4, **kwargs):
```
almarklein: I know this is not entirely part of this PR, but what is a "regular" polygon, and why does it inherit from EllipseVisual?

Eric89GXL: e.g. octagon, hexagon, pentagon. Basically equally spaced divisions. An ellipse is a special case of this (or vice-versa, I guess)

bollu: Yep, It uses the `EllipseVisual` to divide 360 degrees into equi-partitions, by controlling the number of sides of the `EllipseVisual`

almarklein: Can the inheritance be changes so that we have RegularPolygon and EllipseVisual inherit from that? That makes much more sense semantically... (but in a new PR, not this one)

bollu: No, I don't think so, because a RegularPolygon uses the `EllipseVisual` like a circle (overlapping foci). So you can't create an ellipse from a RegularPolygon, because the RegularPolygon constructs a circle, and not an ellipse.

vispy/visuals/visual.py
```diff
@@ -1,34 +1,135 @@
...
+    class PlotLineVisual(visuals.CompoundVisual):
+        def __init__(self, ...):
+            self._line = LineVisual(...)
+            self._point = PointVisual(...)
+            visuals.CompoundVisual.__init__(self, [self._line, self._point])
+
+A compound visual will automatically handle forwarding transform system changes
+and filter attachments to its internally-wrapped visuals. To the user, this
+will appear to behave as a single visual.
+"""
```
almarklein: This looks really nice now.

vispy/visuals/regular_polygon.py
bollu: But, we could perhaps make it a has-a relationship rather than is-a by making the `EllipseVisual` a member of `RegularPolygon`? we're actually perpetrating implementation inheritance here :)

examples/basics/scene/stereo.py
```diff
@@ -8,6 +9,8 @@
 Example demonstrating stereo vision in a scene has an anisotropic aspect
 ratio. This example can be used to test that the cameras behave
 correctly with nested translated/rotated cameras.
+
+NOTE: This example is currently broken!
```
almarklein: I assume it is still possible though? Can we have this example fixed in this PR, since it's a good validation?

Eric89GXL: I'd prefer to have it in another PR, having divided branches like this for
as long as we have is not good... but if you think it is critical we can
wait.

almarklein:
Another epic PR: a whole new scene graph, picking, axis ... wow!

Stuff / examples that fail:
* The fly camera does not react to the mouse. See e.g. volume.py (and similar examples like flipped_axis.py). The fly camera also produces some errors when the keys are used while it is not the active camera.
* In `one_cam_two_scenes.py` the 2D camera is not initialized correctly (its very much zoomed in). Also seen for 2D cam in `one_scene_four_cams.py`
* `line_update.py` is all black for me
* `nested_viewbox.py` does not look right
* Detail (can be fixed later): in `demo/plot/plot.py` you can see how the tick marks are not exactly centered on the gridlines.
* In `demo/scene/scrolling_plots.py` is the magnifyer thing supposed to follow the mouse? (it does not for me)
* `demo/scene/*.py is just freakin' awesome!

vispy/visuals/regular_polygon.py
almarklein: Let's look at this in another issue/PR. It's not the end of the world, but I found the inheritance confusing. Maybe documenting it is sufficient.

examples/basics/scene/stereo.py
almarklein: Another PR is fine.

QuLogic:
> In demo/scene/scrolling_plots.py is the magnifyer thing supposed to follow the mouse? (it does not for me)

This one does not work for me. It's the same issue as the oscilloscope; 10000 is over the max texture size.

Eric89GXL:
The example is supposed to work, but it's not supposed to scroll until you click with the mouse, or activate the mouse wheel. Right now mouse events, if passed to the event handlers, pass to the right place by using picking. I think we probably need two modes: a "simple" mode that passes somehow based on mouse position for `SceneCanvas` (i.e., will probably only ever go to a Viewbox), and an "specific" mode that actually does the draw and picks the visual under the mouse. It's less parsimonious than having the current picking-only system, but it will have much better performance, especially for those types of examples.

almarklein:
What about allowing visuals to "grab" the mouse? Anyway, any of this can be added later.

Eric89GXL:
Well I'm thinking that, in most (90%+) `SceneGraph` cases, interaction via the mouse does not need to be done via picking. In most applications, people want to be able to control the camera with the mouse. We used to just use the mouse position plus some logic regarding the `SceneGraph` item layout to determine which `ViewBox` to send the event to, with very low overhead. It would be nice to have this mode available again, even if it has to be manually switched in some way. Maybe it should be on by default, and if `picking` mode is turned on, it's disabled. We have to think about it.

I think of a `Visual` "grabbing" the mouse as slightly separate, since the `ViewBox` isn't actually a `Visual`. It is a `Node` in the Scenegraph and has `Visual` children (including its background/border), but I don't think that's what you mean... in which case it's a slightly different issue from what I'm thinking of.

michaelaye:
Not sure it's related, but why does the cube get truncated in the https://github.com/vispy/vispy/blob/scenegraph-update/examples/basics/scene/cube.py example after 
camera trigger:

screenshot 2015-07-29 13 01 45

after being displayed correctly at launch:
screenshot 2015-07-29 13 04 01

vispy/scene/canvas.py
campagnola: Not yet. I'm hoping we can eventually support proper picking everywhere, but if not, then we'll need to fill this out.

vispy/scene/canvas.py
campagnola: It's an artifact. I was playing with a few different methods of deciding which item was clicked on, but ultimately ended up using the separate `visuals_at` method instead. I expect these to evolve as we use them more.

vispy/scene/canvas.py
campagnola: Agreed. Ok if I open an issue and kick it down the road?

Eric89GXL: +1 for opening an issue that picking will fail for IPython. It would also be a good place do discuss the idea of having multiple mouse-interaction modes.

vispy/scene/visuals.py
campagnola: It is weird. I wend back and forth on it a few times. I'd be happy for it to change in the future if there is good cause. If the user sets it True, then the object will be rendered using its solid picking ID color.

Eric89GXL: For now maybe we should just make it private, so that users don't think they should use it.

vispy/scene/node.py
campagnola: Docs added (in upcoming sub-PR)

campagnola: https://github.com/vispy/vispy/pull/1030

vispy/scene/widgets/widget.py
campagnola: For `(None, 100)`, the widget will always have a height of 100 and the layout must adjust other items to accommodate this constraint, whereas the width is unconstrained and chosen entirely by the layout. However, this has actually not been implemented yet so I'm hiding the property temporarily (until the Cassowary recommendation in #277 is implemented).

vispy/scene/widgets/widget.py
campagnola: Done.

campagnola:
> * The fly camera does not react to the mouse. See e.g. volume.py (and similar examples like flipped_axis.py). The fly camera also produces some errors when the keys are used while it is not the active camera.
> * In one_cam_two_scenes.py the 2D camera is not initialized correctly (its very much zoomed in). Also seen for 2D cam in one_scene_four_cams.py
> * line_update.py is all black for me
> * nested_viewbox.py does not look right

All fixed in #1030.

> * Detail (can be fixed later): in demo/plot/plot.py you can see how the tick marks are not exactly centered on the gridlines.

The ticks are 2 pixels and the grid lines are 1 pixel. Not much to be done about this except get a better screen :)   (ok, maybe the ticks and grid lines should have the same width)

> * In demo/scene/scrolling_plots.py is the magnifyer thing supposed to follow the mouse? (it does not for me)

No, right now it stays at the location where the mouse wheel was rolled. This was done for my own convenience, but it also happens to make more sense as a real use case.

Eric89GXL:
@almarklein Luke addressed the above comments in #1030, can you look and see if you're happy now? If so, feel free to hit the big green button.

Eric89GXL:
Excellent, the only failure is for size.

vispy/scene/canvas.py
almarklein: Can we add a note to explain when it will work and when not?

vispy/scene/canvas.py
almarklein: That's ok. Note that you *do* have a fallback a few lines up (line 387)

vispy/scene/visuals.py
almarklein: Yes, making it private would be good. But more fundamentally, it seems weird that this is a *state* of the visual. It should always be False, unless during a draw for picking.

almarklein:
> The ticks are 2 pixels and the grid lines are 1 pixel. Not much to be done about this except get a better screen :) (ok, maybe the ticks and grid lines should have the same width)

I'd say ticks and gridlines should either be one pixel, or if wider, they should be antialiased so that their center represents the exact position.

almarklein:
Still some spurious errors when you use the fly-camera keys while it is not yet active, and also when you CTRL+RMB: `File "/home/almar/dev/pylib/vispy/vispy/scene/cameras/fly.py", line 431,
AttributeError: 'SceneMouseEvent' object has no attribute 'map_to_canvas'`. But Ok to fix later.

almarklein:
Thanks @campagnola, @Eric89GXL and everyone else who has contributed to this PR! This makes Vispy so much nicer :D

Issue #277, opened 2014-06-23
ENH: Grid layouts


campagnola:
We will need a way to lay out rectangular widgets in grids that resize nicely with the canvas area. Layouts can be a very complex problem, but hopefully that does not mean they must have a complex API. Some features we might like to have:
- Widgets inside a grid may have minimum, maximum, and preferred sizes
- Should be possible to specify that widgets have sizes relative to each other
- Widgets may span multiple rows / columns
- Adding / removing widgets from the layout causes other widgets to adjust dynamically
- HTML table-like features - cell spacing, padding, borders, ...
- User-adjustable boundaries between some rows / columns
- Widgets can also have min / max aspect ratio (but this might be low priority and difficult to implement)

Example: 4 widgets in a row; the first widget has a preferred size of 100px, and min/max of 50 and 150px. The remaining space is divided up between 3 widgets: 50%, 30%, and 20%. In the second row, a single widget has a maximum height of 100px and occupies columns 1-3.

njsmith:
As I was telling @campagnola at the bar last night, I think these days you need a good reason to do layout using anything *other* than something based on Cassowary. (This is also what the matplotlib folks are trying to figure out how to move to, but they have nastier backcompat issues.)

Basically the idea in Cassowary is, your toolkit/user defines a bunch of arbitrary linear inequality constraints ("the left edge of this button must be to the left of the text inside it, and the right edge must be to the right", "these two axes should be the same width", "the left and right edges of this axes should line up with where the points at positions x1 and x2 fall in this other axis after applying that axes's linear view projection"), and then an incremental constraint solver tells you where to put everything to maximally satisfy the constraints and how things should adapt when the window is resized etc. It almost sounds too good to be true, but it's extremely usable in practice; iOS and OS X have been using this as their main layout approach for years now.

Wikipedia: https://en.wikipedia.org/wiki/Cassowary_%28software%29
Quick overview: http://stacks.11craft.com/cassowary-cocoa-autolayout-and-enaml-constraints.html
List of implementations of the constraint solver (pure Python, Python + C++, Javascript, ...): http://overconstrained.io/
Classic paper: https://constraints.cs.washington.edu/solvers/cassowary-tochi.pdf

Eric89GXL:
Sounds like a good approach to me. @njsmith fancy implementing it here? :)

almarklein:
Can we target it with ctypes?

MatthieuDartiailh:
[...]

PR #1022, opened 2015-07-27
[WIP] Colorbar - vispy.plot

bollu:
Checklist:

- [x] add a `colorbar` function to `PlotWidget`
- [x] separate out border rendering into `BorderVisual`
- [x] integrate border rendering to `ViewBox`
- [x] fix coordinate systems for `ColorBar`
- [x] fix text overlapping
- [x] add an example of using `plotwidget.colorbar` with `vispy.plot`
- [x] change ticks orientation so they are always left-to-right
- [ ] create a new `ColorBarWidget`
- [ ] add an example for `ColorBarWidget`
- [ ] write docs on `PlotWidget.colorbar(..)`
- [ ] write docs on `ColorBarWidget`
- [ ] write test cases for `ColorBarWidget`

### DIFF
--- a/vispy/visuals/colorbar.py
+++ b/vispy/visuals/colorbar.py
@@ -6,11 +6,11 @@
 # Author: Siddharth Bhat
 # -----------------------------------------------------------------------------
 
-from . import Visual, TextVisual, CompoundVisual, PolygonVisual
-from .shaders import ModularProgram, Function
-from ..color import get_colormap, Color
-
 import numpy as np
+
+from . import Visual, TextVisual, CompoundVisual
+from .shaders import Function
+from ..color import get_colormap, Color
 
 VERT_SHADER = """
 attribute vec2 a_position;
@@ -97,6 +97,17 @@ void main() {
 
 
 class _CoreColorBarBorderVisual(Visual):
+    """
+    Visual subclass that renders the borders.
+
+    Note
+    ----
+    This is purely internal.
+    Externally, the ColorBarVisual must be used.
+    This class was separated out to encapsulate rendering information
+    That way, ColorBar simply becomes a CompoundVisual
+    """
+
     def __init__(self, center_pos, halfdim,
                  border_width=1.0,
                  border_color="black",
@@ -110,21 +121,15 @@ class _CoreColorBarBorderVisual(Visual):
 
         Visual.__init__(self, vcode=VERT_SHADER_BORDER,
                         fcode=FRAG_SHADER_BORDER, **kwargs)
-        # self.prepar()
 
     @staticmethod
     def _prepare_transforms(view):
-        # transfrorm = view.transforms.get_transform()
-        print("preparing transforms - corecolorbar")
 
         program = view.shared_program
         program.vert['visual_to_doc'] = \
             view.transforms.get_transform('visual', 'document')
         program.vert['doc_to_render'] = \
             view.transforms.get_transform('document', 'render')
-
-        # total_transform = view.transforms.get_transform()
-        # program.vert['transform'] = total_transform
 
     def _update(self):
         x, y = self._center_pos
@@ -204,16 +209,12 @@ class _CoreColorBarVisual(Visual):
     def __init__(self, center_pos, halfdim,
                  cmap,
                  orientation,
-                 border_width=1.0,
-                 border_color="black",
                  **kwargs):
 
         self._cmap = get_colormap(cmap)
         self._center_pos = center_pos
         self._halfdim = halfdim
         self._orientation = orientation
-        self._border_width = border_width
-        self._border_color = border_color
         self._text_padding = 0
         # setup border rendering
 
@@ -305,16 +306,12 @@ class _CoreColorBarVisual(Visual):
         self._cmap = get_colormap(cmap)
         self._program.frag['color_transform'] = Function(self._cmap.glsl_map)
 
-
     @property
     def text_padding(self):
         return self._text_padding
 
     @staticmethod
     def _prepare_transforms(view):
-        # transfrorm = view.transforms.get_transform()
-        print("preparing transforms - corecolorbar")
-
         # figure out padding by considering the entire transform
         # on the width and height
         program = view.view_program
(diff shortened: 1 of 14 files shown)


bollu:
So, I _finally_ figured this out (hopefully) :) AFAIK, the solution works, and it makes sense to me theoretically.

So, pictures first:

#####  1) default (visual) coordinates with no transforms:

screenshot from 2015-07-29 01-15-54


##### 2) `SceneGraph` coordinates (flipped y axis):

screenshot from 2015-07-29 01-16-11

#####  3) random transforms by attaching to a scene with a turnatable camera:
screenshot from 2015-07-29 01-16-49
screenshot from 2015-07-29 01-17-14

vispy/visuals/colorbar.py
```diff
@@ -436,75 +314,166 @@ def _update(self):
...
+        (halfw, halfh) = halfdim
+
+        visual_to_doc = transforms.get_transform('visual', 'document')
+        doc_to_visual = transforms.get_transform('document', 'visual')
+
+        # doc_widths = visual_to_doc.map(np.array([halfw, halfh, 0, 0],
+        #                                         dtype=np.float32))
+
+        doc_x = visual_to_doc.map(np.array([halfw, 0, 0, 0], dtype=np.float32))
+        doc_y = visual_to_doc.map(np.array([0, halfh, 0, 0], dtype=np.float32))
```
bollu: take the vectors along the x and y axes from visual coordinates to document coordinates.

vispy/visuals/colorbar.py
```diff
@@ -436,75 +314,166 @@ def _update(self):
...
+
+        if doc_x[0] < 0:
+            doc_x *= -1
```
bollu: if the projection along x axis is negative, flip it, since this will be our new "basis" to position

vispy/visuals/colorbar.py
```diff
@@ -436,75 +314,166 @@ def _update(self):
...
+
+        if doc_y[1] < 0:
+            doc_y *= -1
```
bollu: do the same thing with the y axis. Flip if the projection is negative.

vispy/visuals/colorbar.py
```diff
@@ -436,75 +314,166 @@ def _update(self):
...
+
+        # doc_halfw = np.abs(doc_widths[0])
+        # doc_halfh = np.abs(doc_widths[1])
 
-    @staticmethod
-    def _get_anchors(orientation):
         if orientation == "top":
-            return "center", "bottom"
+            doc_perp_vector = -doc_y
```
bollu: Now, since we know the "canonical" directions in the document coordinates, figuring out the placement is dead simple. If you want to push the label "up", push it by `-doc_y`. Ditto with the other directions.

vispy/visuals/colorbar.py
```diff
@@ -436,75 +314,166 @@ def _update(self):
...
-            return "center", "top"
+            doc_perp_vector = -doc_x
+        if orientation == "right":
+            doc_perp_vector = doc_x
+
+        perp_len = np.linalg.norm(doc_perp_vector)
+        doc_perp_vector /= perp_len
+        perp_len += border_width
+        perp_len *= _TEXT_PADDING_FACTOR
+        doc_perp_vector *= perp_len
```
bollu: normalize the vector, add the border width (we're in pixel coordinates, so this is legal), scale up by the original length, add a padding factor `(_TEXT_PADDING_FACTOR = 1.05)`

vispy/visuals/colorbar.py
```diff
@@ -436,75 +314,166 @@ def _update(self):
...
+                                                dtype=np.float32))
+        doc_label_pos = doc_center + doc_perp_vector
+        visual_label_pos = doc_to_visual.map(doc_label_pos)[:3]
+
+        # next, calculate tick positions
+        if orientation in ["top", "bottom"]:
+            # ticks_axis = visual_to_doc.map(np.array([halfw, 0, 0, 0], dtype=np.float32))
+            #if ticks_axis[0] < 0:
+            #    ticks_axis *= -1
+            doc_ticks_pos = [doc_label_pos - doc_x,
```
bollu: Do the same thing for the ticks. We know the canonical positions in document coordinates, so two vector additions give us the tick positions

Eric89GXL:
The `Colorbar`s look good to me, except that I think even for the vertical ones, we want the text on the "axis" labels to be left-to-right (not up-to-down or down-to-up). The "label" for the Colorbar, on the other hand, can be vertical, as it is. So you'll probably have to split those into two separate subvisuals to accomplish it.

bollu:
I changed the direction of the ticks as @Eric89GXL suggested, so the ticks _always_ read left-to-right.

I'm undecided about the position. Maybe even when the orientation is `left`, you would want the ticks on the right (with label on the left)? Although I imagine the more common use-case will be when the labels and the ticks are to the same side of the `ColorBar`.

#### As a visual: 
ticks_orientation_visuals

#### In `vispy.plot`:

colorbar_ticks_plotting

rossant:
@bollu someone has switched to windows 10 ;)

Eric89GXL:
Those LGTM, let's stick with those orientations for now

bollu:
@rossant I was on the "Windows 10 Insider" program, so I've been beta testing it :) 

Unfortunately, the NVIDIA drivers weren't working properly for a while, but they just pushed an update for the new batman game :P All is well with the world.

Update: I've added a `ColorBarWidget`. Right now, I'm just naively positioning it to be at the center of whatever rect you stick in into, but I would like to position it the way `AxisWidget` does based on orientation. @Eric89GXL is that overkill for a first pass?

bollu:
This branch developed merge conflicts? _weird_.

Eric89GXL:
I just merged a branch from @campagnola, that's where the conflicts would come from.

I think as long as the `Widget` makes itself the same size as the plotting area in the "long" direction of the colorbar (based on its orientation), and requests to have itself be short in the "short" direction, it should look okay. Make sense?

bollu:
I'm not sure what you mean by "makes itself the same size as the plotting area". 

Do you mean that the `Widget` that contains the `ColorBar` should be as "long" as the `ColorBar`, and should have a width that's less than the "length" in the other direction?

campagnola:
@Eric89GXL, the `Grid` will be responsible for stretching the color bar and plot area to the same height. To reduce the width of the color bar, we would give it a small stretch value for now (until #277).

bollu:
@Eric89GXL I merged `scenegraph-update` and resolved the merge conflicts. Mergeable again.

bollu:
Also, I'll probably want to do a `rebase -i` for a cleaner commit history. Squashing a few commits and rewording the others.

Eric89GXL:
Yeah, sorry I meant to get `Grid` to do that resizing stuff.

Eric89GXL:
@bollu please close this and open up a new one against `master`, since the big PR has been merged.

PR #1023, opened 2015-07-27
Arrow visual and Bezier lines

sh4wn:
This pull requests contains an extension to the already existing LineVisual: an ArrowVisual. This is a visual with an arrow head which is oriented automatically in the right way.

Supersedes these PRs:

- https://github.com/campagnola/vispy/pull/22
- https://github.com/vispy/vispy/pull/924


Eric89GXL:
Thanks. We'll see what Travis thinks :)

FYI with this many commits you might end up seeing one Travis run fail due to size increases. That one we can selectively ignore in this case.

@rougier you have time to take a look at the code?

Eric89GXL:
@sh4wn you need to update the `test_data_tag = 'test-data-4'` here:

https://github.com/vispy/vispy/blob/master/vispy/testing/image_tester.py#L364

Then hopefully your tests will pass.

sh4wn:
Somehow a test fails to pass here: https://travis-ci.org/vispy/vispy/jobs/73004655

Eric89GXL:
That's a weird one. Can you try setting up a 2.6 virtual environment to see if you can reproduce it? It's possible our `raises` decorator isn't working properly. You could try using `assert_raises` to see if it fixes it.

At the very least that exception could print out the `exc` itself like:
```
        else:
            raise AssertionError("Expected %s, got %s instead (%s)" %
                                (self.exc.__name__, type(exc).__name__), exc)
```
You also have some `flake` errors

sh4wn:
Well, with Python 2.7 it works, don't known yet about 2.6 (Arch Linux doesn't package this version anymore).

I don't see the flake erros by the way?

Eric89GXL:
Maybe your version of `PEP8` is old, see here for errors:

https://travis-ci.org/vispy/vispy/jobs/73166359#L1810

Can you rebase against the latest version of `scenegraph-update`? You then need to manually add `Arrow` to `vispy/scene/visuals.py`, we removed the auto-generation.

sh4wn:
Oh damn, saw your comment too late, I used pull/merge. Rebasing always gives me a lot of merge conflicts and when I fix them and after I use git add it says "Nothing changed, have you used git add?".

Eric89GXL:
Rebasing leads to a cleaner history, so in general it's what I prefer. It also makes things like squashing commits and removing files from history a bit easier if you ever have to do it. Whatever conflicts you get during rebase you should also hit during a merge, so I think it must just be a problem with how you're doing it. It might be worth reading up on a tutorial and learning how to do it in case you end up working with a package that is more strict about merges in the future. But we're okay here :)

Hopefully Travis is happy with this one.

vispy/scene/visuals.py
```diff
@@ -228,6 +228,7 @@ def generate_docstring(subclass, clsname):
 Isoline = create_visual_node(visuals.IsolineVisual)
 Isosurface = create_visual_node(visuals.IsosurfaceVisual)
 Line = create_visual_node(visuals.LineVisual)
+Arrow = create_visual_node(visuals.ArrowVisual)
```
Eric89GXL: Was this file alphabetical until this change?

Eric89GXL: Looks like it almost was -- can you alphabetize at least the simple visuals? `Compound` we might actually want at the top of the entire list since it's a bit different from the others.

sh4wn:
Cool, the tests pass now :)

Eric89GXL:
LGTM, +1 for merge! The only Travis error is related to size, which is fine.

@campagnola I'm not sure what to do with this. It would be nice to have it merged into `scenegraph-update` so that future PRs by @sh4wn will have a nice diff, but I don't want to make merging the big PR more complex than it already is. I'm thinking if `scenegraph-update` won't be ready to merge for a couple of days, it's probably worth merging this one in now. WDYT?

Eric89GXL:
@sh4wn can you close this PR and make it into `master` now instead? Then I can go ahead and merge, assuming Travis is happy.

sh4wn:
Coming up!

PR #1024, opened 2015-07-27
Create SceneGraph Visuals statically rather than dynamically

bollu:
This helps improve auto-completion in IDEs that access static information from the module.
Both Sublime Text and PyCharm benefit from this change. I believe Valloric's `YouCompleteMe` will also benefit from this, as will emacs' `Jedi`, though I haven't tested either (on Windows right now)

One thing is that the docs don't show up (since the docs are dynamically appended to the elements in `create_visual_node`. However, `help(vispy.scene.FooVisual)` works as expected. I was wondering if I should add a one-liner above each declaration.

I also cleaned up the whitespace in the file.


vispy/scene/visuals.py
```diff
@@ -198,25 +198,53 @@ def generate_docstring(subclass, clsname):
...
+# and provide auto-completion. One problem is the fact that
+# Docstrings are _not_ looked up correctly by IDEs, since they
+# are attached programatically in the create_visual_node call.
+# However, help(vispy.scene.FooVisual) still works
+
+Axis = create_visual_node(visuals.AxisVisual)
+Box = create_visual_node(visuals.BoxVisual)
+Cube = create_visual_node(visuals.CubeVisual)
+Ellipse = create_visual_node(visuals.EllipseVisual)
+Gridlines = create_visual_node(visuals.GridLinesVisual)
```
QuLogic: Not `GridLines`?

bollu:
@QuLogic thanks for the catch

almarklein:
Automation is great, but in this case I think manual is better. +1

vispy/scene/visuals.py
```diff
@@ -23,26 +23,26 @@
...
         self.interactive = False
         self._opacity_filter = ColorFilter()
         self.attach(self._opacity_filter)
-        
+
         self._id = VisualNode._next_id
         VisualNode._visual_ids[self._id] = self
         VisualNode._next_id += 1
-        self._picking_filter = PickingFilter(id_=self._id)
+        self._picking_filter = PickingFilter(id=self._id)
```
Eric89GXL: This needs to be `id_`

almarklein: Frankly, I'd rather that it would be just `id`, I know its to prevent hiding the buildin `id` function, but this trailing underscore is ugly IMO.

Eric89GXL: AFAIK it's a standard way to avoid overridding builtins. It's actually pretty useful not to override them e.g. for pylint and introspection purposes (on top of general namespace considerations), so in my mind the functional component is more important than the aesthetic.

bollu: Ah, I'll change the code.

almarklein:
> AFAIK it's a standard way to avoid overridding builtins. It's actually pretty useful not to override them e.g. for pylint and introspection purposes (on top of general namespace considerations), so in my mind the functional component is more important than the aesthetic.

I don't agree in this case, because `id` is quite a common variable name while the use of the `id()` function is rare. But I wont fight it this time :)

Eric89GXL:
True. Regardless of how common each use is, though, you can still get
caught by introspection and pylint problems because of it. So even if
people never use id(), there is still good value in not shadowing it.

Eric89GXL:
Thanks @bollu!

almarklein:
>  you can still get caught by introspection and pylint problems because of it.

This is an example where I think Pylint should not get in the way of how want to write my code :)

Eric89GXL:
Sure, depending on how many silent/hidden errors you want in your code :) I have had problems before not seeing errors in code because of shadowed variables, so there are cases where it is harmful.

campagnola:
Thanks @bollu!

Issue #1025, opened 2015-07-28
scenegraph-update branch: oscilloscope demo fails


michaelaye:
$ python oscilloscope.py
Traceback (most recent call last):
  File "oscilloscope.py", line 198, in <module>
    fft_samples = mic.chunksize \* n_fft_frames
AttributeError: 'MicrophoneRecorder' object has no attribute 'chunksize'
(py27)

Just realized: Why is there no possibility for me to address this issue to the scenegraph-update branch only? Wouldn't branch-connected issues make more sense?

Eric89GXL:
I don't think Github had that feature unfortunately.

I also have that error, but I assumed is due to a problem with that
package. But if you can find a workaround/more general solution that would
be great.

jvtrudel:
Maybe  you could use a label `scenegraph-update branch` or `branch specific`? 
But would it be useful?

QuLogic:
I think it's because you don't have `pyaudio` installed. There's a fake `MicrophoneRecorder` object if it can't be imported, but it doesn't seem to implement everything the "real" one does.

QuLogic:
Please try #1026.

Eric89GXL:
@jvtrudel yeah we could come up with those labels, but I agree the usefulness might be limited as hopefully the `scenegraph-update` branch will be merged very soon.

kmuehlbauer:
Closing. Resolved by #1026.

PR #1026, opened 2015-07-28
Fix oscilloscope demo without pyaudio installed

QuLogic:
See #1025.

I'm still seeing a `RuntimeError: OpenGL got errors (glTexImage2D): GL_INVALID_VALUE` though.

### DIFF
--- a/examples/demo/scene/oscilloscope.py
+++ b/examples/demo/scene/oscilloscope.py
@@ -157,7 +157,7 @@ class Oscilloscope(scene.ScrollingLines):
 rolling_tex = """
 float rolling_texture(vec2 pos) {
     if( pos.x < 0 || pos.x > 1 || pos.y < 0 || pos.y > 1 ) {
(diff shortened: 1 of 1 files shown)


QuLogic:
Ah, `GL_MAX_TEXTURE_SIZE=8192`, but the `ScrollingImage` is 10000. With 8000 it runs, but the top image is empty...

QuLogic:
Oh, the top image works if you input _real_ data. Maybe the non-`pyaudio` sample data should be tweaked a little so it does show up there.

Eric89GXL:
So does this work for you now with and without using `pyaudio`?

QuLogic:
Yea, now the non-`pyaudio` looks like this (I stuck in a sinusoid in there for something interesting):
screenshot from 2015-07-28 18-57-20

Eric89GXL:
Works great, thanks @QuLogic

PR #1030, opened 2015-07-30
Scenegraph update - review

campagnola:
WIP: addressing reviewer comments

### DIFF
--- a/examples/basics/scene/nested_viewbox.py
+++ b/examples/basics/scene/nested_viewbox.py
@@ -87,7 +87,7 @@ vb1.camera.rect = (0, 0, 1, 1)
 vb1.camera.interactive = False
 
 # bottom-left (+y down)
-vb11 = scene.widgets.ViewBox(parent=vb1.scene, name='vb11', 
+vb11 = scene.widgets.ViewBox(parent=vb1.scene, name='vb11', border_width=2e-3, 
                              margin=0.02, border_color='green')
 vb11.pos = 0, 0
 vb11.size = 1, 0.5
@@ -97,7 +97,7 @@ line11 = scene.visuals.Line(pos=pos, color=color, method='gl',
                             parent=vb11.scene)
 
 # top-left (+y up)
-vb12 = scene.widgets.ViewBox(parent=vb1.scene, name='vb12', 
+vb12 = scene.widgets.ViewBox(parent=vb1.scene, name='vb12', border_width=2e-3, 
                              margin=0.02, border_color='blue')
 vb12.pos = 0, 0.5
 vb12.size = 1, 0.5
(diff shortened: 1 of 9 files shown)


Eric89GXL:
Let me know whenever it's good for review

campagnola:
Review-ready, although there are probably some bugs for travis to find..

examples/basics/scene/nested_viewbox.py
```diff
@@ -87,7 +87,7 @@
 vb1.camera.interactive = False
 
 # bottom-left (+y down)
-vb11 = scene.widgets.ViewBox(parent=vb1.scene, name='vb11', 
+vb11 = scene.widgets.ViewBox(parent=vb1.scene, name='vb11', border_width=2e-3,
```
Eric89GXL: you can have a fractional border_width? how come you need it?

Eric89GXL:
Other than my `border_width` confusion, LGTM. We'll hope Travis is happy, too.

Eric89GXL:
FYI I killed the intermediate Travis builds so the last one can run sooner

examples/basics/scene/nested_viewbox.py
campagnola: In this case it is because a viewbox is embedded inside another. One of the uses we envisioned for vispy was to design publication-ready figures using physical units. For example, you might want a plot widget that is 4 cm wide with 1 mm border lines, and these sizes do not necessarily correspond to an integral number of pixels.

Eric89GXL: Ahh, I see. Thus why the `int(...)` was also removed below. Makes sense.

campagnola:
Huh.. I though Travis did that automatically.

campagnola:
Arr.. some test file is opening a window on import.

Eric89GXL:
It will run the commits in the order it receives them, generally only using 5 or so jobs at once. We probably would have had to wait an hour or two for it to get through those intermediate commits before reaching your last one.

Did you push each one as you made it, or all of them at once? It used to be that, if you pushed a bunch of commits all at once, it would only run the last commit. Maybe they changed that behavior, though.

campagnola:
I was pushing them one at a time as I left comments on the main PR.

Eric89GXL:
Ahh, yeah, that would be why. It will save Travis some time if you group commits together, or go cancel the builds as they get started. But it's probably pretty rare you'll push so many commits in such a short span of time.

campagnola:
w00.

Eric89GXL:
Good to go from you, or did you want to deal with the window-opening thing?

campagnola:
False alarm; this is ready.

Eric89GXL:
Is the big PR ready for merge from your end now? I think it is. So if @almarklein agrees with the edits, then we can go ahead and merge.

campagnola:
Yes. If Almar is happy, then we can finally kill the awful beast.

PR #1034, opened 2015-07-31
Bezier curves and arrows

sh4wn:
Add new vispy.geometry.curves module and an ArrowVisual.


Eric89GXL:
Looks like you need another merge

sh4wn:
Pull from master?

Eric89GXL:
Yeah you have a conflict with `master`. I would normally say you should `rebase` but you have merges in there, so I think you have to `merge` again.

sh4wn:
It's able to merge now!

campagnola:
No problem with rebasing after a merge. The only rule is: never rebase a branch that somebody else is working on.

Eric89GXL:
I usually get weird problems rebasing after merges because there are conflicts. I guess you could just skip those merge commits?

campagnola:
I wouldn't skip commits.

Eric89GXL:
How do you deal with the `merge` commits in when rebasing a branch that has merges in it? That's what I don't get... for example, in this branch there are multiple merges with a branch that is now in `upstream/master`, so it seems like you'd have to skip those merges during a rebase against `upstream/master`. What else would you do?

campagnola:
If the merge really does not introduce any code changes, then it can be safely skipped (and in this case, I think git will give a warning and suggest skipping the commit).

That's not always the case, though--merge commits often contain extra changes that were not in either parent branch (especially if there were conflicts during the merge), so by skipping them you run the risk of missing these changes. The subsequent commits might also run into trouble because their changes might be applied against a different code base.

I haven't really had trouble with this, but in general my rule when picking between rebase and merge is to do whatever is easier :)

vispy/visuals/line/arrow.py
```diff
@@ -0,0 +1,275 @@
...
+    'stealth',
+    'curved',
+    'angle_30',
+    'angle_60',
+    'angle_90',
+    'triangle_30',
+    'triangle_60',
+    'triangle_90',
+    'inhibitor_round'
+]
```
Eric89GXL: Let's make this a `tuple` so it's immutable

vispy/visuals/line/arrow.py
```diff
@@ -0,0 +1,275 @@
...
+              with proper joins and endcaps.
+            * "gl" uses OpenGL's built-in line rendering. This is much faster,
+              but produces much lower-quality results and is not guaranteed to
+              obey the requested line width or join/endcap styles.
+
+    antialias : bool
+        Enables or disables antialiasing.
+        For method='gl', this specifies whether to use GL's line smoothing,
+        which may be unavailable or inconsistent on some platforms.
+
```
Eric89GXL: don't need this extra newline, same with a few others below

vispy/visuals/line/arrow.py
```diff
@@ -0,0 +1,275 @@
...
+        * stealth
+        * curved
+        * triangle_30
+        * triangle_60
+        * triangle_90
+        * angle_30
+        * angle_60
+        * angle_90
+        * inhibitor_round
```
Eric89GXL: should indent these four more spaces

Eric89GXL:
@campagnola there seem to be consistent errors in Travis for setting `arrow.parent = None`, any idea what is causing it? The `ArrowVisual` isn't doing anything fancy, so I think it's a deeper problem with the `SceneCanvas`:

https://travis-ci.org/vispy/vispy/jobs/73569457#L1851

vispy/visuals/tests/test_arrows.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+vertices = np.array([
+    [25, 25],
+    [25, 75],
+    [50, 25],
+    [50, 75],
+    [75, 25],
+    [75, 75]
+]).astype('f32')
```
Eric89GXL: Cleaner to do `], dtype=np.float32)` here instead of `]).astype('f32')`. It also avoids a warning for old `numpy`.

campagnola:
> @campagnola there seem to be consistent errors in Travis for setting arrow.parent = None

No idea about this one. It looks like it's happening on `Canvas.__exit__()`. Might still be my fault, in any case..  Perhaps the multiple calls to `c.render()` are causing the framebuffer stack to be cleared incorrectly?

Eric89GXL:
I tried to replicate with a Python2.6 venv on my system and couldn't, so this one might be hard to track down.

@sh4wn for now you can try just creating a new `Canvas` every time inside your loop, instead putting the loop inside the Canvas creation. If that works, please open up an issue with the trimmed stack trace from `Travis`, and a link to the `Travis` log.

sh4wn:
Ok I fixed those latest remarks :) Lets see what Travis thinks.

Eric89GXL:
Wow, that didn't fix it. I guess you can put it back to the more efficient single-creation way, but add this check:

```
if os.getenv('TRAVIS', 'false') == 'true' and sys.version[:3] == '2.6':
    raise SkipTest('Travis fails due to FB stack problem')  # XXX fix this
```

Eric89GXL:
(and please open an issue about it)

sh4wn:
Issue opened: #1042.

Eric89GXL:
LGTM! @sh4wn ready for merge from your end?

sh4wn:
Yes, I'm good to go!

Eric89GXL:
Great, thanks for working through this! It'll be useful to have this in. Now onto the graphs :)

sh4wn:
Awesome!

Issue #1032, opened 2015-07-31
Need 3D sphere/ellipsoid primitive/visual


michaelaye:
So, there are `scene.visuals.Cube` and `scene.visuals.Box`, and IIUC, spheres are hard to do in OpenGL b/c everything is based on triangles, but I'm really in urgent need of a 3D Sphere/Ellipsoid primitive. The only thing I found `scene.visuals.Ellipse` only works in 2D apparently (according to docstring). 
I tried to (ab-) use the `scene.visual.Markers` for that but they don't seem to scale correctly/endlessly with the size parameter? At some point (`size>500` or so) the marker just did not get larger (I compared it visually to an embedded x-axis, so that was my reference.)
Is there a quick work-around I could use for a sphere/ellipsoid in 3D?

campagnola:
You can use `Mesh` and `vispy.geometry.create_sphere()` together. And you might as well just implement it as a subclass of `Mesh`. And while you're at it, you might as well make it into a PR ;)

campagnola:
(correction: a subclass of `vispy.visuals.MeshVisual`)

michaelaye:
ok, will start something and then do a WIP PR if I need any help, okay?

michaelaye:
will it work from current master, or does it require anything from scenegraph-update?

campagnola:
`master` should be fine.  I expect the behemoth PR will be merged in the next few days anyway.

michaelaye:
Darn, it's `resolves` not `solves` for auto-closing. Closing! ;)

Eric89GXL:
I usually just put a one-liner in my PR descriptions saying "Closes #XXX."

michaelaye:
Yes, for some reason I chose to write `solves`, confusing it with `resolves` that also auto-closes the referenced issue.

PR #1035, opened 2015-07-31
[WIP] first commit of Sphere visual. Solves #1032

michaelaye:
What do you guys think?

Some issues I have

* create_sphere does not calculate the outline vertices so the `if edge_color` case is not covered
* I put the default of rows and columns to 30 each to have a smooth-ish appearance. Is that an okay choice or are there more clever ways to do this?
* Question: I added properties for the vertices and faces. I don't think it should harm performance, but maybe you know better?
* Because of the lack of surface features one does not see anything happening when rotating the mouse. Makes me wonder if we should have the default sphere something that avoids that?


michaelaye:
Oh, and a more general GH question: I always struggle to know where to put the `closes #issue` GH magic. Do I already put it in the commit on my own branch? Or do I put it in the pull request? If I would not put it in my original commit, would I have to do a zero-commit later on, just to create that auto-close magic? What's the best strategy here?

michaelaye:
Aand, I already saw some docstring bugs, will update.

michaelaye:
Sorry, not sure what happened here. :( I tried my first rebase, having seen that the scenegraph PR has been merged, and I thought that the idea of rebase is that it makes things cleaner, which doesn't look like it to me here.

michaelaye:
Ok, now I got it, I think. (Had to learn about squashing). ;)

Eric89GXL:
Glad you figured it out, `git` has a steep learning curve for some things :)

You'll need to add your visual also to `vispy/scene/visuals.py`

Eric89GXL:
see the bottom of the file, hopefully it makes sense

vispy/visuals/sphere.py
```diff
@@ -0,0 +1,63 @@
...
+    vertex_colors : ndarray
+        Same as for `MeshVisual` class. See `create_sphere` for vertex ordering.
+    face_colors : ndarray
+        Same as for `MeshVisual` class. See `create_sphere` for vertex ordering.
+    color : Color
+        The `Color` to use when drawing the sphere faces.
+    edge_color : tuple or Color
+        The `Color` to use when drawing the sphere edges. If `None`, then no
+        sphere edges are drawn.
+    """
```
Eric89GXL: don't forget to update docstring with the proper entries for `cols` and `rows`

vispy/visuals/sphere.py
```diff
@@ -0,0 +1,63 @@
...
+    def draw(self, transforms):
+        """Draw the visual
+
+        Parameters
+        ----------
+        transforms : instance of TransformSystem
+            The transforms to use.
+        """
+        MeshVisual.draw(self, transforms)
+        if self._outline:
```
Eric89GXL: if `self._outline is not None` is safer and more explicit

vispy/visuals/sphere.py
```diff
@@ -0,0 +1,63 @@
...
+            set_state(polygon_offset=(1, 1), polygon_offset_fill=True)
+            self._outline.draw(transforms)
```
Eric89GXL: You'll probably want to make this a `CompoundVisual` with two children, the main mesh, and the outline mesh

michaelaye: I just copied the CubeVisual. Does that apply to it as well then?

michaelaye: And as I said, I don't have the input for outline anyway, as create_sphere does not provide it.

Eric89GXL:
Regarding the properties, I think it's fine, but maybe instead we should just give access to the underlying visuals instead somehow. Not sure.

I wouldn't worry too much about the rotating not showing anything. This should be expected for uniform lighting. Maybe add a `Notes` section mentioning this so people aren't surprised.

I can't think of a better default than `30, 30`. It's really going to depend on what people want to do with the spheres.

Can you add a `scene` example so we can see the results? One idea is to make some simplistic visualization of the solar system, with sun and planets, probably with the planet sizes scaled up by some large factor to make up for the fact that they're tiny relative to their distance from the sun.

vispy/visuals/sphere.py
Eric89GXL: It probably should do it that way, yeah. We can refactor that later. Maybe for now don't add the outline? It can always be added later by someone who needs it.

michaelaye: I also was thinking if a default wireframe display would not be nicer than filled faces? Or at least the option for it? It sounds like that could be accomplished with the CompoundVisual? Are there switches where I could then switch off and on their display, depending on what user wants?

michaelaye: Do you have a NotImplementedError maybe instead of removing the outline code? Or better remove it completely?

Eric89GXL: I'd either implement it, or remove it (avoid the NotImplementedError). Can you try making a `CompoundVisual` out of `Cube`? It should be an easier case. Once that works, then it might be easier to do it for the sphere. Maybe we should default to having the outline and filled parts both shown, that way rotating will be visible.

michaelaye: I had the same idea, and tried to combine it with the visuals.Isoline already, but that didn't work out, I only could see the Isolines then. Is there a demo for CompoundVisual, how to use it? I'm scared of it, sounds complex. ;)

Eric89GXL: The simplest one I can think of is `LinePlot`, which just combines `Marker` and `Line`:

https://github.com/vispy/vispy/blob/master/vispy/visuals/line_plot.py

It's hopefully not too scary after looking at that code :)

vispy/visuals/__init__.py
```diff
@@ -31,6 +31,7 @@
 from .regular_polygon import RegularPolygonVisual  # noqa
 from .scrolling_lines import ScrollingLinesVisual  # noqa
 from .spectrogram import SpectrogramVisual  # noqa
+from .sphere import SphereVisual # noqa
```
bollu: there should be two spaces before the `# noqa`. So, the line should look like this:

```python
from .sphere import SphereVisual  # noqa
```

michaelaye:
Since the rebase to the new master, this is failing actually. :(
Here's the traceback:
```python
  File "sphere.py", line 28, in <module>
    app.run()
  File "/Users/klay6683/Dropbox/src/vispy/vispy/app/_default_app.py", line 62, in run
    return default_app.run()
  File "/Users/klay6683/Dropbox/src/vispy/vispy/app/application.py", line 137, in run
  File "/Users/klay6683/Dropbox/src/vispy/vispy/app/backends/_qt.py", line 229, in _vispy_run
    return app.exec_()
  File "/Users/klay6683/Dropbox/src/vispy/vispy/app/backends/_qt.py", line 646, in paintGL
  File "/Users/klay6683/Dropbox/src/vispy/vispy/util/event.py", line 455, in __call__
  File "/Users/klay6683/Dropbox/src/vispy/vispy/util/event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  << caught exception here: >>
  File "/Users/klay6683/Dropbox/src/vispy/vispy/util/event.py", line 471, in _invoke_callback
    cb(event)
  File "/Users/klay6683/Dropbox/src/vispy/vispy/scene/canvas.py", line 206, in on_draw
    self._draw_scene()
  File "/Users/klay6683/Dropbox/src/vispy/vispy/scene/canvas.py", line 252, in _draw_scene
    self.draw_visual(self.scene)
  File "/Users/klay6683/Dropbox/src/vispy/vispy/scene/canvas.py", line 290, in draw_visual
    node.draw()
  File "/Users/klay6683/Dropbox/src/vispy/vispy/scene/visuals.py", line 98, in draw
TypeError: draw() missing 1 required positional argument: 'transforms'
ERROR: Invoking <bound method SceneCanvas.on_draw of <SceneCanvas (PyQt4) at 0x1029887f0>> for DrawEvent
```
Could somebody advise what I need to change?

michaelaye:
Found that some things **DID** change between older master and new scenegraph update. Simplest example `sphere.py` works again with using the new SphereVisual. Sphere is now a CompoundVisual but I don't know what to change to make both the mesh and the filled edges visual? This seems to be the same case for the updated Cube CompoundVisual?

michaelaye:
Found my mistake: had wrong reference to vertices in Sphere visual and not using edge_color in example. Now works!

vispy/visuals/sphere.py
```diff
@@ -0,0 +1,90 @@
...
+    #
+    #     Parameters
+    #     ----------
+    #     transforms : instance of TransformSystem
+    #         The transforms to use.
+    #     """
+    #     MeshVisual.draw(self, transforms)
+    #     if self._outline is not None:
+    #         set_state(polygon_offset=(1, 1), polygon_offset_fill=True)
+    #         self._outline.draw(transforms)
```
Eric89GXL: Go ahead and delete this commented stuff, we shouldn't need it now

Eric89GXL:
LGTM! Glad you were able to figure out the new `Visuals`. @campagnola can you take a look and just verify this is how people should make new visuals?

Eric89GXL:
@michaelaye please fix the Flake errors so that the CIs are happy:

https://travis-ci.org/vispy/vispy/jobs/73663197#L2173

You can do `make flake` or `python make test flake` to test it locally.

vispy/visuals/sphere.py
```diff
@@ -0,0 +1,90 @@
...
+        self.mesh.set_gl_state(polygon_offset_fill=True,
+                               polygon_offset=(1, 1), depth_test=True)
+
+    @property
+    def mesh(self):
+        """The vispy.visuals.MeshVisual that used to fil in.
+        """
+        return self._mesh
+
+    @mesh.setter
```
campagnola: Do we really want to allow setting the mesh / border visuals?
If yes, then you need to remove the old visual and add the new one to the compound.

vispy/visuals/sphere.py
```diff
@@ -0,0 +1,90 @@
...
+        """The vispy.visuals.MeshVisual that used to draw the border.
+        """
+        return self._border
+
+    @border.setter
+    def border(self, border):
+        self._border = border
+
+    @property
+    def vertices(self):
```
campagnola: What are these for?

vispy/visuals/sphere.py
michaelaye: I don't understand what this should be, which old visual? I copied this from the cube visual, it's the same there, and it wasn't before the scenegraph-update PR, so I assumed that is a new way to do things?

vispy/visuals/sphere.py
michaelaye: Just convenience properties, if you mean self.vertices and self.faces. self.mesh. and self.border was also copied from the cube visual, so I thought that's how it should be?
As my properties can't be set, do you think they can be bad?

vispy/visuals/sphere.py
Eric89GXL: It could be wrong there, too. I suspect it is...

The idea is that, if you allow setting `self._mesh`, then you need to remove the existing `self._mesh` from the `CompoundVisual`, and then add the new one, in addition to setting `self._mesh`. Make sense?

Alternatively, just get rid of the `setter` for both `mesh` and `border`, which would be my vote.

vispy/visuals/sphere.py
Eric89GXL: It's good that they can't be set. Do you have a use case in mind for them? In general we try to expose properties for things people have specific uses for. Otherwise it creates unnecessary API and upkeep burden.

michaelaye: Whenever I would need to access to faces or vertices on application level, I would have to do Sphere.mesh.get_vertices() or Sphere.mesh.get_faces() respectively. This is a double-dot access which I thought is frowned upon?

vispy/visuals/sphere.py
michaelaye: > Make sense?
Not yet. I thought line 36 overwrites any derived mesh that might exist in CompoundVisual, so that there's only one mesh? I am confused why there seem to be 2 meshes?

michaelaye: Or, in other words, CompoundVisual's init get's `self` as a handle, so in case it would effect `._mesh` it is still part of the Sphere object, no? Still don't see where are 2 meshes here.

michaelaye: Oh, right. I see it. ;) Well, I just copied this from the cube visual, I thought for some abstract reason it has to be like that?

vispy/visuals/sphere.py
Eric89GXL: In general flat is better than nested, yeah. But here the nesting really
helps make clear whether the vertices come from the mesh or outline. To
disambiguate you'd have to prepend mesh_ to the name anyway, mostly
defeating the purpose.

vispy/visuals/sphere.py
Eric89GXL: I think it's probably wrong there, too

michaelaye:
About the flake errors: When I run `make flake` I get more than 5000 problems all over vispy, see here: https://gist.github.com/michaelaye/ff545054f581a46770b6 so I'm a bit puzzled why travis would not complain about them?

michaelaye:
So, I"m done with fixing the pep8 stuff, have removed the disturbing properties and the commented part. Just let me know what to do about the meshes, I don't get them. I looked at MeshVisual and CompoundVisual and no one seems to set a self.mesh so I don't understand why the `self.mesh.set_gl_state` does not fail, there should not be a `self.mesh` object?

vispy/visuals/sphere.py
Eric89GXL: @michaelaye this `setter` is still here, I thought you meant to remove it?

Eric89GXL:
Read a bit about Python getters and setters. Because we do this:
```
@property
def mesh(self):
    return self._mesh
```
when one does `instance.mesh`, it actually calls that `@property` method under the hood. The `setter` on the other hand gets called if you do `instance.mesh = something`.

vispy/visuals/sphere.py
michaelaye: I only removed the one I added myself. Nobody commented yet definitely if these have to go or not, Luke's commented goes either way, just that the current way it's done is incomplete.

michaelaye:
about the properties, *DOH* sorry, so yes that's obvious. Still don't understand how it _should_ be though.

michaelaye:
So, think I understand now: I would have to remove the subvisual `_mesh` inside the CompoundVisual with `.remove_subvisual` and set it with the new one for the `._mesh` setter and respectively for the `._border` object. I don't know the implications of it, but I can code this if you want to keep the setters. Or just remove all setters for now, what's the vote?

Eric89GXL:
Let's just get rid of the setters -- we can add them later if we need to, but I'd assume YAGNI for now

michaelaye:
After rebasing with current master, I get this Traceback running the simple sphere.py example:

```python
$ python sphere.py
  File "sphere.py", line 16, in <module>
  File "/Users/klay6683/Dropbox/src/vispy/vispy/scene/canvas.py", line 169, in central_widget
    self._central_widget = Widget(size=self.size, parent=self.scene)
  File "/Users/klay6683/Dropbox/src/vispy/vispy/scene/widgets/widget.py", line 49, in __init__
    self._mesh = MeshVisual(color=border_color, mode='triangles')
  File "/Users/klay6683/Dropbox/src/vispy/vispy/visuals/mesh.py", line 210, in __init__
    self.freeze()
AttributeError: 'MeshVisual' object has no attribute 'freeze'
```

Eric89GXL:
Are you sure you updated against the latest `master` and not your local copy? e.g.
```
git fetch upstream
git rebase upstream/master
```

michaelaye:
yep, that's exactly what I executed:
```bash
klay6683 at MacL2938 in ~PYTHONPATH/vispy (solve_1032_sphere_visual●)
$ git rebase upstream/master
First, rewinding head to replay your work on top of it...
Applying: first commit of Sphere visual. Solves #1032
Applying: adding entry in scene/visuals.p for IDEs. Add cols and rows to docstring
Applying: Adapting to changes from new scenegraph update that were visible from CubeVisual. Fixing noqa comment distance. Adding simplest possible sphere example just for easier discussion. Planets to come.
Applying: fixing wrong reference to vertices in Sphere visual and switching on edge_color in example to render the mesh visible on top of filled edges.
Applying: increasing visibility of grid on sphere.
Applying: removing unusued imports
Applying: removing commented stuff and unwanted properties
Applying: removing setters as agreed
```

Eric89GXL:
Just to double-check, you did the `git fetch upstream` before that command, right?

michaelaye:
yep.

Eric89GXL:
Try `make clean` and then do the command again, maybe somehow the `.pyc` files are not being regenerated properly.

If that doesn't work, I'd actually look at the source file at that error line, and compare it to `upstream/master`. Maybe there is some corner case you're hitting, but it really seems unlikely.

This one's pretty weird, I have to admit...

michaelaye:
If Travis is happy, I guess we are good to go? I can squash my commits down to one, if required? Or only remove the last weird one, don't know how I diverged with origin for git, very weird. As one can see there are 0 differences.

Eric89GXL:
Thanks @michaelaye!

michaelaye:
Thanks for your patience and sorry that I was such high maintenance! ;)

PR #1036, opened 2015-07-31
MRG: Minor fixes

Eric89GXL:
A few minor fixes, and allows us to ensure we're staying compatible with old `numpy`.

### DIFF
--- a/vispy/visuals/rectangle.py
+++ b/vispy/visuals/rectangle.py
@@ -75,8 +75,8 @@ class RectangleVisual(PolygonVisual):
 
         num_segments = (radius / hw * 500.).astype(int)
 
-        bias1 = np.full(4, half_width) - radius
-        bias2 = np.full(4, half_height) - radius
+        bias1 = np.ones(4) * half_width - radius
+        bias2 = np.ones(4) * half_height - radius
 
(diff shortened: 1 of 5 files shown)


Eric89GXL:
Ready for review/merge from my end. I re-enabled OpenGL testing on Windows. I think AppVeyor might be mature enough now to handle it better.

almarklein:
LGTM

vispy/visuals/rectangle.py
```diff
@@ -181,7 +181,7 @@ def radius(self, radius):
                                  " (got %s of length %d)" % (type(radius),
                                                              len(radius)))
 
-            if (radius > np.full(4, hw)).all():
+            if (radius > np.ones(4) * hw).all():
```
QuLogic: Why not just `radius > hw`? It should be broadcast to the right shape automatically.

Eric89GXL: Yeah, that's true actually. I wasn't thinking too hard about what was going on. Feel free to open a PR if it bothers you to that level (it's not quite there for me), otherwise we can try to get it on a subsequent one.

QuLogic: A bit too small for a whole PR, I think...

PR #1037, opened 2015-07-31
MRG: Add ability to freeze properties

### LINKED ISSUE #1015: Provide warning when vispy.scene and vispy.visuals is mixed
The correct way to use visuals in the context of the scenegraph is to load the visuals from `vispy.scene.visuals`. However, right now, using a visual from `vispy.visuals` will fail silently with no warning whatsoever. 

A warning / error could be emitted when a raw `Visual` is attached to the scene.

Eric89GXL:
Closes #1015.

@campagnola this does what you were hoping for, I think. Adding the ability to freeze actually showed a few places where we were setting properties that didn't exist, which is nice.

### DIFF
--- a/examples/demo/scene/picking.py
+++ b/examples/demo/scene/picking.py
@@ -24,7 +24,9 @@ t = np.arange(data.shape[1]) * (dt * 1000)
 for i, y in enumerate(data):
     line = plt.plot((t, y), color=colors[i])
     line.interactive = True
+    line.unfreeze()
     line.data_index = i
+    line.freeze()
 
 
 # Build visuals used for cursor
--- a/examples/demo/scene/picking.py
+++ b/examples/demo/scene/picking.py
@@ -24,7 +24,7 @@ t = np.arange(data.shape[1]) * (dt * 1000)
 for i, y in enumerate(data):
     line = plt.plot((t, y), color=colors[i])
     line.interactive = True
-    line.unfreeze()
+    line.unfreeze()  # make it so we can add a new property to the instance
     line.data_index = i
     line.freeze()
 
--- a/vispy/scene/visuals.py
+++ b/vispy/scene/visuals.py
@@ -18,6 +18,7 @@ import weakref
 from .. import visuals
 from .node import Node
 from ..visuals.filters import ColorFilter, PickingFilter
+from ..util import Frozen
 
 
 class VisualNode(Node):
@@ -103,7 +104,9 @@ def create_visual_node(subclass):
 
     # Decide on new class name
     clsname = subclass.__name__
-    assert clsname.endswith('Visual')
+    if not (clsname.endswith('Visual') and issubclass(subclass, Frozen)):
+        raise RuntimeError('Class "%s" must end with Visual, and must '
+                           'subclass Frozen' % clsname)
     clsname = clsname[:-6]
 
     # Generate new docstring based on visual docstring
@@ -121,11 +124,13 @@ def create_visual_node(subclass):
         self._visual_superclass = subclass
 
         subclass.__init__(self, *args, **kwargs)
+        self.unfreeze()
         VisualNode.__init__(self, parent=parent, name=name)
+        self.freeze()
 
     # Create new class
-    cls = type(clsname, (VisualNode, subclass), {'__init__': __init__,
-                                                 '__doc__': doc})
+    cls = type(clsname, (VisualNode, subclass),
+               {'__init__': __init__, '__doc__': doc})
 
     return cls
 
--- a/vispy/scene/visuals.py
+++ b/vispy/scene/visuals.py
@@ -18,7 +18,6 @@ import weakref
 from .. import visuals
 from .node import Node
 from ..visuals.filters import ColorFilter, PickingFilter
-from ..util import Frozen
 
 
 class VisualNode(Node):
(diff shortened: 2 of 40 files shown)


Eric89GXL:
Also, I should mention that this basically affects `SceneCanvas`, our `Visual`s, and our `Widget`s for now.

Ready for review/merge from my end, and the CIs are happy.

examples/demo/scene/picking.py
```diff
@@ -24,7 +24,9 @@
 for i, y in enumerate(data):
     line = plt.plot((t, y), color=colors[i])
     line.interactive = True
+    line.unfreeze()
```
campagnola: An explanation here would be nice.

vispy/util/frozen.py
```diff
@@ -0,0 +1,25 @@
...
+# Class adapted from:
+# http://stackoverflow.com/questions/3603502/
+#        prevent-creating-new-attributes-outside-init
+
+class Frozen(object):
+    __isfrozen = False
+
+    def __setattr__(self, key, value):
+        if self.__isfrozen and not hasattr(self, key):
+            raise AttributeError("%r is not an attribute of %s"
```
campagnola: I would add `"call unfreeze() to allow addition of new attributes"`  or maybe `"see vispy.util.Frozen"`.

vispy/scene/visuals.py
```diff
@@ -103,7 +104,9 @@ def create_visual_node(subclass):
 
     # Decide on new class name
     clsname = subclass.__name__
-    assert clsname.endswith('Visual')
+    if not (clsname.endswith('Visual') and issubclass(subclass, Frozen)):
+        raise RuntimeError('Class "%s" must end with Visual, and must '
```
campagnola: Why not just have VisualNode subclass from Frozen? That covers most of the user-facing API without making any extra demands on the visuals.

vispy/util/frozen.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2015, Vispy Development Team.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+
+
+# Class adapted from:
+# http://stackoverflow.com/questions/3603502/
+#        prevent-creating-new-attributes-outside-init
```
QuLogic: FYI, you do not need the second line for this link to function.

vispy/visuals/isocurve.py
```diff
@@ -96,8 +106,8 @@ def set_data(self, data):
         if _HAS_MPL:
             try:
                 if not (self._X.T.shape == data.shape):
-                    raise
-            except:
+                    raise Exception
+            except Exception:
```
QuLogic: I'm not sure this helps much. :stuck_out_tongue:

vispy/scene/visuals.py
campagnola: Ok, I see BaseVisual subclasses Frozen. Why is this check necessary, then?

vispy/visuals/isocurve.py
Eric89GXL: There is actually a difference, `except:` will even catch `SystemExit` whereas `except Exception:` will not. It might also work for KeyboardInterrupt or whatever that one is called.

vispy/scene/visuals.py
Eric89GXL: It was originally because I had each `Visual` subclass `Frozen` individually, then figured out making `BaseVisual` subclass `Frozen` was simpler. I left this in mostly as a sanity check. It should never really be violated.

vispy/util/frozen.py
Eric89GXL: ahh, good to know

Eric89GXL:
Thanks for the quick reviews, comments addressed. I left the check in for now @campagnola but I can remove it if you think it's overkill.

vispy/visuals/isocurve.py
QuLogic: Ah, good point, but I suppose I was referring more to the use of a very generic `Exception` in the `raise` which sort of ignores that `self._X.T.shape` may or may not raise something (but maybe that's expected.)

vispy/scene/visuals.py
campagnola: How about checking for BaseVisual instead, then? That'd be more sane..

vispy/visuals/isocurve.py
campagnola: I'm staring, but I still can't figure out what this does :)

QuLogic: One might think it could be written as:
```
if not (self._X.T.shape == data.shape):
    self._X, self.Y = np.meshgrid(...)
```
but the bare exception also would catch `AttributeError` on `self._X`, or `<something>.shape`, etc., so is that intended? I don't know, but that's why I don't like bare exception handling...

Eric89GXL: Yeah I *think* that's actually what it's designed to catch, the attribute
not being defined, or as it is now, being None. I'll change the checks to
be explicit as I agree it's much clearer that way.

bollu:
Naive question: I'm assuming VisPy can't use slots because we sometimes inject properties into classes?

Eric89GXL:
I don't know too much about slots, but in that linked StackOverflow post they had some good reasons to avoid using slots for this purpose.

Eric89GXL:
Fixed the two iffy checks @campagnola @sh4wn

bollu:
@Eric89GXL: which SO link are you talking about?

bollu:
> You would want to use `__slots__` if you are going to instantiate a lot (hundreds, thousands) of objects of the same class. `__slots__` only exists as a memory optimization tool.

>  It's __highly discouraged__ to use `__slots__` for constraining attribute creation, and in general you want to avoid it because it breaks pickle, along with some other introspection features of python.


I'm assuming you were referring to this?

campagnola:
That's the one.

campagnola:
@almarklein, @rossant, @rougier, do any of you forsee problems with this?

rougier:
Nothing come to my mind

rossant:
:+1:

PR #1038, opened 2015-08-01
[WIP] Colorbar -  vispy.plot

bollu:
Checklist:

- [x] add a `colorbar` function to `PlotWidget`
- [x] separate out border rendering into `BorderVisual`
- [x] integrate border rendering to `ViewBox`
- [x] fix coordinate systems for `ColorBar`
- [x] fix text overlapping
- [x] add an example of using `plotwidget.colorbar` with `vispy.plot`
- [x] change ticks orientation so they are always left-to-right
- [x] create a new `ColorBarWidget`
- [x] add an example for `ColorBarWidget`
- [x] write docs on `PlotWidget.colorbar(..)`
- [x] write docs on `ColorBarWidget`
- [x] fix overlap in `examples/basics/visuals/colorbar_visual.py`
- [ ] implement `viridis` colormap for VisPy (in separate PR)
- [ ] write test cases for `ColorBarWidget`


bollu:
@Eric89GXL: to continue the discussion we were having, the `ColorBarWidget` should change its widget size, and position itself in the `Grid` relative to the plot?

Eric89GXL:
Yeah, it should place itself the same way the axes do, namely to put themselves in the Grid and let `Grid` deal with the spacing.

bollu:
As far as I can see, the `AxisWidget` does nothing special to place itself. It's the `PlotWidget` that goes through placing the `AxisWidget` and applying the `stretch`. I guess I'll follow the same model for `plotwidget.colorbar(...)`, by creating a `ColorBarWidget` and placing it correctly.

However, I think this will change once we get the new grid system in place, so the components can request placement themselves?

campagnola:
What do you mean by "request placement"?
The widgets will not be able to suggest a position relative to other widgets, but they will be able to suggest size constraints such as a maximum or fixed width.

bollu:
Yep, that was what I meant - suggest a position relative to one another.

Ah, so in that case, the `PlotWidget` will have to handle placement? The new system will be purely for solving size constraints and not positioning constraints? (I realize that the two are intertwined, but I mean that the `Widget` will only have control on size, and not position)

bollu:
Here's what it looks like now with the correct placement implemented in `PlotWidget`

example taken from `examples/basics/plotting/colorbar.py`:

screenshot from 2015-08-02 01-37-21

I should put some data in there with the right colors plus the colorbar, but this was just for a rough idea :)

So, @Eric89GXL - what's next when it comes to plotting + colorbar? I was thinking of the API. and I think that it is quite simple right now. I could have the `clim` be auto-detected form the data given (I think this is possible for "common" data types).

examples/basics/plotting/colorbar.py
```diff
@@ -0,0 +1,30 @@
...
+fig = vp.Fig(size=(800, 400), show=False)
+plot = fig[0, 0]
+
+dimensions = [(200, 10), (10, 200), (10, 200)]
+# note: "bottom" could also be used, but this would
+# conflict with top.
+orientations = ["top", "left", "right"]
+
+
+for i in range(0, len(dimensions)):
```
Eric89GXL: for `dim, ori in zip(dimensions, orientations)` should be cleaner

examples/basics/plotting/colorbar.py
```diff
@@ -0,0 +1,30 @@
...
+    cbar = plot.colorbar(halfdim=dimensions[i],
```
Eric89GXL: I think for the `colorbar` here, the `halfdim` parameter should be hidden from the user, at least at first. It should auto-position and auto-size based on the window size (or really, based on the area allocated to it by `grid`), just like the Axes do

Eric89GXL: `pos` should also be hidden, at least at first

Eric89GXL:
I agree with @QuLogic that the default border thickness should be at most 3 px, maybe 1 or 2.

Why is the "top" one at the bottom of the plot? I'd expect the "bottom" one to be there, and to have a "top" one at the top?

I think it would be great to have the colorbar automatically determine the limits of the data, if there are such data in the plot. It might be tough, though, since the range will be determined by the Visuals in the plot, and there could be several, all with different ranges. So I'm not sure what the best solution is there. For now you can just make people manually specify the cmap and clim to use for the colorbars, we can always add an `'auto'` option later (start simple).

Some other points:

1. I mentioned this inline in the code, but right now the colorbars are some fixed height. This means that if I make the window shorter or taller, the colorbar stays the same size, sometimes getting clipped. I think they should act more like axes, auto-stretching to be the same height as the plot. Eventually we can make it so that users can specify the `pos` and `halfdim`, but we

2. In your example, you should add an image in the middle so the colorbars have some meaning.

Eric89GXL:
(continuing 1 above) ...we'll have to think about the API for that, since there are multiple different ways one might want to specify the size and position. So maybe auto-position and auto-stretch for now is good enough for most people.

bollu:
You're right about the API, and sure - I'll keep `clim` manually assignable for now.

I could have made the `top` orientation be at the top, but I was of the opinion that it would look awkward - so, top and bottom should replace one another (I didn't write code for replacement yet). Thoughts on that?

As for re-sizing, I'll implement that. It seems much nicer that way rather than have the user manually mess about with the dimensions for most use-cases with plotting.

vispy/scene/widgets/colorbar.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        self._update_colorbar()
+
+    def _update_colorbar(self):
+        self._colorbar.pos = self.rect.center
+        self._colorbar.halfdim = ColorBarWidget._calc_halfdim(self.rect,
+                                                              self._colorbar.orientation)  # noqa
+
+    @staticmethod
+    def _calc_halfdim(rect, orientation):
```
bollu: the dimensions of the colorbar in `ColorBarWidget` is auto-calculated based on the dimensions of the rect.

vispy/plot/plotwidget.py
```diff
@@ -338,17 +341,99 @@ def volume(self, vol, clim=None, method='mip', threshold=None,
...
-            
+
         """
         self._configure_3d()
         surf = scene.SurfacePlot(z=zdata, **kwargs)
         self.view.add(surf)
         self.view.camera.set_range()
         return surf
+
+    def colorbar(self, cmap, orientation="",
```
bollu: `pos`, `halfdim` can't be edited anymore from `plotwidget.colorbar(...)`

vispy/plot/plotwidget.py
```diff
@@ -338,17 +341,99 @@ def volume(self, vol, clim=None, method='mip', threshold=None,
...
+        """Show a ColorBar
+
+        Parameters
+        ----------
+        cmap : str | vispy.color.ColorMap
+            Either the name of the ColorMap to be used from the standard
+            set of names (refer to `vispy.color.get_colormap`),
+            or a custom ColorMap object.
+            The ColorMap is used to apply a gradient on the colorbar.
+        orientation : {'left', 'right', 'top', 'bottom'}
```
Eric89GXL: I think the `Widget` instead of having `orientation` as the user-facing parameter, should probably have `location` instead. I think they are pretty much the same thing, but users in `plot` probably want to think about where their colorbar will be placed, whereas at the `scene` level they are probably both orienting it and placing it somewhere themselves.

bollu:
set the default position to `right`

Eric89GXL:
I wouldn't worry about tests for now, since we don't have many tests set up for widgets yet. Let me know when the overlap issue is fixed, and hopefully we can merge.

bollu:
edited colorbar visual example, with no overlap:

screen shot 2015-08-04 at 8 58 31 pm

bollu:
correct colorbars on all sides - 

screen shot 2015-08-04 at 11 33 48 pm

The `top` and `bottom` look thicker than `left` and `right` because of the dimensions of the window (width > height). IMO this is okay, since I doubt you would have multiple colorbars for the same plot.

I personally like the largish gaps between the colorbar and the plot. I'm not sure about universal appeal :)

Eric89GXL:
Those gaps will end up being the responsibility of the grid system, so I wouldn't worry too much about it. Put a nice function in the middle there and we'll probably be good to go! The visual example looks cooler with the new colormap you chose, I think.

bollu:
added an image to `examples/basics/plotting/colorbar.py`:
screen shot 2015-08-05 at 12 39 26 am

I plotted the coolest function I could think of in short notice, which had a broad spectrum of colors that wasn't too bunched up (`sinc` tends to bunch up at x=0, so it doesn't provide a "color range")

Eric89GXL:
Looks good, but could you add a `set_range()` so it zooms in properly?

bollu:
screen shot 2015-08-05 at 12 39 26 am

done :)

@Eric89GXL Something happened to the Scrutinizer build - could you re-run it please?

Eric89GXL:
Oh, I guess that is properly zoomed, given the margin and shape of the image. Can you set the window size so the plotting area is more square?

bollu:
Whoops, wrong image, sorry. Correct one:
screen shot 2015-08-05 at 1 11 02 am

Eric89GXL:
Ahh, cool. Build restarted

bollu:
@Eric89GXL CI is happy!

examples/basics/visuals/colorbar_visual.py
```diff
@@ -81,7 +83,7 @@ def get_mandlebrot_escape_values(width, height):
     return v_get_num_escape_turns(*grid).astype(np.float)
 
 
-def get_vertical_bar():
+def get_vertical_bar(pos, halfdim):
```
Eric89GXL: I think we might want to use `size` instead of `halfdim`, it should be more intuitive for users. The first element is the typically short axis, the second is the typically long axis. For a vertical colorbar, this corresponds to `(w, h)`. Make sense?

bollu: hm, yeah, that does. Will change.

Eric89GXL:
Other than the renaming of the argument (and subsequent `/2.`'s that it will require), LGTM. Let me know when that's done, and I'll take a final look.

examples/basics/visuals/colorbar_visual.py
bollu: So, for a horizontal colorbar, it would be `(h, w)`? because the height is the "minor" axis

bollu: that doesn't particularly make sense to me. How would you auto-detect orientation in that case? `(minor_length, major_length)` doesn't tell you _which_ axis is minor and which is major

Eric89GXL: Yeah. I like the "major" / "minor" terminology, I'd use that in the `docstring` instead of any mention of width/height.

Eric89GXL: I think `halfdim=(x, y)` is not as clear as `size=(major, minor)`, where `major` is the length along the major (typically longer) axis of the colorbar and `minor` is the length along the minor (typically shorter) axis. If we use `size=(major, minor)` everywhere, it's clear what the colorbar should look like.

bollu:
@Eric89GXL I'm going to bed now. The newest build should succeed. I'll review the code again tomorrow. If you see anything that needs to be changed, please do comment :)

Eric89GXL:
Changes LGTM! @campagnola do you have time to have a look?

bollu:
@Eric89GXL , @campagnola: status on these? :)

Eric89GXL:
@campagnola do you have time to have a look?

campagnola:
The screenshot looks awesome!
Reviewing..

vispy/scene/widgets/colorbar.py
```diff
@@ -0,0 +1,144 @@
...
+        self._colorbar.size = \
+            ColorBarWidget.calc_size(self.rect, self._colorbar.orientation)
+
+    @staticmethod
+    def calc_size(rect, orientation):
+        (total_halfx, total_halfy) = rect.center
+
+        # the padding that the colorbar should leave with respect
+        # to the total halfdim
+        MAJOR_AXIS_PADDING = 0.1
```
campagnola: all-caps implies a global variable.

vispy/scene/widgets/colorbar.py
```diff
@@ -0,0 +1,144 @@
...
+        MINOR_AXIS_PADDING = 0.8
+
+        # ratio of minor axis to major axis
+        MINOR_AXIS_RATIO = 0.05
```
campagnola: The aspect ratio constraint doesn't make much sense to me--the width of the colorbar should probably be limited to a narrow range of pixel widths (something like 10-40 px) and should only depend on the amount of horizontal space that is available. Right now if you make the window wide and short, then the colorbar reacts by becoming very thin; the opposite of what I'd expect.

Perhaps we can rethink this after the grid layouts are more fully-featured, though..

vispy/scene/widgets/widget.py
```diff
@@ -5,6 +5,7 @@
 from __future__ import division
 
 import numpy as np
+from vispy.visuals import _BorderVisual
```
campagnola: If `_BorderVisual` is meant to be private, then there is no need for it to be imported to the `vispy.visuals` namespace. However if there's any possibility that it will be useful elsewhere, then it should not be private.

vispy/scene/widgets/widget.py
```diff
@@ -44,9 +45,15 @@ class Widget(Compound):
     def __init__(self, pos=(0, 0), size=(10, 10), border_color=None,
                  border_width=1, bgcolor=None, padding=0, margin=0, **kwargs):
         # For drawing border.
-        # A mesh is required because GL lines cannot be drawn with predictable
-        # shape across all platforms.
-        self._mesh = MeshVisual(color=border_color, mode='triangles')
+        self._border = _BorderVisual(pos=pos,
```
campagnola: By splitting this visual in two, we double the cost of rendering the border/background for each widget. This is a significant cost because there is a fixed overhead for drawing a visual, regardless of its simplicity.

vispy/scene/widgets/widget.py
```diff
@@ -151,13 +157,28 @@ def stretch(self):
 
     @stretch.setter
     def stretch(self, s):
+        self._stretch = s
```
campagnola: should be float(s[0]), float(s[1]) as below.

vispy/scene/widgets/widget.py
```diff
@@ -151,13 +157,28 @@ def stretch(self):
...
+
+        Specifying a fixed size for either axis forces the widget to have a
+        specific size in a layout. Setting either axis to None allows the
+        widget to be resized by the layout.
+        """
+        return self._fixed_size
+
+    @fixed_size.setter
+    def fixed_size(self, s):
         self._stretch = float(s[0]), float(s[1])
```
campagnola: What's this for? stretch != fixed_size.

vispy/scene/widgets/widget.py
```diff
@@ -134,7 +140,7 @@ def inner_rect(self):
         this rectangle.
         """
         m = self.margin + self._border_width + self.padding
-        if not self.border_color.is_blank:
+        if not self._border.border_color.is_blank:
```
campagnola: This is an error (mine) -- the border width is already added in to `m`; no reason to add 1 again.

vispy/visuals/border.py
```diff
@@ -0,0 +1,210 @@
...
+"""
+
+FRAG_SHADER_BORDER = """
+void main() {
+    gl_FragColor = $border_color;
+}
+"""  # noqa
+
+
+class _BorderVisual(Visual):
```
campagnola: This looks more like a general-purpose rectangle visual to me. I suggest renaming to `RectVisual` and including the fill color as well as the border color.  @Eric89GXL what you think?

vispy/visuals/border.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    Visual subclass to display 2D pixel-width borders.
+
+    Parameters
+    ----------
+    pos : tuple (x, y)
+        Position where the colorbar is to be placed with
+        respect to the center of the colorbar
+    halfdim : tuple (half_width, half_height)
```
campagnola: No reason to use `halfdim` here--just use `size` like the widgets do.

vispy/scene/widgets/widget.py
bollu: The thought was that `BorderVisual` would be used by some of vispy's visuals, but that no one "externally" would really need it, so it wouldn't show up in Intellisense / autocompletion.

vispy/visuals/border.py
```diff
@@ -0,0 +1,210 @@
...
+"""  # noqa
+
+
+class _BorderVisual(Visual):
+    """
+    Visual subclass to display 2D pixel-width borders.
+
+    Parameters
+    ----------
+    pos : tuple (x, y)
```
campagnola: To keep consistent with the widgets, this should be the origin corner of the rect instead of its center. This also should also simplify the API and vertex calculations later on. Also, the docs need to be updated since this is no longer a colorbar.

vispy/scene/widgets/widget.py
bollu: Yes, I agree. However, I'm not sure what the neat solution to that is..? Like, it's a code versus performance trade-off right?

vispy/visuals/colorbar.py
```diff
@@ -45,105 +46,67 @@
...
-"""
-
-FRAG_SHADER_BORDER = """
-void main() {
-    gl_FragColor = $border_color;
-}
-"""  # noqa
-
 
 class _CoreColorBarVisual(Visual):
```
campagnola: I don't think this should be private--it's a useful primitive.

vispy/scene/widgets/colorbar.py
bollu: In that case, what's the convention for a constant? I'm used to C/C++ where `ALL_CAPS` are globals and/or constants usually.

vispy/scene/widgets/colorbar.py
bollu: I debated on keeping it pixel-width, but I thought that it would look weird in extreme situations. My personal view on this is that the solution adheres to the principle of least surprise. You resize the window, and the colorbar shrinks along that axis. It doesn't try to maintain pixel-width shape. Plus, the number of times you would throw it into such an extreme situation is very less.

Could you try and decrease the padding and check if that looks nicer to you? 

Of course, pixel-width could also be implemented :) I'm not against the idea in particular.

vispy/scene/widgets/widget.py
bollu: Thanks for the catch. I'll fix this!

vispy/visuals/colorbar.py
```diff
@@ -242,61 +190,22 @@ def cmap(self, cmap):
...
 
+    def _prepare_draw(self, view):
+        self._draw_mode = "triangles"
+        return True
 
 # The padding multiplier that's used to place the text
 # next to the Colorbar. Makes sure the text isn't
 # visually "sticking" to the Colorbar
-_TEXT_PADDING_FACTOR = 1.3
+_TEXT_PADDING_FACTOR = 1.05
```
campagnola: Make this an attribute of ColorBarVisual.

vispy/visuals/border.py
Eric89GXL: We have a `RectangleVisual` already which subclasses `PolygonVisual`, maybe we should just use that.

bollu: That doesn't have pixel-width borders IIRC, so it can't be used for this.

campagnola:
Overall this looks really good--the API and visual output are exactly what we are looking for.
There are some architectural issues that we might want to handle now or leave for a future PR (and I'm sorry I didn't raise these earlier; I was pretty swamped at the time):

* What to do about the issues relating to border, RectVisual, and Widget. Widget itself already has code for drawing the rectangle background+border, and that seems to be sufficient for everything we need widgets to do (and does it in one draw instead of two); I'm not sure why the new _BorderVisual is necessary? Whatever we decide, it probably makes sense to end up with a single RectVisual that draws both the border and fill. This visual could also accept an arbitrary function for the fill color, in which case the core ColorBar visual would become a thin wrapper around RectVisual.

* ColorBar does a lot of work to manage text that duplicates the functionality of AxisVisual. From this perspective, it might make more sense to have ColorBarVisual include only the core color bar and an associated AxisVisual. An immediate benefit of this is that we get the option of drawing ticks next to the color bar. Another option is for ColorBarVisual to include _only_ the color bar itself, whereas ColorBarWidget could just be a Grid that stacks a color bar and axis next to each other. This would probably greatly reduce the code overhead for having color bars, since it would reuse the existing Grid machinery to handle layout issues.

vispy/scene/widgets/colorbar.py
campagnola: If it's local, use all lower-case. If it's a constant local, then it probably makes more sense to make it into an attribute so the user has some option to modify it (but I would not make these into properties because then we would require an API change if we decide to rethink the size management later on).

vispy/scene/widgets/colorbar.py
campagnola: `examples/demo/plotting/colorbar`, for example, looks really nice with the default size. If you male the window half as tall, though, everything looks wrong. However I think we should leave this for now and fix it after we have a proper constraint-solving layout.

vispy/scene/widgets/widget.py
campagnola: Performance almost always gets priority in that tradeoff.
I think 75% of vispy is code acrobatics in the battle for better performance  :)

bollu:
Cool, I'm glad that the API is nice :)

As for the `_BorderVisual`, The reason I implemented it in the first place was because I wanted pixel-width borders for the colorbar. AFAIK, this isn't possible with `RectVisual`, since it uses `PolygonVisual` (that doesn't draw pixel-width borders). I decided to adapt the same code to be used in `Widget`.

However, the new `RectVisual` replacing `_BorderVisual`, along with `ColorBarVisual` being a wrapper sounds solid. However, could this be part of a second iteration on the colorbar?

Ditto with the `AxisVisual`. I guess it was a KISS thing, so I didn't over-complicate it. However, you're entirely right - using `AxisVisual` will give it more power _and_ ease of use, so this is the ideal target. Again, could this be pushed for the next iteration?

I'd like to get this working perfectly - fix the things you pointed out and smoothen the corners a little bit (bounds checking + nicer errors would be nice). After that, work on improvements. Thoughts?

vispy/visuals/colorbar.py
bollu: I would like to invoke YAGNI :) Make it public when we need it?

vispy/scene/widgets/widget.py
bollu: Again, performance improvements to be pushed to the second iteration?

vispy/visuals/border.py
bollu: @campagnola - agreed, it should be changed for consistency. However, I don't see how it would simplify the code - it's just a linear transformation, so it doesn't really affect the problem, right?

bollu:
@campagnola on second thought, about changing `BorderVisual` from `(pos, halfdim)` to `(pos, size)` - the `ColorBarVisual` uses the `(pos, halfdim)` representation as well (which feels like the "natural" way to think of it when talking about major/minor axes IMO). So, changing `BorderVisual` would simply introduce the friction of converting between the two representations with no clear upshot.

vispy/scene/widgets/widget.py
campagnola: This change causes a possible performance regression. We can benchmark this to see if it makes a real difference using `examples/basics/scene/grid_large.py`, however that example appears to be broken at the moment..

campagnola:
Let's leave BorderVisual as it is for now. If this eventually gets merged with RectVisual, then we can reconsider the API.

bollu:
@campagnola - Looks good now? I think we both agreed that `_BorderVisual` would be dealt with later, correct?

@Eric89GXL - CI passes as well, except for the size check in Travis.

campagnola:
The only outstanding issue is the performance regression in `Widget` (examples/basics/scene/grid_large.py runs about 2/3 of its speed in master). Probably easiest to revert the border-related changes in Widget--these don't seem to be necessary for this PR?

bollu:
@campagnola - reverted :) Is it good to go now?

vispy/visuals/border.py
Eric89GXL: @bollu this hasn't been addressed

Eric89GXL:
Otherwise LGTM

vispy/visuals/border.py
bollu: I thought I replied to @campagnola about that - that renaming it would
cause friction in the colorbar code (since it used the center
representation internally) with no real added benefit?

bollu:
@Eric89GXL - like I replied, I think rewriting the `BorderVisual` and the `halfdim` versus `size` is held off for the second iteration.

Eric89GXL:
Okay, please make a note of the remaining todos in a separate issue, then. @campagnola I'll let you do the honors if you're happy

bollu:
@campagnola - good to go? I've created a separate issue for the remaining polish stuff (#1053)

Eric89GXL:
I'll go ahead and merge this, @campagnola if you see additional issues please raise them in #1049

Eric89GXL:
Thanks @bollu!

PR #1043, opened 2015-08-03
[WIP] Graph visual

sh4wn:
This is a new pull request for the upcoming GraphVisual.

Tasks:

- [x] Create compound visual using ArrowVisual and MarkersVisual
- [x] Create API for graph layouts
- Implement Several automatic layouts
    - [x] Random
    - [x] Circular (all nodes on a single circle)
    - [x] Spring (force directed) layout based on the Fruchterman-Reingold algorithm
    - [ ] Maybe more
- [x] Fix layout animation

Issues:

- [ ] Fix arrows below the nodes


Eric89GXL:
@sh4wn note that you can draw inspiration from here for the force-directed stuff:

https://github.com/vispy/vispy/blob/master/examples/demo/scene/force_directed_graph.py

It would be great if you could *greatly* simplify that example in this PR, by internalizing a bunch of the work. WDYT?

examples/basics/visuals/graph.py
```diff
@@ -0,0 +1,68 @@
...
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+
+"""
+This example demonstrates how to visualise a NetworkX graph using the
+GraphVisual.
+"""
+
+import sys
+
+import networkx as nx
```
Eric89GXL: you'll probably need to add `networkx` as something that gets installed on `Travis` for this to pass the example build. Prefer a `conda` package if it's available, otherwise `pip` is fine.

vispy/visuals/graphs/graph.py
```diff
@@ -0,0 +1,197 @@
...
+"""
+
+from ..visual import CompoundVisual
+from ..line import ArrowVisual
+from ..markers import MarkersVisual
+from . import layouts
+
+
+_layouts_map = {
+    'random': layouts.random,
```
Eric89GXL: should probably also have `circular`?

examples/basics/visuals/graph.py
```diff
@@ -0,0 +1,68 @@
...
+    def on_resize(self, event):
+        self.visual.transform.scale = self.visual_size
+
+        vp = (0, 0, self.physical_size[0], self.physical_size[1])
+        self.context.set_viewport(*vp)
+        self.visual.transforms.configure(canvas=self, viewport=vp)
+
+    def on_draw(self, event):
+        gloo.clear('black')
+        self.visual.draw()
```
Eric89GXL: you should add a button-press handler that changes the layout type on the fly

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,46 @@
...
+            if adjacency_mat[edge] == 1 and adjacency_mat[reverse] == 0:
+                arrows.extend([
+                    node_coords[edge[0]],
+                    node_coords[edge[1]]
+                ])
+            elif adjacency_mat[edge] == 0 and adjacency_mat[reverse] == 0:
+                arrows.extend([
+                    node_coords[reverse[0]],
+                    node_coords[reverse[1]]
+                ])
```
Eric89GXL: We will definitely need to vectorize this at some point, it should be possible. But for now it's okay to be explicit :)

Eric89GXL: (but please add a note that it should be a target of future optimization)

Eric89GXL:
Looks like a good start. Let me know when you want additional/deeper feedback.

sh4wn:
Ok, so let's try the rebasing properly with this branch. 

What would be the steps to rebase this branch on the current VisPy master?

The procedure I normally follow:

* git checkout master
* git pull upstream master
* git checkout graph-visual
* git rebase master

After that git status says that the current branch (graph-visual) and the tracking branch (origin/graph-visual) have diverged. This is sort of expected because rebase rewrites commits. But I can't push to that branch anymore (only with -f), so what would be the solution?

sh4wn:
Ah thanks already for the review! Yes simplifying that example looks like a nice goal to me!

Eric89GXL:
That's pretty much what I do. When it says there's a conflict, I do:
```
git mergetool
...
git rebase --continue
```
Every once in a while, if you have a small commit that e.g. fixes a typo or changes some spacing, and the exact changes exist in `master`, you'll need to do `git rebase --skip` instead of `git rebase --continue`, but that's pretty rare.

I have my system configured so that `git mergetool` launches `meld`, an excellent 3-way merge tool. I think it's the default for Ubuntu, actually.

Eric89GXL:
Actually, usually I just do this:
```
git fetch upstream
git rebase upstream/master
```
That way you don't have to deal with your local `master` being in some accidental incorrect state, for example if you made a commit there instead of another branch and forgot.

sh4wn:
Ok, the rebase went fine, but it's more about this message:

    On branch graph-visual
    Your branch and 'origin/graph-visual' have diverged,
    and have 75 and 19 different commits each, respectively.
      (use "git pull" to merge the remote branch into yours)

Pulling right now from my own origin branch seems strange, or should I just force push it?

Eric89GXL:
You should force push. But before doing that, I'd recommend doing:
```
git checkout -b graph-visual-bak origin/graph-visual
git checkout graph-visual
```
Then proceed with
```
git push origin --force graph-visual
```
That way you have a local copy of your remote branch in case you accidentally did something wrong.

Eric89GXL:
BTW the 19 comes from the 19 commits in this PR, the 75 must be the 19 commits in this PR (that show up as different because the hashes have changed) plus what must be 56 commits that have made it into `master` since you originally started your branch.

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,177 @@
...
+
+            yield pos, line_vertices, arrows
+
+    def _sparse_fruchterman_reingold(self, adjacency_mat, directed=False):
+        # Optimal distance between nodes
+        if self.optimal is None:
+            self.optimal = 1 / np.sqrt(self.num_nodes)
+
+        # make sure we have a LIst of Lists representation
+        adjacency_mat = adjacency_mat.tolil()
```
sh4wn: Can I use `tolil()` here? It's actually a SciPy function, so I should probably check if it's available before actually using the sparse solver.

Eric89GXL: `scipy` is a soft dependency, so it would be good to at least check if it's available and throw an error if it's not.

sh4wn:
Force directed layout is implemented (with animation, but the animation is not perfect yet).

Top: Vispy
Bottom: NetworkX with Matplotlib

force-directed-vispy
force-directed-nx

bollu:
@sh4wn out of curiosity, what does it look like with AGG enabled for the lines?

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,212 @@
...
+                    adjacency_mat*distance/self.optimal
+                )
+            ).sum(axis=1)
+
+            # update positions
+            length = np.sqrt((displacement**2).sum(axis=1))
+            length = np.where(length < 0.01, 0.1, length)
+            delta_pos = np.transpose(np.transpose(displacement)*t/length)
+            pos += delta_pos
+            pos = rescale_layout(pos)
```
sh4wn: I think this should be changed somehow, but I'm not sure yet how: It normalizes the node coordinates, but each iteration again. This modifies the current iteration values which may not be desirable (networkx only normalizes them at the end)

If anyone knows some other way to make sure the positions stay within [0, 1], please let me know.

Eric89GXL: what's the problem? is it too slow? or bad in principle?

Eric89GXL:
That example looks nice :)

sh4wn:
@bollu Not tested yet, because the AGG backend doesn't support the 'segment' connect mode. Which means I would need to create a new LineVisual for each line segment (graph edge). With the OpenGL backend I can pass a big list of vertices and tell it to only draw lines between two consecutive vertices.

bollu:
@sh4wn Ah, all right :)

vispy/visuals/graphs/layouts/force_directed.py
sh4wn: Well, I don't think it slow per se (I don't notice any delays), but it somehow feels a bit wrong to alter the values mid iteration. 

Then again, it should scale each point the same with aspect ration maintained, so it don't think it returns incorrect values.

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,212 @@
...
+                np.transpose(delta) * (
+                    self.optimal * self.optimal/distance**2 -
+                    adjacency_mat*distance/self.optimal
+                )
+            ).sum(axis=1)
+
+            # update positions
+            length = np.sqrt((displacement**2).sum(axis=1))
+            length = np.where(length < 0.01, 0.1, length)
+            delta_pos = np.transpose(np.transpose(displacement)*t/length)
```
Eric89GXL: a few `numpy` tricks to learn here. First, `arr.T` is easier to use than `np.transpose(arr)`. Second, you're doing this to get broadcasting, right? You should be able to do this instead:
```
delta_pos = displacement * t / length[:, np.newaxis]
```

vispy/visuals/graphs/layouts/force_directed.py
Eric89GXL: You might be able to change/threshold `delta_pos` instead of changing `pos`, but that would probably lead to weird behaviors, so rescaling here might be the sanest thing to do.

vispy/visuals/graphs/layouts/force_directed.py
sh4wn: I must say, this almost comes directly from NetworkX (forgot to add attribution in the docs). 

But please let me know if things can be improved.

Eric89GXL: Try what I suggested, it should work

sh4wn:
Ok, with 1000 nodes, it is very slow. The sparse solver is definitely faster, but still takes a lot of time to calculate the new layout. The result is an animation with an FPS of about 0.5 (takes about 2 seconds to calculate the new iteration), or maybe even slower.

I'll see if I can profile this a bit with cProfile.

sh4wn:
Profiling shows that the `straight_line_vertices` function to main reason for the slowdown is, so let's see how to optimize this. :)

Eric89GXL:
@sh4wn yes, that function could be much more optimized I think. Is the adjacency matrix sparse, dense, or either?

sh4wn:
I think it's in general sparse (most nodes only have edges with a few other nodes), but there are definitely cases where there are a lot of edges. But this is just based on a feeling of mine, not really based on data.

I do not yet see how NumPy could be used to create a new vertices list based on an adjacency matrix.

Eric89GXL:
I mean, is it an instance of `scipy.sparse.*` or `ndarray`? Or could it be either of those?

Eric89GXL:
And just so the problem is clear, you basically want a fast way to get all the non-zero `(i, j)` pairs from an `ndarray`?

Eric89GXL:
If you have an `ndarray`, you can do:
```
i, j = np.where(adj_mat)
```
`i` then has all the first dimension indices of non-zero values, `j` all the second dimension indices. The values are then `val = adj_mat[i, j]`.

For a sparse matrix, I'd use [to_coo()] (http://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.sparse.coo_matrix.html#scipy.sparse.coo_matrix) to get it in a convenient representation.

If you want to treat the upper and lower triangles separately (I assume you do), then you can [extract the lower and upper triangles] (http://stackoverflow.com/questions/8905501/extract-upper-lower-triangular-part-of-a-numpy-matrix). This will only work for dense (`ndarray`) matrices. Instead of figuring out how to also do it for `coo` matrices, it is probably just be easier to make a mask after you have `i, j` from either method as:
```
mask = (i < j)
```

Make sense?

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,103 @@
...
+        Wether the graph is directed. If this is true it will also generate
+        the vertices for arrows which can be passed to :class:`ArrowVisual`.
+
+    Returns
+    -------
+    vertices : tuple
+        Returns a tuple containing containing (`line_vertices`,
+        `arrow_vertices`)
+    """
+    if adjacency_mat.shape[0] != adjacency_mat.shape[1]:
```
Eric89GXL: `if adjacency_mat.ndim !=2 or ...`

We should also type-check. You might want to put this at the top of the file:
```
try:
    from scipy.sparse import issparse
except ImportError:
    def issparse(x):
        return False
```
Then your type-check here can basically be an attempt to cast it to the proper type before dimension checking:
```
if not issparse(adjacency_mat)) and not isinstance(adjacency_mat, ndarray):
    adjacency_mat = np.array(adjacency_mat, float)
```

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,103 @@
...
+        The upperbound value for the coordinates range
+
+    Returns
+    -------
+    pos : array
+        The rescaled (normalized) coordinates in the range [0, `scale`].
+
+    Notes
+    -----
+    Adapted from NetworkX.
```
Eric89GXL: make a note that this operates in-place (important for devs to know)

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # shift origin to (0,0)
+    lim = 0  # max coordinate for all axes
+    for i in range(pos.shape[1]):
+        pos[:, i] -= pos[:, i].min()
+        lim = max(pos[:, i].max(), lim)
+
+    # rescale to (0,scale) in all directions, preserves aspect
+    for i in range(pos.shape[1]):
+        pos[:, i] *= scale/lim
```
Eric89GXL: Vectorize this function, too. Something like this should work:
```
pos -= pos.min(axis=1)[:, np.newaxis]
pos *= scale / pos.max()
```
Make sense? It's *very* useful in Python / `numpy` / `scipy` to start thinking about operations in vectorized form like this. If you don't feel like you could have gotten to this solution on your own, it's worth taking the time to play around with code and data, looking at what's happening along the way, until you do IMO.

vispy/visuals/graphs/layouts/random.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+
+import numpy as np
+
+from ..util import straight_line_vertices
+
+
+def random(adjacency_mat, directed=False):
+    if adjacency_mat.shape[0] != adjacency_mat.shape[1]:
+        raise ValueError("Adjacency matrix should be square.")
```
Eric89GXL: This check is already done in `straight_line_vertices` so you don't need it here, right?

vispy/visuals/graphs/layouts/random.py
```diff
@@ -0,0 +1,27 @@
...
+
+This layout positions the nodes at random
+"""
+
+import numpy as np
+
+from ..util import straight_line_vertices
+
+
+def random(adjacency_mat, directed=False):
```
Eric89GXL: Need docstring

vispy/visuals/graphs/layouts/force_directed.py
Eric89GXL: Actually, you shouldn't need to convert here at all.

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,212 @@
...
+            # Loop over rows
+            for i in range(adjacency_mat.shape[0]):
+                # difference between this row's node position and all others
+                delta = (pos[i]-pos).T
+                # distance between points
+                distance = np.sqrt((delta**2).sum(axis=0))
+                # enforce minimum distance of 0.01
+                distance = np.where(distance < 0.01, 0.01, distance)
+                # the adjacency matrix row
+                row = np.asarray(adjacency_mat.getrowview(i).toarray())
```
Eric89GXL: You might want to write a simple helper to deal with this, like:
```
def _get_row(adj, idx):
    row = adj[idx]
    if issparse(adj):
        row = row.toarray()
    return row
```

vispy/visuals/graphs/layouts/force_directed.py
Eric89GXL: ...instead, earlier you should ensure that `adjacency_mat` is either `ndarray`, or one of the `sparse` forms that allows row access (maybe any of them, I can't remember?), converting to such a form if necessary e.g. via `to_csr()`.

vispy/visuals/graphs/layouts/circular.py
```diff
@@ -0,0 +1,45 @@
...
+from ..util import straight_line_vertices
+
+
+def circular(adjacency_mat, directed=False):
+    """Places all nodes on a single circle.
+
+    Parameters
+    ----------
+    adjacency_mat : matrix
+        The graph adjacency matrix
```
Eric89GXL: missing `directed` param

If `directed is False`, you should say what part of the matrix is used. Just upper-right, lower-left, or the union of the two?

Eventually we might want to allow the adjacency matrix to actually be weights instead of zeros and ones, too, BTW.

.travis.yml
```diff
@@ -102,7 +102,7 @@ install:
     # Install numpy, flake
     - conda create -n testenv --yes --quiet pip python=$PYTHON > ${REDIRECT_TO};
     - source activate testenv > ${REDIRECT_TO};
-    - conda install --yes --quiet numpy$NUMPY nose pytest > ${REDIRECT_TO};
+    - conda install --yes --quiet numpy$NUMPY nose pytest networkx > ${REDIRECT_TO};
```
Eric89GXL: `networkx` should only be installed for the `DEPS=full` build (line 124)

Eric89GXL: You will probably need to add both `networkx` and `numpy$NUMPY` to the `conda install` line, too, to make sure `networkx` doesn't pull in the wrong version via recommendation

Eric89GXL: ...and you'll also need to modify `appveyor.yml` to add `networkx` there, too

Eric89GXL:
Looks like you're making good progress, I suspect with the suggested vectorization ideas it will get much faster. Let me know if you need help, we can chat about what's going on on `gitter` or Google. I really enjoy optimization :)

vispy/visuals/graphs/layouts/circular.py
sh4wn: If `directed` is false, then it uses the whole matrix (for each combination of (i, j) both (i, j) as (j, i) should be non-zero).

Weights is indeed something we should implement.

sh4wn:
Ok cool, straight_line_vertices is a lot faster now, thanks for the tips :) Most of the code still needs docs, but could you take a look on the NumPy stuff?

With larger networks (1000 nodes) it is still a bit stuttury, about 5 FPS, but this is way better.

Eric89GXL:
There are a lot of `numpy` tricks, and I only know some of them. I'll see if I can find some others :)

Do you know where the slowdowns are currently?

vispy/visuals/graphs/graph.py
```diff
@@ -0,0 +1,216 @@
...
+        for k in self._node_kwargs:
+            if k in kwargs:
+                translated = (self._node_kw_trans[k] if k in
+                              self._node_kw_trans else k)
+
+                node_kwargs[translated] = kwargs.pop(k)
+
+        if len(kwargs) > 0:
+            raise TypeError(
+                "{}.set_data() got invalid keyword arguments: {}".format(
```
Eric89GXL: Although `.format` is stated as the preferred way to do it, it actually seems more common and simpler just to use `%s`. By convention it's what we use, so go ahead and chenge this.

vispy/visuals/graphs/layouts/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Retrieve a graph layout
+
+    Some graph layouts accept extra options. Please refer to their
+    documentation for more information.
+
+    Parameters
+    ----------
+    name : string
+        The name of the layout. The variable :data`AVAILBLE_LAYOUTS`
```
Eric89GXL: I don't *think* you need `:data` in front, the `numpy`/`sphinx` docstring code should automatically find it.

vispy/visuals/graphs/layouts/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+    'random': random,
+    'circular': circular,
+    'force_directed': fruchterman_reingold,
+    'spring_layout': fruchterman_reingold
+}
+
+AVAILABLE_LAYOUTS = _layout_map.keys()
+
+
+def get(name, *args, **kwargs):
```
Eric89GXL: This should at least be `get_layout`. Can we make it private? Or do you think users will need this?

vispy/visuals/graphs/layouts/circular.py
```diff
@@ -0,0 +1,45 @@
...
+    if adjacency_mat.shape[0] != adjacency_mat.shape[1]:
+        raise ValueError("Adjacency matrix should be square.")
+
+    num_nodes = adjacency_mat.shape[0]
+
+    t = np.arange(0, 2.0*np.pi, 2.0*np.pi/num_nodes, dtype=np.float32)
+
+    # Visual coordinate system is between 0 and 1, so generate a circle with
+    # radius 0.5 and center it at the point (0.5, 0.5).
+    node_coords = np.transpose(0.5 * np.array([np.cos(t), np.sin(t)]) + 0.5)
```
Eric89GXL: use `.T` instead of `np.transpose`

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,218 @@
...
+try:
+    from scipy.sparse import issparse
+except ImportError:
+    def issparse(*args, **kwargs):
+        return False
+
+from ..util import straight_line_vertices, rescale_layout
+
+
+class fruchterman_reingold:
```
Eric89GXL: `(object)`

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,148 @@
...
+except ImportError:
+    def issparse(*args, **kwargs):
+        return False
+
+
+def get_edges(adjacency_mat):
+    if issparse(adjacency_mat):
+        func = _sparse_get_edges
+    else:
+        func = _get_edges
```
Eric89GXL: FYI for simple things like this you can do:
```
func = _sparse_get_edges if issparse(adjacency_mat) else _get_edges
```

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,148 @@
...
+
+    for result in func(adjacency_mat):
+        yield result
+
+
+def _sparse_get_edges(adjacency_mat):
+    # Each non zero value in a COO matrix is stored in (row, col, value)
+    # format
+    coo_mat = adjacency_mat.tocoo()
```
Eric89GXL: Converting this repeatedly is going to kill performance. Instead you should just ensure that, at this point in the call, you are always guaranteed e.g. a `coo` matrix, if that's what is fastest

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,148 @@
...
+        _sparse_get_edges(adjacency_mat),
+        _sparse_get_edges(lower)
+    )
+
+
+def _get_directed_edges(adjacency_mat):
+    upper = np.triu(adjacency_mat, k=1)
+    lower = np.tril(adjacency_mat)
+
+    return itertools.chain(get_edges(upper), get_edges(lower))
```
Eric89GXL: in the `_sparse` version you call the `_sparse_get_edges`, but here you call `get_edges`. Either both should call the specific version, or both should call the generic version.

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,148 @@
...
+    Returns
+    -------
+    vertices : tuple
+        Returns a tuple containing containing (`line_vertices`,
+        `arrow_vertices`)
+    """
+
+    if (not issparse(adjacency_mat) and not isinstance(adjacency_mat,
+                                                       np.ndarray)):
+        adjacency_mat = np.array(adjacency_mat, float)
```
Eric89GXL: You can avoid the second conditional by letting `numpy` take care of it, like:
```
if not issparse(adjacency_mat):
    adjacency_mat = np.asarry(adjacency_mat, float)
```
http://docs.scipy.org/doc/numpy/reference/generated/numpy.asarray.html

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,148 @@
...
+
+    if (adjacency_mat.ndim != 2 or adjacency_mat.shape[0] !=
+            adjacency_mat.shape[1]):
+        raise ValueError("Adjacency matrix should be square.")
+
+    arrow_vertices = np.array([])
+
+    edges = np.array(list(get_edges(adjacency_mat)))
```
Eric89GXL: This is going to hurt performance again. Consider a 10,000 element sparse `coo` connectivity matrix. It should be a couple of memory copies to make the array. Instead, what your code will do (because you use generators) is it will go through 10,000 iterations of splitting each of the `i`, `j` arrays, then it will create a list from them, then it will turn those into an array again. This will be much slower than just doing:
```
edges = np.concatenate((adjacency_mat.row[:, np.newaxs],
                        adjacency_mat.col[:, np.newaxis]), axis=-1)
```
I appreciate the idea that you had to write a generally useful `get_edges` function, but in this instance it will hurt performance too much to use it as the size of the connectivity matrix gets large. This is especially true if your `get_edges` function is converting the sparse matrix to `coo` every time, which it might be doing.

I think it would be much faster to just require that, at this point, the input be either sparse `coo` or `ndarray` of floats, and change the code that calls this function to comply with that. Also, this function should be private.

Can all the functions in here be private?

vispy/visuals/graphs/layouts/__init__.py
QuLogic: Not sure if it's necessary either since it's not actually documented, but it should be `:data:` in any case. And `AVAILABLE` is typo'd here.

vispy/visuals/graphs/layouts/__init__.py
sh4wn: I think users will use this too, because it also allows to configure some layouts (for example setting the initial node positions with the Fruchterman-Reingold algorithm)

sh4wn:
Ok, at the cost of additional memory, I now use both a lil_matrix and a coo_matrix version of the adjacency matrix. 

The first one for the Fruchterman-Reingold algorithm, and lil_matrix gives fast access to rows (getrowview, the _get_row function you suggested above is a lot slower). The COO matrix is used for line vertex generation because it allows easy construction of (row, col) pairs and easy construction of upper and lower triangular matrices. 

The time spent in straight_line_vertices is now down to only a few percent, coming from more than 95%. Current call graph from profiler:

screenshot from 2015-08-12 15-44-27

Eric89GXL:
Why use `lil` at all? It should be faster to use `csr` if you plan on extracting rows, right?

Eric89GXL:
And how is the speed with large graphs? Is it smooth? Or is the slowdown in some other function now?

Eric89GXL:
For example, that 17% you lose in `getrowview` might be almost instantaneous for `csr` matrices

vispy/visuals/graphs/graph.py
```diff
@@ -0,0 +1,216 @@
...
+from ..markers import MarkersVisual
+from . import layouts
+
+
+class GraphVisual(CompoundVisual):
+    """Visual for displaying graphs or networks.
+
+    Parameters
+    ----------
+    adjacency_mat : array
```
Eric89GXL: `array or sparse matrix` probably

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,222 @@
...
+            distance = np.sqrt((delta**2).sum(axis=-1))
+            # Enforce minimum distance of 0.01
+            distance = np.where(distance < 0.01, 0.01, distance)
+            # Displacement "force"
+            displacement = np.transpose(
+                np.transpose(delta) * (
+                    self.optimal * self.optimal/distance**2 -
+                    adjacency_mat*distance/self.optimal
+                )
+            ).sum(axis=1)
```
Eric89GXL: should be cleaner to do this with broadcasting instead of `transpose`, but if you absolutely must use `transpose` please use `.T`

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,222 @@
...
+
+            # update positions
+            length = np.sqrt((displacement**2).sum(axis=1))
+            length = np.where(length < 0.01, 0.1, length)
+            delta_pos = displacement * t / length[:, np.newaxis]
+            pos += delta_pos
+            pos = rescale_layout(pos)
```
Eric89GXL: rescale_layout should probably be private, and you don't need to do `pos =` here because it operates inplace

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,122 @@
...
+    upper = triu(adjacency_mat)
+    lower = tril(adjacency_mat)
+
+    return np.concatenate((
+        func(upper),
+        func(lower)
+    ))
+
+
+def straight_line_vertices(adjacency_mat, node_coords, directed=False):
```
Eric89GXL: This should probably also be private

sh4wn:
Well, not entirely, because adjacency_mat[row].toarray() is quite expensive (adjacency_mat in csr format). But maybe we can change the way we calculate the displacement and work around the toarray conversion.

Eric89GXL:
Is `lil` faster, though? That doesn't make much sense, since I think a `lil` will have to traverse the entire list to determine row elements, whereas `csr.getrow(i).toarray()` just has to construct an array.

Eric89GXL:
Oh never mind, I forgot `lil` is row-based. Carry on :)

sh4wn:
Yes `lil` is faster:

screenshot from 2015-08-12 16-41-21

`getrow(i)` and `toarray` in total around 6 ms, against the ~1.3 ms of lil.

Eric89GXL:
So how is the speed with large graphs now? Is it smooth? Or is the slowdown in some other function now?

sh4wn:
It's not bad, but there's definitely a slight delay, and I would guess it's around 8-15 FPS. 5000 nodes is already really slow (~1 FPS)

Eric89GXL:
Where is the slowdown now? It might be possible to optimize further.

Or if you'd rather work on something else, you could address the remaining comments, add some tests (e.g. that ndarray and different sparse types all render), and merge this so you can move on. Up to you.

sh4wn:
The big slow down is in the _sparse_fruchterman_reingold function itself. I think I'll take a look tomorrow on how to improve the performance there.

screenshot from 2015-08-12 16-51-57

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,222 @@
...
+            dtype=adjacency_mat.dtype
+        )
+
+        # The inscrutable (but fast) version
+        # This is still O(V^2)
+        # Could use multilevel methods to speed this up significantly
+        for iteration in range(self.iterations):
+            # Matrix of difference between points
+            for i in range(pos.shape[1]):
+                delta[:, :, i] = pos[:, i, None] - pos[:, i]
```
Eric89GXL: Don't loop, use broadcasting instead:
```
delta[:, :, :] = pos[:, np.newaxis, :] - pos[:, :, np.newaxis]
```
Also `np.newaxis` is preferred to `None` even though it's currently an alias (e.g., it might not be in the future, it is more explicit, etc.)

Eric89GXL: And if that expression isn't correct, there should be one that is. You might need to reconfigure how you initialize and/or think of the `delta` matrix to avoid needing to use `np.transpose` or `np.rollaxis` to swap axes

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,222 @@
...
+
+            # Distance between points
+            distance = np.sqrt((delta**2).sum(axis=-1))
```
Eric89GXL: `delta * delta` is usually faster than `delta ** 2`

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,222 @@
...
+        # Simple cooling scheme.
+        # Linearly step down by dt on each iteration so last iteration is
+        # size dt.
+        dt = t / float(self.iterations+1)
+
+        displacement = np.zeros((self.dim, self.num_nodes))
+        for iteration in range(self.iterations):
+            displacement *= 0
+            # Loop over rows
+            for i in range(adjacency_mat.shape[0]):
```
Eric89GXL: find a way to do this without looping, it will speed things up substantially...

vispy/visuals/graphs/layouts/force_directed.py
```diff
@@ -0,0 +1,222 @@
...
+                row = adjacency_mat.getrowview(i).toarray()
+
+                # displacement "force"
+                displacement[:, i] += (
+                    delta * (
+                        self.optimal *
+                        self.optimal/distance**2 -
+                        row*distance/self.optimal
+                    )
+                ).sum(axis=1)
```
Eric89GXL: I suspect you can boil this down to a `dot` operation between your sparse matrix and an array of `deltas`. This will be fastest if `adjacency_mat` is a `csr` matrix, I think.

Eric89GXL:
I'm also happy to open a PR into your branch/fork to optimize, if you'd rather not deal with it. Depends on how much `numpy` optimization/tricks you'd like to learn :)

vispy/visuals/graphs/layouts/force_directed.py
sh4wn: Cool, I'm starting to understand the broadcasting rules a bit. The above things are now implemented without transpose.

vispy/visuals/graphs/layouts/force_directed.py
Eric89GXL: Do you want to give this bit a shot, or do you want me to open a PR?

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,122 @@
...
+    # Get indices of all non zero values
+    i, j = np.where(adjacency_mat)
+
+    return np.concatenate((i[:, np.newaxis], j[:, np.newaxis]), axis=-1)
+
+
+def _get_directed_edges(adjacency_mat):
+    func = _sparse_get_edges if issparse(adjacency_mat) else _ndarray_get_edges
+    triu = sparse.triu if issparse(adjacency_mat) else np.triu
+    tril = sparse.tril if issparse(adjacency_mat) else np.tril
```
Eric89GXL: once you have more than one (or maybe two) of the same conditional, you should probably split it up:
```
if issparse(adjacency_mat):
   ...
else:
   ...
```

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,122 @@
...
+
+    upper = triu(adjacency_mat)
+    lower = tril(adjacency_mat)
+
+    return np.concatenate((
+        func(upper),
+        func(lower)
+    ))
```
Eric89GXL: This can all fit on a single line

sh4wn:
I'm now using the same "delta pos" calculation method for both sparse and dense matrices. But using adjacency_mat.toarray in _sparse_fruchterman_reingold probably defeats the idea of using separate solvers. It is a bit faster though, but maybe you have another idea?

Eric89GXL:
Yeah just combine the methods, they don't need to be separate anymore really. You can just triage within a single function whether you do `.toarray()` or not.

But please add a note saying that the `toarray()` operation will not work for large sparse arrays (it will cause a memory blowup) so we need to find a way to do it using sparse operations in the future.

BTW what is the speed like now? Any better? Do you know where the slowdown is now?

Eric89GXL:
The example also currently doesn't work for me

sh4wn:
Hmm, do you get an error? Works fine here

examples/basics/visuals/graph.py
```diff
@@ -0,0 +1,76 @@
...
+class Canvas(app.Canvas):
+    def __init__(self):
+        app.Canvas.__init__(self, title="Simple NetworkX Graph",
+                            keys="interactive", size=(600, 600))
+
+        self.graph = nx.fast_gnp_random_graph(100, 0.02)
+
+        self.visual = visuals.GraphVisual(
+            nx.adjacency_matrix(self.graph),
+            layout=layouts.get('force_directed'),
```
Eric89GXL: How do you not get an error here? `get` doesn't exist. I get this, as expected:
```
  File "examples/basics/visuals/graph.py", line 73, in <module>
    win = Canvas()
  File "examples/basics/visuals/graph.py", line 29, in __init__
    layout=layouts.get('force_directed'),
AttributeError: 'module' object has no attribute 'get'
```

sh4wn: Ah it's changed here locally but I didn't commit those changes because I experiment often a bit with the number of nodes and probability of the edge

vispy/visuals/graphs/util.py
```diff
@@ -0,0 +1,122 @@
...
+
+    if directed:
+        arrows = np.array(list(_get_directed_edges(adjacency_mat)))
+        arrow_vertices = node_coords[arrows.ravel()]
+        arrow_vertices = arrow_vertices.reshape((len(arrow_vertices)/2, 4))
+
+    return line_vertices, arrow_vertices
+
+
+def rescale_layout(pos, scale=1):
```
Eric89GXL: this should probably be private

vispy/visuals/graphs/layouts/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+_layout_map = {
+    'random': random,
+    'circular': circular,
+    'force_directed': fruchterman_reingold,
+    'spring_layout': fruchterman_reingold
+}
+
+AVAILABLE_LAYOUTS = _layout_map.keys()
```
Eric89GXL: on Py3k `.keys()` is a generator, so you need to do `list(...)` or `tuple(...)`, and we want it to be immutable so use the latter

vispy/visuals/graphs/layouts/circular.py
```diff
@@ -0,0 +1,46 @@
...
+    Yields
+    ------
+    (node_vertices, line_vertices, arrow_vertices) : tuple
+        Yields the node and line vertices in a tuple. This layout only yields a
+        single time, and has no builtin animation
+    """
+
+    num_nodes = adjacency_mat.shape[0]
+
+    t = np.arange(0, 2.0*np.pi, 2.0*np.pi/num_nodes, dtype=np.float32)
```
Eric89GXL: this would be a good use for `t = np.linpsace(0, 2 * np.pi, num_nodes, endpt=False, dtype=np.float32)`

vispy/visuals/graphs/layouts/circular.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # Visual coordinate system is between 0 and 1, so generate a circle with
```
Eric89GXL: Is it between 0 and 1? I think in general we can make it whatever we want, and centering at 0, 0 with extend -1/+1 seems a bit nicer for doing e.g. `STTransform` on top

Eric89GXL:
Now I get:
```
  File "/home/larsoner/custombuilds/vispy/vispy/visuals/graphs/layouts/force_directed.py", line 197, in _calculate_delta_pos
    )[:, :, np.newaxis]
  File "/home/larsoner/.local/lib/python2.7/site-packages/numpy/matrixlib/defmatrix.py", line 347, in __rmul__
    return N.dot(other, self)
ValueError: shapes (1000,1000,2) and (1000,1000,1) not aligned: 2 (dim 2) != 1000 (dim 1)
ERROR: Invoking <bound method Canvas.animate of <Canvas (PyQt4) at 0x7ff85630b5d0>> for Event
import pdb; pdb.pm()
> /home/larsoner/.local/lib/python2.7/site-packages/numpy/matrixlib/defmatrix.py(347)__rmul__()
-> return N.dot(other, self)
```
This is because:
```
(Pdb) p type(adjacency_mat)
<class 'numpy.matrixlib.defmatrix.matrix'>
```
In other words, `adjacency_mat` is a `matrix` not `ndarray`, so the multiplication operator is a `dot`, not an elementwise multiply.

Eric89GXL:
Does it work for you, though?

sh4wn:
No, I do not encounter any errors when running the example.

Eric89GXL:
What versions of `numpy` / `scipy` are you running? Maybe mine are newer. Are you sure you're running the same version that is up on GitHub, or do you have other local modifications?

vispy/visuals/graphs/layouts/circular.py
sh4wn: Well, in the Graph Example I am using STTranform, and when using -1 to 1 some nodes are rendered outside the canvas.

Eric89GXL: You can change the `STTransform` you are using to compensate for that, though, right?

Can you look at what other visuals do by default? I think they are centered at `(0, 0)` so we should try to be consistent if we can.

sh4wn:
Sorry for the absence lately, I've been feeling really sick the past few days. But let's try to get this PR merged before the end of GSoC, and leave optimization for now.

These are the library versions:

```
Python 3.4.3 (default, Mar 25 2015, 17:13:50) 
[GCC 4.9.2 20150304 (prerelease)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> import scipy
>>> import networkx
>>> print(numpy.__version__, scipy.__version__, networkx.__version__)
1.9.2 0.16.0 1.10
```

Eric89GXL:
Okay, that's reasonable. Go ahead and fix the remaining issues pointed out above. Then I might make a PR into your branch to fix any issues I see at my end.

sh4wn:
Hmm, when trying to display a non-directed graph, I get this error:

```
  File "/home/lucas/programming/python/vispy/examples/basics/visuals/graph.py", line 78, in <module>
    app.run()
  File "/home/lucas/programming/python/vispy/vispy/app/_default_app.py", line 62, in run
    return default_app.run()
  File "/home/lucas/programming/python/vispy/vispy/app/application.py", line 137, in run
  File "/home/lucas/programming/python/vispy/vispy/app/backends/_qt.py", line 229, in _vispy_run
    return app.exec_()
  File "/home/lucas/programming/python/vispy/vispy/app/backends/_qt.py", line 646, in paintGL
  File "/home/lucas/programming/python/vispy/vispy/util/event.py", line 455, in __call__
  File "/home/lucas/programming/python/vispy/vispy/util/event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  << caught exception here: >>
  File "/home/lucas/programming/python/vispy/vispy/util/event.py", line 471, in _invoke_callback
    cb(event)
  File "/home/lucas/programming/python/vispy/examples/basics/visuals/graph.py", line 65, in on_draw
    self.visual.draw()
  File "/home/lucas/programming/python/vispy/vispy/visuals/visual.py", line 588, in draw
    v.draw()
  File "/home/lucas/programming/python/vispy/vispy/visuals/visual.py", line 588, in draw
    v.draw()
  File "/home/lucas/programming/python/vispy/vispy/visuals/visual.py", line 432, in draw
  File "/home/lucas/programming/python/vispy/vispy/visuals/line/arrow.py", line 76, in _prepare_draw
    v = self._prepare_vertex_data()
  File "/home/lucas/programming/python/vispy/vispy/visuals/line/arrow.py", line 93, in _prepare_vertex_data
    v['v1'] = arrows[:, 0:2]
IndexError: too many indices for array
```

With this as graph:

```python
self.graph = nx.fast_gnp_random_graph(3000, 0.0006)
```

examples/basics/visuals/graph.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def on_draw(self, event):
+        gloo.clear('black')
+        self.visual.draw()
+
+    def animate(self, event):
+        ready = self.visual.animate_layout()
+
+        if ready:
+            self.timer.disconnect(self.animate)
```
Eric89GXL: This example works well on my machine now, there must have been some error because I was on `networkx` version 1.9 instead of 1.10. It makes me wonder if it used to give a sparse matrix and now it gives dense, or vice-versa. We need to check that both code paths work before merging...

examples/basics/visuals/graph.py
```diff
@@ -0,0 +1,79 @@
...
+class Canvas(app.Canvas):
+    def __init__(self):
+        app.Canvas.__init__(self, title="Simple NetworkX Graph",
+                            keys="interactive", size=(600, 600))
+
+        self.graph = nx.fast_gnp_random_graph(1000, 0.0006, directed=True)
+
+        self.visual = visuals.GraphVisual(
+            # np.asarray(nx.to_numpy_matrix(self.graph)),
+            nx.adjacency_matrix(self.graph),
```
Eric89GXL: for example, changing this to `...).todense(),` causes a failure on my machine:
```
  File "/home/larsoner/custombuilds/vispy/vispy/visuals/graphs/layouts/force_directed.py", line 94, in _fruchterman_reingold
    dtype='f32'
  File "/home/larsoner/.local/lib/python2.7/site-packages/numpy/core/numeric.py", line 469, in asarray
    return array(a, dtype, copy=False, order=order)
TypeError: data type "f32" not understood
ERROR: Invoking <bound method Canvas.animate of <Canvas (PyQt4) at 0x7f82a09b0550>> for Event
```

Eric89GXL:
Let me know if you're not feeling well enough to make the additional changes for merge. If so I can take over. I hope you feel better soon!

examples/basics/visuals/graph.py
Eric89GXL: And using an older `numpy` I get this error:
```
  File "/home/larsoner/custombuilds/vispy/vispy/visuals/graphs/layouts/force_directed.py", line 121, in _fruchterman_reingold
    delta_pos = self._calculate_delta_pos(adjacency_mat, pos, delta, t)
  File "/home/larsoner/custombuilds/vispy/vispy/visuals/graphs/layouts/force_directed.py", line 197, in _calculate_delta_pos
    )[:, :, np.newaxis]
  File "/usr/lib/python2.7/dist-packages/numpy/matrixlib/defmatrix.py", line 347, in __rmul__
    return N.dot(other, self)
ValueError: objects are not aligned
```

Eric89GXL: ...and with the newer Python with your latest change, same thing but more informative error:
```
  File "/home/larsoner/custombuilds/vispy/vispy/visuals/graphs/layouts/force_directed.py", line 121, in _fruchterman_reingold
    delta_pos = self._calculate_delta_pos(adjacency_mat, pos, delta, t)
  File "/home/larsoner/custombuilds/vispy/vispy/visuals/graphs/layouts/force_directed.py", line 197, in _calculate_delta_pos
    )[:, :, np.newaxis]
  File "/home/larsoner/.local/lib/python2.7/site-packages/numpy/matrixlib/defmatrix.py", line 349, in __rmul__
    return N.dot(other, self)
ValueError: shapes (1000,1000,2) and (1000,1000,1) not aligned: 2 (dim 2) != 1000 (dim 1)
ERROR: Invoking <bound method Canvas.animate of <Canvas (PyQt4) at 0x7f682f8a3510>> for Event
```

examples/basics/visuals/graph.py
Eric89GXL: Can you also print something here that says it has converged / finished? It would be cool if the number of iterations could be `'auto'`, where you calculate the difference between the previous layout and the current one, and raise `StopIteration` once it has sufficiently converged.

examples/basics/visuals/graph.py
sh4wn: I think it's something with small differences between np.matrix and np.array. todense() returns a matrix, while something like `np.asarray(nx.to_numpy_matrix(self.graph))` works fine.

Eric89GXL: Right, that's what it is -- and the matrix operates differently from an array with the `*` operator, see my comment:

https://github.com/vispy/vispy/pull/1043#issuecomment-130733060

So the code should probably ensure that, if it's a matrix, it gets cast to an array so that this error doesn't happen.

examples/basics/visuals/graph.py
sh4wn: Well, the algorithm has a builtin 'cooldown' which limits the displacement more and more with each iteration. I'm not sure how we would automatically calculate the required "temperature".

sh4wn:
Ok, ugh it's slowly getting better, I'll see if I can finish up a few last things this evening. But I'll be mostly in my bed for now.

Eric89GXL:
@sh4wn if you don't mind, I could open a PR into your branch with some fixes (e.g., addressing my comments and maybe a couple of efficiency tweaks) if that would help. That way it could be more or less ready to go soon. I hope you feel better soon.

sh4wn:
Yes, I saw the PR. I'll merge it soon. The next study year starts next week, so I'm currently preparing a few thing for that, but I'll continue this work the coming months (although probably with varying levels of available time)!

Eric89GXL:
@sh4wn now that you're a couple weeks in, how does the workload look? Will you have time for this? If not, I can try to bring it up to shape for merge.

sh4wn:
Yeah I should reserve some time for this, but past month a lot of weekends were also already filled with activities. This week is also quite packed, but I think I'll have some time next week!

Eric89GXL:
@sh4wn what is the status here?

sh4wn:
I have a week off from the 19th of october till the 25th. I want to finish this pull request in that week!

sh4wn:
Ok, what would need to be done to get this merged?

I was thinking to remove all special code for sparse matrices, add some tests, and check the documentation. Any more ideas?

Eric89GXL:
IIRC getting the examples and tests working should do it. Then you or others can iterate in subsequent PRs.

Eric89GXL:
FYI I'm in the process of making Travis happy due to changes unrelated to this PR in #1110

Eric89GXL:
@sh4wn there are some conflicts with `master` now, can you fix them by rebase or merge?

sh4wn:
Hmm why do the tests do not find the networkx module in Travis?

QuLogic:
The minimal deps build does not (and should not) install `networkx`. You need to make it a soft dependency.

The full deps builds _do_ find `networkx`, but are simply failing.

sh4wn:
Hmm yes, maybe there are too many factors involved and testing it with a predefined seed.

Eric89GXL:
@sh4wn are you done working on this / should I take over?

sh4wn:
Yeah sorry, I'm completely swamped with study assignments right now, and I don't expect to change that in the coming months, sorry! If you would like to take over, please go ahead!

Eric89GXL:
@astrofrog do you think we should try to get this in the release? I *think* I might have time to tweak this to get it mergeable in the next week.

astrofrog:
@Eric89GXL - I would classify this as 'would be nice but not critical', so if it can't be wrapped up in the next couple of weeks, it's probably not worth holding up the release much longer?

QuLogic:
Did this get replaced?

Eric89GXL:
https://github.com/vispy/vispy/pull/1223

PR #1044, opened 2015-08-06
[ENH] Add Viridis sampled at 128 points

bollu:
I sampled `viridis` from the matplotlib commit pointed out by @QuLogic [available here](https://github.com/matplotlib/matplotlib/pull/4707/).

For some reason, VisPy freezes when there are too many samples in a `ColorArray`. I did the simplest thing I could think of, and just sampled the original data to get 128 evenly spaced colors.

Does anyone have any idea as to how to fix the freeze? If not, I'll open an issue.


Eric89GXL:
Do you know what the generating function is for `viridis`? If so, then we can actually program it into the shader and not have to stuff it in as an array.

bollu:
@Eric89GXL Like I mentioned on `gitter`, I don't. The only code samples I could find are just sampled versions of the actual function.

almarklein:
I would be surprised if we can write the viridis cmap with a function in GLSL... As I understood, they have this program with which you can create new colormaps subject to certain constraints. The output is probably just the 128 sample colors.

Eric89GXL:
Have you tried this? Does it work? If so, we might as well make it the default.

Can you also add a note in the code saying that you downsampled the original set of points (256?) down to 128 for speed reasons?

campagnola:
Why are we having performance issues with 256 points?

Eric89GXL:
Not sure, it freezes for @bollu.

One problem is that the colormap shaders are constructed using conditionals instead of using a 1D texture. We should enhance them to use 1D texture lookup instead, it should be much faster, and allow for more points.

campagnola:
+1

bollu:
@Eric89GXL Should I implement that? It seems simple enough (in principle) ? And sure, I'll add a note.

@campagnola - VisPy simply freezes if I use a large number of points. I'm not sure where, since I can't even `C-c` for a stack trace. The only other time I've seen such behaviour is with Requests, which suppresses all Exceptions (including `KeyboardInterrupt`, thereby rendering it unkillable).

almarklein:
Wait, the data is not upload via texture? How do you get the data there without one?

Eric89GXL:
Conditionals in the shader. It made some sense for small numbers of points, but was never really intended for long term / high numbers of points use.

almarklein:
Ah, you mean that we generate an if-statement for each element in the colormap? Wow :)  Conditionals mean code branching, which is bad for parallelism. I would not be surprised if things would be faster if we used multiplications instead (`val = float(condition) * x + (1.0 - float(condition)) * y`). But using a texture would be the proper way to go.

campagnola:
On Tue, Aug 11, 2015 at 6:08 AM, Almar Klein <notifications@github.com>
wrote:

> Ah, you mean that we generate an if-statement for each element in the
> colormap? Wow :) Conditionals mean code branching, which is bad for
> parallelism. I would not be surprised if things would be faster if we used
> multiplications instead (val = float(condition) * x + (1.0 -
> float(condition)) * y). But using a texture would be the proper way to go.
>
With conditionals, every core in an SMP must wait long enough for the
slowest comparison. Multiplication would be slightly slower because every
core would have to go through _every_ comparison, even if no cores are
going to use any of the remaining comparisons. A texture lookup should be
much faster all around--it's true that texture lookups take a long time,
but they parallelize well so you generally get good core utilization.

rougier:
Texture lookup are costly. If you can find a simple function (with no if statement) it'll faster.

bollu:
I added info about the origin of the colormap, plus the fact that it was sampled.

Eric89GXL:
Can you add a note in the code saying that we should instead change how the code is done to use a texture lookup method? That way we won't need to subsample their original set of points. Did they have 256 sample points in their code?

bollu:
Done. And yes, they have 256 samples.

Eric89GXL:
BTW did you try the colormap out in any examples? It might be worth making it our default.

njsmith:
There is code to generate that colormap, but it probably doesn't make sense to run it as a shader.

The hard part isn't the math -- I mean, the math is somewhat hairy, it's specified as a bezier with 6 control points in CAM02-UCS space, so you need [the control points](https://github.com/BIDS/colormap/blob/master/option_d.py#L6-L9) and the [bezier formulas](https://github.com/matplotlib/viscm/blob/master/viscm/bezierbuilder.py) to get the curve and then to convert CAM02-UCS -> CIECAM02 -> XYZ -> sRGB you need a [whole](https://github.com/njsmith/colorspacious/blob/master/colorspacious/luoetal2006.py) [bunch](https://github.com/njsmith/colorspacious/blob/master/colorspacious/luoetal2006.py) [of](https://github.com/njsmith/colorspacious/blob/master/colorspacious/ciecam02.py) [formulas](https://github.com/njsmith/colorspacious/blob/master/colorspacious/basics.py), but [someone claims to have already implemented that part in GLSL](https://github.com/libretro/common-shaders/blob/master/test/lab/misc/cam02-ucs-reverse.cg). And CAM02-UCS is a super-useful space to have around -- it lets you easily do things like pick evenly spaced line colors, or invent new colormaps (like the circular colormap example in our SciPy talk), or render one variable to lightness and one variable to hue and combine these in a perceptually sound way.

But for viridis, the hard part is that the Bezier is not intrinsically parametrized by arclength, yet we want our colormap points to be uniformly spaced. So you can't just look up the color value for some point like 0.3 in isolation; instead what we do is [discretize the curve, compute arclength along the the discrete curve, and use this to decide on each point's color](https://github.com/matplotlib/viscm/blob/master/viscm/bezierbuilder.py#L58-L80). And you definitely don't want to be redoing that for every fragment, so...

(Or maybe you could do even more math to analytically reparametrize the Bezier by arclength. Probably doable, actually, since it's just some polynomials. But I'd probably just go with the lookup table ;-).)

njsmith:
btw you might also want the other colormaps linked in that commit -- a lot of people seem to like `inferno` in particular :-)

Eric89GXL:
Thanks for the info @njsmith, good to know it would be too much work to do in the shaders :) I've also added `inferno` to the list of colormaps we should add.

bollu:
@Eric89GXL merge this, and re-implement colormaps in a separate PR? Should I open an issue?

bollu:
@njsmith - noob question: Don't Bezier curves get unstable at endpoints with upwards of 4-5 control points? (Isn't that one of the reasons why people prefer B-splines for longer chains)?

njsmith:
Beats me, I'm not an expert. This was an interactive editing tool, though, so instability wasn't really a concern -- if you don't like what you see, just drag the points around until you do :-).

Eric89GXL:
Sorry to make you keep changing a simple PR, but maybe include all 256 points in the file, but create the ColorMap with `_viridis_data[::2]`? That way the next PR won't need to modify `_viridis_data` at least.

bollu:
@Eric89GXL - Done. If CI passes, this is merge-able

Eric89GXL:
Great, thanks! Please open an issue about making Colormaps have 1D texture support for more efficient / extensible operation. And also put in there we should change our default to `viridis` once it's been Texture-ized.

bollu:
I think 1D textures aren't part of pure openGL2.1? (Just like 3D textures) IIRC? Neither does OpenGL ES 2.0

Eric89GXL:
We have a 3D texture emulation mode using 2D Textures under the hood, in theory we should be able to do the same thing for 1D textures

PR #1046, opened 2015-08-09
Fix installation

dseuss:
Some shaders were not copied during the installation process, which made some examples crash.


Eric89GXL:
@dseuss thanks for the fix. cc @sh4wn when we add non-`.py` files to the project, we need to make sure they get copied in `setup.py`

sh4wn:
Ah good one, totally forgot about that.

Issue #1017, opened 2015-07-23
ImageVisual: bicubic interpolation


kmuehlbauer:
Since `texture2D` only knows 'linear' and 'nearest' interpolation, I searched through VisPy for bicubic interpolation. Only thing I found is this: https://github.com/vispy/vispy/blob/edebe3931d9ae5ea734091848c793593d65a0e41/vispy/glsl/build-spatial-filters.py
I tried to implement the Bicubic function within the ImageVisual but failed. 

So I decided to implement the bicubic interpolation in the ImageVisual fragment shader from scratch (via 16 texture2D calls).  Before I finalize this I want to ask if there might be other Visuals or components within VisPy which would benefit and where would be the best location for implementation?

`texture2D` seems to be the right location, but I'm not really sure on how to implement the `bicubic` interpolation as an addition to the two existing 'hardware' interpolation.  

However, I will have a PR ready tomorrow (hopefully) as a starting point.

Any comments appreciated!

rougier:
Have a look at https://github.com/glumpy/glumpy/blob/master/examples/interpolations.py
and https://github.com/glumpy/glumpy/blob/master/glumpy/library/misc/spatial-filters.frag

This used the computed kernel from spatial-filters.py and is used to offer 16 image interpolation (including bicubic)

kmuehlbauer:
Yes, actually I tried to get this running properly, but it fails, and I didn't have the time yet to track this down. And I won't find time to further dive into this in the next two weeks, so this is no urgent matter. 

Here is what I got with the Bicubic implementation suggested by @rougier:
![test1](https://cloud.githubusercontent.com/assets/5821660/8879152/e0fce972-322f-11e5-9607-2c063b0bd7d5.jpg)

This is with the bicubic implementation from scratch:
![test2](https://cloud.githubusercontent.com/assets/5821660/8879146/d9f0f18c-322f-11e5-8f25-95f2d10ca416.jpg)

Maybe I did something wrong, feeding the kernel-data and image-data to the shader. I simply did this in ImageVisual:

```
self.shared_program['u_kernel'] = np.load(os.path.join(_data_dir, 'spatial-filters.npy'))
self.shared_program['u_data'] = self._texture
self.shared_program['u_shape'] = self._data.shape[1], self._data.shape[0]
self._texture.interpolation = 'nearest'
```
And in the fragment shader:
```
gl_FragColor =  Bicubic(u_data, u_shape, texcoord);
```

The implementation from scratch is like this:

[...]

PR #1048, opened 2015-08-11
FIX: add proper offset to spatial-filter fragment shader

kmuehlbauer:
This PR fixes an offset issue within the spatial-filters fragment shader. See discussion in #1017.


kmuehlbauer:
In #1017 @rougier suggested an example showing all interpolations in one application. It could go with this PR or separate. Suggestions?

rougier:
LGTM

Eric89GXL:
It seems simple enough to add to this PR if you have the time to do it

rougier:
This PR can be merged now. The other one may need more comments.

Eric89GXL:
Works for me. Thanks for the fix @kmuehlbauer

PR #1049, opened 2015-08-11
[WIP] Layout grids using Cassowary

bollu:
- [ ] Get baseline implementation working (just layout, no padding / margins)
- [ ] Implement margins
- [ ] Implement padding

- [ ] document code
- [ ] Regular test cases
- [ ] Pathological extreme test cases


bollu:
@Eric89GXL  - I'll have to change CI to install cassowary, correct? would it automatically install (since it's listed as a dependency)?

Eric89GXL:
Don't add a dependency, put it in `vispy/externals` like we do with other small packages. @QuLogic it should be straightforward to use your decoupling system, right?

QuLogic:
Evidently, since @bollu seems to have figured it out :wink: . However, you also need to add cassowary to Travis on the minimal deps build.

bollu:
@QuLogic - by that, you mean that Cassowary should be installed using `pip`?

Eric89GXL:
On the full build, yeah, because that's where we test the packaging version by deleting `_bundled` entirely. If you look at Travis you will see the related failure.

bollu:
@Eric89GXL - I edited the `.travis.yml` right, didn't I? The builds pass, except for the size check.

QuLogic:
There are also `flake8` errors.

bollu:
@QuLogic - ah, yes. I wasn't worried about my own code (I was just checking if I added cassowary correctly). However, it looks like the cassowary files under `vispy/ext/_bundled/cassowary/` fail `flake`. How do I remove these files from the `flake` check?

QuLogic:
`flake8` checks are in `vispy/testing/_runners.py`.

Eric89GXL:
You should be able to add an exception in there somewhere, there are
existing ones for other bundled libraries

.travis.yml
```diff
@@ -123,6 +123,7 @@ install:
       elif [ "${DEPS}" == "full" ]; then
         conda install --yes --quiet pyqt pyopengl;
         pip install -q pyglet;
+        pip install -q cassowary;
```
Eric89GXL: not here, this belongs on line 133

examples/basics/plotting/grid-basic.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
```
Eric89GXL: name it `grid_basic` instead of `grid-basic`. It's better to use underscores than hyphens in python filenames to facilitate importing

vispy/plot/plotwidget.py
```diff
@@ -42,9 +42,9 @@ def __init__(self, *args, **kwargs):
         super(PlotWidget, self).__init__(*args, **kwargs)
         self.grid = self.add_grid(spacing=0, margin=10)
         self.title = scene.Label("", font_size=16)
-        self.title.stretch = (1, 0.1)
-        self.grid.add_widget(self.title, row=0, col=2)
-        self.view = self.grid.add_view(row=1, col=2, border_color='grey')
```
Eric89GXL: you're going to put all these back eventually, right? I think the existing API of setting `stretch` is good, maybe we should replace `min_width` by `min_size` to be consistent, though.

vispy/scene/widgets/grid.py
```diff
@@ -10,6 +10,9 @@
 from .widget import Widget
 from ...util.np_backport import nanmean
 
+from vispy.ext.cassowary import (SimplexSolver, expression,
```
Eric89GXL: always do a relative import within a package, so `from ...ext.cassowary import ...`

vispy/scene/widgets/grid.py
```diff
@@ -108,16 +111,20 @@ def add_widget(self, widget=None, row=None, col=None, row_span=1,
...
-        
+
         # update stretch based on colspan/rowspan
         stretch = list(widget.stretch)
-        stretch[0] = col_span if stretch[0] is None else stretch[0]
-        stretch[1] = row_span if stretch[1] is None else stretch[1]
-        widget.stretch = stretch
-        
+        # stretch[0] = col_span if stretch[0] is None else stretch[0]
+        # stretch[1] = row_span if stretch[1] is None else stretch[1]
```
Eric89GXL: I think this was actually correct before...?

vispy/scene/widgets/widget.py
```diff
@@ -64,8 +64,11 @@ def __init__(self, pos=(0, 0), size=(10, 10), border_color=None,
         
         # layout interaction
         # todo: use Cassowary; see #277 
-        self._fixed_size = (None, None)
+        self._width_limits = [0, None]
+        self._height_limits = [0, None]
         self._stretch = (None, None)
+        self.var_w = self.var_h = None  # HACK
```
Eric89GXL: what is `var_w` and why is it public?

vispy/scene/widgets/widget.py
```diff
@@ -115,6 +118,79 @@ def size(self, s):
...
+        self.min_width = self.max_width = fixed_width
+
+    @property
+    def height(self):
+        """The height of this widget"""
+        return self._size[1]
+
+    @height.setter
+    def height(self, height):
+        self.size = (self._size[0], height)
```
Eric89GXL: I think if someone sets a `height`, what it should really do is set a fixed size

vispy/scene/widgets/widget.py
```diff
@@ -115,6 +118,79 @@ def size(self, s):
...
+    def max_height(self, max_height):
+        if max_height is None:
+            self._height_limits[1] = None
+            return
+
+        self._height_limits[1] = float(max_height)
+
+    def fixed_height(self, fixed_height):
+        self.min_height = fixed_height
+        self.max_height = fixed_height
```
Eric89GXL: I think the names should probably be `height_min`, `height_max`, `height_fixed`, `height_stretch`, and `width_min`, `width_max`, `width_fixed`, and `width_stretch`. Right now we have some of them split by `width` / `height` and others grouped (like `stretch`). We should be consistent.

vispy/scene/widgets/grid.py
```diff
@@ -184,62 +191,174 @@ def __repr__(self):
...
+    #         row, col, rspan, cspan, ch = self._grid_widgets[key]
 
-            # Translate the origin of the node to the corner of the area
-            # ch.transform.reset()
-            # ch.transform.translate((colstart[col], rowstart[row]))
-            ch.pos = colstart[col], rowstart[row]
+    #         # Translate the origin of the node to the corner of the area
+    #         # ch.transform.reset()
+    #         # ch.transform.translate((colstart[col], rowstart[row]))
+    #         ch.pos = colstart[col], rowstart[row]
```
Eric89GXL: just delete all this old stuff instead of commenting, you shouldn't need it

vispy/scene/widgets/widget.py
bollu: Like the line says, hack :) I was messing about with the constraint solving when I was trying this out. I'll rebase and squash these commits so it doesn't show up in history

vispy/scene/widgets/widget.py
bollu: Yeah, that makes sense. Will change

vispy/scene/widgets/widget.py
Eric89GXL: No need to squash

bollu:
@Eric89GXL Merging #1038 before this will help a lot, so that the colorbar code can use the new system as well.

Eric89GXL:
Good point, you should be good to rebase now

bollu:
@Eric89GXL - Travis seems to be acting super weird? It gives the errors as

```
E     File "/home/travis/build/vispy/vispy/vispy/scene/widgets/grid.py", line 133
E       for (key, val) in self._grid_widgets.items()
E         ^
E   SyntaxError: invalid syntax

```

where the total block of code is this:

```
        self._grid_widgets = {key: val
                              for (key, val) in self._grid_widgets.items()
                              if val[-1] != widget}

```

which is syntactically right, and gives no errors / flake violations locally. What's going on?

Eric89GXL:
that's on Python 2.6 probably, where you can't construct `dict` that way

bollu:
Ah, yeah, you're right. Hm, I didn't know that it doesn't work in python 2.6. 

I guess this is the right way to do it, then?

```
d = dict((key, value) for (key, value) in iterable)
```

Eric89GXL:
I can never remember the right way, but that might be it. Might be worth setting up a 2.6 VM on your end to test it out.

rossant:
is 2.6 support really mandatory?

Eric89GXL:
It's usually not too difficult to maintain, so I figure why not.

bollu:
@campagnola: what exactly is `spacing` in your implementation of the `Grid`? I'm not really able to understand in what units the spacing is specified

campagnola:
spacing is the number of pixels between cells in the grid.

Eric89GXL:
@bollu can you add an ASCII representation of a simple (2 col 1 row with
the top cell having colspan 2 maybe?) showing what the different terms
(spacing, padding, margin) mean? It would be very helpful for devs and
users alike.

bollu:
@Eric89GXL Will do.

I've been thinking about the current implementation, and it's different from @campagnola's previous implementation. Let's take an example to illustrate what I mean:

Consider that we want this layout:
```
+-----------+
|  label    |
+---+-------+
| y | None  |
+---+-------+
```

The `None` means that there's no widget there.

Now, in the old implementation, this would be

```
label: (row, col): (0, 0) | (rowspan, colspan): (1, 2)
y: (row, col): (1, 0) | (rowspan, colspan): (1, 1)
```

This will cause the `y` to be displayed properly, since it doesn't span the entire column. The old algorithm splits the total region into "unit" area (taking the stretch factor into account), and then assigns multiples of the "unit" area to the widgets.

Since the new implementation uses Cassowary to solve linear inequalities, it creates a system of equations that is supposed to correspond to the desired output.  The same system

```
label: (row, col): (0, 0) | (rowspan, colspan): (1, 2)
y: (row, col): (1, 0) | (rowspan, colspan): (1, 1)
```

would generate the system of equations:

```
w_label == total_w
w_y == total_w

... (plus other equations for height and stretch)
```

This is because the implementation goes row-by-row, summing up the widths of all elements in a row, and constraining that to be equal to the total width. So, the current algorithm will layout like this:

```
+-----------+
|  label    |
+-----------+
|     y     |
+-----------+
```

To prevent this from happening, we would need to create a "dummy" widget whose only purpose is to take up space (and add a term to the constraint equation) so that the `y` is laid out right. So, the input would need to be of the form:

```
label: (row, col): (0, 0) | (rowspan, colspan): (1, 2)
y: (row, col): (1, 0) | (rowspan, colspan): (1, 1)
dummy: (row, col): (1, 1) | (rowspan, colspan): (1, 1)
```

Which would yield the layout

```
+-------------+
|    label    |
+-----+-------+
|  y  | Dummy |
+-----+-------+
```

The difference was that before, the `None` was implicit, and the user didn't have to specify an extra `Widget`. Now, the user is forced to specify one just to ensure the right layout.

I can think of a solution: 

If a widget is added, such that it doesn't ensure "spanning" of the row/column, we add dummy widgets automatically to ensure spanning. These would get overwritten when a widget is added in the "dummy" positions.

However, this doesn't seem very clean.

Another approach would be to simply add "space filling" terms into the constraint equation. I'm not sure what's the right way to solve this.

I'm of the opinion that the way it works right now would be unintuitive to people who try and use the system for the first time, even though it makes sense once you grok it.

Thoughts?

bollu:
I found a pretty pathological counterexample to the soundness of the approach we designed
while refactoring `PlotWidget`.


Consider a grid like so, where `1, 2, 3, 4` are the 4 widgets placed


```
+-----------+---+ 
|    1      | 2 |
|           |   |
|---+-------+   |
|   | Empty |   |
| 3 +-------+---+
|   |      4    |
+---+-----------+
```

```
1 is at (0, 1)
2 is at (0, 1)
3 is at (1, 0)
4 is at (1, 1)
```

clearly:

```
1_w + 2_w == total_w
3_w + 4_w == total_w

1_h + 3_h == total_h
2_h + 4_h == total_h
```

So the constraints would be satisfied, but the situation is clearly not what someone would want, since it
creates a "hole" in the middle - it's counter intuitive and shouldn't behave that way.


However, there is no constraint that dictates that such a situation cannot occur, since
the grid doesn't really impose a "grid" structure, while @campagnola's implementation does.

However, this has the benefit of allowing columns or rows to _not_ line up, which is more flexible.

The solution to this would be to impose that they "line up", i.e

```
col_elements[0].pos_x == col_elements[1].pos_x ==  ... == col_elements[1].pos_x
row_elements[1].pos_y == row_elements[2].pos_y == ... == row_elements[2].pos_y
```

This should interact with `min/max_width` and `min/max_height` correctly. I think you can do that
by setting the constraint weights correctly, but I haven't put careful thought into it.

It's quite unfortunate that I only ran into this quite late, since this sort of thing
doesn't pop up with simplistic examples.

Eric89GXL:
I think the original formulation we talked about should still solve this, but you are thinking of one detail of the problem differently from how I did. In your examples, if a grid cell does not yet contain anything (`None` or `empty` above), then *you omit it from the equations*. I think this is incorrect.

Once you set up the grid cells, all cells should have a `stretch` of 1 and a `min=0, max=inf` type behavior. It doesn't matter if they include anything or not. And then once you fully specify the system, you're all set. Both of those examples above will work. This will require a couple minor changes to the plotting code to e.g. set a `max=0` to a cell we don't want to display until we need it (like the Y axis), but will work. I think this is the behavior users should expect -- If I construct a 3x3 grid and put one thing in the middle of it without saying anything else, I'd expect it to take up 1/3 of the space vertically and horizontally and be centered.

So this should make the process of grid solving simpler and take care of the problems you mention above. It also makes the user be more explicit about what they want. In case they do actually want to construct a 3x3 grid but have the middle cell take up the entire space, they will have to do more than just add one widget to make it happen, they will have to change constraints on the other cells. This seems good to me.

Make sense?

campagnola:
> This is because the implementation goes row-by-row, summing up the widths of all elements in a row, and constraining that to be equal to the total width. . .To prevent this from happening, we would need to create a "dummy" widget whose only purpose is to take up space (and add a term to the constraint equation) so that the y is laid out right. 

That sounds correct to me. Either the user needs to add a spacer to the layout if they want it explicitly empty, or they need to give the widgets appropriate geometry constraints such as "label.width = 2 * y.width" or "y.max_width = ...".

campagnola:
> If a widget is added, such that it doesn't ensure "spanning" of the row/column, we add dummy widgets automatically to ensure spanning. These would get overwritten when a widget is added in the "dummy" positions.

I think this kind of situation is rare enough that we probably don't need to worry about it. Most people will fill their layouts completely, or leave open cells that are otherwise constrained, like:
```
+---+-------------+
|   |             |
| y |   viewbox   |
|   |             |
+---+-------------+
|   |     x       |
+---+-------------+
```

campagnola:
```
+-----------+---+ 
|    1      | 2 |
|           |   |
|---+-------+   |
|   | Empty |   |
| 3 +-------+---+
|   |      4    |
+---+-----------+
```
In this example, widget 2 is at position (0, 2), not (0, 1). It is generally expected that rows / columns will be aligned--this is how most UI toolkit layouts work, and how HTML tables work. It's also very important for building form-like interfaces where you require widgets to be aligned across multiple rows. Consider a similar example:
```
+-----+-----------+
| 1   |     2     |
+-----+----+------+
|   3      |   4  |
+----------+------+
```
In this case we have two options:
1) All widgets are in a single grid that is 2 rows x 3 columns, and widgets 2/3 have colspan=2.
2) The layout is constructed from one top-level grid (with 1 column and 2 rows) and two nested grids (with 2 columns and 1 row each). In this case the columns would not need to be aligned between the two nested grids.

Eric89GXL:
I think the solution I described will handle these cases okay, do you
agree?

campagnola:
If a grid cell is empty, I think it should be omitted from the constraints. For example if I add widgets to columns 1 and 3 in a 1x3 grid, then column 2 should shrink to nothing unless some other constraint forces it to become larger. If a widget is added to column 2 then the column will expand, and if the widget is subsequently removed then the column will collapse. This is again how most UI toolkits and HTML tables work, so it's likely to be the expected behavior.

We can still keep the default stretch values for widgets, though--when a widget is added to the grid, its stretch values are automatically set equal to the row/column span values, unless otherwise specified.

Eric89GXL:
Ahh, it won't, but I thought of a modification that should solve it. Hang on.

Eric89GXL:
I think if we solve for the width of each column, and the height of each row in pixels, we should be all set. Then it's a matter of formulating our equations in terms of these, row by row and column by column.

Let's do this one, as a `2x3` with a couple of them having `rowspan=2`:
```
+-----+-----------+
| 1   |     2     |
+-----+----+------+
|   3      |   4  |
+----------+------+
```
The first equations as always will be that our widths and heights must add up:
```
c0_w + c1_w + c2_w = grid_w
r0_h + r1_h = grid_h
```
And by default, the stretches of the ones with rowspans=2 will have horizontal stretch 2. So the stretch equations would be:
```
c0_w / 1. = (c1_w + c2_w) / 2.
(c0_w + c1_w) / 2. = c2_w
r0_h = r1_h
```
So that will work correctly, I think.

Now let's go back to this one:
```
+-----------+---+ 
|    1      | 2 |
|           |   |
|---+-------+   |
|   | Empty |   |
| 3 +-------+---+
|   |      4    |
+---+-----------+
```
Again the mandatory standard ones about summation:
```
c0_w + c1_w + c2_w = grid_w
r0_h + r1_h + r2_h = grid_h
```
And from the stretches:
```
(c0_w + c1_w) / 2. = c2_w / 1. [from row 0]
c0_w / 1. = c1_w / 1. = c2_w / 1. [from row 1]
c0_w / 1. = (c1_w + c2_w) / 2. [from row 2]
... [similar things for the rows]
```

I think this should work for all the cases we've talked about. And the max/min widths of each cell, which I haven't put in here, can also easily be expressed in terms of row and column sizes as well.

This way, the job of the `cassowary` solver is simply to solve for the geometry of the grid, which is completely specified by the width of each column and height of each row. Then the actual equations for the system get constructed on the basis of which rows and columns are spanned, requested stretch, and min/max size. This seems like the simplest solution to me.

Eric89GXL:
@campagnola in this system, your shrink-to-nothing-if-empty preference would be a matter of what the default constraints for each unoccupied grid cell would be. For example, setting the stretch=0 until occupied by a widget, or probably better the max_size=0 until occupied, would do it.

Eric89GXL:
Although the `max_size=0` or `stretch=0` constraints would break this example from above (the `Empty` would collapse), we can always tell people to add a `Spacer` dummy widget to fix those cases I suppose:
```
+-----------+---+ 
|    1      | 2 |
|           |   |
|---+-------+   |
|   | Empty |   |
| 3 +-------+---+
|   |      4    |
+---+-----------+
```

campagnola:
That's my thought--it is rare that we actually want empty space in the layout; most people want the widgets to fill all of the available space. If you want an empty space, then it needs to be specified explicitly somehow.

campagnola:
@bollu, how does the Cassowary algorithm handle overconstrained systems?

bollu:
@campagnola - cassowary throws a "required failure" exception if the
constraints are unsolvable. Otherwise, redundant constraints it doesn't
care about.

bollu:
So, If I understand correctly, we divide the grid of dimension `(rowspan x colspan)` into cells of dimensions `(1 x 1)`. Then, we formulate all equations in terms of a grouping of the unit cells, with `max_width = 0, max_height = 0` for empty cells, so that they "collapse". Correct?

campagnola:
There are situations when an empty cell will not collapse, such as when there are items in adjacent cells forcing both the row height and column width to be nonzero. I think this means that the cell should not have any constraints, but I'm not quite clear on what the implementation should look like.

The overconstraint issue will take some work to deal with. I had assumed the algorithm would just provide a best possible fit rather than an exception.. For example: what happens when two adjacent widgets have different requirements for their height? What happens when the total minimum size of all widgets exceeds the grid size? How do we deal with widgets that have multiple size constraints, like minimum width + maximum width + stretch?

campagnola:
Ok, I am reading https://constraints.cs.washington.edu/solvers/cassowary-tochi.pdf
and it looks like we can have "preference" constraints, which solves the overconstraint issue.

bollu:
@campagnola - yes, constraints have different strengths - `WEAK`, `MEDIUM` and `REQUIRED`, along with weights to resolve two constraints of the same strength. So, the resolution is not hard assuming we know the order in which we want constraints to be solved. However, the question is that of figuring out the order of constraint solving.

Eric89GXL:
I think we had a plan in place for which constraints would get which
requirements by default, right? (I assume this is what you mean by
"order".) We should stick with that plan and see how well it works. Look
back through gitter and you should find it.

bollu:
Yep, that's right. What I said was aimed at @campagnola, since he was
wondering about resolution.

Eric89GXL:
Great -- @bollu can you come up with an example script that makes the layouts we've mentioned here, creating a new `vispy` canvas for each of them? It would be excellent as a demonstration of how the algorithm will behave. You can just fill the cells with a `Spacer` widget (to be created) with different background colors for demonstration purposes.

bollu:
@Eric89GXL on it!

bollu:
Also, you can see what the current implementation of `PlotWidget` looks like - it's slightly more verbose (since you have to specify a lot more), but it is also arguably cleaner _because_ it is explicit.

Also, do we have a consensus on wether cells should collapse or not? I'm still unclear about that.

Eric89GXL:
Yeah, just add a weak `max_size=0` until a grid cell is filled, that should take care of it

Eric89GXL:
You might also need to omit the non-filled cells from the `stretch` equivalencies, too

bollu:
The `stretch` would automatically get omitted because we set the `max_width` because of constraint strength :)

Eric89GXL:
I'm not sure what you mean by that, can you clarify? I think for all (filled) cells by default we want the max, min, and stretch constraints always active, and let `cassowary` deal with reconciling them. That would make the empty cells an exception, in that `stretch` constraints wouldn't be added for them. Are you saying you don't currently have `stretch` constraints active when `max/min` is set? I don't think that's what we want...

Eric89GXL:
Maybe instead of "omitted" what you mean is it will be *ignored* or be lower-precedence than the `max_size` one...?

bollu:
Nope, I decided to actually just skip adding the constraint anyway, because I _know_ I want the constraint to be ignored. So, if you setup a non-trivial `min_width`, `max_width`, `min_height`, `max_height`, the corresponding `stretch` equation isn't added. I'm sorry I wasn't clear on that.

I thought it would:

1. make the intent clear that we want stretch only if the defaults are being obeyed
2. reduce the workload on the constraint solver

Eric89GXL:
I don't think that will work for all/many use cases. You could omit `stretch` if `max=min` if you really want, but if `max > min` you actually still need it in there.

bollu:
@Eric89GXL - right, IIRC, I made it activate only when the two are equal (since there's no need for constraining in that case). If not, there's an implementation bug that I'll have to change.

Again, I apologise for not being specific :)

Eric89GXL:
Cool, let me know when you have that example going. Hopefully it's not too much work given the implementation you have. I think it's okay for `PlotWidget` to be more explicit about its constraints, too.

bollu:
[Link to the examples](https://gist.github.com/bollu/47f90d800ad9eb866847)

Eric89GXL:
Can you add them as part of this PR instead? They can all be part of the same example file, like `vispy/basics/scene/grid_layouts.py` or so. That way I can see the latest change set, too.

bollu:
Considering this example:

```
+-----+-----------+
| 1   |     2     |
+-----+----+------+
|   3      |   4  |
+----------+------+
```

We could _also_ choose to have 

```
1.pos = (0, 0) | 1.stretch = (1, 1)
2.pos = (1, 0) | 2.stretch  = (2, 1)
3.pos = (0, 1) | 3.stretch = (2, 1)
4.pos = (1, 1) | 4.stretch = (1, 1)
```

To clarify what you're saying, this should _not_ work, since each row/col cell should have the same edge boundaries, and this would not be respect in this case, correct?

The correct way to construct this layout would be to have 3 columns per row, and then have
```
1.colspan = 1
2.colspan = 2
3.colspan = 2
4.colspan = 1
```

So, well, I'm still confused - in the "right" implementation, what should a `2x2` grid with 
```
1.pos = (0, 0) | 1.stretch = (1, 1)
2.pos = (1, 0) | 2.stretch  = (2, 1)
3.pos = (0, 1) | 3.stretch = (2, 1)
4.pos = (1, 1) | 4.stretch = (1, 1)
```
yield? What should the placement be like in this case?

@campagnola  - I'm guessing it was to prevent this that your code

Eric89GXL:
I think with incompatible statements like that, we can live with whatever
cassowary provides. It'll probably make them equal size, which is fine.

Eric89GXL:
In other words, if users request something impossible it's probably okay
whatever the result is, at least for this first PR. Let's focus on getting
the correctly constructed / reasonable examples working first, and worry
about dealing with bad specifications once the good ones work the way we
want. I suspect whatever cassowary returns might be reasonable enough, and
we can always expand it later to give a nice warning or correct things if
something is impossible.

Do the properly constructed examples work yet?

It would also be good to add another simple example: a 2x2 grid of total
size e.g. 600x600 and equal stretches, where one of the cells has max size
400 and min size 200 in each dimension. This will allow people to see that
those take precedence over stretch, for example, and test that the system
works in that case, too.

Eric89GXL:
(And yes, that first example you mention above should not happen with a 2x2
grid, only a 2x3 with colspan=2 for two cells.)

bollu:
@Eric89GXL - Nope, I'm writing the code to divide it into grids right now.

campagnola:
> I think with incompatible statements like that, we can live with whatever
cassowary provides. It'll probably make them equal size, which is fine.

+1

Eric89GXL:
@bollu what is the status here?

bollu:
I've been pushing inconsistently onto this. The algorithm works now (AFAIK), but the layout algorithm changes mean that the `PlotWidget` needs to be slightly re-written. I'm booked till Saturday, unfortunately  :( I'll make time for this.

bollu:
@Eric89GXL: could you look at the `examples/basics/plotting/plot.py` and tell me how it looks? (In terms of both code and visually)

examples/basics/plotting/grid_holed.py
```diff
@@ -0,0 +1,44 @@
...
+# -----------------------------------------------------------------------------
+"""
++-----------+---+
+|    1      | 2 |
+|           |   |
+|---+-------+   |
+|   | Empty |   |
+| 3 +-------+---+
+|   |      4    |
++---+-----------+
```
Eric89GXL: I don't get this when I run it, do you?

examples/basics/plotting/grid_two_row_two_col_.py
```diff
@@ -0,0 +1,41 @@
...
+# -----------------------------------------------------------------------------
+# Copyright (c) 2015, Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
++-----+-----------+
+| 1   |     2     |
++-----+----+------+
+|   3      |   4  |
++----------+------+
```
Eric89GXL: I don't think this layout should be possible with two rows and two cols. Two rows and *three* cols, yes. But with two cols only, it implies the width of the column changes from row 0 to row 1, which is wrong.

examples/basics/plotting/grid_x_y_viewbox.py
```diff
@@ -0,0 +1,47 @@
...
+
+canvas = scene.SceneCanvas(keys='interactive')
+canvas.size = 600, 600
+canvas.show()
+
+grid = canvas.central_widget.add_grid()
+
+widget_y_axis = grid.add_widget(row=0, col=0)
+widget_y_axis.bgcolor = "#999999"
+widget_y_axis.stretch = (0.2, 1)
```
Eric89GXL: I think we'll typically want to do this by setting it to a fixed width (and fixed height for x-axis), not by setting the stretch.

Eric89GXL:
`plot.py` looks good! The y- and x-axes stay the same horizontal and vertical sizes, and the viewbox gets bigger, which is what I'd expect. Now onto fixing the other examples?

vispy/scene/widgets/grid.py
```diff
@@ -31,6 +31,21 @@ def __init__(self, spacing=6, **kwargs):
...
         self._n_added = 0
         self._default_class = ViewBox  # what to add when __getitem__ is used
+        self._solver = None
+        self._need_solver_recreate = True
+
+        # width and height of the Rect used to place child widgets
+        self.var_w = Variable("w_rect")
+        self.var_h = Variable("h_rect")
+        self.var_x = Variable("x_rect")
+        self.var_y = Variable("x_rect")
```
Eric89GXL: Why are you giving variables to the `x` and `y` positions? You should only need to solve for the column widths and row heights, as we've discussed a couple of times previously.

Eric89GXL: (the `x` and `y` positions can be derived from the column widths and row heights given the colspans, rowspans, and locations in the grid, so it's wrong to add them as variables)

examples/basics/plotting/grid_holed.py
bollu: Nope, this one is broken. I'll have to fix it. I was focusing on getting "actual" plotting working

examples/basics/plotting/grid_two_row_two_col_.py
bollu: Hm, you're right. I'm not "gridding" it. I'll fix this

vispy/scene/widgets/grid.py
bollu: Yep, I know :) It simplified life to implement it as such, so I was doing it that way. Since the x and y positions were purely equations without multiple constraints, it's as easy as taking the x and y constraint equations and just using them as code

Eric89GXL: I think it might be safer *not* to do that. Otherwise the next person to edit the code might think of them as actual variables that are solved for in some meaningful way by the solver, and try to make dependencies between them. Removing them makes it clearer what the solver is solving for, i.e. the row and column sizes, nothing more or less.

bollu: Huh, I suppose you're right. All right, I'll rewrite this.

examples/basics/plotting/grid_two_row_two_col_.py
bollu: Wait, question: If the column width does not change, then is the entire column "locked" to the same width? Didn't we agree on flexible widths without collapse?

Eric89GXL: No flexible column width. A given column should be the same width in every row.

bollu: The same for rows? Every element in the row must have the same height?

Eric89GXL: Yes. So the example shown above here most have 3 columns. To end up looking
that way.

Eric89GXL: But of course something that spans multiple rows will have the combined
height of the rows it is contained by.

bollu:
@Eric89GXL: Taking everything into account, I added some more code. All examples run as expected locally. The only thing left to do is to change the `x` and `y` computation. I'll change that once you okay this. 

Could you test it out and tell me what you think?

examples/basics/plotting/grid_holed.py
```diff
@@ -0,0 +1,47 @@
...
+canvas.size = 600, 600
+canvas.bgcolor = "#000000"
+canvas.show()
+
+grid = canvas.central_widget.add_grid()
+
+
+widget_top_left = grid.add_widget(row=0, col=0, col_span=2)
+widget_top_left.bgcolor = "#ffffff"
+widget_top_left.stretch = (2, 1)
```
Eric89GXL: you shouldn't need to manually set the stretch, by default stretch should equal the number of rows (vstretch) or columns (hstretch), so initializing with a col_span should do it automatically

examples/basics/plotting/grid_uneven_col.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
```
Eric89GXL: I don't thnik we really need this example given that we have the "hole" one above

examples/basics/plotting/grid_x_y_viewbox.py
```diff
@@ -0,0 +1,47 @@
...
+# -----------------------------------------------------------------------------
+
+"""
++---+-------------+
+|   |             |
+| y |   viewbox   |
+|   |             |
++---+-------------+
+| sp|     x       |
++---+-------------+
```
Eric89GXL: Looks good. But for some strange reason on my system it doesn't always recalculate / reposition if I resize the window multiple times in succession. Is it always reliable for you? Could just be an event-system bug, so probably not a show-stopper.

examples/basics/plotting/grid_x_y_viewbox.py
Eric89GXL: please change this

examples/basics/plotting/plot.py
```diff
@@ -11,13 +11,13 @@
 
 fig = vp.Fig(size=(600, 500), show=False)
 
-# Plot the target square wave shape
```
Eric89GXL: This example does not work properly for me, does it work for you?

vispy/plot/plotwidget.py
```diff
@@ -359,72 +371,3 @@ def surface(self, zdata, **kwargs):
         self.view.add(surf)
         self.view.camera.set_range()
         return surf
-
-    def colorbar(self, cmap, position="right",
```
Eric89GXL: Did you mean to delete all this?

examples/basics/plotting/grid_uneven_col.py
bollu: I suppose. I want to keep it around to drive home the difference between `stretch` versus `row_span` and `col_span`

examples/basics/plotting/grid_holed.py
bollu: oh, right :) I got rid of that part in the code to reduce the number of things to suspect. I'll add that back again

vispy/scene/widgets/widget.py
```diff
@@ -63,9 +63,13 @@ def __init__(self, pos=(0, 0), size=(10, 10), border_color=None,
         self._size = 100, 100
 
         # layout interaction
-        # todo: use Cassowary; see #277
-        self._fixed_size = (None, None)
-        self._stretch = (None, None)
+        self._width_limits = (0, None)
+        self._height_limits = (0, None)
+        self._stretch = (1, 1)
```
Eric89GXL: you can make `stretch` a `kwarg`, that way when the `Widget` is created by the grid system it can set it appropriately based on the stretch

examples/basics/plotting/grid_uneven_col.py
Eric89GXL: but if you modify `stretch` to auto-depend (by default) on `colspan` and `rowspan`, then you won't need `stretch` in either example, only `*_span`

examples/basics/plotting/grid_x_y_viewbox.py
bollu: Try it after the new push

examples/basics/plotting/plot.py
bollu: Try it after the new push :) I accidentally rewrote correct code.

vispy/plot/plotwidget.py
bollu: Nope, I'm going to re-integrate this once I know that baseline `Plot` works

vispy/scene/widgets/widget.py
bollu: Oh, right. That's nicer aesthetically. Will do

bollu:
@Eric89GXL could you try again, please?

examples/basics/plotting/grid_holed.py
```diff
@@ -0,0 +1,42 @@
...
+
+canvas = scene.SceneCanvas(keys='interactive')
+canvas.size = 600, 600
+canvas.bgcolor = "#000000"
+canvas.show()
+
+grid = canvas.central_widget.add_grid()
+
+
+widget_top_left = grid.add_widget(row=0, col=0, col_span=2)
```
Eric89GXL: can you add (or use if it exists already) the `bgcolor` kwarg? It would make this example even simpler / shorter

examples/basics/plotting/grid_x_y_viewbox.py
Eric89GXL: still pending

examples/basics/plotting/plot.py
```diff
@@ -35,25 +35,21 @@
...
 labels = [vp.Label('n=%d' % plot_nvals[i], color=colors[i], anchor_x='left')
           for i in range(len(plot_nvals))]
 boxgrid = box.add_grid()
 for i, label in enumerate(labels):
-    boxgrid.add_widget(label, row=i, col=0)
-hspacer2 = vp.Widget()
-hspacer2.stretch = (4, 1)
-boxgrid.add_widget(hspacer2, row=0, col=1)
+    label_widget = boxgrid.add_widget(label, row=i, col=0)
+
```
Eric89GXL: try resizing the plot to something tiny. I get an error. We should probably make the plotting area of the widget have min size (1, 1) instead of (0, 0). It's a silly size anyway, but that way we might avoid an error.

Eric89GXL:
Otherwise examples look reasonable

examples/basics/plotting/plot.py
bollu: ah, right. I'll fix that

bollu:
What about the speed? The way it creates constraints is naive right now. I think at least some of it can be pruned (arguably at the cost of clarity)

Eric89GXL:
The speed seems fine on my system. Let's wait to see if someone has a problem with it before reducing clarity of the code.

bollu:
`test_perspective_render` is failing under AppVeyour and Travis. I'm pretty sure I didn't touch these. Do I merge upstream changes?

Eric89GXL:
I think you should be able to rebase, no?

Eric89GXL:
It's also possible that some small difference in `scene` caused by the new layout (e.g., being off by a pixel or two) is just barely driving the test out of configuration. Can you run the test locally and see what the output is like? There is also a test auditing mode, but I can't remember offhand what the command-line option is for it.

bollu:
Sure :) I remember adding a comment about auditing mode so it's easier to
find

And yeah, rebasing should work in this case

bollu:
@Eric89GXL - I'm sorry for having this drag, but, is it possible to merge this once I fix the tests? I think that will provide me some momentum so I can split remaining work (eg. re-integrating colorbars) into separate pull requests.

examples/basics/plotting/grid_x_y_viewbox.py
campagnola: +1 on this--the ability to mix stretch-based and unit-based constraints is one of the primary motivations for including Cassowary. It'd be great to see that in action.

campagnola:
This is looking good, @bollu. Don't worry about rushing, this is an important one  :)

bollu:
@campagnola - Haha, thanks for the motivation :) this needs some final spit and polish to be ready to be merged IMO, as long as it hasn't diverged too much from `master`. I'm hoping to have this done by new year's!

vispy/scene/widgets/grid.py
```diff
@@ -183,63 +288,238 @@ def __repr__(self):
...
+        for hs in height_grid:
+            for h in hs:
+                solver.add_constraint(h >= 0, strength=REQUIRED)
+
+        for (_, val) in grid_widgets.items():
+            (y, x, ys, xs, widget) = val
+
+            for ws in width_grid[y:y+ys]:
+                total_w = np.sum(ws[x:x+xs])
+                # assert(total_w is not None)
```
bollu: this is completely unexplainable by me - *adding* this assert somehow makes `examples/basics/plotting/plot.py` work, *without* tripping the assert itself.

However, without the assert,`plot.py` fails with a `NoneType` error. 

I don't understand how the `assert` could doesn't even throw an `AssertionError` or something. It simply works for no discernible reason. The only thing I can think of is some sort of race condition, but I can't see anything obvious like that in this case.

AFAIK, `np.sum` returns `0` if the array is empty so that shouldn't be a problem. [Docs here](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.sum.html)

bollu: ping @Eric89GXL  @campagnola  with respect to the weird error. Can you reproduce it?

This is pretty much the last thing other than some housekeeping stopping this merge as far as I can see

rougier: Got the same problem but it is random.

Fix seems to be to replace line 385 in simplex_solver:
```
if e != None
```
by
```
if e is not None:
```

bollu: Hm, thanks. So it's a `cassowary` error.
I guess I'll send a PR to `cassowary` about this .

Thanks for the help @rougier!

bollu: any explanation as to why it works?

rougier: Yes, ```!=``` requires a comparator ```__eq__``` while the ``ìs not``` does not.

bollu: Ah, I see. `is` compares identity, correct? (in the sense of "is it the same object")?

rougier: Yes, and since `None` is unique, this is the way to "compare" to `None`

Eric89GXL:
@bollu if you rebase this on `upstream/master` you should have the image tester tweaks you needed

bollu:
Thanks for merging the other branch :) will rebase off of master

bollu:
So, I was checking out the test failures, and it's being caused because of the change in the layout algorithm. For example:
screen shot 2016-01-04 at 2 40 15 pm

Even though the output is the same, the layout is different since the test is using `grid` for the tests.

The output *looks* the same, but I'm not qualified enough to judge.

So either I could spend time replicating the original exactly, or I could pass the test based on the fact that they're actually passing.

What do I do in this case?

Eric89GXL:
Functionally this implies your PR changes the default behavior/look/spacing
of (at least) Grid. Can you set the defaults for grid such that it passes /
has the old look? Or are the effective new defaults superior in some
convincing way?

Eric89GXL:
Actually it looks like the original doesn't occupy the entire Canvas, is
that correct? That seems inferior to your PR's output to me. If so, we'll
have to decide of we want to make this backward incompatible change (could
break people's existing code).

bollu:
Yep. The original doesn't. I'll actually check and see how much code this
breaks.

bollu:
@Eric89GXL - do I keep the "new" implementation that uses all of the rendering area?

Eric89GXL:
How many tests does it break? If it's only a few, we can hope it won't break that many people's code. Although maybe it can be considered a bug fix to use the entire area...

@campagnola any opinion?

bollu:
@Eric89GXL - AFAICT only `test_perspective_render` is failing from looking at the logs

Eric89GXL:
That seems reasonable to me. I'd say go ahead and try to finish up the PR,
taking that as a bug fix. Even if it breaks a little bit of other people's
code the workaround should be simple of the system works properly.

bollu:
@campagnola - do you think it's alright to break some code for "the greater good"? :)

@Eric89GXL - I'll cleanup the code a little and ping you for a code review? I'm as eager to get this merged into the mainline as you are

Eric89GXL:
Sounds good to me, let me know when it's ready

vispy/scene/widgets/widget.py
```diff
@@ -63,9 +63,13 @@ def __init__(self, pos=(0, 0), size=(10, 10), border_color=None,
         self._size = 100, 100
 
         # layout interaction
-        # todo: use Cassowary; see #277
-        self._fixed_size = (None, None)
+        self._width_limits = (1, None)
```
bollu: keeping this as a tuple introduces a decent number of annoyances, but I chose the annoyance over having two separate fields (`min` and `max`). Should I change this? I'm unable to decide

Eric89GXL: if it's private you can make it a list and it will probably be much less annoying because it's mutable

campagnola:
I think now is a good time to break with the old grid behavior, so long as nothing new can be considered a regression.

vispy/scene/widgets/widget.py
```diff
@@ -63,9 +63,13 @@ def __init__(self, pos=(0, 0), size=(10, 10), border_color=None,
...
+        self._height_limits = (1, None)
         self._stretch = (None, None)
+        # used by the constraint solver
+        # in grid - these will be Cassowary Variable
+        self.var_w = self.var_h = None
+        self.var_x = self.var_y = None
```
campagnola: Should these be private attributes?

examples/basics/plotting/grid_x_y_viewbox.py
campagnola: I just tried setting:

    widget_y_axis.width_min = 50
    widget_y_axis.width_max = 50
    widget_x_axis.height_min = 50
    widget_x_axis.height_max = 50

..and it seems to work well :)  Can you write that into the example in place of `x_axis.stretch` and `y_axis.stretch`?

examples/basics/plotting/plot.py
campagnola: Bump

bollu: Ah yes, making them private was on my todo list :)

There's a contract that I'm not sure how to express properly - the
GridWidget uses the var_* of its children, even though the var_* are
private.

In C++, I could use "friend" to express that relationship. I'm not sure
what the right was is in Python

Eric89GXL: It's okay for `GridWidget` to access private attributes of its children. The public/private distinction via leading `_` in our classes is mostly to distinguish what should be publicly accessible / minimally change functionality, versus things that only devs / other classes should access, and can be changed at will during development (assuming it's done correctly).

bollu:
Any recommendations on how to test the new API? As in, do you have any particular test cases in mind that you'd like to see?

Eric89GXL:
I think the examples that we've talked about already are pretty good test cases. Plus ensuring the existing examples still look the same (or preferably better) would be good.

bollu:
@Eric89GXL - did I screw up a rebase or some such? My build is failing at `test_arrow_transform_draw` at a non-compiling vertex shader. I'll check it again once I wake up by checking my history, but I'm off to bed now and was wondering if you could take a look if you're free :)

Eric89GXL:
It doesn't look like you pushed a rebase. Did you? At least of a week ago (https://github.com/vispy/vispy/pull/1143), there was no problem. You could try enabling Travis-CI builds for your own fork and pushing a dummy branch with a trivial change (relative to `upstream/master`) and see if it passes.

bollu:
Ah yes, that's a good idea.
I'll try that tonight

bollu:
@Eric89GXL - I'm currently happy with where the interface is right now, though I'd like to have tons more testing of the code. I have a feeling the constraints will break in certain situations, and I want to rewrite the cassowary implementation to be faster (and less exception-throwy), but this works for now. It's also decently fast (but I'd like it to be faster).

Could you please add a new tag to the `HEAD` of `vispy/test-data`? (the data that fixed the `test_perspective_render` has been pushed, but the testing rig doesn't download it due to the lack of a tag).

once this passes CI, can we have a code review @campagnola ?

Eric89GXL:
Yes please do not optimize speed for now, subsequent PR for that is better. `test-data-5` tag created.

bollu:
All the python27 builds succeeded, while it looks like the python34 builds will fail. I looked at the logs, but it looks like weird things that shouldn't fail are failing..?

bollu:
the appveyour build is failing because of `test_arrow_transform_draw`. I'm not too sure what's happening, but it is completely orthogonal to my changes

vispy/ext/cassowary.py
```diff
@@ -0,0 +1,12 @@
...
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+
+"""
+Handle loading cassowary package from system or from the bundled copy
+"""
+
+try:
+    from ._bundled.cassowary import *  # noqa
+except ImportError:
+    import cassowary  # noqa
```
Eric89GXL: these two import statements are not equivalent, you want `from cassowary import *` on this line

Eric89GXL:
You can add a skip for the failing AppVeyor test with something like:
```
# XXX fix AppVeyor eventually
if  os.getenv('APPVEYOR', '') == 'true':
    raise SkipTest('AppVeyor has unknown failure')
```

examples/basics/plotting/grid_uneven_col.py
```diff
@@ -0,0 +1,44 @@
...
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
+columns ->
+  0     1    2
+0 +-----+----+------+
+  | 1   |  2 |  2   |
+1 +-----+----+------+
+  |   3 |  3 |   4  |
+2 +-----+----+------+
```
Eric89GXL: if "columns" goes 0,1,2 then "rows" should only go 0,1

Also the numbers should probably be centered on the rows and columns

And the "1" in the upper left does not line up with the "3" in the lower left

vispy/scene/widgets/grid.py
```diff
@@ -183,63 +258,239 @@ def __repr__(self):
...
-    def _update_child_widgets(self):
-        # Resize all widgets in this grid to share space.
-        n_rows, n_cols = self.grid_size
-        if n_rows == 0 or n_cols == 0:
-            return
+    @staticmethod
+    def _add_total_width_constraints(solver, width_grid, _var_w):
+        for (y, ws) in enumerate(width_grid):
+            width_expr = expression.Expression()
+            for (x, w) in enumerate(ws):
```
Eric89GXL: don't need to enumerate, you don't use the vars

vispy/scene/widgets/grid.py
```diff
@@ -183,63 +258,239 @@ def __repr__(self):
...
+    @staticmethod
+    def _add_stretch_constraints(solver, width_grid, height_grid,
+                                 grid_widgets, widget_grid):
+        xmax = len(height_grid)
+        ymax = len(width_grid)
+
+        stretch_widths = [[] for _ in range(0, ymax)]
+        stretch_heights = [[] for _ in range(0, xmax)]
+
+        for (_, val) in grid_widgets.items():
```
Eric89GXL: cleaner as `for val in grid_widgets.values()`

and you can even do `for y, x, ys, xs, widget in grid_widgets.values():` and save a line below

vispy/visuals/colorbar.py
```diff
@@ -593,6 +593,7 @@ def label(self):
 
     @label.setter
     def label(self, label):
+        assert()
```
Eric89GXL: ?

examples/basics/plotting/grid_uneven_col.py
bollu: Ah yeah, I should cleanup the docstrings. Will do. I have to center the labels as well

vispy/scene/widgets/grid.py
bollu: Yeah, this was with respect to the older code where I was using the `y` and `x` to position.

vispy/scene/widgets/grid.py
bollu: I decided against destructuring it in the same line because it was way too much visual noise in the same line for my taste. I could change t if you want me to.

vispy/visuals/colorbar.py
bollu: regression on my part. Will fix.

I'll make a separate PR adding a new test for this - this regression should have been caught, I'm not sure how it wasn't.

bollu:
What do you feel about the performance of the constraint solving? is it acceptable? were you able to find examples that broke it?

I'd love to have a collection of "extreme" layouts just to know where the breaking points of the implementation are

bollu:
Any idea why it fails at this?

```
Py3.4, DEPS=minimal, TEST=standard

The command "if [ "${DEPS}" == "minimal" ]; then echo "Size difference ${SIZE_DIFF} kB"; if git log --format=%B -n 2 | grep -q "\[size skip\]"; then echo "Skipping size test"; else test ${SIZE_DIFF} -lt 100; fi; fi;" exited with 1.

```

Eric89GXL:
You have so many changed lines and/or commits that it causes a > 100 kB increase in repo size. It's acceptable for this PR.

I wouldn't worry about all the extreme cases right now. If we can get the examples we've listed already to work properly we should merge. It will never be perfect, and in the meantime, people can't use it (or extend it themselves) very easily, so we should try to get it in instead.

I'll try to check the performance soon, but last time I tried (months ago) it was at least acceptable. So assuming you didn't do anything to slow it down significantly recently it should be fine. Again performance improvements can and should probably be done in subsequent PRs where the changes and potential breakages are easier to track.

bollu:
no, performance hasn't degraded since then - as far as I've noticed at least. It actually got a little faster by making most constraints stricter.

Also, the AppVeyor tests are behaving bizarrely - it looks like every arrow related test is failing for some reason. I'm not sure why

bollu:
@Eric89GXL - All builds pass, except for the size check :)

Eric89GXL:
Great! I'll go ahead and try the examples and merge if it's all working well. We can iterate in subsequent PRs.

examples/basics/plotting/grid_holed.py
```diff
@@ -0,0 +1,43 @@
...
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
++---+-------+---+
+| 1 |   1   | 2 |
+|---+-------+---+
+| 3 | Empty | 2 |
++---+-------+---+
+| 3 |   4   | 4 |
++---+-------+---+
```
Eric89GXL: This one collapses to be effectively a 2x2. I guess that's expected based on our constraint defaults? I originally expected the solution to be as depicted here, where the `stretch` of the 1x2 and 2x1 cells pushes it to have some empty space in the middle.

examples/basics/plotting/grid_uneven_col.py
```diff
@@ -0,0 +1,40 @@
...
+# Copyright (c) 2015, Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
+   c0    c1    c2
+r0 +-----+-----+-----+
+   |  1  |  2  |  2  |
+r1 +-----+-----+-----+
+   |  3  |  3  |  4  |
+   +-----+-----+-----+
```
Eric89GXL: okay there is definitely some bug here. I get a 2x2 again. This one should definitely be as depicted above, where 2 and 3 are twice as wide as 1 and 4. It suggests the `stretch` parameters is either not set properly, or not respected properly during constraint solving, because 2 and 3 should have twice as much stretch / preferred width as 1 and 4.

bollu:
Cool :) It's a relief to have this finally merged with master

examples/basics/plotting/grid_x_y_viewbox.py
```diff
@@ -0,0 +1,48 @@
...
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+
+"""
++---+-------------+
+|   |             |
+| y |   viewbox   |
+|   |             |
++---+-------------+
+| sp|     x       |
```
Eric89GXL: Now this works, but I found a new bug: try making the window small enough that it's less than 50x50. We get some nasty error. That shouldn't happen. We might need to do a `try` clause if there isn't a cleaner way.

examples/basics/plotting/ipython_fig_playground.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2015, Vispy Development Team.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+"""
+Bare bones plotting region that can be used with the python
+interpreter as a playground.
+
+Run with
+python -i ipython-base.py
```
Eric89GXL: docstring is wrong

examples/basics/plotting/plot.py
```diff
@@ -11,13 +11,13 @@
 
 fig = vp.Fig(size=(600, 500), show=False)
 
-# Plot the target square wave shape 
+# Plot the target square wave shape
```
Eric89GXL: I get the left side of the ylabel text cut off, do you?

screenshot from 2016-02-18 14-15-31

Otherwise it works well

examples/basics/plotting/plot_colorbars.py
```diff
@@ -0,0 +1,16 @@
...
+"""
+Show use of colorbars in plot
+"""
+import vispy.plot as vp
+
+fig = vp.Fig(size=(600, 500), show=False)
+plotwidget = fig[0, 0]
+
+plotwidget.plot([(x, x**2) for x in range(0, 100)])
+plotwidget.colorbar(position="top", cmap="viridis")
```
Eric89GXL: argh, I get an error with `cmap=viridis` because of `IF statement nested too deeply`, but I think we already have an issue open (and maybe you started a PR?) for fixing it using texture lookups

Eric89GXL: for now, though, can you change it to something that should work for everyone?

Eric89GXL:
Can't merge yet unfortunately, examples still don't work as expected :(

examples/basics/plotting/grid_uneven_col.py
Eric89GXL: In other words, a `col_span=2` should set the horizontal stretch = 2. Does that not happen automatically right now?

examples/basics/plotting/plot_colorbars.py
bollu: yes, I'll change it to the default colormap - I changed it to `viridis` because it looks cool

examples/basics/plotting/plot.py
bollu: hm, nope. I think it's because there is no padding on the left side - I'll add some.

examples/basics/plotting/grid_x_y_viewbox.py
bollu: Yes, this was what I meant by saying that Cassowary breaks too easily :/ I'm not a fan of completely black-holing every Cassowary exception though. Is there a saner alternative?

Eric89GXL: One option is to set the min/max size constraints to be the level just below mandatory. Do you think that would work? They would need to be the only things on that level, though, so not sure.

Actually, can you add to the docstring somewhere which level each constraint is placed? It's important to say it somewhere for the next person who goes to make changes.

bollu: yes, that may work - but from what I've seen (empirically), `REQUIRED` constraints are faster to solve than anything else - so reducing the strength could screw up the performance. It's worth a shot.

And all right, will document it :)

bollu:
can you try it now? The performance sucks though :/

Eric89GXL:
Will try this week. Is that because you changed the priority levels of the
min/max?

bollu:
Nope, I added a `try/catch` block around the entire solver code. It feels *ugly* to reduce constraint strengths and mess with their semantic meaning

Eric89GXL:
Did it need to be around the entire thing? Would be better just to catch
the falling lines if there are a small number of potential ones.

bollu:
@Eric89GXL better for performance or neatness? I found it elegant to have a single catch block, since the cassowary updates are incremental

Eric89GXL:
Better from a safe-programming perspective. It's best to catch only the lines you expect to cause an error so that you don't accidentally catch errors that you don't mean to. For a simple example, if you put a ton of code in a `try/except ValueError`, and somewhere inside there your code does something like `int('foo')` you will *always* hit an exception. What you meant to catch were probably errors related to something else, but because you had a bunch of lines all in the same try/except, you caught something else you didn't mean to, and have no great way to detect it.

vispy/ext/_bundled/cassowary/tableau.py
```diff
@@ -0,0 +1,109 @@
...
+        varset = self.columns[oldVar]
+        for v in varset:
+            row = self.rows[v]
+            row.substitute_out(oldVar, expr, v, self)
+            if v.is_restricted and row.constant < 0.0:
+                self.infeasible_rows.add(v)
+
+        if oldVar.is_external:
+            self.external_rows.add(oldVar)
+            try:
```
Eric89GXL: For each try/except clauses, please write a comment explaining why it can happen. Otherwise it's difficult for the next person to make changes.

vispy/scene/widgets/grid.py
```diff
@@ -183,63 +258,242 @@ def __repr__(self):
...
+                if row == 0:
+                    y = 0
+                else:
+                    y = np.sum(value_vectorized(self._height_grid[col][0:row]))
+
+                widget.size = (width, height)
+                widget.pos = (x, y)
+
+        except RequiredFailure as e:
+                print("cassowary failure\n%s" % e)
```
Eric89GXL: We probably don't really want to print the errors, we should catch them and do something intelligent with them. We could probably just say in the docs that un-solvable systems will lead to no change in layout...?

bollu:
@Eric89GXL - I would up weakening all constraints to `STRONG` to make sure that cassowary *never* fails (yes this is ugly, but I'm not sure a better compromise exists). Also, I fixed a bug in `axis` that would cause it to error out if the dimensions of the plot is too small.

Do you think you have time test out the changes sometime?

Eric89GXL:
Do the max/min size constraints still get set to a level that is higher (more important) than any other constraint? That would be ideal.

bollu:
Yes, they are still the strongest of all constraints

examples/basics/plotting/grid_holed.py
Eric89GXL: This is still broken

examples/basics/plotting/grid_uneven_col.py
Eric89GXL: still broken

Eric89GXL:
> Do you think you have time test out the changes sometime?

It looks like not all of the previous comments have been addressed

bollu:
@Eric89GXL - whoops, you're right, I'd missed the regression. Fixed it now. Anything else that I missed?

Eric89GXL:
You didn't push any commits...?

Eric89GXL:
(To see if you missed anything, it's easiest to look through the changes on GitHub and look at comments that are still visible / haven't been auto-collapsed, and make sure you've taken care of them)

bollu:
@Eric89GXL - I did push [this commit] didn't I?(https://github.com/vispy/vispy/pull/1049/commits/3d30c411d2a867d6dfa8786d108af8dc6fc16646)

(make initial stretch to None, not 1) was the commit message, `3d30c411d2a867d6dfa8786d108af8dc6fc16646` was the hash

Eric89GXL:
It must have been delayed somehow -- `git fetch bollu` didn't update locally and no commit appeared on GitHub, even though it's there now. Strange.

vispy/ext/_bundled/cassowary/tableau.py
Eric89GXL: this comment hasn't been addressed

Eric89GXL:
The examples now work, which is great!

Could you double-check that all previous comments have been addressed? I found one that hadn't.

vispy/ext/_bundled/cassowary/tableau.py
bollu: This is the wrong file, because this is internal cassowary code, right?

vispy/scene/widgets/grid.py
```diff
@@ -109,15 +128,71 @@ def add_widget(self, widget=None, row=None, col=None, row_span=1,
...
         self._next_cell = [row, col+col_span]
 
+        widget._var_w = Variable("w-(row: %s | col: %s)" % (row, col))
+        widget._var_h = Variable("h-(row: %s | col: %s)" % (row, col))
+
         # update stretch based on colspan/rowspan
+        # usually, if you make something consume more grids or columns,
+        # you also want it to actually *take it up*, ratio wise.
+        # otherwise, it will never *use* the extra rows and columns,
+        # thereby collapsing the extras to 0.
```
bollu: I've added the comments about Cassowary here

vispy/ext/_bundled/cassowary/simplex_solver.py
```diff
@@ -0,0 +1,612 @@
...
+        z_row = self.rows[self.objective]
+
+        e_vars = self.error_vars.get(cn)
+        # print("e_vars ==", e_vars)
+        if e_vars:
+            for cv in e_vars:
+                try:
+                    z_row.add_expression(self.rows[cv], -cn.weight * cn.strength, self.objective, self)
+                    # print('add expression', self.rows[cv])
+                except KeyError:
```
Eric89GXL: when does this happen?

vispy/ext/_bundled/cassowary/simplex_solver.py
```diff
@@ -0,0 +1,612 @@
...
+            return
+
+        try:
+            for basic_var in self.columns[minus_error_var]:
+                expr = self.rows[basic_var]
+                c = expr.coefficient_for(minus_error_var)
+                expr.constant = expr.constant + (c * delta)
+                if basic_var.is_restricted and expr.constant < 0:
+                    self.infeasible_rows.add(basic_var)
+        except KeyError:
```
Eric89GXL: When does this happen?

vispy/ext/_bundled/cassowary/simplex_solver.py
Eric89GXL: Ahh never mind, this is inside `cassowary`, as you say below

vispy/ext/_bundled/cassowary/simplex_solver.py
Eric89GXL: same here, ignore

vispy/ext/_bundled/cassowary/tableau.py
Eric89GXL: yeah it's fine

vispy/ext/_bundled/cassowary/simplex_solver.py
bollu: AFAIK, it happens if you try to remove a constraint that hasn't been added (or has already been removed)

Eric89GXL:
LGTM, I'll go ahead and merge once the CIs come back happy. Thanks for sticking with it @bollu

bollu:
@Eric89GXL - heh, np. My pleasure :) feels good to see this (almost) merged!

bollu:
@Eric89GXL - CI passes. The only failing test is the size test, which you'd said was okay, correct?

```
The command "if [ "${DEPS}" == "minimal" ]; then echo "Size difference ${SIZE_DIFF} kB"; if git log --format=%B -n 2 | grep -q "\[size skip\]"; then echo "Skipping size test"; else test ${SIZE_DIFF} -lt 100; fi; fi;" exited with 1.
```

Eric89GXL:
Yep, that one failing is fine. Thanks heaps @bollu, it will be great to have this functionality

bollu:
Just glad to see this merged :) Any ideas on how to improve performance of this now? I'd personally like it to be fluid on resizing which it isn't (for the `plot` example at least on my laptop)

Eric89GXL:
I think there are higher-priority issues to work on. Resizing in most applications is a rare enough operation that there are bigger, more-often-used performance bottlenecks currently in the package (e.g., collection-izing the `scene` system).

BUT if you really wanted to work on this specifically, I'd do some line profiling to figure out why it's slow. It's possible the `cassowary` people already have some ideas of where the bottlenecks in the Python implementation are. If operations are done in a lot of loops, then vectorization might help, even if it leads to some redundant operations.

bollu:
yes, I should do that. Shall I open an issue about performance so we can move the discussion there?

Eric89GXL:
We already have enough issues open related to performance :) Seriously though let's assume YAGNI until a user hits a use case where rapid resizing is necessary, and hopefully they will be motivated to improve it.

bollu:
sounds good to me :)

campagnola:
Hooray! Congrats on completing this epic PR, @bollu!

PR #1050, opened 2015-08-12
add spatial-filters example in examples/basic/gloo

kmuehlbauer:
PR suggested in #1017 and as follow up to #1048.
 
The added example shows the available 17 interpolations in spatial-filters.frag . Toggling with 'Left' and 'Right' arrow keys. The actual used filter is shown in the window title.

### DIFF
--- a/examples/basics/gloo/spatial_filters.py
+++ b/examples/basics/gloo/spatial_filters.py
@@ -90,7 +90,7 @@ class Canvas(app.Canvas):
         self.texture = gloo.Texture2D(I, interpolation='nearest')
 
         self.kernel = gloo.Texture2D(np.load(os.path.join(_data_dir, 'spatial-filters.npy')),
-                                     interpolation='nearest')
+                                     interpolation='nearest', internalformat='r16f')
 
         self.program['u_texture'] = self.texture
         self.program['u_kernel'] = self.kernel
--- a/examples/basics/gloo/spatial_filters.py
+++ b/examples/basics/gloo/spatial_filters.py
@@ -5,10 +5,9 @@ Example demonstrating spatial filtering using spatial-filters fragment shader.
 Left and Right Arrow Keys toggle through available filters.
 """
 
-import os
 import numpy as np
 
-from vispy.io import _data_dir
+from vispy.io import load_spatial_filters
 from vispy import gloo
 from vispy import app
 
@@ -20,7 +19,7 @@ I[1::2, 2] = 0.5
 I[2, 2] = 1.0
 
 # loading interpolation kernel
-kernel = np.load(os.path.join(_data_dir, 'spatial-filters.npy'))
+kernel = load_spatial_filters()
 
 # A simple texture quad
 data = np.zeros(4, dtype=[('a_position', np.float32, 2),
--- a/examples/basics/gloo/spatial_filters.py
+++ b/examples/basics/gloo/spatial_filters.py
@@ -125,8 +125,7 @@ class Canvas(app.Canvas):
         self.program['u_shape'] = I.shape[1], I.shape[0]
         self.program['u_interpolation'] = 0
 
-        self.names = ["Nearest"]
-        self.names.extend(names)
+        self.names = tuple(["Nearest"]) + names
 
         self.title = 'Spatial Filtering using %s Filter' % \
                      self.names[int(self.program['u_interpolation'])]
--- a/examples/basics/gloo/spatial_filters.py
+++ b/examples/basics/gloo/spatial_filters.py
@@ -148,9 +148,9 @@ class Canvas(app.Canvas):
                 step = -1
             self.program['u_interpolation'] = \
                 (self.program['u_interpolation'] + step) % 17
-        self.title = 'Spatial Filtering using %s Filter' % \
-                     self.names[int(self.program['u_interpolation'])]
-        self.update()
+            self.title = 'Spatial Filtering using %s Filter' % \
+                         self.names[int(self.program['u_interpolation'])]
+            self.update()
 
     def on_draw(self, event):
         gloo.clear(color=True, depth=True)
--- a/examples/basics/gloo/spatial_filters.py
+++ b/examples/basics/gloo/spatial_filters.py
@@ -55,54 +55,54 @@ void main()
     // using if-statements here is inefficient
     // used here just for selection of interpolation function
     if (u_interpolation < 0.5) {
-        gl_FragColor = Nearest(u_texture, u_shape, v_texcoord);
-    }
-    else if (u_interpolation < 1.5) {
         gl_FragColor = Bilinear(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 2.5) {
+    else if (u_interpolation < 1.5) {
         gl_FragColor = Hanning(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 3.5) {
+    else if (u_interpolation < 2.5) {
         gl_FragColor = Hamming(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 4.5) {
+    else if (u_interpolation < 3.5) {
         gl_FragColor = Hermite(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 5.5) {
+    else if (u_interpolation < 4.5) {
         gl_FragColor = Kaiser(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 6.5) {
+    else if (u_interpolation < 5.5) {
         gl_FragColor = Quadric(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 7.5) {
+    else if (u_interpolation < 6.5) {
         gl_FragColor = Bicubic(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 8.5) {
+    else if (u_interpolation < 7.5) {
         gl_FragColor = CatRom(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 9.5) {
+    else if (u_interpolation < 8.5) {
         gl_FragColor = Mitchell(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 10.5) {
+    else if (u_interpolation < 9.5) {
         gl_FragColor = Spline16(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 11.5) {
+    else if (u_interpolation < 10.5) {
         gl_FragColor = Spline36(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 12.5) {
+    else if (u_interpolation < 11.5) {
         gl_FragColor = Gaussian(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 13.5) {
+    else if (u_interpolation < 12.5) {
         gl_FragColor = Bessel(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 14.5) {
+    else if (u_interpolation < 13.5) {
         gl_FragColor = Sinc(u_texture, u_shape, v_texcoord);
     }
-    else if (u_interpolation < 15.5) {
+    else if (u_interpolation < 14.5) {
         gl_FragColor = Lanczos(u_texture, u_shape, v_texcoord);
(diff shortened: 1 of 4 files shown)


examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# vispy: gallery 2
```
rougier: Is it relevant ?

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,165 @@
...
+"""
+Example demonstrating spatial filtering using spatial-filters fragment shader.
+Left and Right Arrow Keys toggle through available filters.
```
rougier: Remove capitals maybe

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,165 @@
...
+"""
+
+import os
+import numpy as np
+
+from vispy.util.transforms import ortho
```
rougier: Unnecessary import

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,165 @@
...
+from vispy.io import _data_dir
```
rougier: Unnecessary import

kmuehlbauer:
How do I correctly invoke the testing at my machine to find the errors before travis does?

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,165 @@
...
+from vispy import gloo
+from vispy import app
+
+
+# Image to be displayed
+W, H = 100, 100
+I = np.random.uniform(0,1,(5,5,3)).astype(np.float32)
```
rougier: Maybe a non random image would be better suited to show actual interpolation.  Your central point in a 5x5 was fine.

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,165 @@
...
+class Canvas(app.Canvas):
+
+    def __init__(self):
+        app.Canvas.__init__(self, keys='interactive', size=((W*5), (H*5)))
+
+        self.program = gloo.Program(VERT_SHADER, FRAG_SHADER)
+        self.texture = gloo.Texture2D(I, interpolation='nearest')
+
+        self.kernel = gloo.Texture2D(np.load(os.path.join(_data_dir, 'spatial-filters.npy')),
+                                     interpolation='nearest', internalformat='r16f')
```
rougier: Maybe add a comment to relate the r16f to the relevant issue on github

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,165 @@
...
+#include "misc/spatial-filters.frag"
+uniform sampler2D u_texture;
+uniform vec2      u_shape;
+uniform float     u_interpolation;
+varying vec2 v_texcoord;
+void main()
+{
+    //gl_FragColor = texture2D(u_texture, v_texcoord);
+    //gl_FragColor.a = 1.0;
+
```
rougier: Here you can add a comment saying those tests are currently inefficient even it is not really the point here.

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,165 @@
...
+            else:
+                step = -1
+            self.program['u_interpolation'] = \
+                (self.program['u_interpolation'] + step) % 17
+        self.title = 'Spatial Filtering using %s Filter' % \
+                     self.names[int(self.program['u_interpolation'])]
+
+        self.update()
+
+    def on_resize(self, event):
```
rougier: This is unnecessary if you use a (-1,-1) x (+1,+1) quad without the ortho

rougier:
@Eric89GXL might know the answer to your question

examples/basics/gloo/spatial_filters.py
kmuehlbauer: _data_dir import is needed for the kernel loading, isn't it?

Eric89GXL:
You can do `python make test` and it will give you a list of options, e.g. `python make test flake`

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,163 @@
...
+
+# create 5x5 matrix with border pixels 0, center pixels 1
+# and other pixels 0.5
+I = np.zeros(25).reshape((5, 5)).astype(np.float32)
+I[1:4, 1::2] = 0.5
+I[1::2, 2] = 0.5
+I[2, 2] = 1.0
+
+# loading interpolation kernel
+kernel = np.load(os.path.join(_data_dir, 'spatial-filters.npy'))
```
Eric89GXL: I agree this is not so nice -- users shoudn't need private functions/attributes (names preceded by `_`) to do things. Perhaps we should add a `vispy.io.get_spatial_filters()` function.

kmuehlbauer:
Yes, I had to install flake8 and friends first, works now.

examples/basics/gloo/spatial_filters.py
kmuehlbauer: @Eric89GXL: Agreed, I copied this verbatim from glumpy. 

Could this go into `vispy.io.datasets.py` as `load_spatial_filters()`, as there are already some `load_`functions?

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,163 @@
...
+        self.program['u_texture'] = self.texture
+        self.program['u_kernel'] = self.kernel
+        self.program['u_shape'] = I.shape[1], I.shape[0]
+        self.program['u_interpolation'] = 0
+
+        self.names = [
+            "Nearest", "Bilinear", "Hanning", "Hamming",
+            "Hermite", "Kaiser", "Quadric", "Bicubic",
+            "CatRom", "Mitchell", "Spline16", "Spline36",
+            "Gaussian", "Bessel", "Sinc", "Lanczos", "Blackman"]
```
Eric89GXL: we could have a tuple of these names returned by `get_spatial_filters`, too, which would be nice

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,163 @@
...
+        if event.key in ['Left', 'Right']:
+            if event.key == 'Right':
+                step = 1
+            else:
+                step = -1
+            self.program['u_interpolation'] = \
+                (self.program['u_interpolation'] + step) % 17
+        self.title = 'Spatial Filtering using %s Filter' % \
+                     self.names[int(self.program['u_interpolation'])]
+        self.update()
```
Eric89GXL: these can be indented as well, as you only need an update if the interp has changed

examples/basics/gloo/spatial_filters.py
Eric89GXL: yeah sure, +1 for following whatever naming scheme we use there

examples/basics/gloo/spatial_filters.py
kmuehlbauer: Ahh, ja...

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,162 @@
...
+
+# create 5x5 matrix with border pixels 0, center pixels 1
+# and other pixels 0.5
+I = np.zeros(25).reshape((5, 5)).astype(np.float32)
+I[1:4, 1::2] = 0.5
+I[1::2, 2] = 0.5
+I[2, 2] = 1.0
+
+# loading interpolation kernel
+kernel = load_spatial_filters()
```
Eric89GXL: so `kernel, names = load_spatial_filters()`

examples/basics/gloo/spatial_filters.py
kmuehlbauer: Do you mean as this:  `tuple(["Nearest", "Bilinear",...])`?

Eric89GXL: equivalently `('Nearest', 'Bilinear', ...)`

Eric89GXL: tuple is preferable to list here because it's immutable

kmuehlbauer: The kernel has 16 interpolations implemented, "Nearest" doesn't have its own kernel.

Eric89GXL: I'm not quite sure I understand your point...

My point is that the `ndarray` you get from `load_spatial_filters.py` can be used with 16 different filters. When loading it, it would be useful to have that list provided, as:
```
kernels, names = load_spatial_filters()
```
where `kernels` is an `ndarray`, and `names` is a tuple of the names that can be used in `glsl` code. Make sense?

kmuehlbauer: Perfectly sense, but I had to fiddle with the names to also have "Nearest" inside in the example file.

vispy/io/datasets.py
```diff
@@ -33,3 +33,20 @@ def load_crate():
...
+
+
+def load_spatial_filters():
+    """Load spatial-filters kernel
+
+    Returns
+    -------
+    kernel : array
+        16x1024 16 interpolation kernel with length 1024 each.
+    names : list with respective interpolation names
```
Eric89GXL: numpy docstring format
```
names : list of str
    Respective interpolation names, plus "Nearest" which does not require
    a filter but can still be used.
```
Then you can have `"Nearest"` at the end of the list, no fiddling required later.

vispy/io/datasets.py
```diff
@@ -33,3 +33,20 @@ def load_crate():
...
+    """
+    names = ["Bilinear", "Hanning", "Hamming", "Hermite",
```
Eric89GXL: please make this a `tuple` so it's immutable

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,158 @@
...
+        self.names = names
+
+        self.title = 'Spatial Filtering using %s Filter' % \
+                     self.names[int(self.program['u_interpolation'])]
+
+        self.program.bind(gloo.VertexBuffer(data))
+
+        gloo.set_clear_color('white')
+
+        self.show()
```
Eric89GXL: somewhere you need a `self.context.set_viewport(512, 512)` call

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,158 @@
...
+        self.program['u_interpolation'] = 16
+
+        self.names = names
+
+        self.title = 'Spatial Filtering using %s Filter' % \
+                     self.names[int(self.program['u_interpolation'])]
+
+        self.program.bind(gloo.VertexBuffer(data))
+
+        gloo.set_clear_color('white')
```
Eric89GXL: the newer style is to use `self.context.` instead of `gloo.` for these commands

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,158 @@
...
+            else:
+                step = -1
+            self.program['u_interpolation'] = \
+                (self.program['u_interpolation'] + step) % 17
+            self.title = 'Spatial Filtering using %s Filter' % \
+                         self.names[int(self.program['u_interpolation'])]
+            self.update()
+
+    def on_draw(self, event):
+        gloo.clear(color=True, depth=True)
```
Eric89GXL: same here

examples/basics/gloo/spatial_filters.py
```diff
@@ -0,0 +1,158 @@
...
+        self.program.draw('triangle_strip')
+
```
Eric89GXL: it also might be worth adding an `on_resize` handler for when the window gets resized

examples/basics/gloo/spatial_filters.py
kmuehlbauer: only `self.context.set_viewport(0, 0, 512, 512)` works

Eric89GXL: yeah that's the one :)

Eric89GXL:
Good to go from your end?

kmuehlbauer:
If @rougier has nothing to add, it is good for me.

Eric89GXL:
@rougier want to have another look?

rougier:
Nope, good for me.

PR #1051, opened 2015-08-12
WIP: add interpolation to ImageVisual and respective example

kmuehlbauer:
As discussed in #1017:

This is a first attempt to implement spatial filtering to image visual.
The example works, but the handling with the "hardware" modes ('nearest', 'linear') is still broken. 

For every interpolation a dedicated lookup_function is created and fit to the shader_program as needed. I'm open for any suggestions to improve this. I already thought about subclassing ImageVisual, but that should not be mandatory.

There is sure a lot of room for improvement.

### DIFF
--- a/examples/basics/scene/image.py
+++ b/examples/basics/scene/image.py
@@ -33,7 +33,7 @@ view.camera = scene.PanZoomCamera(aspect=1)
 # flip y-axis to have correct aligment
 view.camera.flip = (0, 1, 0)
 view.camera.set_range()
-view.camera.zoom(0.1, (250,200))
+view.camera.zoom(0.1, (250, 200))
 
 # get interpolation functions from Image
 names = image.interpolation_functions
--- a/examples/basics/scene/image.py
+++ b/examples/basics/scene/image.py
@@ -33,6 +33,7 @@ view.camera = scene.PanZoomCamera(aspect=1)
 # flip y-axis to have correct aligment
 view.camera.flip = (0, 1, 0)
 view.camera.set_range()
+view.camera.zoom(0.1, (250,200))
 
 # get interpolation functions from Image
 names = image.interpolation_functions
--- a/examples/basics/scene/image.py
+++ b/examples/basics/scene/image.py
@@ -10,30 +10,7 @@ Simple use of SceneCanvas to display an Image.
 import sys
 from vispy import scene
 from vispy import app
-import numpy as np
-
-
-def get_image():
-    """Load an image from the demo-data repository if possible. Otherwise,
-    just return a randomly generated image.
-    """
-    from vispy.io import load_data_file, read_png
-
-    try:
-        return read_png(load_data_file('mona_lisa/mona_lisa_sm.png'))
-    except Exception as exc:
-        # fall back to random image
-        print("Error loading demo image data: %r" % exc)
-
-    # generate random image
-    image = np.random.normal(size=(100, 100, 3))
-    image[20:80, 20:80] += 3.
-    image[50] += 3.
-    image[:, 50] += 3.
-    image = ((image - image.min()) *
-             (253. / (image.max() - image.min()))).astype(np.ubyte)
-
-    return image
+from vispy.io import load_data_file, read_png
 
 canvas = scene.SceneCanvas(keys='interactive')
 canvas.size = 800, 600
@@ -43,8 +20,10 @@ canvas.show()
 view = canvas.central_widget.add_view()
 
 # Create the image
+img_data = read_png(load_data_file('mona_lisa/mona_lisa_sm.png'))
 interpolation = 'Nearest'
-image = scene.visuals.Image(get_image(), interpolation=interpolation,
+
+image = scene.visuals.Image(img_data, interpolation=interpolation,
                             parent=view.scene, method='subdivide')
 
 canvas.title = 'Spatial Filtering using %s Filter' % interpolation
@@ -59,7 +38,6 @@ view.camera.set_range()
 names = image.interpolation_functions
 names = list(names)
 names.sort()
-print(names)
 act = 17
 
 
@@ -72,11 +50,11 @@ def on_key_press(event):
             step = 1
         else:
             step = -1
-    act = (act + step) % len(names)
-    interpolation = names[act]
-    image.interpolation = interpolation
-    canvas.title = 'Spatial Filtering using %s Filter' % interpolation
-    canvas.update()
+        act = (act + step) % len(names)
+        interpolation = names[act]
+        image.interpolation = interpolation
+        canvas.title = 'Spatial Filtering using %s Filter' % interpolation
+        canvas.update()
 
 
 if __name__ == '__main__' and sys.flags.interactive == 0:
--- a/examples/basics/scene/image.py
+++ b/examples/basics/scene/image.py
@@ -11,6 +11,7 @@ import sys
 from vispy import scene
 from vispy import app
 import numpy as np
+from itertools import cycle
 
 canvas = scene.SceneCanvas(keys='interactive')
 canvas.size = 800, 600
@@ -20,12 +21,29 @@ canvas.show()
 view = canvas.central_widget.add_view()
 
 # Create the image
-img_data = np.random.normal(size=(100, 100, 3), loc=128,
-                            scale=50).astype(np.ubyte)
-image = scene.visuals.Image(img_data, parent=view.scene)
-
+#img_data = np.random.normal(size=(100, 100, 3), loc=128,
+#                            scale=50).astype(np.ubyte)
+img_data = np.zeros(25).reshape((5, 5)).astype(np.float32)
+img_data[1:4, 1::2] = 0.5
+img_data[1::2, 2] = 0.5
+img_data[2, 2] = 1.0
+ipol = 'Nearest'
+image = scene.visuals.Image(img_data, ipol=ipol, parent=view.scene)
+canvas.title = 'Spatial Filtering using %s Filter' % ipol
 # Set 2D camera (the camera will scale to the contents in the scene)
 view.camera = scene.PanZoomCamera(aspect=1)
+view.camera.set_range()
+
+names = cycle(image._names)
+
+# Implement key presses
+@canvas.events.key_press.connect
+def on_key_press(event):
+    ipol = next(names)
+    image.ipol = ipol
+    canvas.title = 'Spatial Filtering using %s Filter' % ipol
+    canvas.update()
+
 
 if __name__ == '__main__' and sys.flags.interactive == 0:
     app.run()
--- a/examples/basics/scene/image.py
+++ b/examples/basics/scene/image.py
@@ -11,7 +11,6 @@ import sys
 from vispy import scene
 from vispy import app
 import numpy as np
-from itertools import cycle
 
 canvas = scene.SceneCanvas(keys='interactive')
 canvas.size = 800, 600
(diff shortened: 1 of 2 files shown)


examples/basics/scene/image.py
```diff
@@ -20,12 +21,29 @@
...
-image = scene.visuals.Image(img_data, parent=view.scene)
-
+#img_data = np.random.normal(size=(100, 100, 3), loc=128,
+#                            scale=50).astype(np.ubyte)
+img_data = np.zeros(25).reshape((5, 5)).astype(np.float32)
+img_data[1:4, 1::2] = 0.5
+img_data[1::2, 2] = 0.5
+img_data[2, 2] = 1.0
+ipol = 'Nearest'
+image = scene.visuals.Image(img_data, ipol=ipol, parent=view.scene)
```
Eric89GXL: It should probably be `interp`, `interp_method`, or `interpolation`. I think I prefer the last one.

vispy/visuals/image.py
```diff
@@ -67,15 +68,33 @@
...
         texcoord.y < 0.0 || texcoord.y > 1.0) {
             discard;
         }
-        return texture2D($texture, texcoord); 
+        return texture2D($texture, texcoord);
     }"""
 
+def create_ipol_frag():
+    from ..io import load_spatial_filters
+    from ..ext.ordereddict import OrderedDict
```
Eric89GXL: you shouldn't have to nest either one of these -- don't if you don't have to

examples/basics/scene/image.py
kmuehlbauer: +1 for `interpolation`

Eric89GXL:
I agree you shouldn't have to subclass `ImageVisual`, this seems like a good approach to me

examples/basics/scene/image.py
Eric89GXL: okay just use that one, it's really the most complete/Pythonic name anyway

vispy/visuals/image.py
kmuehlbauer: What do you mean by `nest`?

Eric89GXL: these imports are nested (within) the function `create_ipol_frag`, but there is no reason to put them here instead of at the top of the file with the rest of the imports

Eric89GXL: One good reason to have all imports at the top is it makes it easy to answer the question, "what other functions/modules does this module depend on?". Usually imports should only be nested if necessary, e.g. if a dependency is optional or there is a circular import problem.

kmuehlbauer: Ah, yes, I wasn't sure if that function will stay this way, so I have it there for the time being. This will go into import section when finalizing. Thanks for catching.

kmuehlbauer:
@Eric89GXL 

This works now with all available interpolation modes. To distinguish between the two "hw" modes, which are bound to Texture2D I used `_texture_interpolation` and `_interpolation` for the parameter.

The interpolation names (`_names`) should get a more suitable name and made public.

I am totally new to GLSL, so there sure is improvement potential on the shader code side. @campagnola, could you please have a look?

examples/basics/scene/image.py
```diff
@@ -20,12 +20,41 @@
...
+                            parent=view.scene)
+canvas.title = 'Spatial Filtering using %s Filter' % interpolation
 
 # Set 2D camera (the camera will scale to the contents in the scene)
 view.camera = scene.PanZoomCamera(aspect=1)
+view.camera.set_range()
+
+# hack for now, should rename "names" to something useful
+# and make public
+names = image._names
```
Eric89GXL: I don't think we want people to get it from the object, but instead from some global variable `INTERPOLATION_FILTERS` in some sensible place, or from some function. Right now you could get this by doing:
```
from vispy.io import load_spatial_filters
names = load_spatial_filters()[1]
```
It isn't quite as nice as a constant var imported from somewhere, but it would work without needing private vars.

Eric89GXL:
GLSL LGTM. The example runs nicely except for this:
```
INFO: Variable u_kernel is not an active uniform
INFO: Variable u_shape is not an active uniform
```
You should put the setting of those variables in a conditional.

But actually I'm wondering if you could just *always* use the Filter code instead. I think there are filters for `Nearest` and `Linear` (right?), so you could just set the GL interp to `nearest` and always use the same code snippet for the drawing. It would simplify the code and alleviate this issue, with a hopefully negligible performance cost.

examples/basics/scene/image.py
kmuehlbauer: I thought about doing exactly what you suggested, but then we would miss 'nearest' and 'linear'. I added the two to the names-tuple inside ImageVisual and also to the Function-dictionary.

Eric89GXL: Maybe we should add those to the list of names that get returned by that function. I thought we already added one of them to the end at least, might as well add the other.

Eric89GXL:
Actually I just realized that you'll probably still get warnings about those variables even if you use the same bit of shader code because the GPU compiler will probably optimize them out. So never mind about using the same code, just put the setting of those variables in a conditional instead.

kmuehlbauer:
Thought about that too, but I have had issues here with textures and precision. 

What is the difference between the 'linear' and the shaders 'Bilinear'. The shader function looks_up from a 1D-texture, but the other?

Eric89GXL:
I assumed those were the same thing, actually.

examples/basics/scene/image.py
kmuehlbauer: Yes, we have 'Nearest' at the end of the function. Difference is, that there exists a function-declaration in the spatial-filters. frag for it. What we can think of is, using this also for 'linear'.

kmuehlbauer:
I thought, that the 'linear' is built-in hardware function and may have different output in terms of precision.

examples/basics/scene/image.py
Eric89GXL: Since they are going to differ, then we could put a `INTERPOLATION_FILTERS = (...)` with the complete list in `scene/image.py` that gets imported into `scene/`, then people can do:
```
from vispy.scene import INTERPOLATION_FILTERS
```
The constant is slightly nicer than having to import and run a function anyway.

Eric89GXL:
Yeah let's not worry about it for now. If you add the `INTERPOLATION_FILTERS` constant and put the variable-setting in a conditional, it should be good to go.

kmuehlbauer:
I can do as suggested, but if someone subclasses ImageVisual and removes some of the interpolation functions, he would also have to change the `INTERPOLATION_FILTERS` constant.

Eric89GXL:
That's true. Although I think it's not terribly likely that someone would do that, we can leave it on the object, then. Just make the attribute public as a tuple, and fix the conditional on the variable setting for the shaders, and you should be good, then.

kmuehlbauer:
The thing with the conditional doesn't work. The 'Nearest' function in spatial-filters.frag want to have the u_kernel and u_shape even if it is not using it. Or do I have something misunderstood.

kmuehlbauer:
Then I would leave the 'Nearest' function out of ImageVisual because it its represented by 'nearest' anyway. What do you think?

Eric89GXL:
Take a look at what the `Nearest` does, it will not actually use `shape` or `u_kernel` and thus they will both be optimized out:

https://github.com/vispy/vispy/blob/master/vispy/glsl/misc/spatial-filters.frag#L274

Do you not get those `INFO` messages when you run?

Eric89GXL:
I'm not sure what you mean by the 'Nearest' comment. Do you mean right now it has both 'Nearest' and 'nearest' in it?

kmuehlbauer:
Yes, I do, but I get a shader compilation error, if I do not set those variables for 'Nearest`.

kmuehlbauer:
Yes, at the moment is has both, which is doubled ;-)

kmuehlbauer:
I'll look into it tomorrow.

Eric89GXL:
Hmm, let's fix both problems, then. Have the mode be named `Nearest` (for consistency with the other names), and don't include 'nearest'. But don't actually use the GLSL interpolation code when the mode is `'Nearest'`, just use the hardware mode. Then, in that mode and `linear`, you don't need to set those vars. Make sense?

BTW do we also have `'Bilinear'` and `'linear'`? Those are likely the same thing and we can do the same thing, call it `Bilinear` but use the hardware mode.

Eric89GXL:
I think code for `Nearest` and `Bilinear` is included in `spatial_filters.py` for completeness and simplicity of switching modes, but in this case it's easy for us to switch out the GLSL for the hardware mode, so let's just do it.

kmuehlbauer:
+1

kmuehlbauer:
@Eric89GXL, There is one flaw using the example, if you switch back and forth repeatedly between one hardware mode and one shader mode, the shader mode brakes somehow. I haven't tracked down the problems source yet.

examples/basics/scene/image.py
```diff
@@ -12,6 +12,29 @@
...
+
+    # generate random image
+    image = np.random.normal(size=(100, 100, 3))
+    image[20:80, 20:80] += 3.
+    image[50] += 3.
+    image[:, 50] += 3.
+    image = ((image - image.min()) *
+             (253. / (image.max() - image.min()))).astype(np.ubyte)
+
+    return image
```
Eric89GXL: I think for simplicity of the example we should just load the `mona_lisa` image. We don't really do this safety-checking in other examples...

vispy/visuals/image.py
```diff
@@ -116,36 +138,64 @@ class ImageVisual(Visual):
...
+                 interpolation='Nearest', **kwargs):
         self._data = None
-        self._interpolation = 'nearest'
+
+        # check texture interpolation
+        self._interpolation = interpolation
+        if self._interpolation == 'Bilinear':
+            texture_interpolation = 'linear'
+        else:
+            texture_interpolation = 'nearest'
```
Eric89GXL: can you check to make sure that the `interpolation` is in the tuple of accepted values? Also, for consistency we should probably have them be all lower case, e.g. `spline16` instead of `Spline16`

Eric89GXL:
I'm not sure why you'd get that switching error, but I have a feeling it's not due to your PR, but some bug with `gloo` somewhere. If you don't want to solve it here you don't have to, you can open an issue for it. But it would be great if you could find it.

examples/basics/scene/image.py
kmuehlbauer: OK, will remove that function and just put the read_png into the main script.

vispy/visuals/image.py
kmuehlbauer: Yes, I'll add a check here. 

The lowercase is another thing. Lowercase would collide with spatial_filter.frag function names. If have no really cool idea how to solve this.

kmuehlbauer:
I'm afraid that I'm not yet capable of that, but If you have some pointers on how to do this I can try. If it isn't solved I'll open an issue.

vispy/visuals/image.py
Eric89GXL: make the API that faces the user all lower case, but when you go to actually use the shader code, capitalize the first letter

kmuehlbauer:
@Eric89GXL I had to move several code blocks, to get correct checking. Haven't found the bug yet...

Eric89GXL:
I wouldn't worry about it. Just open an issue, then we can get this in and you can move on to more fun things :)

kmuehlbauer:
@Eric89GXL, ping @campagnola 

I've thought about the whole interpolation implementation. I'm not sure anymore that this is finally what should be achieved (cluttering visuals with functionality). 

So I took the time to read more of the visuals and scene code and found the ability to attach/detach filters (and/or add VisualComponents)

The whole interpolation implementation could be moved from the `ImageVisual` to the filters section. We could have a Interpolation class within filters, which holds the interpolation kernel and stuff and changes the `_data_lookup_fn` of the parent visual (ImageVisual) according to the `_build_interpolation` function here in this PR.

I have this working here, but like to get your impressions on that. The benefit would be, that this goes to more modularity (which I think was intended here). The ImageVisual could stay leaner. The only problem I see so far, is that there would be the need to change parameters of the parent visual.

Eric89GXL:
I thought about having it be a filter initially, but I'm not sure how it would work, since it operates on a texture, not on e.g. the `gl_FragColor` itself. Because of this, it would really only work with `ImageVisual`, right?

kmuehlbauer:
Yes, for now it would work only with ImageVisual. Or with any other Visual which has a '_data_lookup_fn' similar to ImageVisual.

The idea seems appealing to have this working on gl_FragColor. Would that be worth a try?

Eric89GXL:
I don't think you can do it. In fact that's not actually the level at which you want to operate with this type of operation. Consider taking your image and rotating (in 2 or 3 D) the image a bit, and you'll see that operating at the Texture interpolation level is different from operating at the `gl_FragColor` level. Maybe you'd eventually want to add something similar where you blur the entire scene, but I think in OpenGL that needs to be done by rendering to a Framebuffer/Texture and operating on that result.

You are right, though, that anything that accesses a Texture could potentially want this functionality. I'm not sure the best way to expose that in a general way, though. So I think for now it's fine to have it on ImageVisual, and just make a note that the code could eventually be refactored to be more generally usable on arbitrary texture accesses. We can always refactor it later assuming the public API is sensible, which it is at this point.

kmuehlbauer:
Yes, this is what a quick internet search was giving. 

So I'll just add a note at the `_build_interpolation` that this could be considered moving to some filter function later.

Eric89GXL:
Sounds good, let me know when it's good to go from your end and I'll take a last look

kmuehlbauer:
@Eric89GXL, I think I'm done so far. For the switching glitch I already opened an issue here #1061 

And here is [a gist](https://gist.github.com/kmuehlbauer/6e2cb27c55655d4bdef8) for reference with the first attempt to make this a filter:

Eric89GXL:
It's a bit hard to see the effect because the image is so small. Maybe try sub-selecting just the face part of the image? WDYT?

kmuehlbauer:
Like this?

Eric89GXL:
LGTM, yeah. I'll take a final look and merge if the CIs are happy

examples/basics/scene/image.py
```diff
@@ -20,12 +20,43 @@
...
+                            parent=view.scene, method='subdivide')
+
+canvas.title = 'Spatial Filtering using %s Filter' % interpolation
 
 # Set 2D camera (the camera will scale to the contents in the scene)
 view.camera = scene.PanZoomCamera(aspect=1)
+# flip y-axis to have correct aligment
+view.camera.flip = (0, 1, 0)
+view.camera.set_range()
+view.camera.zoom(0.1, (250,200))
```
Eric89GXL: style error, missing whitespace after `,`

kmuehlbauer:
gnarrrrf

examples/basics/scene/image.py
kmuehlbauer: gnarrrf

Eric89GXL:
Muuuuuuuuuuch better. I should be able to merge this time :)

Eric89GXL:
Thanks @kmuehlbauer

PR #1069, opened 2015-08-19
FIX: change kernel texture internalformat and interpolation 

### LINKED ISSUE #1068: ImageVisual: attaching Isoline Filter - interpolation problem
Since ImageVisual is now capable of several interpolation methods, I tried to setup a isoline creation in pure glsl. 

Therefore I took advantage of @rougier https://github.com/glumpy/glumpy/blob/master/examples/isocurves.py for isoline tracing. I adapted this to have isocolor, isowidth and isolevel as parameter to the `Isoline` class. The example can sweep (via Left/Right key) trough the available interpolation functions (plus `bicubic2D` because I want to show the differences) and add/remove isolines with Up/Down. If you remove the comments from the `curve` IsocurveVisual you will get those CPU computed isolines on top.

So far the preamble. The problem you see, when zooming in a bit and switching the interpolations. On my systems only the `bicubic2D` interpolator accounts for clear, smooth isolines. In recent discussions texture precision was claimed as possible issue.

My humble request is, please test this self contained program https://gist.github.com/kmuehlbauer/db681d889cec36c56394 and write your experiences. 

Here are two images which show the behaviour I experience. The differences between the interpolated images are not seen to the human eye, but the isoline trace routine clearly gets them.

This is bicubic interpolation from 1DTexture
isoline_bicubic
This is bicubic2D interpolation (implementation from scratch)
isoline_bicubic2d

Thanks for the patience!

kmuehlbauer:
to fix precision issues, fixes #1068

This fixes problems with ImageVisual using spatial filtering using 1D kernel lookups.

### DIFF
--- a/examples/basics/scene/contour.py
+++ b/examples/basics/scene/contour.py
@@ -8,14 +8,13 @@
 Simple use of SceneCanvas to display an Image.
 """
 import sys
-import numpy as np
 
 from vispy import scene, app, visuals
 from vispy.visuals.filters import IsolineFilter
 from vispy.io import load_data_file, read_png
 
 canvas = scene.SceneCanvas(keys='interactive')
-canvas.size = 800, 600
+canvas.size = 600, 800
 canvas.show()
 
 # Set up a viewbox to display the image with interactive pan/zoom
@@ -27,9 +26,7 @@ image = scene.visuals.Image(img_data, interpolation=interpolation,
                             parent=view.scene)
 image.transform = visuals.transforms.STTransform(translate=(0, 0, 0.5))
 
-levels = np.linspace(img_data.min(), img_data.max(), 13)[1:-1]
-color_lev = 'white'
-level = 0
+level = 10
 iso = IsolineFilter(level=level, width=1., color='black')
 image.attach(iso)
 
@@ -37,8 +34,11 @@ image.attach(iso)
 view.camera = scene.PanZoomCamera(aspect=1)
 # flip y-axis to have correct aligment
 view.camera.flip = (0, 1, 0)
-view.camera.set_range()
+# select face part
+view.camera.rect = (160, 130, 240, 200)
 
+canvas.title = ('Spatial Filtering using %s Filter - Isoline %d level'
+                % (image.interpolation, iso.level))
 
 # get interpolation functions from Image
 names = image.interpolation_functions
--- a/examples/basics/scene/contour.py
+++ b/examples/basics/scene/contour.py
@@ -1,0 +1,116 @@
+# -*- coding: utf-8 -*-
+# vispy: gallery 30
+# -----------------------------------------------------------------------------
+# Copyright (c) 2015, Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
+Simple use of SceneCanvas to display an Image.
+"""
+import sys
+from vispy import scene
+from vispy import app
+from vispy import visuals
+from vispy.util.filter import gaussian_filter
+from vispy.visuals.filters import Isoline
+import numpy as np
+from vispy.io import load_data_file, read_png
+
+canvas = scene.SceneCanvas(keys='interactive')
+canvas.size = 800, 600
+canvas.show()
+
+# Set up a viewbox to display the image with interactive pan/zoom
+view = canvas.central_widget.add_view()
+
+# Create the image
+interpolation = 'bicubic'
+np.random.seed(1000)
+img_data = np.random.normal(size=(100, 100), loc=50, scale=150)
+img_data = gaussian_filter(img_data, (4, 4, 0)).astype(np.float32)
+
+img_data = np.zeros(25).reshape((5, 5)).astype(np.float32)
+img_data[1:4, 1::2] = 0.5
+img_data[1::2, 2] = 0.5
+img_data[2, 2] = 1.0
+
+#img_data = read_png(load_data_file('mona_lisa/mona_lisa_sm.png'))
+#img_data = np.dot(img_data[...,:3], [0.299, 0.587, 0.144])
+#img_data = img_data[...,0] * img_data[...,1] * img_data[...,2]
+#img_data /= 3.
+#image = scene.visuals.Image(img_data, interpolation=interpolation,
+#                            parent=view.scene, method='subdivide')
+
+img_data1 = np.zeros((100, 100))
+img_data1[0,0] = -0.5
+img_data1[99,99] = +0.5
+image = scene.visuals.Image(img_data, interpolation=interpolation,
+                            parent=view.scene, method='subdivide')
+
+#image = scene.visuals.Contour(img_data, interpolation='bicubic1',
+#                            parent=view.scene, method='subdivide',
+#                            levels=2, width=1.0, color='yellow')
+
+image.transform = visuals.transforms.STTransform(translate=(0, 0, 0.5))
+
+levels = np.linspace(img_data.min(), img_data.max(), num=13, endpoint=True)[1:-1]
+color_lev = 'white'
+# Create isocurve, make a child of the image to ensure the two are always
+# aligned.
+#curve = scene.visuals.Isocurve(img_data, levels=levels, color_lev=color_lev,
+#                              parent=view.scene, method='subdivide')
+
+iso = Isoline(level=10, width=1., color='black')
+
+canvas.title = 'Spatial Filtering using %s Filter' % interpolation
+
+# Set 2D camera (the camera will scale to the contents in the scene)
+view.camera = scene.PanZoomCamera(aspect=1)
+# flip y-axis to have correct aligment
+view.camera.flip = (0, 1, 0)
+view.camera.set_range()
+
+# get interpolation functions from Image
+names = image.interpolation_functions
+names = list(names)
+names.sort()
+print(names)
+act = 17
+
+act = -1
+
+level = 2
+first = True
+
+
+# Implement key presses
+@canvas.events.key_press.connect
+def on_key_press(event):
+    global act, level, first, interpolation
+    if event.key in ['Left', 'Right']:
+        if event.key == 'Right':
+            step = 1
+        else:
+            step = -1
+        act = (act + step) % len(names)
+        interpolation = names[act]
+        image.interpolation = interpolation
+
+    if event.key in ['Up', 'Down']:
+        if first:
+            image.attach(iso)
+            first = False
+        if event.key == 'Up':
+            level += 1
+        else:
+            if level > 1:
+                level -= 1
+        #levels = np.linspace(img_data.min(), img_data.max(), num=level+1, endpoint=True)[1:-1]
+        #curve.levels = levels
+        iso.level = level
+    canvas.title = 'Spatial Filtering using %s Filter - Isoline %d level' % (interpolation, level)
+    canvas.update()
+
+
+if __name__ == '__main__' and sys.flags.interactive == 0:
+    app.run()
--- a/examples/basics/scene/contour.py
+++ b/examples/basics/scene/contour.py
@@ -24,10 +24,8 @@ interpolation = 'bicubic'
 img_data = read_png(load_data_file('mona_lisa/mona_lisa_sm.png'))
 image = scene.visuals.Image(img_data, interpolation=interpolation,
                             parent=view.scene, method='impostor')
-
 level = 10
-iso = IsolineFilter(level=2, width=1., color='white')
-image.attach(iso)
+iso = IsolineFilter(level=level, width=1., color='white')
 
 # Set 2D camera (the camera will scale to the contents in the scene)
 view.camera = scene.PanZoomCamera(aspect=1)
@@ -61,6 +59,14 @@ def on_key_press(event):
     canvas.title = ('Spatial Filtering using %s Filter - Isoline %d level'
                     % (image.interpolation, iso.level))
     canvas.update()
+
+
+# attaching of isoline filter via timer
+def on_timer1(event):
+    image.attach(iso)
+    canvas.update()
+
+timer1 = app.Timer('auto', iterations=1, connect=on_timer1, start=True)
 
 if __name__ == '__main__' and sys.flags.interactive == 0:
     app.run()
--- a/examples/basics/scene/contour.py
+++ b/examples/basics/scene/contour.py
@@ -34,7 +34,7 @@ img_data[1:4, 1::2] = 0.5
 img_data[1::2, 2] = 0.5
 img_data[2, 2] = 1.0
 
-#img_data = read_png(load_data_file('mona_lisa/mona_lisa_sm.png'))
+img_data = read_png(load_data_file('mona_lisa/mona_lisa_sm.png'))
 #img_data = np.dot(img_data[...,:3], [0.299, 0.587, 0.144])
 #img_data = img_data[...,0] * img_data[...,1] * img_data[...,2]
 #img_data /= 3.
--- a/examples/basics/scene/contour.py
+++ b/examples/basics/scene/contour.py
@@ -23,16 +23,55 @@ canvas.show()
 # Set up a viewbox to display the image with interactive pan/zoom
 view = canvas.central_widget.add_view()
 
+
+t = 2.23547566289471
+
+import struct, array
+
+#myarray1 = array.array('f')
+myarray = array.array('f')
+myarray.append(t)
+
+def unpack_helper(data):
+    str = data.tostring()
+    size = len(data) * 4
+    fmt = "!%dB" % size
+    print(fmt, size)
+    #size = struct.calcsize(fmt)
+    return struct.unpack(fmt, str)
+
+def pack_helper(data):
+    return struct.pack("!%sf" % len(data), *data)
+
+
+p = myarray.tostring()
+#p = struct.pack('!f', t)
+u = struct.unpack('!4B', p)
+print(t,p,u)
+
 # Create the image
 interpolation = 'bicubic'
 np.random.seed(1000)
 img_data = np.random.normal(size=(100, 100), loc=50, scale=150)
 img_data = gaussian_filter(img_data, (4, 4, 0)).astype(np.float32)
 
-img_data = np.zeros(25).reshape((5, 5)).astype(np.float32)
+
+
+img_data = np.zeros(25).reshape((5, 5)).astype(np.float32)# + 1.223
 img_data[1:4, 1::2] = 0.5
 img_data[1::2, 2] = 0.5
 img_data[2, 2] = 1.0
+
+myarray = array.array('f', img_data.ravel())
+#mystring = myarray.tostring()
+
+#d = mystring
+#p = pack_helper(mystring)
+#print(p)
+print(myarray)
+u = unpack_helper(myarray)
+print(u)
+
 
 img_data = read_png(load_data_file('mona_lisa/mona_lisa_sm.png'))
 #img_data = np.dot(img_data[...,:3], [0.299, 0.587, 0.144])
@@ -69,6 +108,7 @@ view.camera = scene.PanZoomCamera(aspect=1)
 # flip y-axis to have correct aligment
 view.camera.flip = (0, 1, 0)
 view.camera.set_range()
+
 
 # get interpolation functions from Image
 names = image.interpolation_functions
@@ -110,7 +150,6 @@ def on_key_press(event):
         iso.level = level
     canvas.title = 'Spatial Filtering using %s Filter - Isoline %d level' % (interpolation, level)
     canvas.update()
-
 
 if __name__ == '__main__' and sys.flags.interactive == 0:
     app.run()
--- a/examples/basics/scene/contour.py
+++ b/examples/basics/scene/contour.py
@@ -9,7 +9,7 @@ Simple use of SceneCanvas to display an Image.
 """
 import sys
 
-from vispy import scene, app, visuals
+from vispy import scene, app
 from vispy.visuals.filters import IsolineFilter
 from vispy.io import load_data_file, read_png
 
@@ -23,11 +23,10 @@ view = canvas.central_widget.add_view()
 interpolation = 'bicubic'
 img_data = read_png(load_data_file('mona_lisa/mona_lisa_sm.png'))
 image = scene.visuals.Image(img_data, interpolation=interpolation,
-                            parent=view.scene)
-image.transform = visuals.transforms.STTransform(translate=(0, 0, 0.5))
+                            parent=view.scene, method='impostor')
 
 level = 10
-iso = IsolineFilter(level=level, width=1., color='black')
+iso = IsolineFilter(level=2, width=1., color='white')
 image.attach(iso)
 
 # Set 2D camera (the camera will scale to the contents in the scene)
--- a/vispy/glsl/build-spatial-filters.py
+++ b/vispy/glsl/build-spatial-filters.py
@@ -129,29 +129,30 @@ class SpatialFilter(object):
         code += '    float w, w_sum = 0.0;\n'
         code += '    vec4 r = vec4(0.0,0.0,0.0,0.0);\n'
         for i in range(n):
-            code += '    w = texture2D(kernel, vec2(%f+(x/%.1f),index) ).r;\n' % (1.0 - (i + 1) / float(n), n)  # noqa
+            code += '    w = unpack_interpolate(kernel, vec2(%f+(x/%.1f), index));\n' % (1.0 - (i + 1) / float(n), n)  # noqa
             code += '    w = w*kernel_scale + kernel_bias;\n'  # noqa
             # code += '   w_sum += w;'
             code += '    r += c%d * w;\n' % i
-            code += '    w = texture2D(kernel, vec2(%f-(x/%.1f),index) ).r;\n' % ((i+1)/float(n), n)  # noqa
+            code += '    w = unpack_interpolate(kernel, vec2(%f-(x/%.1f), index));\n' % ((i+1)/float(n), n)  # noqa
             code += '    w = w*kernel_scale + kernel_bias;\n'
             # code += '   w_sum += w;'
             code += '    r += c%d * w;\n' % (i + n)
         # code += '    return r/w_sum;\n'
         code += '    return r;\n'
         code += '}\n'
+        code += "\n"
         code += 'vec4\n'
         code += '%s' % filter_2
-        code += '(sampler2D texture, sampler2D kernel, float index, vec2 uv, vec2 pixel )\n'  # noqa
+        code += '(sampler2D texture, sampler2D kernel, float index, vec2 uv, vec2 pixel)\n'  # noqa
         code += '{\n'
-        code += '    vec2 texel = uv/pixel - vec2(0.5,0.5) ;\n'
+        code += '    vec2 texel = uv/pixel - vec2(0.5, 0.5) ;\n'
         code += '    vec2 f = fract(texel);\n'
-        code += '    texel = (texel-fract(texel)+vec2(0.001,0.001))*pixel;\n'
+        code += '    texel = (texel-fract(texel) + vec2(0.001, 0.001)) * pixel;\n'  # noqa
         for i in range(2*n):
             code += '    vec4 t%d = %s(kernel, index, f.x,\n' % (i, filter_1)
             for j in range(2*n):
                 x, y = (-n+1+j, -n+1+i)
-                code += '        texture2D( texture, texel + vec2(%d,%d)*pixel),\n' % (x, y)  # noqa
+                code += '        texture2D( texture, texel + vec2(%d, %d) * pixel),\n' % (x, y)  # noqa
 
             # Remove last trailing',' and close function call
             code = code[:-2] + ');\n'
@@ -198,8 +199,8 @@ class Nearest(SpatialFilter):
     def _get_code(self):
         self.build_LUT()
         code = 'vec4\n'
-        code += 'interpolate( sampler2D texture, sampler1D kernel, vec2 uv, vec2 pixel )\n'  # noqa
-        code += '{\n   return texture2D( texture, uv );\n}\n'
+        code += 'interpolate(sampler2D texture, sampler1D kernel, vec2 uv, vec2 pixel)\n'  # noqa
+        code += '{\n   return texture2D(texture, uv);\n}\n'
         return code
     code = property(_get_code, doc='''filter functions code''')
 
@@ -646,7 +647,7 @@ for i, f in enumerate(filters):
 bias = K.min()
 scale = K.max()-K.min()
 K = (K-bias)/scale
-np.save("spatial-filters.npy", K.astype(np.float32))
+# np.save("spatial-filters.npy", K.astype(np.float32))
 
 print("// ------------------------------------")
 print("// Automatically generated, do not edit")
@@ -654,8 +655,41 @@ print("// ------------------------------------")
 print("")
 print("const float kernel_bias  = %f;" % bias)
 print("const float kernel_scale = %f;" % scale)
+print("const float kernel_size = %f;" % n)
+print("const float M_2_m23 = 0.00000011920928955078125;")
 print("uniform sampler2D u_kernel;")
 print("")
+
+code = 'float\n'
+code += 'unpack(vec4 rgba)\n'
+code += '{\n'
+code += '\t// return rgba.r;  // uncomment this for r32f debugging\n'
+code += '\trgba.rgba = rgba.abgr * 255;\n'
+code += '\tfloat sign = 1.0 - step(128.0,rgba[0])*2.0;\n'
+code += '\tfloat exponent = 2.0 * mod(rgba[0],128.0) + ' \
+        'step(128.0,rgba[1]) - 127.0;\n'
+code += '\tfloat mantissa = mod(rgba[1],128.0)*65536.0 + rgba[2]*256.0 + ' \
+        'rgba[3] + float(0x800000);\n'
+code += '\treturn sign * exp2(exponent) * (mantissa * M_2_m23);\n'
+code += '}\n'
+print(code.expandtabs(4))
+
+code = 'float\n'
+code += 'unpack_interpolate(sampler2D kernel, vec2 uv)\n'
+code += '{\n'
+code += '\t// return texture2D(kernel, uv).r; ' \
+        '//uncomment this for r32f debug without interpolation\n'
+code += '\tfloat kpixel = 1/kernel_size;\n'
+code += '\tfloat u = uv.x / kpixel;\n'
+code += '\tfloat v = uv.y;\n'
+code += '\tfloat uf = fract(u);\n'
+code += '\tu = (u - uf) * kpixel;\n'
+code += '\n'
+code += '\tfloat d0 = unpack(texture2D(kernel, vec2(u, v)));\n'
+code += '\tfloat d1 = unpack(texture2D(kernel, vec2(u + 1 * kpixel, v)));\n'
+code += '\treturn mix(d0, d1, uf);\n'
+code += '}\n'
+print(code.expandtabs(4))
 
 F = SpatialFilter(1.0)
 print(F.filter_code())
--- a/vispy/glsl/build-spatial-filters.py
+++ b/vispy/glsl/build-spatial-filters.py
@@ -656,16 +656,29 @@ print("")
 print("const float kernel_bias  = %f;" % bias)
 print("const float kernel_scale = %f;" % scale)
 print("const float kernel_size = %f;" % n)
+print("const vec4 bits = vec4(1.0, 1.0/256.0, 1.0/(256.0*256.0), 1.0/(256.0*256.0*256.0));")  # noqa
 print("uniform sampler2D u_kernel;")
 print("")
 
 code = 'float\n'
-code += 'unpack(vec4 rgba)\n'
+code += 'unpack_unit(vec4 rgba)\n'
+code += '{\n'
+code += '\t// return rgba.r;  // uncomment this for r32f debugging\n'
+code += '\treturn dot(rgba, bits);\n'
+code += '}\n'
+print(code.expandtabs(4))
+
+code = 'float\n'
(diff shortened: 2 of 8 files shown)


Eric89GXL:
Looks reasonable to me.

This won't be compatible with WebGL unfortunately. What we could do instead is pack 32 bits of information in a standard RGBA texture by splitting it up. We do something very similar to fit two 16 bit numbers in a standard RGBA texture for text rendering. It might be worth doing that here, as we would be able to maintain high precision even in WebGL. @kmuehlbauer you up for trying that method?

Eric89GXL:
In the `sdf` renderer for text, it's done here:

https://github.com/vispy/vispy/blob/master/vispy/visuals/text/_sdf.py#L68

kmuehlbauer:
I'll give it a try, but may come back with a bunch of questions...

Eric89GXL:
This might do it for you:

http://stackoverflow.com/questions/18453302/how-do-you-pack-one-32bit-int-into-4-8bit-ints-in-glsl-webgl

Eric89GXL:
Ahh no, that's a 32 bit integer, you want a 32 bit float packed into 4 8-bit ints, hmm...

Eric89GXL:
Perhaps this instead:

http://www.gamedev.net/topic/486847-encoding-16-and-32-bit-floating-point-value-into-rgba-byte-texture/

kmuehlbauer:
To understand a bit more, the 32bit floats in a Texture2D is then internally converted to 16bit float to comply with WEBGL?

Eric89GXL:
No, WebGL can only handle standard 8-bit textures. So you need to use the internalformat RGBA, which is 4 bytes (stored internally as 4 8-bit unsigned integers). So if what you really want to store is one 32-bit float per "pixel" in the texture like you do here, what you need to do is break that 32-bit float into 4 8-bit numbers for storage, and upon retrieval, convert those 4 8-bit numbers back into your desired 32-bit number.

In the text code, what it does is store two 16-bit numbers in a standard RGBA texture by breaking up each 16-bit number into 2 8-bit numbers. Make sense?

kmuehlbauer:
The normal approach as in the gamedev-link above works, but the precision isn't good enough. So this won't work.

But if we try to encode/decode the binary IEEE representation of the float instead, then it should be possible to enhance precision within the shader. At least this stackoverflow  post claims this.

http://stackoverflow.com/questions/7059962/how-do-i-convert-a-vec4-rgba-value-to-a-float

So, I implemented this approach within ImageVisual (encoding) and fragment shader decoding. And it works actually very well. There are some glitches on the isolines, but that come into play first when the zoom level is really high. I have prepared four images to show the differences:

This is actual implementation using kernel texture with 'r16f' and 'nearest'.
float16_nearest

This is actual implementation using kernel texture with 'r16f' and 'linear'.
float16_linear

This is suggested implementation using kernel texture with 'r32f' and 'linear'. Doesn't work for WebGL.
float32_linear

This is using packing/unpacking float to 'rgba8' kernel texture and 'linear'. Here you see also some outlines of the original image pixels.
float_rgba_linear

This is using packing/unpacking float to 'rgba8' kernel texture and 'nearest'.
float_rgba_nearest

As it is seen, the last one outperforms the other implementation (except 'r32f') with respect to isoline smoothness. 

Comments on the stackoverflow post also claim that this approach also has some problems (which we actually see). But we could take this as a starter to refine and possibly fix those errors.

For this PR the question is, where should it be implemented? Can Texture2D be extended for that. WDYT?

rougier:
One problem with this technique is that we don't have linear interpolation anymore, do we ?

Eric89GXL:
Not at the hardware level, you'd need to do it yourself in glsl instead

kmuehlbauer:
Yes we would have to choose 'nearest' for the kernel texture interpolation. Why is that a problem?

Eric89GXL:
What does r32f look like with nearest? I'm curious if it looks worse than
linear.

We can do the linear interpolation ourselves after conversion to float if
we need to, so if linear looks better for r32f it might look better
implemented in glsl for our packed floats.

rougier:
The glitches you see probably come from this nearest interpolation.

kmuehlbauer:
Have a look, the glitches are more pronounced when using 'linear' in the packed implementation.

rougier:
Yes, because interpolating the float encoding won't work. You have to emulate bilinear filtering in the shader (without using the kernel of course since the goal is to use this bilnear interpolation to extract kernel values for filters)

kmuehlbauer:
@Eric89GXL the '32f', 'nearest" is same to packed 'rgba8', 'nearest'. Does that mean that the packing/unpacking is working correctly?

kmuehlbauer:
@rougier sorry, this took a while to settle, yes we have to work on the unpacked float, not on the 'rgba8'.

Eric89GXL:
`r32f` `nearest` being the same as `rgb8` `nearest` does indicate your packing and unpacking is working correctly, which is great.

@kmuehlbauer I think we can create a `linear` interpolation in GLSL ourselves while having `nearest` as the hardware mode for `rgb8`. In other words, the procedure would be to 1) look up the 4 necessary pixels in the texture, 2) unpack all four of those `rgb8` values to `float`, then 3) do the linear interpolation properly from those four points. That way we can have `linear` interpolation on the floats working properly while using `rgb8` textures for storage. Does that make sense?

kmuehlbauer:
@Eric89GXL yes, this seems like it should work.

Eric89GXL:
You up for giving it a go? You might be able to look at @rougier's interpolation GLSL code a bit to make it easier.

kmuehlbauer:
Yes, having it on the screen right now. Had to put in the unpack there anyway.  

The implementation for bicubic in @rougier spatial-filters for instance is like that:

1. get the texel
2. get integral and fractional part of texel
3. get one row of texture (4 pixel colors) from the image texture 
4. feed this row to kernel lookup function
5. get 4 colors from kernel texture (4 texture calls), combine with weights, scale and bias and return interpolated color
6. go to 3. and do for 3 more rows
7. do 5. with the 4 row colors and get final interpolated color

In 5. when the kernel texture call is made, it is assumed that those lookups come from an interpolated texture, which we don't have. So it would take 'just' take two kernel texture calls (one at texel-integral part and one at texel integral part +1). Then we could `mix()` those two together with the fractional part of the texel. Right?

Eric89GXL:
Just about -- I think it actually takes 4 calls, not 2, to do the linear interpolation because it's bilinear interpolation. In other words you do a weighting of four points, not two, to get the linearly interpolated value.

kmuehlbauer:
@Eric89GXL  yes, but the kernel texture is just 1D. The 4 point thing is in 7.

Eric89GXL:
ahhhh, you're right, we just need the 1D texture lookup to be enhanced -- sounds good to me

kmuehlbauer:
This is the call where the kernel texture is called:

```
float w, w_sum = 0.0;
vec4 r = vec4(0.0,0.0,0.0,0.0);
w = texture2D(kernel, vec2(0.500000+(x/2.0), index)).r;
w = w*kernel_scale + kernel_bias;
r += c0 * w;
```
c0 is one row color from the image texture.

As I understand, the vec2(0.50000+(x/2.0), index) is the position from which the kernel value is gotten, index indicates the interpolation (bicubic etc).

But I'm lost a bit between which kernel texture coordinates I have to interpolate. @rougier do you have an idea, or won't it go anyway?

I get results but they indicate that my coordinates of the kernel texture are wrong somehow.

kmuehlbauer:
You can dig here a little, if you want to have a look at the WIP code.
https://github.com/kmuehlbauer/vispy/tree/texturepacking

rougier:
The 1d coordinate is `float u = 0.500000+(x/2.0)`, so you have to interpolate between `(u-fract(u)) and (u-fract(u) + 1)`

kmuehlbauer:
Seems to work, I get smooth isolines without glitches, but they appear to have "corners" at the pixel boundaries of the image pixels.

Eric89GXL:
Can you post an image, I can't quite picture that. But it sounds like we're on the right track at least.

rougier:
I do not know how you did the interpolation, but what I wrote might be wrong. I think you need the texure width at some point:
```
float u = (0.500000+(x/2.0)) * tex_width
float u1 =  (u-fract(u)) / tex_width;
float u2 =  (u-fract(u) + 1.0) / tex_width;
```

kmuehlbauer:
The colors also aren't correct so the isolines are not in the right place. Maybe this is interconnected.

isoline_corners

kmuehlbauer:
The colors are more brighter. Can it be connected to the mix function, I actually have the fract(u) as mixing parameter @rougier.

kmuehlbauer:
tex_width from kernel texture? Is it somewhere in spatial-filters (pixels?)

kmuehlbauer:
@rougier this is the adapted code in spatial-filters just for bicubic and some others.

```
float lerp(sampler2D kernel, vec2 uv) {
    vec2 texel = uv;
    float u = uv.x;
    float v = uv.y;
    float uf = fract(u);

    float d0 = unpack(texture2D(kernel, vec2(u - uf, v)));
    float d1 = unpack(texture2D(kernel, vec2(u - uf + 1, v)));

    return mix(d0, d1, uf);
}

vec4
filter1D_radius2( sampler2D kernel, float index, float x, vec4 c0, vec4 c1, vec4 c2, vec4 c3 )
{
    float w, w_sum = 0.0;
    vec4 r = vec4(0.0,0.0,0.0,0.0);
    w = lerp(kernel, vec2(0.500000+(x/2.0),index));//.r;
    w = w*kernel_scale + kernel_bias;
    r += c0 * w;
    w = lerp(kernel, vec2(0.500000-(x/2.0),index));//.r;
    w = w*kernel_scale + kernel_bias;
    r += c2 * w;
    w = lerp(kernel, vec2(0.000000+(x/2.0),index));//.r;
    w = w*kernel_scale + kernel_bias;
    r += c1 * w;
    w = lerp(kernel, vec2(1.000000-(x/2.0),index));//.r;
    w = w*kernel_scale + kernel_bias;
    r += c3 * w;
    return r;
}
```

kmuehlbauer:
I've made two images, one with the bicubic interpolation using 'r32f', 'linear' and the second with packed 'rgba' and the new approach. As you can see, the image is brighter. Maybe someone has an idea, where this might be located or what to check. 
monalisa_normal_bicubic
monalisa_normal_unpack

Eric89GXL:
Can you link into the right part of your code in your branch? Or even better, just push your changes to this branch. (I notice your branch is `master` -- this is one reason I suspect most people generally develop on something other than `master`. You can have `master` always be the one that should work, and sandbox your WIP changes somewhere else.)

Eric89GXL:
It looks to my eye like the colors are about twice as bright as they should be. You can check by making a simple image that goes from black to white from left to right, and looking to see if the interpolated version goes to white at the middle instead of the end.

kmuehlbauer:
@Eric89GXL yes, originally, I thought the small change in the beginning was good in master.

You can have a look in the branch I linked above (texturepacking). I'll push the latest version in a minute.

Eric89GXL:
I saw you linked that branch above, but it's harder to do the comparison without setting it up manually, etc. We might as well use GitHub for what it's good at, making nice diffs and facilitating discussion in the lines of code itself :)

Eric89GXL:
They don't copy the `test` folder, either. They copy everything else because they need it.

Maybe you didn't notice, but I actually have `h5io` structured so that `_h5io.py` is the only file with code in it. Other than that there is just the `__init__.py`. So any updates to h5io will be in that single file...

Eric89GXL:
Compare:

https://github.com/scikit-learn/scikit-learn/tree/master/sklearn/externals/joblib
https://github.com/joblib/joblib/tree/master/joblib

Eric89GXL:
Rats, wrong thread, ignore those :)

kmuehlbauer:
What do you suggest then? Should I open another PR with the other branch?

Eric89GXL:
Sure, that would be fine. But it would be a little bit better not to open another PR since that way all the discussion will stay in the same thread (this one). So pushing your changes to your `master` branch would be best. In the meantime you could set up some temporary master if you need something fully functional like `git checkout -b master-real upstream/master`. But if that's too much of a pain, another PR is also okay.

kmuehlbauer:
OK, the changes are on master now...

vispy/glsl/misc/spatial-filters.frag
```diff
@@ -6,6 +6,15 @@ const float kernel_bias  = -0.234377;
 const float kernel_scale = 1.241974;
 uniform sampler2D u_kernel;
 
+float unpack(vec4 rgba) {
+    //return rgba.r;
+    rgba.rgba = rgba.abgr * 255;
+    float sign = 1.0 - step(128.0,rgba[0])*2.0;
+    float exponent = 2.0 * mod(rgba[0],128.0) + step(128.0,rgba[1]) - 127.0;
+    float mantissa = mod(rgba[1],128.0)*65536.0 + rgba[2]*256.0 + rgba[3] + float(0x800000);
+    return sign * exp2(exponent) * (mantissa * exp2(-23.0 ));
```
Eric89GXL: this looks like it would have been a pain to figure out!

vispy/visuals/image.py
```diff
@@ -87,6 +87,21 @@
...
     }"""
 
+import array, struct
+
+def floatToFixed(f):
+    str = array.array('f', f.ravel()).tostring()
+    print(len(str))
+    size = f.size * 4
+    fmt = "!%dB" % size
+    print(fmt, size, f.shape + (4,))
```
Eric89GXL: There is probably a one-liner way to do this with `numpy` directly, like `np.fromstring(np.array(f, np.float32).tostring(), 'b')` or so

vispy/glsl/misc/spatial-filters.frag
kmuehlbauer: We can test this by testing with a normal image texture2D.

vispy/glsl/misc/spatial-filters.frag
```diff
@@ -34,29 +44,60 @@ filter2D_radius1(sampler2D texture, sampler2D kernel, float index, vec2 uv, vec2
     return filter1D_radius1(kernel, index, f.y, t0, t1);
 }
 
+float lerp(sampler2D kernel, vec2 uv) {
```
Eric89GXL: just a point of curiosity, why `lerp`? Linear interp? Maybe we should make more explicit like `unpack_interpolate(...)`

vispy/glsl/misc/spatial-filters.frag
kmuehlbauer: In the stackoverflow they said that there might be glitches with some values. I'll try your black and white thing first.

vispy/glsl/misc/spatial-filters.frag
```diff
@@ -34,29 +44,60 @@ filter2D_radius1(sampler2D texture, sampler2D kernel, float index, vec2 uv, vec2
...
+    vec2 texel = uv;
+    float u = texel.x;
+    float v = texel.y;
+    float uf = fract(u);
+
+    float d0 = unpack(texture2D(kernel, vec2(u - uf, v)));
+    float d1 = unpack(texture2D(kernel, vec2(u - uf + 1, v)));
+    return mix(d0, d1, uf);
```
Eric89GXL: I expected there to be some factor of two problem in this function, but I don't see one...

I'll try playing with the code to see if I can figure it out.

vispy/glsl/misc/spatial-filters.frag
Eric89GXL: Yeah, it doesn't look like your test image is the result of glitchy values, but something more systematic

vispy/glsl/misc/spatial-filters.frag
Eric89GXL: If I have this just return d0, I get something very bright. If I just return d1, I get something black. So I think the error is in here somewhere.

Eric89GXL: Hmm, maybe not...

kmuehlbauer:
This is what you had in mind...

linear_gradient

vispy/glsl/misc/spatial-filters.frag
kmuehlbauer: yeah sure, this is work in progress, I'll clean up if this gets ever solved ...

vispy/visuals/image.py
kmuehlbauer: This would be very nice, I tried with numpy but got no solution, so I took this one which worked for testing.

Eric89GXL:
Interesting, that's not even monochrome :) It's also off by a factor of about two. Some clues I suppose but I don't know what they mean exactly.

Eric89GXL:
I mean grayscale not monochrome

kmuehlbauer:
This is the cubehelix colormap which is appended in the image visual...

vispy/glsl/misc/spatial-filters.frag
```diff
@@ -34,29 +44,60 @@ filter2D_radius1(sampler2D texture, sampler2D kernel, float index, vec2 uv, vec2
     return filter1D_radius1(kernel, index, f.y, t0, t1);
 }
 
+float lerp(sampler2D kernel, vec2 uv) {
+    vec2 texel = uv;
```
kmuehlbauer: multiply by 2 here, then it is almost done.

Eric89GXL:
Ahh, okay. So in that case, it does look like there is some factor of two appearing somewhere. But if I change all the filters over to the new format, they actually all appear to work, but some have a factor of two (or more) brightness problem, while others are correct (e.g., bilinear is correct).

I can send a PR into your branch if you want to see what I'm talking about.

vispy/glsl/misc/spatial-filters.frag
Eric89GXL: I tried that, but it makes some of the other filters worse.

kmuehlbauer:
It seems, that this also changes the row-index into the kernel which selects the interpolation function. So this may be just coincidence that this works. Also all functioning interpolations seems the same to me.

kmuehlbauer:
uv : texcoord
pixel : 1/shape
index : selects interpolation within kernel texture

This is the way of the texcoord to the final coord into the kernel texture. Ideas?  
```
uv/pixel - vec2(0.5,0.5);
# we use this fractional part
vec2 f = fract(texel); 
# and only the first element
x = f.x
# this is the first of four texels which will be called 
texel = vec2(0.500000+(x/2.0),index)
# 1D coord
float u = texel.x;
# interpolation index
float v = texel.y;
# fractional part 
float uf = fract(u);
# get first of two kernel values
float d0 = unpack(texture2D(kernel, vec2(u - uf, v)));
float d1 = unpack(texture2D(kernel, vec2(u - uf + 1, v)));
color = mix(d0, d1, uf);
```

Eric89GXL:
@kmuehlbauer can you hop on `gitter`? maybe we can work it out there

kmuehlbauer:
yes sure, but where to go there?

Eric89GXL:
@rougier the artifacts are gone now, thanks to your advice about the `width` division issue.

@kmuehlbauer left to do:

1. clean up warnings during example
2. look where else `spatial-filters.frag` is used, those places/examples will need to be updated
3. improve example aesthetics, probably by zooming in a bit so the interp effect can be seen, and add some isolines (10 again?)

Let me know when that's done and I'll have another look. Thanks for hashing out the solution with us!

kmuehlbauer:
After some rest I did test the linear texture interpolation with 'rf32'. With the monalisa the eye can be betrayed, but the 5x5 matrix shows the differences:

This is using 'r32f' kernel texture with 'nearest' and the glsl linear kernel interpolation:
float_r32f_nearest_interpolate

This is using 'r32f' kernel texture with 'nearest' (you would see the glitches when zoomin in):
float_r32f_nearest

This is using 'r32f' kernel texture with 'linear' (no glitches at all)
float_r32f_linear

So there is still something wrong with the linear kernel texture interpolation (`unpacking_interpolation`). 

One thing is with the `pixel` that we always use the `pixel.x` even we interpolate in the y-direction. This is for bicubic (and some others): First the four rows of the 4x4 pixel matrix are interpolated, last the output of the 4 rows are treated the same way. I already read this here where this spatial filter approach is described but can't figure out where the problem might be located: http://http.developer.nvidia.com/GPUGems/gpugems_ch24.html. 

@rougier where do we make wrong assumptions?

rougier:
I'm a bit lost now on what you're testing now. Di you see my post above about the ``tex_width`` stuff ?

From what I understood, I would say this is wrong:
```
float d0 = unpack(texture2D(kernel, vec2(u - uf, v)));
float d1 = unpack(texture2D(kernel, vec2(u - uf + 1, v)));
```

Instead, I would write:

```
float d0 = unpack(texture2D(kernel, vec2((u - uf)*pixel, v)));
float d1 = unpack(texture2D(kernel, vec2((u - uf + 1)*pixel, v)));
```

kmuehlbauer:
Yes, I have it, we don not need the `pixel` from the image texture, we need it from the kernel texture. I don't know if we have access to this value (1024) within the shader, so I just hardcoded it for the moment.

kmuehlbauer:
@ rougier Hang on for a minute, all push my changes in a minute.

kmuehlbauer:
@Eric89GXL @rougier It works now also with the packed filter, but further testing wouldn't be that bad.

Also we have to incorporate these changes in `build-spatial-filters.py`, and recreate the `spatial-filters.frag before merging.

rougier:
Nice. Can you post the result on Mona Lisa for example ?

kmuehlbauer:
To compare, here is the image using the kernel with rgba packed floats and nearest with the linear kernel interpolation in the shader:

float_packed_rgba

kmuehlbauer:
@rougier do you meant this monalisa zoom with isoline? You can compare to the others above.

rougier:
I mean something like this (zoomed detail from [isocurves-lena.py](https://github.com/glumpy/glumpy/blob/master/examples/isocurves-lena.py) in glumpy)

isocurves-lena

kmuehlbauer:
This is monalisas left eye (or the right one, if you look at it):
monalisas_left_eye

kmuehlbauer:
@Eric89GXL @rougier 

This is ready for code and functional review and testing I think. If all is good, I could do the last TODO next week.

TODO:
- change build-spatial-filters.py and recreate spatial-filters.frag

Further improvements (maybe later PR)
- optimize unpack function
- make isoline function 3d
- more ideas concerning the rgba-packing and texture stuff?

vispy/glsl/misc/spatial-filters.frag
```diff
@@ -4,23 +4,47 @@
...
+    rgba.rgba = rgba.abgr * 255;
+    float sign = 1.0 - step(128.0,rgba[0])*2.0;
+    float exponent = 2.0 * mod(rgba[0],128.0) + step(128.0,rgba[1]) - 127.0;
+    float mantissa = mod(rgba[1],128.0)*65536.0 + rgba[2]*256.0 + rgba[3] + float(0x800000);
+    return sign * exp2(exponent) * (mantissa * exp2(-23.0));
+}
+
+float unpack_interpolate(sampler2D kernel, vec2 uv) {
+    //return texture2D(kernel, uv).r; //uncomment this for r32f debug without interpolation
+    float kpixel = 1/kernel_size;
```
Eric89GXL: So my error was thinking `shape` was related to the `kernel` when really it was related to the data to be interpolated. Makes sense

Eric89GXL:
There are some rendering artifacts for `method='subdivide'` in `Image`, but I'm not sure what would be required to fix them. Perhaps we should use `method='impostor'` instead...?

examples/basics/scene/contour.py
```diff
@@ -0,0 +1,67 @@
...
+canvas.show()
+
+# Set up a viewbox to display the image with interactive pan/zoom
+view = canvas.central_widget.add_view()
+
+interpolation = 'bicubic'
+img_data = read_png(load_data_file('mona_lisa/mona_lisa_sm.png'))
+image = scene.visuals.Image(img_data, interpolation=interpolation,
+                            parent=view.scene)
+image.transform = visuals.transforms.STTransform(translate=(0, 0, 0.5))
```
Eric89GXL: I don't actually need this transform for the example to run, can you remove it and the import of `visuals` above?

vispy/glsl/misc/spatial-filters.frag
```diff
@@ -4,23 +4,47 @@
 
 const float kernel_bias  = -0.234377;
 const float kernel_scale = 1.241974;
+const float kernel_size = 1024;
```
Eric89GXL: might as well make this `const float kpixel = 1. / 1024.`

vispy/glsl/misc/spatial-filters.frag
```diff
@@ -4,23 +4,47 @@
...
 uniform sampler2D u_kernel;
 
+float unpack(vec4 rgba) {
+    //return rgba.r;  // uncomment this for r32f debugging
+    rgba.rgba = rgba.abgr * 255;
+    float sign = 1.0 - step(128.0,rgba[0])*2.0;
+    float exponent = 2.0 * mod(rgba[0],128.0) + step(128.0,rgba[1]) - 127.0;
+    float mantissa = mod(rgba[1],128.0)*65536.0 + rgba[2]*256.0 + rgba[3] + float(0x800000);
+    return sign * exp2(exponent) * (mantissa * exp2(-23.0));
```
Eric89GXL: pou could make `exp2(-23.0)` a constant to save a little bit of computation

vispy/glsl/misc/spatial-filters.frag
```diff
@@ -4,23 +4,47 @@
...
+float unpack_interpolate(sampler2D kernel, vec2 uv) {
+    //return texture2D(kernel, uv).r; //uncomment this for r32f debug without interpolation
+    float kpixel = 1/kernel_size;
+    float u = uv.x / kpixel;
+    float v = uv.y;
+    float uf = fract(u);
+    u = (u - uf) * kpixel;
+
+    float d0 = unpack(texture2D(kernel, vec2(u, v)));
+    float d1 = unpack(texture2D(kernel, vec2(u + 1 * kpixel, v)));
```
Eric89GXL: no need for the `1 * ` here

rougier: I would keep it to suggest  "1 pixel"

Eric89GXL:
A few minor things:

* ~~It looks like in the example isoline 1 level does not show any isoline. Shouldn't it show one at the middle intensity of the image?~~ Once I turned `antialias=0., color='white'` I could see some specks of color, so I guess there just aren't that many colors at the midpoint.

* The default choices of `width` and `antialias` do not make the line actually show up in the expected color. Try it with `color='white'` and you'll see what I mean, everything shows up gray. Either the GLSL for the antialiasing should be tweaked, or `width` and `antialias` should be changed so that if someone picks a color, they see that color in the lines. If this turns out to be difficult, it could be done in a separate PR.

Other than that and my minor comments, LGTM!

vispy/glsl/misc/spatial-filters.frag
Eric89GXL: For clarity I guess?

kmuehlbauer:
Just to clarify:

* `level` - the number of 'regions', which are divided by **one** isoline. This means the number of isolines is `numiso = level - 1`. Default value should be `level=2`.
* `width` - linewidth in screen pixels. Default value is 1.
* `antialiasing` - amount of shading at the isoline edge in screen pixels. Default is 1.

Regarding the colors. There was a problem with wrongly using alpha with the foreground color. I fixed this, so the isoline color is correctly mixed with the background color. 

Anyway, if you chose `white` with the above default values, you would get more or less `lightgrey` lines. Only for straight lines, you would get the one white pixel in the middle. I didn't find a solution to this, also black lines show up as `darkgrey`. Therefore I decided to let the user apply `width` and `antialiasing` to fit them to their needs.

kmuehlbauer:
I'm currently adding the changes to `build-spatial-filters.py`.

We have `spatial-filters.frag` currently  used in:

* example `spatial_filters.py` https://github.com/vispy/vispy/blob/master/examples/basics/gloo/spatial_filters.py
* glsl shader `sdf-glyph.frag` https://github.com/vispy/vispy/blob/master/vispy/glsl/collections/sdf-glyph.frag
* ImageVisual `image.py`

The sdf-glyph.frag isn't use anywhere, but the TextVisual has a reimplementation of the CatRom interpolator using the kernel. The changes in `build-spatial-filters.py` do only reflect the changes for ImageVisual.

@rougier, @Eric89GXL how should this be tackled? Shouldn't there be some check if packed or unpacked is used?

Eric89GXL:
Can you update the text visual to use your enhanced code?

Eric89GXL:
In other words, always use packed

kmuehlbauer:
Could we have `io.load_spatial_filters()` return kernel or packed kernel8 via parameter?

Eric89GXL:
Sure, that would be fine. But I'd prefer if for now everything used the packed one instead so we maintain ES support, though. Does TextVisual use the r32f version?

kmuehlbauer:
@Eric89GXL It seems, that TextVisual and its underlying code just loads the kernel data, without using the interpolation code in `spatial-filters.frag` (because it uses it's own implementation). I've found no reference to 'r32f' but found 'rgba'. 

So, for the time being only the spatial-filters.py example and ImageVisual use the spatial-filters.frag.

Eric89GXL:
Great. Eventually we do want `TextVisual` to use the same filtering code instead of its own code -- the `TextVisual` code was in `vispy` before the spatial-filters stuff. We can do that as a separate PR, though, unless you want to change it here.

Is this ready to go from your end now?

kmuehlbauer:
No, not yet, I have to finalize the changes in `build-spatial-filters.py`. This is the script which is actually creating the `spatial-filters.frag`

Eric89GXL:
Cool. I think for now we can try just having the packed version, it's easy enough to add back the unpacked version if we really need it.

kmuehlbauer:
OK, this took some while to clean up.

* packed version is now default in spatial-filters.frag
* `io.load_spatial_filters()` returns packed kernel by default
* isoline filter works now only on the luminance of the parent visual, no fuzz with the parent visual color_transform 
* minor changes to improve example performance 
* minor changes in ImageVisual (comments and code readability)

Should be ready from my end now.

kmuehlbauer:
@Eric89GXL 

Out of curiosity I tested this approach again (http://www.gamedev.net/topic/486847-encoding-16-and-32-bit-floating-point-value-into-rgba-byte-texture/)

Obviously I had something wrong at the early stages of this implementation, because this works now also as expected. So we might change the pack/unpack function to this more readable and faster implementation.

The decoding function in the shader looks like this:
```
float unpack(vec4 rgba)
{
  rgba = rgba.abgr * 255;
  const vec4 bitSh = vec4(1.0/(256.0*256.0*256.0), 1.0/(256.0*256.0), 1.0/256.0, 1.0);
  return(dot(rgba, bitSh));
}
```

Eric89GXL:
If you could change it to the faster implementation that would be great.
But that doesn't look equivalent to me, it looks like that would store only
floats between zero and one. But maybe that's what we need here and not
something that works for arbitrary magnitude floats?

Eric89GXL:
Or bounded by zero and about 255, rather.

kmuehlbauer:
This is the python side packing, so far:

```
def pack_float_to_rgba(value):
        pack = np.zeros(value.shape + (4,))
        vec = [1., 256., 256., 256.]
        for i, v in enumerate(vec):
            value, pack[..., i] = np.modf(value * v)
        return pack
```

kmuehlbauer:
And yes, this works for [0,1] floats. This is what texture2D is using, right?

Eric89GXL:
The current packing function is more general, in that it can store any 32-bit float. The new version should only work for floats in the range (0, 1). That's okay for this particular application since the values are all in the range (0, 1) (I think?), but we should probably keep the other packing function around in case we ever want the full range of floating point values.

vispy/io/datasets.py
```diff
@@ -51,4 +54,18 @@ def load_spatial_filters():
              "Mitchell", "Spline16", "Spline36", "Gaussian",
              "Bessel", "Sinc", "Lanczos", "Blackman", "Nearest")
 
-    return (np.load(op.join(DATA_DIR, 'spatial-filters.npy')), names)
+    def pack_float_to_rgba(value):
+        pack = np.zeros(value.shape + (4,))
+        vec = [1., 256., 256., 256.]
+        for i, v in enumerate(vec):
+            value, pack[..., i] = np.modf(value * v)
+        return pack
```
Eric89GXL: don't nest this function

vispy/io/datasets.py
```diff
@@ -51,4 +54,18 @@ def load_spatial_filters():
              "Mitchell", "Spline16", "Spline36", "Gaussian",
              "Bessel", "Sinc", "Lanczos", "Blackman", "Nearest")
 
-    return (np.load(op.join(DATA_DIR, 'spatial-filters.npy')), names)
+    def pack_float_to_rgba(value):
+        pack = np.zeros(value.shape + (4,))
```
Eric89GXL: `... (4,), dtype=np.ubyte)` will make it so you don't need to cast later

vispy/glsl/build-spatial-filters.py
```diff
@@ -654,9 +655,37 @@ def weight(self, x):
...
+print("const float kernel_size = %f;" % n)
 print("uniform sampler2D u_kernel;")
 print("")
 
+code = 'float\n'
+code += 'unpack(vec4 rgba)\n'
+code += '{\n'
+code += '\t// return rgba.r;  // uncomment this for r32f debugging\n'
+code += '\trgba.rgba = rgba.abgr * 255;\n'
+code += '\tconst vec4 bits = vec4(1.0/(256.0*256.0*256.0), 1.0/(256.0*256.0), 1.0/256.0, 1.0);\n'
```
Eric89GXL: you can move this outside the function to make it clear it's constant regardless of the input

vispy/glsl/build-spatial-filters.py
```diff
@@ -654,9 +655,37 @@ def weight(self, x):
 print("")
 print("const float kernel_bias  = %f;" % bias)
 print("const float kernel_scale = %f;" % scale)
+print("const float kernel_size = %f;" % n)
 print("uniform sampler2D u_kernel;")
 print("")
 
+code = 'float\n'
+code += 'unpack(vec4 rgba)\n'
```
Eric89GXL: let's call this `unpack_unit` (and same thing for the `pack` functiona above) because it only packs/unpacks things in the unit range, not any float

vispy/glsl/build-spatial-filters.py
```diff
@@ -654,9 +655,37 @@ def weight(self, x):
...
+code += '{\n'
+code += '\t// return rgba.r;  // uncomment this for r32f debugging\n'
+code += '\trgba.rgba = rgba.abgr * 255;\n'
```
Eric89GXL: The one thing I don't get is why the `rgba.abgr` above gets multiplied by 255. That makes me think the maximum value should be 255 + 255 / 256. + 255 / (256 * 256) + 255 / (256 * 256 * 256), which is about 256. Maybe the above `rgba.rgba = rgba.abgr` shouldn't be multiplied by 255?

kmuehlbauer:
Yes, I agree. On the python-side we could just extend the `packed`-parameter:

```
def load_spatial_filters(packed='simple'):

    if packed = 'simple':
        kernel = pack_float_to_rgba(kernel).astype(np.ubyte)
    elif packed ='ieee':
        kernel = np.fromstring(kernel.tostring(), np.ubyte).reshape((kernel.shape + (4,)))
    elif packed = 'off':
        kernel = kernel
```

But how should the selection be made on the shader side?

Eric89GXL:
I think the spatial filters code could just always use `simple` / `unit`, we shouldn't need to pack them the less efficient but more general way. We probably want to just keep around the old function/code somewhere for when we need it.

kmuehlbauer:
The `pack/unpack_ieee` functions are located next to the `pack/unpack_unit`. So they can be activated as necessary.

vispy/io/datasets.py
```diff
@@ -35,13 +35,43 @@ def load_crate():
...
-def load_spatial_filters():
+def pack_unit(value):
+    """Packs float values between [0,1] into 4 unsigned int8
+
+    Returns
+    -------
+    pack: array
+        packed interpolation kernel
+    """
+    print(value.max(), value.min())
```
Eric89GXL: leftover from debug

Eric89GXL:
Other than the debugging leftover, LGTM. Ready to merge from your end? If so I'll give the example one last try and merge if everything is good. Thanks for finding a more efficient implementation BTW, I think it should be substantially faster.

Eric89GXL:
Examples look great, except for one minor thing. Every even number of levels, one of the levels seems to be duplicated (it is whiter than the rest). Do you see that, too? I wonder if there is some bug with the calculation of levels...?

kmuehlbauer:
This is intended behaviour. If you have an even number of levels, you have an uneven number of isolines. The thick isoline divides the luminance exactly in the middle.

kmuehlbauer:
Documentation could possibly be a bit more verbose, but that can be made in a later pr. Also we could think about some testing functions (pack/unpack). Just mention it here, to not forget about it.

Apart from that ready from my end.

Eric89GXL:
Great, I'll merge once the CIs are happy. Do you want to take a stab at changing TextVisual to use this in a subsequent PR? If not, I can do it.

kmuehlbauer:
I'll have a look, then...

Eric89GXL:
Agh, example error:

https://travis-ci.org/vispy/vispy/jobs/77162247#L1798

Those variables should only be set if the filter is not `Nearest`, because otherwise they are optimized out by the compiler. Alternatively, you can just get the test to pass by using something other than a hardware mode as the first / default mode, although that's a little less clean :)

Eric89GXL:
(examples shouldn't have such warning messages in them, thus the error)

kmuehlbauer:
This also doesn't finally solve the problem.

Eric89GXL:
Works okay over here, we'll see if Travis is happy

Eric89GXL:
Thanks for working through this @kmuehlbauer!

PR #1070, opened 2015-08-20
ENH: Implement icosphere for uniform sphere tessellation

### LINKED ISSUE #1040: ENH: More uniform tessellation of spheres
Right now our sphere tessellation is done using latitute-longitute type calculations, which are not uniform. I have some code for more even tessellation of sphere that I or someone else could add if they feel like it.

pradeep-pyro:
Fixes #1040

### DIFF
--- a/examples/basics/scene/sphere.py
+++ b/examples/basics/scene/sphere.py
@@ -17,7 +17,7 @@ view = canvas.central_widget.add_view()
 view.camera = 'arcball'
 view.padding = 100
 
-sphere = scene.visuals.Sphere(radius=1, parent=view.scene,
+sphere = scene.visuals.Sphere(radius=1, method='ico', parent=view.scene,
                               edge_color='black')
 
 if __name__ == '__main__' and sys.flags.interactive == 0:
--- a/examples/basics/scene/sphere.py
+++ b/examples/basics/scene/sphere.py
@@ -19,19 +19,20 @@ canvas = scene.SceneCanvas(keys='interactive', bgcolor='white',
 view = canvas.central_widget.add_view()
 view.camera = 'arcball'
 
-sphere1 = scene.visuals.Sphere(radius=1, method='latlon', parent=view.scene,
+sphere1 = scene.visuals.Sphere(radius=1, method='latitude', parent=view.scene,
                                edge_color='black')
 
 sphere2 = scene.visuals.Sphere(radius=1, method='ico', parent=view.scene,
                                edge_color='black')
 
-sphere3 = scene.visuals.Sphere(radius=1, method='cube', parent=view.scene,
+sphere3 = scene.visuals.Sphere(radius=1, rows=10, cols=10, depth=10,
+                               method='cube', parent=view.scene,
                                edge_color='black')
 
 sphere1.transform = STTransform(translate=[-2.5, 0, 0])
 sphere3.transform = STTransform(translate=[2.5, 0, 0])
 
-view.camera.set_range(x=[-3,3])
+view.camera.set_range(x=[-3, 3])
 
 if __name__ == '__main__' and sys.flags.interactive == 0:
     canvas.app.run()
--- a/examples/basics/scene/sphere.py
+++ b/examples/basics/scene/sphere.py
@@ -11,14 +11,27 @@ This example demonstrates how to create a sphere.
 import sys
 
 from vispy import scene
-canvas = scene.SceneCanvas(keys='interactive', size=(800, 600), show=True)
+from vispy.visuals.transforms import STTransform
+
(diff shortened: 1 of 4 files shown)


vispy/geometry/generation.py
```diff
@@ -352,7 +334,100 @@ def create_sphere(rows, cols, radius=1.0, offset=True):
...
+    cols : int
+        Number of columns (for method='latlon').
+    radius : float
+        Sphere radius.
+    offset : bool
+        Rotate each row by half a column (for method='latlon').
+    subdivisions : int
+        Number of subdivisions to perform (for method='icosphere')
+    method : str
+        Method for generating sphere. Accepts 'latlon' for latitude-longitude, and 'icosphere' for
```
Eric89GXL: let's just go with `ico` instead of `icosphere` -- it would probably need to be `icosahedron` to be the full name anyway, and that's too long

Eric89GXL:
There are a few style errors, check Travis. Also, could you update one of the examples, and maybe one of the tests, to use the new method? The ico should look nicer than the latlong anyway.

Eric89GXL:
Otherwise, LGTM

rougier:
By the way, make me thing the cube sphere might be a good addition:

Eric89GXL:
Do you just take a cube and set the coordinates to all be the same distance from the center, or is the transformation more complex than that?

rougier:
Yes, take a tesselated cube, compute distance from each vertex to center and scale them such that it is 1 (or radius).

Eric89GXL:
@pradeep-pyro you up for adding that one? Sounds like it might be straightforward.

pradeep-pyro:
Ok, I will fix the style issues later today. Yeah, I can add the cube sphere as well.

pradeep-pyro:
PTAL.
For the cube sphere, we will need another argument `depth` in `create_sphere()`for passing it to `create_box()`. And this must be added to `SphereVisual` as well. Shall I go ahead or would it make the function needlessly complex? Another way is to make it a separate function.

Eric89GXL:
I think one more argument is probably okay. I can't think of a nicer solution.

pradeep-pyro:
Done, I also updated the sphere example to demo all three types of tessellation.

vispy/geometry/generation.py
```diff
@@ -355,6 +337,117 @@ def create_sphere(rows, cols, radius=1.0, offset=True):
...
+                  subdivisions=3, method='latlon'):
+    """Create a sphere
+    Parameters
+    ----------
+    rows : int
+        Number of rows (for method='latlon' and 'cube').
+    cols : int
+        Number of columns (for method='latlon' and 'cube').
+    depth : int
+        Number of depth segments (for method='latlon' and 'cube').
```
Eric89GXL: `depth` is not used for `latlon`

vispy/geometry/generation.py
```diff
@@ -355,6 +337,117 @@ def create_sphere(rows, cols, radius=1.0, offset=True):
...
+    return MeshData(vertices=verts, faces=faces)
+
+
+def create_sphere(rows=10, cols=10, depth=10, radius=1.0, offset=True,
+                  subdivisions=3, method='latlon'):
+    """Create a sphere
+    Parameters
+    ----------
+    rows : int
+        Number of rows (for method='latlon' and 'cube').
```
Eric89GXL: I prefer `latitude` to `latlon`, it conveys the same idea using a whole word with only a couple more characters

examples/basics/scene/sphere.py
```diff
@@ -11,14 +11,27 @@
...
-sphere = scene.visuals.Sphere(radius=1, parent=view.scene,
-                              edge_color='black')
+sphere1 = scene.visuals.Sphere(radius=1, method='latlon', parent=view.scene,
+                               edge_color='black')
+
+sphere2 = scene.visuals.Sphere(radius=1, method='ico', parent=view.scene,
+                               edge_color='black')
+
+sphere3 = scene.visuals.Sphere(radius=1, method='cube', parent=view.scene,
+                               edge_color='black')
```
Eric89GXL: the `cube` has more triangles than the others, it's probably worth changing the params to make it look more similar

Eric89GXL:
Other than my minor comment and Travis's unnhappiness (https://travis-ci.org/vispy/vispy/jobs/76833422), LGTM. BTW you can run style tests locally with `make flake` or `python make flake`.

pradeep-pyro:
Yeah, I missed to check the style on one of the files, fixed now. PTAL.

Eric89GXL:
Thanks @pradeep-pyro!

PR #1075, opened 2015-08-26
make TextVisual use spatial-filter.frag with packed kernel

kmuehlbauer:
This is a follow up to #1069.

TextVisual uses now the reworked filtering with float packed rgba interpolation kernel. 

* Examples are working
* Test is passing (it doesn't test this part because of font_size)


rougier:
LGTM

Eric89GXL:
Examples still look great. A PR with more red than green, hooray! Thanks @kmuehlbauer

Issue #78, opened 2014-01-16
Headless mode


Eric89GXL:
It just occurred to me that people have mentioned wanting headless plotting as a feature. Currently Travis does not currently test this case, but it would be very easy to do so by adding a switch to toggle the start of the `Xvfb` service. Is headless mode currently expected to work? If so, I'll add it to Travis. If not, then this issue can stay open until someone manages to add support for it.

tedivm:
Any status on this, or ideas on how to render things headless on linux?

Eric89GXL:
The easy way is to use `Xvfb` if possible. Travis-CI does it and it works well.

The hard way is probably to get EGL working properly. We don't have it fully working yet AFAIK, but @almarklein might know how far away we are.

tedivm:
I'm running Xvfb but everything we render is coming back blank. It's good to know someone else has this working though, as it means I actually have some hope :-)

Eric89GXL:
Take a look at `.travis.yml`, you can see what it does. It manages to run the `visual` tests, which rely on reading the pixel buffers, so it must not be rendering all blank :)

tedivm:
@Eric89GXL thanks for the help- turns out my issue was a mismatch in a shared library causing segfaults. Still tracking down the exact issue but the fact that someone has got this working is useful.

Eric89GXL:
Fixed by #1077

PR #1077, opened 2015-09-01
Offscreen rendering with OSMesa

julienr:
This adds a new backend for OSMesa [1], which allows vispy to do offscreen rendering on Linux (and other platform that support mesa/llvmpipe) without a X server nor a graphic card. This is a (partial?) fix to #78 .

I created a docker image which you can use to test the backend. See here for instructions :
https://github.com/julienr/vispy_osmesa_docker

I am not sure the code is production ready but I am opening this PR so we can discuss this implementation.

[1] http://www.mesa3d.org/osmesa.html


Eric89GXL:
This looks like a great start. I didn't realize you could create a Mesa context without an X server running. If I have the Mesa libs installed on Ubuntu for example, do I already have OSMesa support? Or is there something additional to install?

julienr:
On 14.04, you need `libosmesa6` and `libgles2-mesa`.
If you checkout my *osmesa* branch, you can then run the example osmesa script with :

```
VISPY_GL_LIB=/usr/lib/x86_64-linux-gnu/mesa-egl/libGLESv2.so.2.0.0 \
        OSMESA_LIBRARY=/usr/lib/x86_64-linux-gnu/libOSMesa.so.6.5.3 \
        python examples/offscreen/simple_osmesa.py
```

The problem is that I don't think the ubuntu osmesa package use llvmpipe and the rendering is **very** slow (took 3-5 minutes on my machine).

So I ended up compiling my own osmesa. You can find the commands I used [here](https://github.com/julienr/vispy_osmesa_docker/blob/master/src/_setup/2_build_osmesa.sh). With osmesa + llvmpipe, the rendering is really fast.

Eric89GXL:
Do you really need to set those env vars? It seems like ctypes should find them okay. We have an EGL backend that uses libGLESv2 already, so you might be able to do whatever is done there to find the library.

julienr:
You are right, you don't need to specify the `OSMESA_LIBRARY` (I just got used to it because I compiled mine to /opt).

But if I do not specify `VISPY_GL_LIB`, vispy tries to find the `GL` library which will :
- Pickup the OpenGL library from the X driver if there is an X server on the system. This will work.
- End with a ```RuntimeError: Could not load OpenGL library, gl cannot be used``` if the system has no X server

I think the problem is that ```gloo/gl/gl2.py``` wants to find a `GL` lib but we want `GLESv2`. Since gl2 is imported before the backend, the backend cannot request a different GL library (we would need to change `os.environ['VISPY_GL_LIB']`).

Eric89GXL:
Have you tried using `vispy.use(...)` at the top of your example to set both the GL library and the app at the same time? We could try to make that work. Or actually, when doing `use_app('osmesa')`, it should probably call `gl.use_gl(...)` with the proper arguments to get it to look for GLES.

julienr:
I dug a bit deeper and it seems OSMesa's libGLESv2 is mostly an alias to desktop OpenGL [1]. This means if I use the es2 vispy backend, the shaders fail to compile because of some GLES specifics like "precision highp int". So osmesa should use the gl2 backend.

Long story short, when using OSMesa, `libGL` is actually contained in `libOSMesa` [2]. So we have to use `VISPY_GL_LIB=libOSMesa.so` . I think the best way to do this is to add a special case for OSMesa in `vispy.use` where we set `os.environ['VISPY_GL_LIB']`. This is what julienr@77a1880 does.

With those changes, the env vars are no longer necessary if you have `libOSMesa.so` in your lib path.

[1] http://marc.info/?l=mesa3d-dev&m=134313453229954
[2] http://www.paraview.org/pipermail/paraview/2010-February/015988.html

Eric89GXL:
have you run `python make test osmesa`? It would be good if it passed. Then we could get the proper packges on the `full` version of Travis to actually have it run the tests.

julienr:
@Eric89GXL I got most tests to pass. I think the only remaining issue is with the dpi testing. Since osmesa doesn't use X, we can't call `xdpyinfo` to get DPI, and test_dpi fails because it calls `get_dpi(raise_error=True)`. This also generates a lot of `WARNING: could not determine DPI`.

I am not quite sure what DPI means in the context of offscreen rendering and what should be done about it. I see two solutions :
1. In `vispy.util.dpi.get_dpi`, test if we are using the OSMesa backend and return some fixed value (96)
2. Exclude osmesa from `test_dpi`. This won't fix the numerous `could not determine DPI` warnings though.

I would go for solution 1, but I'm not sure what all the ramifications could be.

Eric89GXL:
We have to pick some default DPI (users can always override it) so 96 is reasonable.

julienr:
I made the DPI changes. All tests seem to pass with `python make test osmesa`.

Next step is to get travis to run osmesa tests. It seems like the ubuntu osmesa packages are too old to include llvmpipe and therefore rendering takes ages. So I think the only option is to compile the latest version of mesa for travis (which is running Ubuntu 12.04 it seems). I am not sure if it is easier to build a conda package for osmesa or compile in a 12.04 VM and then upload a tarball to travis. Thoughts ?

Eric89GXL:
Let's go with the source-built tarball instead. That way if we ever move away from Anaconda we can still run the test suite just fine. Do you know if the library will be picked up properly without having to set an env var when running `make test osmesa`? Maybe you'll have to set `LD_LIBRARY_PATH` or so, which wouldn't be too bad I guess.

julienr:
Any idea why Travis didn't run the tests on my latest commit that modifies .travis.yml ?

QuLogic:
You need to add a semicolon at the end of the lines in the `if` (multiple lines are merged into one.) Not sure if that's the exact reason why Travis didn't run though.

julienr:
@QuLogic That was it (and also, it doesn't like comments in if clauses for the same reason) ! Thanks !

julienr:
There is a remaining issue for which I don't have a good solution.

On my test system (Ubuntu 14.04), I compiled OSMesa manually so it uses LLVMPipe and installed to a prefix (`/opt/osmesa_llvmpipe`).

So I have the following libs :
```
~|⇒ ls /opt/osmesa_llvmpipe/lib 
libglapi.la  libglapi.so.0      libGLU.a   libGLU.so    libGLU.so.1.3.1  libOSMesa.so    libOSMesa.so.8.0.0
libglapi.so  libglapi.so.0.0.0  libGLU.la  libGLU.so.1  libOSMesa.la     libOSMesa.so.8  pkgconfig
```

Now, I also have normal desktop opengl on this system with my libGL in `/usr/lib/x86_64-linux-gnu/libGL.so`.

The problem is : if I want to use osmesa, I add `/opt/osmesa_llvmpipe/lib` to my `LD_LIBRARY_PATH`. But after doing this, the desktop backends (Qt4) that requires libGL fails to load because ldd picks up my OSMesa `libglapi.so` as shown in the following example :

```
~|⇒ LIBRARY_PATH=/opt/osmesa_llvmpipe/lib:$LIBRARY_PATH \
  LD_LIBRARY_PATH=/opt/osmesa_llvmpipe/lib:$LIBRARY_PATH \
  python make test unit

/usr/lib/x86_64-linux-gnu/mesa/libGL.so.1: undefined symbol: _glapi_tls_Dispatch
...
  File "/home/julien/anaconda/envs/vispy/lib/python2.7/ctypes/__init__.py", line 365, in __init__
type: /usr/lib/x86_64-linux-gnu/mesa/libGL.so.1: undefined symbol: _glapi_tls_Dispatch
```

And indeed, if I run ldd on `libGL` with my OSMesa prefix, it picks up the wrong libglapi :
```
~|⇒ LIBRARY_PATH=/opt/osmesa_llvmpipe/lib:$LIBRARY_PATH LD_LIBRARY_PATH=/opt/osmesa_llvm/lib:$LIBRARY_PATH ldd /usr/lib/x86_64-linux-gnu/libGL.so
	linux-vdso.so.1 =>  (0x00007fff895f7000)
	libglapi.so.0 => /opt/osmesa_llvmpipe/lib/libglapi.so.0 (0x00007f1f36a0b000)
```

For comparison, without the prefix :
```
~|⇒ ldd /usr/lib/x86_64-linux-gnu/libGL.so 
	linux-vdso.so.1 =>  (0x00007fffc9368000)
	libglapi.so.0 => /usr/lib/x86_64-linux-gnu/libglapi.so.0 (0x00007f1c44c31000)
```

I think the conclusion is that on a system with only OSMesa, this works fine. If you have both a desktop libGL and OSMesa, you'll have to play with your `LIBRARY_PATH` to get things to work. I don't think we can do much about it since it's a linker/ctypes issue.

Maybe the best way to go is to write a doc page about using the OSMesa backend ?

Eric89GXL:
So OSMesa doesn't provide a `libGL.so`? Can you have it compile a dummy one, or one that wraps to `libOSMesa.so`? I'm not familiar with how this stuff works, but I'm surprised it doesn't provide that.

vispy/ext/osmesa.py
```diff
@@ -0,0 +1,94 @@
...
+    # VISPY_GL_LIB=/opt/osmesa_llvmpipe/lib/libGLESv2.so \
+    #   LD_LIBRARY_PATH=/opt/osmesa_llvmpipe/lib/ \
+    #   OSMESA_LIBRARY=/opt/osmesa_llvmpipe/lib/libOSMesa.so \
+    #   python vispy/ext/osmesa.py
+    context = OSMesaCreateContext()
+    w, h = 640, 480
+    pixels = allocate_pixels_buffer(w, h)
+    ok = OSMesaMakeCurrent(context, pixels, 640, 480)
+    assert ok, 'Failed to OSMesaMakeCurrent'
+    assert OSMesaGetCurrentContext().value == context.value
```
Eric89GXL: `assert` statements are skipped when doing `python -O`, so it's better to do a formal check and `raise` for these

Eric89GXL: Actually, maybe we should get rid of these debugging lines now that your code is working?

vispy/ext/osmesa.py
```diff
@@ -0,0 +1,94 @@
...
+# Copyright (c) 2015, Vispy Development Team.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+
+"""A ctypes-based API to OSMesa
+"""
+from __future__ import print_function
+import os
+import ctypes
+from ctypes import c_int as _c_int, c_uint as _c_uint, c_void_p
+from vispy.gloo import gl
```
Eric89GXL: things in `ext` should be independent of `vispy`, so you should just sub in the constants here like:
```
GL_RGBA = 6408
GL_UNSIGNED_BYTE = 5121
```

vispy/util/osmesa_gl.py
```diff
@@ -0,0 +1,20 @@
...
+    so gl2 picks up libOSMesa.so as the OpenGL library.
+
+    This modification must be done before vispy.gloo is imported for the
+    first time.
+    """
+    if 'VISPY_GL_LIB' in os.environ:
+        logger.warning('VISPY_GL_LIB is ignored when using OSMesa. Use '
+                       'OSMESA_LIBRARY instead.')
+    if 'OSMESA_LIBRARY' in os.environ:
+        os.environ['VISPY_GL_LIB'] = os.environ['OSMESA_LIBRARY']
```
Eric89GXL: can just be `os.environ['VISPY_GL_LIB'] = os.getenv('OSMESA_LIBRARY', 'libOSMesa.so')`

vispy/util/wrappers.py
```diff
@@ -77,6 +78,11 @@ def use(app=None, gl=None):
         app = 'headless'
         gl = 'webgl'
 
+    if app == 'osmesa':
+        from vispy.util.osmesa_gl import fix_osmesa_gl_lib
+        fix_osmesa_gl_lib()
+        assert gl is None, "Do not specify GL when using osmesa"
```
Eric89GXL: replace `assert` statement

.travis.yml
```diff
@@ -169,11 +176,21 @@ install:
         pip install -q PySDL2;
       fi;
 
+    # Install OSMesa
+    - wget https://github.com/julienr/vispy_osmesa_docker_12_04/releases/download/0.1/osmesa.tar.bz2 -O /tmp/osmesa.tar.bz2
```
Eric89GXL: Can you add a PR to this repo to make an `osmesa` directory with this tar inside it:

https://github.com/vispy/demo-data

Might want to call it `osmesa/osmesa_VERSON_12_04.tar.bz2` or so.

examples/offscreen/simple_osmesa.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+# vispy: testskip
+# Copyright (c) 2015, Vispy Development Team.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+"""
+This is a simple osmesa example that produce an image of a cube
+
+If you have both osmesa and normal (X) OpenGL installed, execute with
+something like the following to pickup the OSMesa libraries :
```
Eric89GXL: `libraries::`, plus newline between this and the literal below it, will make it render properly in ReStructuredText

.travis.yml
```diff
@@ -42,6 +42,13 @@ matrix:
           packages:
             - *mesa_apt
             - *full_apt
+    - env: PYTHON=2.7 DEPS=full TEST=osmesa IPYTHON="=2" NUMPY="=1.7"
+      addons:
+        apt:
+          sources:
+            - *full_sources
+          packages:
+            - *full_apt
```
Eric89GXL: It's unfortunate that we need to make a Travis run specifically for OSMesa. It would also be nice if one could have both OSMesa and normal GL on a system and switch between them, but maybe that's overkill. It sounds like you've tried all reasonable solutions to get that working, though, so we can live with it. Can you add a comment here saying it needs its own special run for such-and-such reasons?

vispy/ext/osmesa.py
julienr: I think this particular piece of code can be useful to debug OSMesa-related issues. We can tell user to just try to run `python vispy/ext/osmesa.py` to test if the basic OSMesa functionality is at least working. I could put this in a separate file or remove it altogether if you think it's not useful.

julienr:
> So OSMesa doesn't provide a libGL.so? Can you have it compile a dummy one, or one that wraps to > libOSMesa.so? I'm not familiar with how this stuff works, but I'm surprised it doesn't provide that.

I do not think it does (at least I haven't found out how), but I think this is intentional so you can distinguish between the desktop OpenGL lib (libGL) and the offscreen-rendering one (libOSMesa). 

`libOSMesa` contains `libGL` + the `OSMesaXXX` functions. You could symlink `libGL` to `libOSMesa`, but I don't think it would solve the conflicting `libglapi` problem. 

I think one way to have both Desktop and Offscreen OpenGL on one system and not have the linker confused is to have both libraries come from the same mesa build. Since Ubuntu 12.04 ships with a very slow osmesa (because it is not using llvmpipe), one option would be to build the whole mesa ourselves on Ubuntu 12.04, but I fear this will turn into a dependency mess...

.travis.yml
julienr: https://github.com/vispy/demo-data/pull/6

I'll update this url when the demo-data PR gets accepted.

Eric89GXL:
@julienr is this good to go from your end now? Sorry for the slow responses.

julienr:
@Eric89GXL I think this is a good start. I would say this still need some real world testing, but maybe the best way to do that is to have it merged so people can use it. 

I think writing some doc (basically summarizing this discussion with some gotchas) would be useful. I'm not quite sure where the best place to put this is. In ``backends/_osmesa.py`` ?
Anyway, maybe I can add another PR for the doc so you can merge this.

Eric89GXL:
Agreed, let's get it in so people can try it.

Eric89GXL:
Thanks @julienr!

julienr:
@Eric89GXL Thanks for the review & merge

PR #1079, opened 2015-09-05
added examples for scene.visuals

skidzo:


campagnola:
We would like to avoid adding redundant examples. Is there something described by these examples that is not already covered elsewhere?

skidzo:
SceneVisuals describes how to use scene.visuals.Visual() there is already one example, for a Cube.

This Example is derived from that example, but it also demonstrates how one can generate his own primives, combine them, define their transformations, etc...

I will continue updating it, as i have started writing a visualization backend for our Multibody System Dynamics Software... see DocBO/mubosym 

So in short, No, actually this file is missing badly! I have carefully checked all the examples and i think the location in basics/scene/visuals does not exist yet....

PR #1087, opened 2015-09-24
Add blending to improve the look of the vector field plot.

drufat:
This makes the vector field plot look nicer.

### DIFF
--- a/examples/demo/gloo/quiver.py
+++ b/examples/demo/gloo/quiver.py
@@ -261,13 +261,9 @@ void main()
 
 
 canvas = app.Canvas(size=(2*512, 2*512), keys='interactive')
-
-
-@canvas.connect
-def on_initialize(event):
-    gl.glEnable(gl.GL_BLEND)
-    gl.glBlendFunc(gl.GL_SRC_ALPHA, gl.GL_ONE_MINUS_SRC_ALPHA)
-    gl.glBlendEquation(gl.GL_FUNC_ADD)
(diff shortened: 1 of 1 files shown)


examples/demo/gloo/quiver.py
```diff
@@ -264,6 +264,13 @@
 
 
 @canvas.connect
+def on_initialize(event):
+    gl.glEnable(gl.GL_BLEND)
+    gl.glBlendFunc(gl.GL_SRC_ALPHA, gl.GL_ONE_MINUS_SRC_ALPHA)
+    gl.glBlendEquation(gl.GL_FUNC_ADD)
```
Eric89GXL: better to use `set_state` here instead. So it would be:
```
canvas.context.set_state(blend=True, blend_func=('src_alpha', 'one_minus_src_alpha'), blend_equation='func_add'))
```

drufat:
Thanks for the suggestion. I changed it accordingly.

examples/demo/gloo/quiver.py
```diff
@@ -6,7 +6,7 @@
 # -----------------------------------------------------------------------------
 
 from vispy import app, gloo
-
+from vispy.gloo import gl
```
Eric89GXL: you no longer need this line then

drufat:
Oops. Fixed that. Unneeded import is now removed.

Eric89GXL:
Looks great, thanks!

Issue #1092, opened 2015-10-05
Event for unicode characters not just hardware keys.


alexjc:
Currently it's not possible to determine what character the users pressed (in their own locale), only a standard keycode is returned.  The GLFW backend for example has support with a function called `glfwSetCharCallback` which returns a stream of unicode.

Do you want this kind of support in VisPy? It makes things possible like text-based input or mapping by key name (in a locale correct way) rather than by keycode.  If so, I can add the events and GLFW support relatively easily.

alexjc:
I made a prototype that I'm currently using for an interactive console. See #1093 for my thoughts on implementing this. It's much simpler than I expected!

Eric89GXL:
PR merged

PR #1093, opened 2015-10-05
GLFW support for localized characters in key events

alexjc:
Prototype for #1092: Hooked up GLFW's character callback to pipe unicode events into the canvas.

### DIFF
--- a/vispy/app/backends/_glfw.py
+++ b/vispy/app/backends/_glfw.py
@@ -281,6 +281,7 @@ class CanvasBackend(BaseCanvasBackend):
         glfw.glfwSetWindowRefreshCallback(self._id, self._on_draw)
         glfw.glfwSetWindowSizeCallback(self._id, self._on_resize)
         glfw.glfwSetKeyCallback(self._id, self._on_key_press)
+        glfw.glfwSetCharCallback(self._id, self._on_key_char)
         glfw.glfwSetMouseButtonCallback(self._id, self._on_mouse_button)
         glfw.glfwSetScrollCallback(self._id, self._on_mouse_scroll)
         glfw.glfwSetCursorPosCallback(self._id, self._on_mouse_motion)
@@ -433,6 +434,9 @@ class CanvasBackend(BaseCanvasBackend):
         if self._vispy_canvas is None:
             return
         self._vispy_mouse_move(pos=(x, y), modifiers=self._mod)
+
+    def _on_key_char(self, _id, text):
+        self._vispy_canvas.events.key_char(text=chr(text))
 
     def _on_key_press(self, _id, key, scancode, action, mod):
         if self._vispy_canvas is None:
--- a/vispy/app/backends/_glfw.py
+++ b/vispy/app/backends/_glfw.py
@@ -463,7 +463,8 @@ class CanvasBackend(BaseCanvasBackend):
             fun(key=key, text=text, modifiers=self._mod)
 
     def _on_key_char(self, _id, text):
-        # Repeat strokes (frequency configured at OS) are sent here only, no events.
+        # Repeat strokes (frequency configured at OS) are sent here only,
+        # no regular _on_key_press events. Currently ignored!
         if len(self._next_key_events) == 0:
             return
 
--- a/vispy/app/backends/_glfw.py
+++ b/vispy/app/backends/_glfw.py
@@ -296,6 +296,7 @@ class CanvasBackend(BaseCanvasBackend):
 
         # Init
         self._initialized = True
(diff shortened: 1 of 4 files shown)


Eric89GXL:
I wonder if instead of doing this we should just always take the translated key. @alexjc can you see what you get for different backends? I suspect something like Qt will probably give the actual character after being processed by the layout, whereas something like GLFW will not. If that's the case, then we should probably unify them all to take the "final" version of the key instead of the low-level one.

alexjc:
+1 for always taking the translated character for key-presses, in theory!

In practice I hope it works out :-)  Since mapping keys to unicode is a many-to-one relationship, translating back and forth between the two event streams is going to be difficult (replicating some OS features?) and I'm not sure GLFW provides any help for this.

One approach that may work is to merge two event streams in the Vispy backend, for example packaging up the last received unicode character with the corresponding key event—assuming they are dispatched by GLFW in the right order (unicode first).  Then, some meta key_press events will have `text=None` until the OS sends a character.

Looking into it now...

alexjc:
GLFW sends the unicode character only *after* the last key press event, which will make it harder to merge the event streams. Here's an example of me typing Ø (with ALT, SHIFT, O).

```
_on_key_press 342 (alt)
_on_key_press 340 (shift)
_on_key_press 83 (o)
_on_key_char Ø
_on_key_press 83 (o)
_on_key_press 340 (shift)
_on_key_press 342 (alt)
```

PyQt5 does not deal with unicode at all apparently, it only sends ASCII but they are adapted to the correct locale at least.

To fix GLFW's backend to be more like PyQt5 without adding a new event, I could send non-text `key_press` events immediately (e.g. Shift, Alt, Ctrl, Cmd, etc.) then for events with text I wait for the next unicode character then package that up into the final event.

How does that sound?

Eric89GXL:
That sounds reasonable to me. Is that what you have implemented in the last couple of commits?

alexjc:
Yes, it's done and the changes are down to one file. I've been testing it with an interactive console and it's been robust for me. Review and more testing welcome!

Eric89GXL:
Somewhere in the `app` tests we do send fake events do different backends, you could try to do it with GLFW if you're really inclined to do it. But I'm okay to just let the user experience test this one.

vispy/app/backends/_glfw.py
```diff
@@ -447,7 +449,18 @@ def _on_key_press(self, _id, key, scancode, action, mod):
...
+        # this event contains a character we store all other data and dispatch
+        # it once the final unicode character is sent shortly after.
+        if text != '' and action == glfw.GLFW_PRESS:
+            self._next_key_events.append((fun, key, self._mod))
+        else:
+            fun(key=key, text=text, modifiers=self._mod)
+
+    def _on_key_char(self, _id, text):
+        (fun, key, mod) = self._next_key_events.pop(0)
+        fun(key=key, text=chr(text), modifiers=mod)
```
almarklein: This means all key events but the first are discarded, I assume this is intended?

alexjc: @almarklein No it's not intended, and when testing it doesn't behave that way here. Did I miss something? :-)

That code just delays all key press events which have a character associated with them (storing them in a list), then when GLFW sends a character event it will use the correct unicode character and dispatch the event shortly after.

One bug I've just noticed is that the `text` for GLFW_PRESS will be locale-correct but for GLFW_RELEASE it will not be...

almarklein: Ah, you pop and append: its a queue. My mistake!

alexjc:
OK, fixed a few things that came up in the review and another bug with repeat keys.

When you keep your finger pressed, the OS sends multiple characters but no new events. How do you want to handle those? Currently they are ignored. To handle those elegantly I think it would need a new event type.

alexjc:
I'm pretty happy with this change now. Hopefully new tests pass with minor to docs & comments.

@Eric89GXL @almarklein I need a decision for repeat keys: they are currently ignored and Vispy's current events do not seem to have been designed with them in mind. Repeat keys are events sent at an OS-configured frequency when the key is left pressed. (I notice that browsers don't use repeat keys for typing in text-boxes either, so I'd suggest we just ignore them.)

Eric89GXL:
Let's just continue to ignore then for now, and address it in a subsequent
PR if someone has interest in it. I'll review this shortly.

alexjc:
Thanks! I believe the test failure is not related to this change. Can you confirm?

Eric89GXL:
Yep, unrelated. LGTM, thanks!

Issue #748, opened 2015-03-16
Implement textures for meshes


rossant:
Not sure what the API could be like (need to provide the texture and UV mapping).

bollu:
@campagnola, @Eric89GXL do we have a plan for this? It would be nice if this could be bundled with all the sweeping changes we have right now

Eric89GXL:
I think this can be tackled after the big merge. So you can start working
on it now but we shouldn't put it in Luke's PR.

campagnola:
+1
The less we can bundle, the better. All we really need is for the examples to run and unit tests to pass.
Sorry I've been absent lately; will be back home in a few days.

kmuehlbauer:
@djhoese What has to be done here in addition to what'sn already implemented in #1444?

djhoese:
I guess nothing. The TextureFilter does everything that this issue originally wanted (if I had to guess).

PR #1097, opened 2015-10-07
Textured globe example

Jozhogg:
Not intended for merge. For a project I was working on recently I thought I would try vispy as I was interested in using the `vispy.scene` sub package. The goal of the project was to display some data on a textured 3D globe. Unfortunately, the `MeshVisual` class doesn't seem to support custom shaders or texturing at the moment.

This PR contains my hack to add textures to a globe by mapping the normals to points on a Plate Carrée map ("the texture file for the map was too large for me to upload to github but any Plate Carée map should do just fine"). I was wondering if there is a better way to do this in vispy at the moment? And also whether there are plans to add more flexibility with shaders and/or adding texture options to the `MeshVisual` class?

### DIFF
--- a/examples/demo/scene/textured_globe.py
+++ b/examples/demo/scene/textured_globe.py
@@ -1,0 +1,18 @@
+import sys
+import numpy as np
+from vispy.scene import visuals
+from vispy import scene
+
+canvas = scene.SceneCanvas(keys='interactive', show=True, bgcolor='k')
+
+view = canvas.central_widget.add_view()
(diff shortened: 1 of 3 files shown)


Eric89GXL:
This is probably a dup of #748, adding textures to meshes, which indeed has not been implemented yet. Does that issue seem to cover your use case?

Jozhogg:
Yeah, looks like that's what I was looking for. Thanks for pointing it out.

PR #1101, opened 2015-10-09
Fix grid_widget when Fig gets a single element

gouarin:
This PR i a fix when we use Fig class to create a grid.

There is a bug if you write

```
f = Fig()
ax = f[0]
...
```

Then the cspan is 0. The PR solves the problem by giving a default size to 1 for the column.

This PR adds also a _row_widgets that defines the widgets that take all the row and need to be expand when the column size of the grid has changed.

The idea is to have the same behavior if you do

```
f = Fig()
ax1 = f[0]
...
ax2 = f[1, :2]
```

or 

```
f = Fig()
ax1 = f[0, :2]
...
ax2 = f[1]
```

I think that we have to change also the stretch of the widget when we change the size of the cspan of the _grid_widgets in _update_child_widgets but if I do that I have a "maximum recursion depth exceeded" because modified stretch call again _update_child_widgets.

### DIFF
--- a/vispy/scene/widgets/grid.py
+++ b/vispy/scene/widgets/grid.py
@@ -38,7 +38,7 @@ class Grid(Widget):
         if not isinstance(idxs, tuple):
             idxs = (idxs,)
         if len(idxs) == 1:
-            idxs = idxs + (slice(None),)
+            idxs = idxs + (slice(0, 1, None),)
         elif len(idxs) != 2:
             raise ValueError('Incorrect index: %s' % (idxs,))
         lims = np.empty((2, 2), int)


Eric89GXL:
I don't think we want to make a special category called "row widgets". This makes for exceptional behavior that is different from other widgets, we'd also have to do it with columns, etc. So I think we'll have to find another suitable solution. Now that I think about it, we might want to wait until #1049 is merged before moving forward with a fix, since it would be good to fit the fix into that framework. Will that work for you?

gouarin:
I think you right. I do this patch thinking that the implementation was not very good and need to be improved. Cassowary will do the job.

So perhaps, it could be great to push only the fisrt commit to be able to use f[0] ?

Eric89GXL:
Yeah probably. Does that first commit fix the problem?

gouarin:
Yes. We enforce the widget to have a cspan equal to 1. 

Previously, the Slice(None) gave cspan equal to 0.

Eric89GXL:
Can you revert your last commit so the CIs (Travis, AppVeyor) can run on that commit? You can use `git rebase -i HEAD~2` to do it, but be careful with it.

djhoese:
I did a quick test by modifying `examples/basics/plotting/plot.py` to request `Fig[0]` and I did see this error with current master. Including these changes seemed to fix it. Let's merge this...5 years later!

(I know the tests are failing but they don't seem to be related to this and this is a small fix. I'll revert if it is still a problem.)

Issue #1063, opened 2015-08-17
Issues with HiDPI displays (using PyQt5)


astrofrog:
I tried the ``examples/basics/plotting/plot.py`` example with PyQt5 on a HiDPI screen, and this is the output:

<img width="714" alt="screen shot 2015-08-17 at 9 44 12 pm" src="https://cloud.githubusercontent.com/assets/314716/9314429/2012512c-4529-11e5-8038-0eb95385440d.png">

It looks like some of the plotting code must be making assumptions about a fixed non-HiDPI DPI value.

nippoo:
Blarp. I went through and fixed everything on GLFW by adding a `physical_size` attribute in https://github.com/vispy/vispy/pull/787 so the examples should be HiDPI-capable... fingers crossed, it should just be a matter of making PyQt5 set the correct `physical_size` attribute as the framebuffer size. Looks like this is the issue here but it might be more subtle than that. I won't have time for 2 weeks...

astrofrog:
Fixed in #1103, thanks @theWatchmen!

PR #1103, opened 2015-10-17
Added physical_size to Qt backend to manage different fb size on Retina displays

theWatchmen:
This is a fix for  #1063.
In resizeGL Qt reports only the physical size of the framebuffer. To mirror the behaviour of the GLFW backend, the physical_size is now stored and can be accessed when the resize event fires, so tha the correct physical size of the frame buffer is set on HiDPI displays. Only tested with PyQt5 on Mac.

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -283,6 +283,7 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
         # Qt supports OS double-click events, so we set this here to
         # avoid double events
         self._double_click_supported = True
+        self._physical_size = p.size
 
     def _vispy_warmup(self):
         etime = time() + 0.25
@@ -300,6 +301,14 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
     def _vispy_set_size(self, w, h):
         # Set size of the widget or window
         self.resize(w, h)
+
+    def _vispy_set_physical_size(self, w, h):
+        self._physical_size = (w, h)
+
+    def _vispy_get_physical_size(self):
+        if self._vispy_canvas is None:
+            return
+        return self._physical_size
 
     def _vispy_set_position(self, x, y):
         # Set location of the widget or window. May have no effect for widgets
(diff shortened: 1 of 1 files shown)


Eric89GXL:
Hmm, looks like there are some CI errors, but I don't think they're related to your change (which otherwise looks good). I'll test your changes locally, where I only have PyQt4.

Eric89GXL:
@QuLogic do you have time to take a look (12 changed lines) to see if this looks reasonable from your understanding of PyQt5?

Eric89GXL:
FYI @theWatchmen if you edit your description or make a comment saying "Closes #1063" GitHub will magically close the associated issue when this gets merged, which is satisfying.

vispy/app/backends/_qt.py
```diff
@@ -636,7 +645,8 @@ def initializeGL(self):
     def resizeGL(self, w, h):
         if self._vispy_canvas is None:
             return
-        self._vispy_canvas.events.resize(size=(w, h))
+        self._vispy_set_physical_size(w, h)
+        self._vispy_canvas.events.resize(size=(self.width(), self.height()), physical_size=(w, h))
```
Eric89GXL: PEP8 this needs to be shorter, you can run `python make flake` from the `vispy` root to see the error. But otherwise tests pass, which is great.

Eric89GXL:
(ping me once the line is fixed and the CIs are happy and I'll merge)

theWatchmen:
@Eric89GXL it should be fixed now. On a side note, the command for checking the syntax is `python make test flake`, it took me a while to figure that out :)

theWatchmen:
Closes #1103

Eric89GXL:
LGTM. Have you tested that resizing the windows works properly on your Retina machine?

theWatchmen:
Just tested it on the application I am working on, the behaviour seems consistent with the glfw backend.

Eric89GXL:
Great, thanks for the fix @theWatchmen!

PR #1104, opened 2015-10-18
Added GridMesh visual

inclement:
This PR is the same as the previous one at https://github.com/vispy/vispy/pull/959, which I managed to mess up somehow.

It includes everything from before, plus the final fixes that I never did, plus some small changes to work with changes in vispy in the last few months. So, it should be ready for re-review.


Eric89GXL:
LGTM, thanks for addressing the last issues. It looks like there is a small problem in the example though, in that the spheres don't occlude one another properly. Do you have that problem, too? Maybe depth testing needs to be turned on for the visual?

inclement:
I don't see any problem with the sphere occlusion, do you have a screenshot of it?

Eric89GXL:
I get one set of spheres always in front even if they aren't supposed to be. But I don't think that's your bug; we have an open issue for that somewhere. Ready for merge from your end, then?

inclement:
Yep, I think it's good to go from my end.

Eric89GXL:
Great, thanks @inclement!

PR #1107, opened 2015-10-18
Made some shaders work with ES2

inclement:
Without these fixes, the shaders don't work on Android, so I think (but am not certain) it's an ES2 compatibility thing.


Eric89GXL:
It's likely, and these make sense for fixes. I'll merge and hopefully I can get the CIs happy in a separate PR, since they're unhappy about unrelated changes.

almarklein:
For the record: yes, GLES prohibits implicit type casting. I *think* that even some desktop drivers don't allow it ...

PR #1110, opened 2015-10-19
FIX: Minor fixes

Eric89GXL:
Fixes a couple problems with the CIs getting more strict. We'll see if they're happy.

### DIFF
--- a/examples/basics/visuals/arrows_quiver.py
+++ b/examples/basics/visuals/arrows_quiver.py
@@ -62,7 +62,8 @@ class Canvas(app.Canvas):
         self.visual.transforms.configure(canvas=self, viewport=vp)
 
     def rotate_arrows(self, point_towards):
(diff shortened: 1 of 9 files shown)


Eric89GXL:
This should "fix" the CIs -- I had to add a test skip because for some unknown reason an `ArrowVisual` test was failing for old numpy. I tried for a while to fix it but couldn't, so I'll have to take a look later. But it's important for our CIs to work, so probably worth merging with the skip. Ready for review/merge from my end.

Eric89GXL:
@rossant do you have a minute to take a quick look at this? It would be good to get it in so the currently open PRs could be tested properly.

rossant:
LGTM, although I'm not that familiar with that part of the code.

Eric89GXL:
Okay, I'll go ahead and merge. If others want to chime in I'm happy to make a follow-up PR.

PR #1111, opened 2015-10-19
Fixed a bug when using rectangular vector field

dionhaefner:
When calculating the row and column indices, there was a division by shape.y instead of shape.x, causing bugs when using a non-square vector field.

### DIFF
--- a/examples/demo/scene/flow_lines.py
+++ b/examples/demo/scene/flow_lines.py
@@ -36,7 +36,7 @@ class VectorFieldVisual(visuals.Visual):
         dist = index.y * seg_len;
         
         vec2 local;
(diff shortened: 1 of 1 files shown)


Eric89GXL:
The change makes sense, but I'm having a hard time actually seeing the difference in the example. How should it manifest?

dionhaefner:
In the example, as is, there is no difference. Since the field is evaluated on a 100x100 array, it does not matter whether we divide by the shape in x or y direction. Trouble starts when people are inserting their own vector field, which may not have square dimensions anymore (as it was the case for me).

Eric89GXL:
Ahh, I see what you mean.

Eric89GXL:
Thanks @dionhaefner

PR #1112, opened 2015-10-22
FIX: cubehelix issues

kmuehlbauer:
While co-working with VisPy and matplotlib I wondered why the VisPy cubehelix implementation misses the green parts of the colormap. First I thought this is with the standard cubehelix implementation and tried to change `start`, `rot` and other keywords. Nothing happened, this was due to `cubehelix` was already instantiated within `colormap`-module and no pure class any more. 

The colors were not influenced by this change, so I found that **rgb** was in the wrong order **rbg**. After changing this everything works as expected (as far I can see).


Eric89GXL:
Whoops! Thanks.

PR #1115, opened 2015-10-28
write a function to create a colormap texture

### LINKED ISSUE #1052: Make Colormap shader efficient by using 1D Textures
As discussed in [this pull request thread](https://github.com/vispy/vispy/pull/1044), the current implementation of Colormaps generate a chain of `if-then-else` which is very expensive as the number of control points increases.

Ideally, the colormap gradient is computed once, and is stored as a 1D texture which is used to map texture coordinates to the corresponding color for the `cmap.glsl_map` function that maps texcoords to colors.

bollu:
fixes #1052 

I'm planning on generating a 1D texture for any `Colormap` so that colormap indexing can be reduced to texture lookup

### DIFF
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -156,6 +156,10 @@ def _glsl_step(controls=None):
     return """vec4 colormap(float t) {\n%s\n}""" % s
 
 
+def _create_texture_for_cmap(controls=None, colors):
+    pass
+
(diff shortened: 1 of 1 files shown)


Eric89GXL:
Just a stub so I'll close, @bollu feel free to open another PR if you have time to take this up again

PR #1121, opened 2015-10-29
Call view_changed when camera.depth_value is set

julienr:
This calls `camera.view_changed` when `camera.depth_value` is set. Otherwise, one has to wait for the next view refresh to see the result.


Eric89GXL:
Thanks @julienr!

PR #1125, opened 2015-11-06
Update volume.py with an top left axis view

PennyQ:
Add an extra axis view in the left top to show the rotation of the volume visual. The location and scale of this axis is relative to the `Viewbox` coordinates in the GUI rather than the internal scene. Now only available for the `TurntableCamera`.

image

### DIFF
--- a/examples/basics/scene/volume.py
+++ b/examples/basics/scene/volume.py
@@ -32,6 +32,7 @@ import numpy as np
 
 from vispy import app, scene, io
 from vispy.color import get_colormaps, BaseColormap
+from vispy.visuals.transforms import STTransform
 
 # Read volume
 vol1 = np.load(io.load_data_file('volume/stent.npz'))['arr_0']
@@ -64,6 +65,12 @@ cam2 = scene.cameras.TurntableCamera(parent=view.scene, fov=fov,
 cam3 = scene.cameras.ArcballCamera(parent=view.scene, fov=fov, name='Arcball')
 view.camera = cam2  # Select turntable at first
 
+# Create an XYZAxis visual
+axis = scene.visuals.XYZAxis(parent=view)
+s = STTransform(translate=(50, 50), scale=(50, 50, 50, 1))
+affine = s.as_matrix()
+axis.transform = affine
+
 
 # create colormaps that work well for translucent and additive volume rendering
 class TransFire(BaseColormap):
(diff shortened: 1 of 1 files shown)


examples/basics/scene/volume.py
```diff
@@ -64,6 +64,12 @@
 cam3 = scene.cameras.ArcballCamera(parent=view.scene, fov=fov, name='Arcball')
 view.camera = cam2  # Select turntable at first
 
+# Create an XYZAxis visual
+axis = scene.visuals.XYZAxis(parent=view)
+s = STTransform(translate=(50, 50), scale=(50, 50, 50, 1))
+affine = s.as_matrix()
+print affine
```
Eric89GXL: don't need this statement, leftover from debugging?

examples/basics/scene/volume.py
```diff
@@ -64,6 +64,12 @@
...
+axis.transform = affine
```
Eric89GXL: have you tried doing `axis.transform = s` instead? Hopefully you don't have to do the `.as_matrix()` to get it to work...?

Eric89GXL: ahh, maybe you do because later you need to rotate, scale, and translate?

examples/basics/scene/volume.py
```diff
@@ -64,6 +64,12 @@
 cam3 = scene.cameras.ArcballCamera(parent=view.scene, fov=fov, name='Arcball')
 view.camera = cam2  # Select turntable at first
 
+# Create an XYZAxis visual
+axis = scene.visuals.XYZAxis(parent=view)
+s = STTransform(translate=(50, 50), scale=(50, 50, 50, 1))
```
Eric89GXL: I get an error here:
```
NameError: name 'STTransform' is not defined
```
You probably need an `import` line above.

PennyQ:
Ha, thanks. I added the `import` line and it should work well now. 

I tried to set a `MatrixTransform` for the `XYZAxis` visual directly and initialize the 'scale' and 'translate' at the beginning, but the axis failed to display before you drag your mouse, so `.as_matrix()` is the only solution I found that could work well. Although it looks silly :)

almarklein:
Can we (and should we) put the functionality to sync woth the camera inside the XYZAxis visual?

Eric89GXL:
@almarklein +1 from me

PennyQ:
@almarklein that would be a great idea :)

Eric89GXL:
@PennyQ can you clean up the example? If so then we can merge it, and later add the functionality to XYZAxis visual

PennyQ:
@Eric89GXL Yes I did the rebase, anything else for it?

Eric89GXL:
Yeah, see comments in the code above.

examples/basics/scene/volume.py
PennyQ: Yes!

PennyQ:
@Eric89GXL Done, one check failed, can you show me how to fix it or it doesn't matter ? thanks!

Eric89GXL:
You have PEP8 errors:

https://travis-ci.org/vispy/vispy/jobs/131732075#L2428

Eric89GXL:
Works great, thanks @PennyQ

PR #1130, opened 2015-11-13
camera for syncing plots by x or y axis with example

opekar:
- based on chat with @chongxi see https://gitter.im/vispy/vispy?at=560a98eeff22c70f6fab265f
- camera generalized for x,y axis
- example included

### DIFF
--- a/examples/basics/plotting/sync_camera_plot.py
+++ b/examples/basics/plotting/sync_camera_plot.py
@@ -20,8 +20,7 @@ y2 = np.sin(x) + 0.2 * np.random.randn(len(x))
 sp1 = fig[0, 0]
 sp2 = fig[1, 0]
 line1 = sp1.plot((x, y1), color='#00b4b2', width=1, title='I/V Curve',
-        xlabel='Current (pA)', ylabel='Membrane Potential (mV)',
-        face_color='w')
+    xlabel='Current (pA)', ylabel='Membrane Potential (mV)', face_color='w')
 line2 = sp2.plot((x, y2), color='#ff0000')
 
 #note: for complete sync is better connect cameras directly
--- a/examples/basics/plotting/sync_camera_plot.py
+++ b/examples/basics/plotting/sync_camera_plot.py
@@ -9,9 +9,7 @@ Example of syngronization of two plots during zooming
 import numpy as np
 
 from vispy import plot as vp
-from vispy.io import load_data_file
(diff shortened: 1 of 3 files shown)


Eric89GXL:
@campagnola do you have time to look?

campagnola:
Doesn't PanZoom already support linking?

campagnola:
(and if the answer is no, then I would suggest building this feature into PanZoomCamera rather than creating a new camera)

almarklein:
I think this implements linking of x and y individually, which canbe useful. But would indeed be good to implement this on the standard PanZoom camera.

opekar:
thanks for taking a look on this merge request. I choose this solution because the functionality is well separated, but I'll be happy to change it to something which fits more to vispy architecture.

* purpose is to link of x and y individually
* my usecase is to x-link four 2D charts and let the y range different in each of the chart.

__a)__ The best interface I like is
```
cam.link(cam2, x_sync_only=True)
```
but for that few things would need to be changed in BaseCamera, considering this minor functionality, it's maybe not worth to make the code more complicated for that

__b)__ if x/y sync functionality will be in  PanZoom camera, meaning I will need to add 2 optional constructor parameters, and implement the  set_state method as in [this PR](https://github.com/opekar/vispy/commit/542f10dd9d40d328cbd2c0612c3fe6f31059d2d1#diff-ef06a427730810d86afd7cbe6d226ec1R10)

But isnt  __b)__ bit messy? still one extra camera instance is needed as with the [current solution](https://github.com/opekar/vispy/commit/542f10dd9d40d328cbd2c0612c3fe6f31059d2d1#diff-59012afb976f455d4b0576ec0b1681bdR29)

let me know if you really prefer __b)__

Eric89GXL:
@opekar what about `cam.link(cam2, sync='xyz')` where `sync` can be any string of length 1-3 with the elements `x`, `y`, or `z`. Can that work?

opekar:
@Eric89GXL  looks more intuitive, when I have spare time, I'll try to implement that.

djhoese:
I think this is now available through the `axis` keyword argument.

https://github.com/vispy/vispy/blob/91430ea4cfa7abf57819b88416fdab51226a9b6b/vispy/scene/cameras/base_camera.py#L383-L398

I'm going to close this because of that. If I'm missing something comment here and we can reopen this.

PR #1132, opened 2015-11-19
Removed call to non-existent Widget.remove_parent

inclement:
I'm not sure if there are other problems related to this (like setting the parent in the first place - but other places do reference self.parent so maybe that's expected), but the method doesn't exist so this at least prevents an actual crash.


Eric89GXL:
I'm not sure either... but this is a good fix at least.

PR #1136, opened 2015-11-26
Update bundled six to 1.8.0 and allow it to be frozen

### LINKED ISSUE #1074: Bundled six fails to import when frozen
When I freeze a simple vispy program, I can't import vispy because the import logic in `vispy.ext.six` fails when frozen. I have been trying to figure out how to fix this, but I couldn't in a reasonable amount of time so I will post my progress here.

Here is the script I am using to freeze the program (`freeze.py`):

```python
from cx_Freeze import setup, Executable

setup_kwargs = {
    'scripts': ['program.py'],
    'executables': [Executable("program.py")],
}

setup(**setup_kwargs)

```

Here is the one-line "program" that just imports vispy (`program.py`):

```python
import vispy
```

To freeze and run, just install `vispy` and `cx_Freeze` and type:

```sh
$ python freeze.py build_exe
$ ./build/exe.linux-x86_64-3.4/program 
```

(Your build folder is going to be different on a non-Linux machine or if you don't use Python 3.4).

This fails when vispy tries to do `imp.find_module('vispy.ext._bundled.six')`. I tried importing `six` like some of the other bundled packages if `hasattr(sys, "frozen")` is true in `vispy/ext/six.py`...:

```python
...

if hasattr(sys, "frozen"):
    try:
        from ._bundled.six import *  # noqa
    except ImportError:
        from six import *  # noqa
else:
    _import_six()

...
```

...but that didn't work. I got this error:

```
ImportError: No module named 'vispy.ext.six.moves'; 'vispy.ext.six' is not a package
```

I know @QuLogic did the bundling; do you have any ideas?

QuLogic:
Six 1.8.0 appears to be the latest version available on Debian stable. Ubuntu and Fedora are both on 1.9.0. I changed the shim to use `importlib` where available (2.7+) which should allow vispy to be frozen.

@jdreaver Can you see if this works? Right now, it seems like you need to freeze `six` as well, even though the bundled copy is there. I haven't figured out why it can't see it, but I think this is a good enough work-around for the time being.

Fixes #1074 hopefully.

### DIFF
--- a/vispy/ext/_bundled/png.py
+++ b/vispy/ext/_bundled/png.py
@@ -45,8 +45,8 @@ import struct
 import zlib
 # http://www.python.org/doc/2.4.4/lib/module-warnings.html
 import warnings
-from .six.moves import map as imap
-from .six import string_types
+from ..six.moves import map as imap
+from ..six import string_types
 import itertools
 
 
--- a/vispy/ext/_bundled/six.py
+++ b/vispy/ext/_bundled/six.py
@@ -1,6 +1,6 @@
 """Utilities for writing code that runs on Python 2 and 3"""
 
-# Copyright (c) 2010-2013 Benjamin Peterson
+# Copyright (c) 2010-2014 Benjamin Peterson
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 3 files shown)


jdreaver:
Hey, that removes the six error! I had to manually specify to include `vispy.ext._bundled.six`, but the fact that it works is what matters.

Of course, I am getting the glsl and `io._data` errors because of extra data files, but those are separate issues. Thanks a lot for your work!

Eric89GXL:
Thanks @QuLogic! Ready to go from your end?

QuLogic:
Pretty much; I don't know why cx_Freeze doesn't include the bundled `six` by default, but I think it's not a big deal to have to specify it manually.

Eric89GXL:
Great, thanks

PR #1139, opened 2015-12-03
bugfix vispy.io.read_mesh fails because of numpy bug #823

bdvd:
By some .obj files (for example obj files generated with FreeCAD) the read_mesh function fails because of a numpy bug (in numpy repository under bug 823, 4366 and 4384). 

Because handling the numpy bug would be too complex, I corrected the code here.

By using this modification I was able to read all obj files.

### DIFF
--- a/vispy/geometry/calculations.py
+++ b/vispy/geometry/calculations.py
@@ -75,7 +75,7 @@ def _calculate_normals(rr, tris):
     nn = np.zeros((npts, 3))
     for verts in tris.T:  # note this only loops 3x (number of verts per tri)
         for idx in range(3):  # x, y, z
-            nn[:, idx] += np.bincount(verts, tri_nn[:, idx], minlength=npts)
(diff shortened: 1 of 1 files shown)


Eric89GXL:
What was the type before? `int64` or something else? I ask because if it wasn't `int32` or `int64`, then we'd probably want to fix the bug where `tris` gets created instead. And if it was `int64` and it worked with `int32` I'd be confused.

bdvd:
Sorry. I forgot to copy the error message. It was uint32. It's just the problem of conversion between signed and unsigned.

Error message:
Cannot cast array data from dtype('uint32') to dtype('int32') according to the rule 'safe'

Eric89GXL:
Cool, thanks. Can you fix the PEP8 error?

https://travis-ci.org/vispy/vispy/jobs/94717480#L2286

bdvd:
Done.

Eric89GXL:
Thanks @bdvd

PR #1143, opened 2015-12-19
Add touch event on OS X

rossant:
This adds preliminary multitouch support on OS X (Qt backend). It works fine on my end so it should be ready to go.

cc @Eric89GXL @nippoo

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -248,6 +248,10 @@ class ApplicationBackend(BaseApplicationBackend):
 
 # ------------------------------------------------------------------ canvas ---
 
+def _get_qpoint_pos(pos):
+    """Return the coordinates of a QPointF object."""
+    return pos.x(), pos.y()
+
 
 class QtBaseCanvasBackend(BaseCanvasBackend):
     """Base functionality of Qt backend. No OpenGL Stuff."""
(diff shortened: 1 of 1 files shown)


campagnola:
Generates a lot of `WARNING: Got touch without getting TouchBegin for id 171` messages for me..

rossant:
What version of OS X? What example have you run?

campagnola:
Kubuntu 14.04 (let's say it's roughly equivalent to OSX 12.5 ;)
I get this error running any of the examples (multiple errors printed while touching the screen).

rossant:
Ah, I didn't even realized this could be used on touchscreen... (I did this PR for the OS X trackpad)
It looks like a bug in Ubuntu or Qt. Could you try the fix in the first link below? Otherwise, should we just restrict this on OS X for now? We can add support for touchscreen later.

https://gabrbedd.wordpress.com/2012/07/10/getting-multitouch-qt-on-ubuntu-12-04/
https://bugs.launchpad.net/ubuntu/+source/qt4-x11/+bug/1007847

campagnola:
No luck from those. It appears to be a bug with the combination of QGLWidget and XInput, so it should be enough to disable it for Linux (only lines 293/294 need to be disabled).

rossant:
ok @campagnola, done. This is ready to merge if travis is happy.

rossant:
done

campagnola:
Thanks @rossant! Anyone else want a peek before merge?

Eric89GXL:
Cool! Looks like it's actually a pretty simple bit of code. LGTM +1 for merge

Eric89GXL:
Thanks @rossant

PR #1145, opened 2015-12-24
BUG: Type casting error in agg_path_collection.py

nmayorov:
I had troubles running examples `collections\tiger.py`and `collections\chloropleth.py` because of a small problem with type casting in newer versions of numpy.

So I hope this tiny fix is helpful.


Eric89GXL:
Makes sense, thanks @nmayorov

PR #1152, opened 2016-01-13
top and right orientations in AxisWidget

allebacco:
When instantiating AxisWidget class with *orientation='right'* or *orientation='top'*, the function AxisWidget.on_resize raises an exception because AxisWidget._axis_ends supports only 'left' and 'button' orientations.
This PR adds support for 'right' and 'top' orientations in function _axis_ends of AxisWidget class.

### DIFF
--- a/vispy/scene/widgets/axis.py
+++ b/vispy/scene/widgets/axis.py
@@ -55,7 +55,8 @@ class AxisWidget(Widget):
         elif self.orientation == 'top':
             return np.array([[r.left, r.top], [r.right, r.top]])
         else:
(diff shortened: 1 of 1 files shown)


Eric89GXL:
LGTM. @bollu will this interact with your PR at all?

bollu:
No, it shouldn't :) I should rebase off of this actually, so it gets included in my PR #1049

Eric89GXL:
@allebacco can you fix the PEP8 error that Travis reports? Then we can merge

allebacco:
@Eric89GXL Sorry for the issue, it should be fixed.

Eric89GXL:
Thanks @allebacco

PR #1161, opened 2016-01-24
allow conflicting library, if the user has chosen one

### LINKED ISSUE #1144: PyQt4 and PySide
My python interpreter has both PySide and PyQT4 libraries simultaneously. When I run my application in debug mode, an exception raises:

```
  File "C:\Program Files (x86)\JetBrains\PyCharm Community Edition 4.5.4\helpers\pydev\pydevd.py", line 2358, in <module>
    globals = debugger.run(setup['file'], None, None, is_module)
  File "C:\Program Files (x86)\JetBrains\PyCharm Community Edition 4.5.4\helpers\pydev\pydevd.py", line 1778, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "D:/simple_circle.py", line 55, in <module>
    Canvas(title="Graph").native.show()
  File "D:/simple_circle.py", line 35, in __init__
    **kwargs)
  File "d:\vispy\vispy\app\canvas.py", line 169, in __init__
    self._app = use_app(call_reuse=False)
  File "d:\vispy\vispy\app\_default_app.py", line 47, in use_app
    default_app = Application(backend_name)
  File "d:\vispy\vispy\app\application.py", line 49, in __init__
    self._use(backend_name)
  File "d:\vispy\vispy\app\application.py", line 238, in _use
    'PyQt' % [b[0] for b in CORE_BACKENDS])
RuntimeError: Could not import any of the backends. You need to install any of ['PyQt4', 'PyQt5', 'PySide', 'Pyglet', 'Glfw', 'SDL2', 'wx', 'EGL', 'osmesa']. We recommend PyQt.
```

It seems running an application from PyCharm in debug mode sys.modules imports both “PySide.QtCore” и “PyQt4.QtCore”. Without debug mode – only “PyQt4.QtCore”, thus the application runs properly.

code:

``` python
import numpy as np
from vispy import gloo, app
from vispy.gloo import set_viewport, set_state, clear

vert = """
void main (void) {
    gl_Position = vec4(a_position,1.0);
    gl_PointSize = 100.0;
}
"""

frag = """
void main()
{
    float r = length(gl_PointCoord - vec2(0.5, 0.5));
    if (r < 0.5)
        gl_FragColor = vec4(1.0, 0.0, 0.0, 1.0);
    else
        gl_FragColor = vec4(0.0, 1.0, 0.0, 1.0);
}
"""

class Canvas(app.Canvas):

    def __init__(self, **kwargs):
        # Initialize the canvas for real
        app.Canvas.__init__(self, keys='interactive', size=(512, 512),
                            **kwargs)
        self.program = gloo.Program(vert, frag)
        self.program['a_position'] = np.array([[0.3, 0.3, 0.0]], dtype=np.float32)

        set_viewport(0, 0, *self.physical_size)
        set_state(clear_color='white')

        self.show()

    def on_resize(self, event):
        set_viewport(0, 0, *event.physical_size)

    def on_draw(self, event):
        clear(color=True)
        self.program.draw('points')

if __name__ == '__main__':
    import sys
    from PyQt4 import QtGui
    qapp = QtGui.QApplication(sys.argv)
    Canvas(title="Graph").native.show()
    sys.exit(qapp.exec_())
```

ombschervister:
Fix #1144
If interpreter has both PySide and PyQT4 simultaneously and I chose one by app.use_app then print warning instead raise exception

### DIFF
--- a/vispy/app/application.py
+++ b/vispy/app/application.py
@@ -164,6 +164,9 @@ class Application(object):
         # with requires_application()
         test_name = os.getenv('_VISPY_TESTING_APP', None)
 
+        if backend_name is not None:
+            backends.SELECTED_BY_USER_BACKEND = backend_name.lower()
+
         # Check whether the given name is valid
         if backend_name is not None:
             if backend_name.lower() == 'default':
--- a/vispy/app/backends/__init__.py
+++ b/vispy/app/backends/__init__.py
@@ -46,5 +46,8 @@ BACKENDMAP = dict([(be[0].lower(), be) for be in BACKENDS])
 # List of attempted backends. For logging.
 TRIED_BACKENDS = []
 
+# For allow conflicted libraries if user select some one.
+SELECTED_BY_USER_BACKEND = None
+
 # Flag for _pyside, _pyqt4 and _qt modules to communicate.
 qt_lib = None
--- a/vispy/app/backends/_pyqt4.py
+++ b/vispy/app/backends/_pyqt4.py
@@ -17,8 +17,14 @@ try:
     for lib in ['PySide', 'PyQt5']:
         lib += '.QtCore'
         if lib in sys.modules:
-            raise RuntimeError("Refusing to import PyQt4 because %s is "
-                               "already imported." % lib)
+            # If user choose this backend then load it, print warning about conflict.
+            if backends.SELECTED_BY_USER_BACKEND == 'pyqt4':
+                logger.warning('It was found a library "%s" that conflicts with "%s"'
+                               % (lib, 'PyQt4.QtCore'))
+            # else pass import.
+            else:
+                raise RuntimeError("Refusing to import PyQt4 because %s is "
+                                   "already imported." % lib)
     # Try importing (QtOpenGL first to fail without import QtCore)
     if not USE_EGL:
         from PyQt4 import QtOpenGL  # noqa
--- a/vispy/app/backends/_pyside.py
+++ b/vispy/app/backends/_pyside.py
@@ -17,8 +17,14 @@ try:
     for lib in ['PyQt4', 'PyQt5']:
         lib += '.QtCore'
         if lib in sys.modules:
-            raise RuntimeError("Refusing to import PySide because %s is "
-                               "already imported." % lib)
+            # If user choose this backend then load it, print warning about conflict.
+            if backends.SELECTED_BY_USER_BACKEND == 'pyside':
+                logger.warning('It was found a library "%s" that conflicts with "%s"'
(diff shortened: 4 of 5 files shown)


ombschervister:
If the user has installed PySide and PyQt4, it is better to issue warnings than force the user to delete one of the libraries (or refuse vispy). Do you think so?

QuLogic:
But it's not checking whether they're both installed; it's checking whether one or the other has already been _loaded_. You can't mix the two.

ombschervister:
But the problem is that in debug mode I always find PyQt4 and Pyside already imported.
Moreover, all widgets are working properly, only vispy raise an error. If I delete the tests from vispy - even vispy works.

If PySide have not some checks about conflicts with PyQt4, why vispy have? This is due to some specific problem, which occurs only in vispy?

almarklein:
I've seen nasty segfaults when mixing PySide and PyQt. Having both imported at the same time seems like a bug, is there anything that you can do about that? Why would sublime import both?

If this is really a problem, we could do something similar to what you suggest is this PR, but I'd suggest a `RAISE_ERROR_ON_MIXING_QT_LIBS` variable that is True by default.

astrofrog:
In other packages, such as Glue (http://www.glue-viz.org) I've taken the approach of using a helper module to manage all this:

https://github.com/glue-viz/qt-helpers

Using this, you can import anything needed from Qt using e.g.

```
from qt_helpers import QtGui, QtCore
```

Whether to use PyQt4, PyQt5, or PySide can be fixed by an environment variable, otherwise one is chosen. Once one of the libraries is chosen, the other ones can no longer be imported.

Feel free to use ``qt_helpers.py`` in VisPy if it would help!

rossant:
@astrofrog do you know if there are conda packages of PyQt5 that work on all systems (Windows, OS X, Linux, 64-bit, Python 2 and Python 3)? Personally that's the only reason why I haven't switched to PyQt5 yet.

MatthieuDartiailh:
Spyder is now using https://github.com/spyder-ide/qtpy as Qt shim.

astrofrog:
@rossant - I usually try and find a PyQt5 package from one of the user conda channels that works, but haven't found one channel that consistently gives an up-to-date package that works on all platforms. I think the Continuum team will likely make an official PyQt5 conda package at some point:

https://github.com/ContinuumIO/anaconda-issues/issues/654#issuecomment-181874077

(I'll cc @msarahan in case there are any updates from the conda side!)

Eric89GXL:
Closing for now since it seems like using `qt_helpers.py` might be the way to go, but @ombschervister let me know if you disagree

Issue #1163, opened 2016-01-24
unicode in shader


ombschervister:
If comments in glsl-code has unicode symbols then shader compilation error raises.
Can you enable remove_comments function in gloo.preprocessor.preprocess?

```
WARNING: Traceback (most recent call last):
  File "D:/work/rexlab/test.py", line 49, in <module>
    app.run()
  File "D:\work\vispy\vispy\app\_default_app.py", line 62, in run
    return default_app.run()
  File "D:\work\vispy\vispy\app\application.py", line 137, in run
    return self._backend._vispy_run()
  File "D:\work\vispy\vispy\app\backends\_qt.py", line 232, in _vispy_run
    return app.exec_()
  File "D:\work\vispy\vispy\app\backends\_qt.py", line 434, in event
    out = super(QtBaseCanvasBackend, self).event(ev)
  File "D:\work\vispy\vispy\app\backends\_qt.py", line 434, in event
    out = super(QtBaseCanvasBackend, self).event(ev)
  File "D:\work\vispy\vispy\app\backends\_qt.py", line 706, in paintGL
    self._vispy_canvas.events.draw(region=None)
  File "D:\work\vispy\vispy\util\event.py", line 455, in __call__
    self._invoke_callback(cb, event)
  File "D:\work\vispy\vispy\util\event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  << caught exception here: >>
  File "D:\work\vispy\vispy\util\event.py", line 471, in _invoke_callback
    cb(event)
  File "D:/work/rexlab/test.py", line 45, in on_draw
    self.program.draw('triangle_strip')
  File "D:\work\vispy\vispy\gloo\program.py", line 470, in draw
    canvas.context.flush_commands()
  File "D:\work\vispy\vispy\gloo\context.py", line 171, in flush_commands
    self.glir.flush(self.shared.parser)
  File "D:\work\vispy\vispy\gloo\glir.py", line 232, in flush
    self._shared.flush(parser)
  File "D:\work\vispy\vispy\gloo\glir.py", line 146, in flush
    parser.parse(self._filter(self.clear(), parser))
  File "D:\work\vispy\vispy\gloo\glir.py", line 458, in parse
    self._parse(command)
  File "D:\work\vispy\vispy\gloo\glir.py", line 436, in _parse
    ob.set_shaders(*args)
  File "D:\work\vispy\vispy\gloo\glir.py", line 631, in set_shaders
[...]

PR #1164, opened 2016-01-24
enable remove_comments in shader preprocessor

ombschervister:
To #1163 for allow comments on own language

### DIFF
--- a/vispy/gloo/preprocessor.py
+++ b/vispy/gloo/preprocessor.py
@@ -65,6 +65,6 @@ def preprocess(code):
     """Preprocess a code by removing comments, version and merging includes."""
 
     if code:
-        #code = remove_comments(code)
+        code = remove_comments(code)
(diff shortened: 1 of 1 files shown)


campagnola:
I think it is currently disable because it messes with the line numbers in error reports. Can you check whether this is true?
If so, perhaps we can just replace commented lines with blank lines.

ombschervister:
Ok, I will do it

ombschervister:
Yes, you are right about incorrect line number in error reports. And i tried to fix it in this pull request.

But there is incorrect line number because adding "#version 120" too.

Eric89GXL:
@ombschervister is this ready to go from your end? Can you add a small test that would fail on `master` (e.g., with unicode in a GLSL comment) but passes now?

Eric89GXL:
Also, you need to adjust one of the existing tests, since it fails now

PR #1179, opened 2016-02-23
Fix bug that caused settable Isosurface.color property to not work

astrofrog:
In master, if one creates an ``IsosurfaceVisual``, the ``color`` property which is normally settable for ``MeshVisual`` is not settable for ``IsosurfaceVisual`` because ``set_data`` doesn't accept colors:

```
In [2]: surface.color = (1, 0.5, 0.3, 1.0)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-c4c805007590> in <module>()
----> 1 surface.color = (1, 0.5, 0.3, 1.0)

/Users/tom/tmp/vispy/vispy/util/frozen.py in __setattr__(self, key, value)
     15                                  '"unfreeze()" to allow addition of new '
     16                                  'attributes' % (key, self))
---> 17         object.__setattr__(self, key, value)
     18 
     19     def freeze(self):

/Users/tom/tmp/vispy/vispy/visuals/mesh.py in color(self, c)
    278     @color.setter
    279     def color(self, c):
--> 280         self.set_data(color=c)
    281 
    282     def mesh_data_changed(self):

TypeError: set_data() got an unexpected keyword argument 'color'
```

I just went ahead and made it possible to set the colors via ``set_data``, which solves the issue.

### DIFF
--- a/vispy/visuals/isosurface.py
+++ b/vispy/visuals/isosurface.py
@@ -6,6 +6,7 @@ from __future__ import division
 
 from .mesh import MeshVisual
 from ..geometry.isosurface import isosurface
+from ..color import Color
 
 
 class IsosurfaceVisual(MeshVisual):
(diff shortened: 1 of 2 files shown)


astrofrog:
I expanded this to make sure that when changing the colors, we update the MeshVisual without recomputing the vertices/faces. This is now ready for review.

astrofrog:
The AppVeyor failure appears to be unrelated.

Eric89GXL:
Looks great, thanks for the fix

PR #1180, opened 2016-02-23
Fix isosurface function for non-contiguous arrays

astrofrog:
The ``vispy.geometry.isosurface.isosurface(...)`` function makes use of the ``strides`` property of arrays, and we therefore need to make sure that the data is contiguous.


astrofrog:
The AppVeyor failure appears to be unrelated

Eric89GXL:
Thanks @astrofrog

Issue #1155, opened 2016-01-17
allow more than three sided polygons in `vispy.io.writemesh()`


glyg:
Hi,

Wavefront OBJ file format allows faces with more than 3 vertices, according to [wikipedia](https://en.wikipedia.org/wiki/Wavefront_.obj_file#Face_Elements).

The `WavefrontWriter` class currently only supports triangular faces, due to the reshape happening here:

https://github.com/vispy/vispy/blob/master/vispy/io/wavefront.py#L307

```python
        # Reshape faces
        Nfaces = faces.size // 3
        faces = faces.reshape((Nfaces, 3))
```

I think commenting out these lines, and being more specific in the docstring regarding the format of the `faces` argument is enough to allow that, as `writefaces` doesn't explicitly expects `val` to have 3 elements.

I think the users understand they need to pass `faces` such that it iterates over the faces any way, so even for triangles, `faces` is already passed a a (NFaces, 3) ndarray. 

To ensure backward compat, maybe the above lines can be wrapped in a try block?

glyg:
A quick follow up:

I did the above on my local repo, tests are passing and I can write polygonal meshes in .OBJ, which are parsed correctly by blender.

Here is a simple example:

```python
from vispy.io import write_mesh

vertices = np.array([[ 0.5  ,  1.375,  0.   ],
                     [ 0.5  ,  0.625,  0.   ],
                     [ 3.25 ,  1.   ,  0.   ],
                     [ 1.   ,  0.375,  0.   ],
                     [ 2.   ,  0.375,  0.   ],
                     [ 1.5  ,  0.625,  0.   ],
                     [ 1.5  ,  1.375,  0.   ],
                     [ 1.   ,  1.625,  0.   ],
                     [ 2.   ,  1.625,  0.   ]])

faces = np.array([[1, 0, 7, 6, 5, 3],
                  [4, 5, 6, 8, 2]])
write_mesh('vp_test.obj', vertices=vertices,
           faces=faces, normals=None, 
           texcoords=None, overwrite=True)
```

Here is the generated file:
```
# Wavefront OBJ file
# Created by vispy.
#
# unnamed object
# 9 vertices, no texcords, no normals, 2 faces

v 0.5 1.375 0.0
v 0.5 0.625 0.0
v 3.25 1.0 0.0
v 1.0 0.375 0.0
v 2.0 0.375 0.0
v 1.5 0.625 0.0
v 1.5 1.375 0.0
v 1.0 1.625 0.0
[...]

PR #1186, opened 2016-03-14
Added support for writing non-triangular meshes in .OBJ files -- GH #1155

glyg:
Here we go.

Not sure the tests are correctly or sensibly written. Also not sure if one can (and should) test if a warning is raised.

Waiting for comments!

### DIFF
--- a/vispy/io/wavefront.py
+++ b/vispy/io/wavefront.py
@@ -79,7 +79,9 @@ class WavefrontReader(object):
         # Done
         t0 = time.time()
         mesh = reader.finish()
-        logger.debug('reading mesh took ' + str(time.time() - t0) + ' seconds')
+        logger.debug('reading mesh took ' +
+                     str(time.time() - t0) +
+                     ' seconds')
         return mesh
 
     def readLine(self):
@@ -303,10 +305,14 @@ class WavefrontWriter(object):
         if faces is None:
             faces = np.arange(len(vertices))
 
-        # Reshape faces
-        Nfaces = faces.size // 3
-        faces = faces.reshape((Nfaces, 3))
-
+        try:
+            # Reshape faces
+            Nfaces = faces.size // 3
+            faces = faces.reshape((Nfaces, 3))
+        except ValueError:
+            logger.warning(
+                '''Faces doesn't appear to be triangular, '''
+                '''be advised the file cannot be read back in vispy''')
         # Number of vertices
         N = vertices.shape[0]
 
--- a/vispy/io/wavefront.py
+++ b/vispy/io/wavefront.py
@@ -319,7 +319,7 @@ class WavefrontWriter(object):
             is_triangular = np.array([len(f) == 3
                                       for f in faces])
             if not(np.all(is_triangular)):
-                log.warning('''Faces doesn't appear to be triangular,
+                logger.warning('''Faces doesn't appear to be triangular,
                 be advised the file cannot be read back in vispy''')
         # Number of vertices
         N = vertices.shape[0]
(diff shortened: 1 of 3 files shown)


vispy/io/wavefront.py
```diff
@@ -303,10 +305,13 @@ def writeMesh(self, vertices, faces, normals, values, name=''):
...
-        Nfaces = faces.size // 3
-        faces = faces.reshape((Nfaces, 3))
-
+        try:
+            # Reshape faces
+            Nfaces = faces.size // 3
+            faces = faces.reshape((Nfaces, 3))
+        except ValueError:
+            logger.warning('''Faces doesn't appear to be triangular, '''
+                           '''be advised the file cannot be read back''')
```
almarklein: Can you please add "... be read back *in Vispy*" here, to make it explicit that this is a vispy thing?

almarklein:
Other than that, this LGTM

glyg:
Fixed the warning and squashed the commits

Thx for the review

vispy/io/wavefront.py
```diff
@@ -303,10 +305,14 @@ def writeMesh(self, vertices, faces, normals, values, name=''):
...
             faces = np.arange(len(vertices))
 
-        # Reshape faces
-        Nfaces = faces.size // 3
-        faces = faces.reshape((Nfaces, 3))
-
+        try:
+            # Reshape faces
+            Nfaces = faces.size // 3
+            faces = faces.reshape((Nfaces, 3))
```
Eric89GXL: hmm... but couldn't you have a non-triangular file that just happened to be divisibly by 3, thus making this "check" not correct?

glyg:
Damn you're right!
Now I'm confused.

My understanding is that the reshape has to happen primarily if the `faces` argument was not provided, in which case it's created just above the reshape, and needs reshaping... Now if the user provides the `face` argument, one can assume it will be correctly shaped? 

Said otherwise, if `faces` is already there, why reshape it without telling the user?

Another point to consider: in the provided test, `faces` is a 1D array of lists, so it's even worse, because you'll end up with strange triples of lists after the reshape, which will raise a `TypeError` further down the road, when increasing the index by 1...

Eric89GXL:
I am a bit far removed from the code at this point to be able to tell you definitively. Can you tell from the tests what is supposed to happen?

Instead of implicitly trying to reshape, could we add or modify a `kwarg` that would default to doing a reshape (to match current behavior) but could be turned off (for your use case)?

glyg:
I rewrote the test with three polygonal faces, and it does rise a `TypeError`

```python
/home/guillaume/anaconda3/envs/py35/lib/python3.5/site-packages/vispy-0.5.0.dev0-py3.5.egg/vispy/io/wavefront.py in writeFace(self, val, what)
    278         """
    279         # OBJ counts from 1
--> 280         val = [v + 1 for v in val]
    281         # Make string
    282         if self._hasValues and self._hasNormals:

/home/guillaume/anaconda3/envs/py35/lib/python3.5/site-packages/vispy-0.5.0.dev0-py3.5.egg/vispy/io/wavefront.py in <listcomp>(.0)
    278         """
    279         # OBJ counts from 1
--> 280         val = [v + 1 for v in val]
    281         # Make string
    282         if self._hasValues and self._hasNormals:

TypeError: can only concatenate list (not "int") to list
```

If a `kwarg` is acceptable, it is definitely the easier way to go.
I'll modify the code to see how it rolls.

vispy/io/wavefront.py
QuLogic: Also, why waste the space for triple quotes if you aren't going to use its main benefit (unescaped newlines).

glyg: fair point! I'll fix that too

Eric89GXL:
Thanks @glyg

glyg:
thanks to you all! Hope I can be of more use in the future!

Issue #1175, opened 2016-02-23
Depth order incorrect with isosurface visual


astrofrog:
If I run the following simple example for an isosurface:

```python
import sys
import numpy as np

from vispy import app, scene

# Create a canvas with a 3D viewport
canvas = scene.SceneCanvas(keys='interactive')
view = canvas.central_widget.add_view()


data = np.arange(1000).reshape(10, 10, 10)

surface = scene.visuals.Isosurface(data, level=data.max()/4.,
                                   color=(0.5, 0.6, 1, 1.0), shading=None,
                                   parent=view.scene)

surface2 = scene.visuals.Isosurface(data, level=data.max()/2.,
                                   color=(1.0, 0.2, 0.2, 1), shading=None,
                                   parent=view.scene)


axis = scene.visuals.XYZAxis(parent=view.scene)

cam = scene.TurntableCamera(elevation=30, azimuth=30)
cam.set_range((-10, 10), (-10, 10), (-10, 10))
view.camera = cam
```

I get the right output from some viewing angles:

![screen shot 2016-02-23 at 12 28 19 pm](https://cloud.githubusercontent.com/assets/314716/13251590/24a5bf0e-da29-11e5-90ef-3cd202a1fd39.png)

but if I turn it around, and look from a point of view where the red should be in front of the blue, there are artifacts:

![screen shot 2016-02-23 at 12 28 26 pm](https://cloud.githubusercontent.com/assets/314716/13251600/36082426-da29-11e5-92f8-7a04d6ccb175.png)

rougier:
Maybe depth_test is off ?

astrofrog:
@rougier - I don't think so: ``IsoSurfaceVisual`` uses ``MeshVisual`` which includes this line:

```python
        self.set_gl_state('translucent', depth_test=True,
                          cull_face=False)
```

astrofrog:
@rougier - just out of curiosity, do you also see this issue? I'm trying to figure out if all my issues are somehow related to OpenGL version?

rougier:
Yes, I got this issue. It's hard to tell where is the problem (I've becoming unfamiliar with scene stuff).

astrofrog:
This appears to be fixed by setting depth_size to 24:

canvas = scene.SceneCanvas(keys='interactive', config={'depth_size': 24})

as suggested by @PennyQ in #1174. As such, this is a duplicate issue of #896, so I'll close it.

Issue #1174, opened 2016-02-23
Missing tiles in rendering of isosurface example


astrofrog:
Running the ``examples/basic/scene/isosurface.py`` example produces the following output for me with the latest developer version of VisPy:

![screen shot 2016-02-23 at 12 22 25 pm](https://cloud.githubusercontent.com/assets/314716/13251448/43442af0-da28-11e5-8c18-0e506db488cd.png)

That doesn't seem quite right since by default one shouldn't be seeing through the main isosurface. It looks like some tiles are missing?

PennyQ:
@astrofrog Yeah, this could be fixed by setting 
`canvas = scene.SceneCanvas(keys='interactive', config={'depth_size': 24})`
here is a issue for reference https://github.com/vispy/vispy/issues/896

astrofrog:
@PennyQ - ok thanks! I'll close this as a duplicate of #896

Issue #896, opened 2015-05-09
Mesh Visualization


SimonDanisch:
Hi,
next example isn't working:
![image](https://cloud.githubusercontent.com/assets/1010467/7551912/c616bec4-f69a-11e4-9113-d9f4009a07ad.png)
This looks definitely like wrong normals.
Is this a similar problem as in #892 ?

```Julia
# -*- coding: utf-8 -*-
# Copyright (c) 2014, Vispy Development Team.
# Distributed under the (new) BSD License. See LICENSE.txt for more info.

"""
Simple demonstration of Mesh visual.
"""
import sys
import numpy as np
from vispy import scene
from vispy import app, gloo, visuals, io, geometry
from vispy.geometry import create_sphere
from vispy.visuals.transforms import (STTransform, AffineTransform,
                                      ChainTransform)

canvas      = scene.SceneCanvas(keys='interactive')
canvas.view = canvas.central_widget.add_view()

verts, faces, normals, nothin = io.read_mesh("Rider.obj")
mesh = scene.visuals.Mesh(vertices=verts, shading='smooth', faces=faces)

canvas.view.add(mesh)
canvas.view.camera = scene.TurntableCamera()

canvas.view.camera.set_range((-20, 20), (-20, 20), (-20, 20))
canvas.show()

if __name__ == '__main__':
    if sys.flags.interactive != 1:
        canvas.app.run()
```
Best,
Simon

Eric89GXL:
Did this work at some point previously? Could show the same problem with our `triceratops` object? If not, could you upload the model somewhere?

Do you know if this worked at all in a previous version of `vispy`?

SimonDanisch:
that's my second try with vispy, so I don't know about previous states ;)
Do you mean  the tutorial\visuals\05_viewer_location.py example?
It doesn't seem to use the newest API. When I use the mesh in my code snippet it looks different but definitely weird.
![image](https://cloud.githubusercontent.com/assets/1010467/7553367/605c8230-f6f4-11e4-909e-c6ebdf43ba32.png)

I got the model fromturbosquid:

Eric89GXL:
Yeah, and in case we need a simpler example:
```
>>> from vispy import io, plot
>>> fname = io.load_data_file('orig/triceratops.obj.gz')
>>> fig = plot.Fig()
>>> fig[0, 0].mesh(*io.read_mesh(fname)[:2])
```
I get the same behavior there. @campagnola any idea what might be causing this?

[...]

PR #1190, opened 2016-03-18
Set default depth to 24 bits 

rougier:
Several problems have been reported when using 16 bits depth (see #1174, #1175, #896).
This PR sets default to 24.


almarklein:
Thanks Nico :)

Issue #1189, opened 2016-03-16
plot 3D data


bendichter:
Is there a way to plot 3D data using the plot interface? From the documentation, ax.plot is limited to 2D data and I cannot find a function for plotting data in x,y,z.

Thanks!

liubenyuan:
Hi,

I build a simple (naive) `plot3d` using `vispy.gloo`, it was simply modified from [Getting started with Vispy](http://ipython-books.github.io/featured-06/), the code can be download at

[02-plot3d.py](https://github.com/liubenyuan/vispy-tutorial/blob/master/gloo/examples/02-plot3d.py)

**BTW,** It would be greate if there is some tutorial on how to build your own **high-level** plotting command from the ground up, i.e., how to write code from `gloo` to `visual` to `scene` and finally to `plot` or `mpl_plot` ? so that I could export my figure for publication purpose.

byliu

```python

import numpy as np
from vispy import app, gloo
from vispy.util.transforms import perspective, translate, rotate

vertex = """
uniform mat4   u_model;         // Model matrix
uniform mat4   u_view;          // View matrix
uniform mat4   u_projection;    // Projection matrix
attribute vec3 a_position;
void main()
{
    gl_Position = u_projection * u_view * u_model * vec4(a_position, 1.0);
}
"""

# The other shader we need to create is the fragment shader.
# It lets us control the pixels' color.
fragment = """
void main()
{
    gl_FragColor = vec4(0.0, 0.0, 0.0, 1.0);
}
"""


class Canvas(app.Canvas):

    def __init__(self, data, theta=37.5, phi=75, z=6):
        # In order to display a window, we need to create a Canvas.
        app.Canvas.__init__(self, size=(800, 400), title='plot3d',
                            keys='interactive')

        # build shader program
        prog = gloo.Program(vert=vertex, frag=fragment)

        # initialize 3D view
        view = np.eye(4, dtype=np.float32)
        model = np.eye(4, dtype=np.float32)
        projection = np.eye(4, dtype=np.float32)

        view = translate((0, 0, -z))
        prog['u_model'] = model
        prog['u_view'] = view
        prog['u_projection'] = projection
        prog['a_position'] = data

        # bind
        self.program = prog
        self.theta = theta
        self.phi = phi

        # config
        gloo.set_viewport(0, 0, *self.physical_size)
        gloo.set_clear_color('white')
        gloo.set_state('opaque')

        # show the canvas
        self.show()

    def on_resize(self, event):
        """
[...]

PR #1191, opened 2016-03-19
enable 3D plot in LinePlotVisual, add a demo using plot3d

liubenyuan:
#1189 

In `gloo.visuals`, both `LineVisual` and `MarkersVisual` have 3D capability, but in `LinePlotVisual`, 3D is disabled, which makes some typical visual task such as `plot3d`, `scatter3d` not available. This patch enables this feature, and provides a simple demo on using `LinePlotVisual` as a `plot3d` or `scatter3d` command in `examples/basic/visuals`.

### DIFF
--- a/examples/basics/visuals/line_plot3d.py
+++ b/examples/basics/visuals/line_plot3d.py
@@ -2,13 +2,15 @@
 """ plot3d using existing visuals : LinePlotVisual """
 
 import numpy as np
+import sys
+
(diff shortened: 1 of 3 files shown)


Eric89GXL:
Thanks @liubenyuan

Juanlu001:
Looking forward to a new release that incorporates this :)

PR #1193, opened 2016-03-23
Fix boolean uniforms in GLIR

airhorns:
I was unable to get boolean uniforms to work in my shaders until I made these changes. I have somehow managed to lose my code that could reproduce the error, sorry :(. What kind of test can I add that will verify this doesn't break in the future? I'm super new to working with GL and vispy so I'm not quite sure what's appropriate for getting this merged, so any help would be much appreciated!


Eric89GXL:
> What kind of test can I add that will verify this doesn't break in the future?

GLIR tests are a bit difficult to write. The easiest thing might be to do some draw call that depends on the `bool` value being correct, and check the result. Set it to the opposite value, and check that the result has changed.

Eric89GXL:
Merging for now, thanks @airhorns

PR #1195, opened 2016-03-25
doc: Add URL link for Qt in README.rst

root-nishi:
Make Qt point to http://www.qt.io, as new python developers may not be aware of it.


Eric89GXL:
Thanks for the link

QuLogic:
Should've been https...

PR #1199, opened 2016-04-01
Axis label for AxisVisual class

astrofrog:
It would be nice to have an easy way to add an axis label (in addition to the tick labels) for the AxisVisual class. This PR is a very rough implementation of that. Currently this looks like:

screen shot 2016-04-01 at 4 24 56 pm

The text is always lined up with the screen. There are still several things to do to make this acceptable in my opinion. For example we need to choose a spacing from the axis to the label that depends on the font size and can be overriden by the user.

However, even then it might not be very pretty. Does anyone have suggestions on how we could do this better?

cc @PennyQ

### DIFF
--- a/examples/basics/visuals/line.py
+++ b/examples/basics/visuals/line.py
@@ -7,6 +7,9 @@
 """
 Demonstration of various features of Line visual.
 """
+
+from __future__ import division
+
 import sys
 import numpy as np
 
@@ -28,7 +31,7 @@ color[:, 1] = color[::-1, 0]
 connect = np.empty((N-1, 2), np.int32)
 connect[:, 0] = np.arange(N-1)
 connect[:, 1] = connect[:, 0] + 1
-connect[N/2, 1] = N/2  # put a break in the middle
+connect[N//2, 1] = N//2  # put a break in the middle
 
 
 class Canvas(app.Canvas):
--- a/examples/basics/scene/axes_plot.py
+++ b/examples/basics/scene/axes_plot.py
@@ -18,28 +18,29 @@ grid.spacing = 0
 
 title = scene.Label("Plot Title", color='white')
 title.height_max = 40
-grid.add_widget(title, row=0, col=0, col_span=3)
+grid.add_widget(title, row=0, col=0, col_span=2)
 
-yaxis = scene.AxisWidget(orientation='left')
-yaxis.width_max = 40
-grid.add_widget(yaxis, row=1, col=1)
+yaxis = scene.AxisWidget(orientation='left',
+                         axis_label='Y Axis',
+                         axis_font_size=12,
+                         axis_label_margin=50,
+                         tick_label_margin=5)
+yaxis.width_max = 80
+grid.add_widget(yaxis, row=1, col=0)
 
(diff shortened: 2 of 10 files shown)


Eric89GXL:
@astrofrog did you get this working? If it works decently we could get it into 0.5.

astrofrog:
@Eric89GXL - sorry for the delay, I finally got back to this. I added a number of options to control the appearance of the axes. Could you or someone else review this?

One thing I'd like to improve in future is the orientation of the axis labels, which should be parallel to the axes themselves I think. Shouldn't be too hard to do, but would prefer to leave it to a separate pull request.

astrofrog:
I decided to just implement axis label rotation:

screen shot 2016-09-15 at 3 14 08 pm

examples/basics/visuals/line.py
```diff
@@ -7,6 +7,9 @@
 """
 Demonstration of various features of Line visual.
 """
+
+from __future__ import division
```
Eric89GXL: I don't think you need to actually do this import to use `//`

Eric89GXL:
This looks reasonable to me. What example do you suggest I run to test it?

astrofrog:
@Eric89GXL - I'm not sure which existing example would be best to test it - was using a custom script to try it out. But there might be some existing examples we should check to make sure they still work. Will try and investigate.

examples/basics/visuals/line.py
astrofrog: Ah I wasn't sure if you supported Python 2.6. Still, I prefer to do this to be explicit to make / behave the same on 2.x and 3.x

astrofrog:
@Eric89GXL - you can try ``examples/basics/scene/axes_plot.py`` to make sure I didn't break anything. That example adds the axes labels manually rather than using the new functionality in this PR.

kmuehlbauer:
Good work! I tested with the `axes_plot.py` and had to make some changes there. @astrofrog I'll have a PR with these changes pointing to your fork ready in a minute . Said that, the example works. Maybe we can have also an example only for the `AxisVisual` showing the possibilities?

Minor thing: The `tick_label_margin` defaults now to 15, before was 5. I'm not sure what to take best as default, but 15 is IMHO a bit too far away.

astrofrog:
@kmuehlbauer - thanks for updating the example! I've changed the default tick_label_margin back to 5. I will try and add an example in a separate PR showing how to use AxisVisual to construct 3D axes, so maybe that will be enough?

kmuehlbauer:
@astrofrog Cool. A 3D example would be really nice. Hopefully this can be merged soon, @Eric89GXL?

astrofrog:
@Eric89GXL - it might make sense to change the coveralls preferences to set the status on the PR rather than commenting - what do you think?

kmuehlbauer:
@astrofrog :grinning:  Make some noise!

astrofrog:
We need to change coveralls to set the PR status instead of commenting...

astrofrog:
I've just realized I have access to the coveralls preferences, so I've disabled commenting

astrofrog:
@Eric89GXL - I think this is ready to go

Eric89GXL:
Thanks @astrofrog

Issue #1203, opened 2016-04-19
A few minor bugs


scottpaine:
I've been working with this package for about a month now building my own methods to visualize 2D datasets and am loving the fast speeds and intuitive methods for using canvases/visuals.

That being said, I've found two bugs.

**First**, in ColorBarVisual, setting the size doesn't actually set the size - it sets the halfdims:
```python
@size.setter
    def size(self, size):
        ...
        if self.orientation in ["top", "bottom"]:
            (width, height) = size
        else:
            (height, width) = size
        ...
        self._halfdim = (width, height)
        self._update()
```
I think that the halfdims should be `width // 2, height // 2`, but in either case if I make a visual called `cbar` and say `cbar.size = (200, 100)`, I end up with a visual that is 400 x 200 pixels.

**Second**, in ImageVisual, resetting the `clim` doesn't redraw the image. When setting the `clim`, the bool `_need_vertex_update` is set to true:

```python
@clim.setter
def clim(self, clim):
        ...
        self._need_vertex_update = True
        self.update()
```

In `_prepare_draw`, this makes the visual call `_build_vertex_data`, which itself doesn't use `clim` at all:

```python
def _prepare_draw(self, view):
        ...
        if self._need_vertex_update:
            self._build_vertex_data()
        ...
```

The only method in ImageVisual which does use the `clim` is in `_build_texture`:

```python
def _build_texture(self):
        ...
        if data.ndim == 2 or data.shape[2] == 1:
            # deal with clim on CPU b/c of texture depth limits :(
            # can eventually do this by simulating 32-bit float... maybe
            clim = self._clim
            if isinstance(clim, string_types) and clim == 'auto':
                clim = np.min(data), np.max(data)
            clim = np.asarray(clim, dtype=np.float32)
            data = data - clim[0]  # not inplace so we don't modify orig data
            if clim[1] - clim[0] > 0:
                data /= clim[1] - clim[0]
            else:
                data[:] = 1 if data[0, 0] != 0 else 0
            self._clim = np.array(clim)

        self._texture.set_data(data)
        self._need_texture_upload = False
```

And this is called from `_prepare_draw` when `_need_texture_upload` is `True`. So, I think the fix to this is easy - instead of setting `_need_vertex_update` to `True`, `_need_texture_upload` needs to be set when resetting the `clim`.

**Third**, in PlotWidget, calling the method `colorbar` with `position="right"` causes a bad drawing. This mainly comes from what happens when assigning with `right`:

```python
[...]

PR #1217, opened 2016-05-16
Minor bugfixes

scottpaine:
Fixes for issues raised in [#1203](https://github.com/vispy/vispy/issues/1203)

### DIFF
--- a/vispy/visuals/colorbar.py
+++ b/vispy/visuals/colorbar.py
@@ -679,5 +679,5 @@ class ColorBarVisual(CompoundVisual):
         elif height == 0.:
             raise ValueError("height must be non-zero, not %s" % (height, ))
 
-        self._halfdim = (width / 2, height / 2)
+        self._halfdim = (width / 2., height / 2.)
         self._update()
--- a/vispy/visuals/colorbar.py
+++ b/vispy/visuals/colorbar.py
@@ -679,5 +679,5 @@ class ColorBarVisual(CompoundVisual):
         elif height == 0.:
             raise ValueError("height must be non-zero, not %s" % (height, ))
 
-        self._halfdim = (width, height)
+        self._halfdim = (width / 2, height / 2)
         self._update()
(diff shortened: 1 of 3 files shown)


vispy/visuals/colorbar.py
```diff
@@ -679,5 +679,5 @@ def size(self, size):
         elif height == 0.:
             raise ValueError("height must be non-zero, not %s" % (height, ))
 
-        self._halfdim = (width, height)
+        self._halfdim = (width / 2, height / 2)
```
Eric89GXL: This might need to be `//` (integer division) for Python 2, WDYT?

scottpaine: I figured since `size` as a property is returned as `(halfw * 2., halfh * 2.)` (or `(halfh * 2., halfw * 2.)`), if you want odd pixel amounts or a size that isn't in pixels you need to assume that `width`/`height` are floats. It might actually be good to convert them to floats so that Python 2 doesn't break them when doing math.

Eric89GXL: sure, to get them to be floats just change it to `/ 2.`

scottpaine: If they're passed in as ints in Python 2 and then you do `/ 2` on them, it'll return the integer value for `/ 2`. What I mean is more like this:

```python
...
if self.orientation in ["top", "bottom"]:
            (width, height) = size
        else:
            (height, width) = size
width = float(width)
height = float(height)
...
```

That way, Python 2 will be happy if you do `/ 2` on them and return the right value. Python 3 just plain won't care.

Eric89GXL: Note the `.` at the end of the `2.` in my quote above -- it's in the code not as part of the sentence. If you do `x / 2.` the result will be a float in Python2 and Python3

scottpaine: Didn't look close enough! Committing now.

Eric89GXL: I also didn't put any punctuation at the end of my sentence, so I didn't write goodly enough :)

Eric89GXL:
Thanks @scottpaine

PR #1220, opened 2016-05-18
BUG: fixed a bug that caused a mesh to disappear if the color was set using the .color attribute

astrofrog:
This is because previously, set_data would get called with just the color, but this resulted in MeshData being reinitialized with all parameters set to None.

### DIFF
--- a/vispy/visuals/mesh.py
+++ b/vispy/visuals/mesh.py
@@ -277,7 +277,9 @@ class MeshVisual(Visual):
 
     @color.setter
     def color(self, c):
-        self.set_data(color=c)
+        if c is not None:
+            self._color = Color(c)
+        self.mesh_data_changed()
 
     def mesh_data_changed(self):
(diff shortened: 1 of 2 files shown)


Eric89GXL:
LGTM, feel free to merge if the CIs are happy. If you get the same Travis error here (3.4 examples) as in your other PR, go ahead and merge and we can fix the CI separately.

Eric89GXL:
`vispy/visuals/tests/test_mesh.py:14:5: F841 local variable 'vol' is assigned to but never used`

astrofrog:
@Eric89GXL - oops, this was copied from another test, will fix.

Eric89GXL:
Do you edit with `pyflakes` and `pep8` on? It can help spot those

astrofrog:
@Eric89GXL - only PEP8 by default, but I should change that :)

PR #1223, opened 2016-05-19
ENH: Graph visual

Eric89GXL:
We'll see if the CIs are happy. I think that's all that's left to be done. The rest could be subsequent PRs/improvements.


Eric89GXL:
Phew, CIs are good

PR #1224, opened 2016-05-19
API: Remove mpl_plot

### LINKED ISSUE #1099: Matplotlib Backend not found
How do you get the matplotlib backend to work? Should I be setting this in the matplotlibrc file, or elsewhere?

Eric89GXL:
In the interest of maintainability, I think we need to remove the `mpl` backend. It doesn't add too much for people, and adds maintenance burden for us. It might be good to trim down in some other, places, too, not sure.

Closes #1099.


Eric89GXL:
I'll merge tomorrow unless anyone objects

PR #1225, opened 2016-05-19
Add a points selection example

PennyQ:
Here is a script for selecting scatter points along ray casting direction in 3D viewer, with control by the KeyPress 1~4 to enter the selection mode, which allows user draw a selection region ( with different shapes as free lasso, rectangle, ellipse and picking) on the canvas and get points within that region.

The selected points will be highlighted with red color as shown below :)

image

### DIFF
--- a/examples/basics/scene/points_selection.py
+++ b/examples/basics/scene/points_selection.py
@@ -1,0 +1,279 @@
+"""
+    3D Points Picking & select tool based on Vispy library
+    Author: Penny Xuran Qian (NAOC/SAO) pennyqxr@gmail.com
+
+    Controls:
+    * 1 - free lasso select
+    * 2 - rectangle select
+    * 3 - ellipse select
+    * 4 - point picking
+
+    press again to switch between select and view mode
+
+"""
+
+import sys
+import numpy as np
+
+from PyQt4 import QtGui
+from vispy import scene
+from matplotlib import path
+
+
+def rectangle_vertice(center, height, width):
+    # Borrow from _generate_vertices in vispy/visuals/rectangle.py
+
+    half_height = height / 2.
+    half_width = width / 2.
+
+    bias1 = np.ones(4) * half_width
+    bias2 = np.ones(4) * half_height
+
+    corner1 = np.empty([1, 3], dtype=np.float32)
+    corner2 = np.empty([1, 3], dtype=np.float32)
+    corner3 = np.empty([1, 3], dtype=np.float32)
+    corner4 = np.empty([1, 3], dtype=np.float32)
+
+    corner1[:, 0] = center[0] - bias1[0]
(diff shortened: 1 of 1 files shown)


examples/basics/scene/points_selection.py
```diff
@@ -0,0 +1,271 @@
...
+                mask = [selection_path.contains_points(data)]
+
+                self.selected = mask
+                self.mark_selected()
+
+                # Reset lasso
+                self.line_pos = []  # TODO: Empty pos input is not allowed for line_visual
+                self.line.set_data(np.array(self.line_pos))
+                self.line.update()
+
```
PennyQ: Here will pop up an `ValueError`  that the `LineVisual` doesn't accept empty input data as pos, is there any other way to do the reset for the `LineVisual` ? Thanks!

The error info:
```
...
File "/Users/penny/anaconda/lib/python2.7/site-packages/vispy-0.5.0.dev0-py2.7.egg/vispy/visuals/line/line.py", line 312, in _prepare_draw
    self._pos_vbo.set_data(pos)
  File "/Users/penny/anaconda/lib/python2.7/site-packages/vispy-0.5.0.dev0-py2.7.egg/vispy/gloo/buffer.py", line 192, in set_data
    data = self._prepare_data(data, **kwargs)
  File "/Users/penny/anaconda/lib/python2.7/site-packages/vispy-0.5.0.dev0-py2.7.egg/vispy/gloo/buffer.py", line 451, in _prepare_data
    % (self._last_dim, c))
ValueError: Last dimension should be 2 not 1
```

Eric89GXL:
@PennyQ this seems useful but should probably use the `vispy.app` system instead of relying specifically on PyQt

examples/basics/scene/points_selection.py
```diff
@@ -0,0 +1,279 @@
...
+            self.select_id = event.text
+
+    def on_mouse_press(self, event):
+        # Realize picking functionality and set origin mouse pos
+
+        if event.button == 1 and self.select_flag:
+            if self.select_id == '4':
+                # Ray intersection on the CPU to highlight selected point(s)
+
+                data = self.tr.map(self.data)[:, :2]
```
mattangus, line 211: I'm trying to adapt your example into my own code. `self.tr.map` seems to return homogeneous coordinates. You need to change this to:

```
data = self.tr.map(self.data)
data = data[:, :2] / data[:, 2:3]
```

From testing `data[:, 2]` and `data[:, 3]` are (approximately) the same, so these are 4d homogeneous coords for 2d points.

djhoese:
I've added this to the SciPy 2021 Sprint project in case someone wants to pick it up again. We could try modifying the existing "picking.py" example and merge it with what's been done here while also taking in some of the comments above.

PR #1227, opened 2016-05-23
Fix isosurface function for use with Numpy 1.12.0.dev

astrofrog:
Before this, j was a one-element array and Numpy raised the error:

```
TypeError: only integer scalar arrays can be converted to a scalar index
```


Eric89GXL:
Thanks @astrofrog

PR #1234, opened 2016-06-05
add hide to alias show(False) to Canvas

chemelnucfin:
"Hide" was my first instinct to hide the canvas, instead of show(False).


Eric89GXL:
The problem as I see it is that there are many such potential aliases, e.g. `draw=paint=update`, `show(False)=hide`. Putting them all in violates the "one clear way" principle without much benefit.

PR #1244, opened 2016-06-20
Fix line width related documentation error

drewm1980:
I believe this is a trivial sync of the documentation with the implementation.  

If it's not, then the code should probably be fixed up by someone more familiar with subtleties of how rendering is supposed to work in vispy.


Eric89GXL:
Thanks @drewm1980

drewm1980:
No, thank YOU @Eric89GXL !

PR #1254, opened 2016-07-07
fix face/vertex colors

ktritz:
using vispy.plot mesh: vertex and face colors are now applied to mesh.
shading can be specified instead of always 'smooth'.

The mesh object in the plot widget was ignoring the face and vertex colors set during the Init by creating meshdata without these colors applied. It also applied a default 'smooth' shading, which doesn't work with face_colors.


astrofrog:
@ktritz - can you rebase? Some of the Travis failures may be fixed now.

ktritz:
Looks like I'm still getting doc string issues with the Travis builds.

vispy/plot/plotwidget.py
```diff
@@ -236,7 +236,8 @@ def mesh(self, vertices=None, faces=None, vertex_colors=None,
         meshdata : MeshData | None
             Meshdata to use. If not None, then vertices, faces, and fname
             must be None.
-
+        shading : str | None
+            Shading to use.
         Returns
```
kmuehlbauer: @ktritz Regarding the docstring violation, there might be an empty line missing before `Returns`.

ktritz:
Looks like adding a blank line before the Return in the docstring has fixed everything. This should be a pretty easy merge.

vispy/plot/plotwidget.py
```diff
@@ -222,7 +222,7 @@ def image(self, data, cmap='cubehelix', clim='auto', fg_color=None):
 
     def mesh(self, vertices=None, faces=None, vertex_colors=None,
              face_colors=None, color=(0.5, 0.5, 1.), fname=None,
-             meshdata=None):
+             meshdata=None, shading=None):
```
Eric89GXL: the old default was effectively `'smooth'` so it should stay that way, no?

ktritz: The problem with the 'smooth' default is that it prevents face_colors from working. Sure, people can explicitly set smooth=None to enable face_colors, but this might be confusing to some people if they try to modify the face_colors and see no effect.
However, if people think that's ok, It's not a big deal to me.
Let me know, and I'll modify the code accordingly.

Eric89GXL: ...or `shading='auto'` rather

ktritz: Or did you mean I should add the appropriate code to handle the 'auto' keyword as you specify?

Eric89GXL: You would need to add it

ktritz: Got it. I'll give it a go.

vispy/plot/plotwidget.py
```diff
@@ -243,13 +243,19 @@ def mesh(self, vertices=None, faces=None, vertex_colors=None,
         meshdata : MeshData | None
             Meshdata to use. If not None, then vertices, faces, and fname
             must be None.
+        shading : str | 'auto'
```
Eric89GXL: 'auto' is a `str`, so no need to put it up here

docstring should be more like (please fill in missing values if you can):

Shading to use, can be XXX, YYY, or ZZZ. Default ('auto') will use None if face_colors is set, and 'smooth' otherwise.

vispy/plot/plotwidget.py
```diff
@@ -243,13 +243,19 @@ def mesh(self, vertices=None, faces=None, vertex_colors=None,
...
 
         Returns
         -------
         mesh : instance of Mesh
             The mesh.
         """
         self._configure_3d()
+        shading = 'smooth'
+        if face_colors is not None:
+            shading = None
```
Eric89GXL: think about what happens if a user passes a `shading` param, you are ignoring it here

nest this in an `if shading == 'auto'` and you should be good

vispy/plot/plotwidget.py
```diff
@@ -260,10 +266,11 @@ def mesh(self, vertices=None, faces=None, vertex_colors=None,
                 raise ValueError('vertices, faces, and fname must be None if '
                                  'fname is not None')
         else:
-            meshdata = MeshData(vertices, faces)
+            meshdata = MeshData(vertices, faces, face_colors=face_colors,
+                                vertex_colors=vertex_colors)
```
Eric89GXL: Do you need to do this? It seems like the lines below (instantiation of the Mesh class) will already set these parameters.

ktritz: For some reason, face and vertex colors must be set in MeshData, or it doesn't work. I'm not sure if scene.Mesh just isn't applying the keywords properly, or if there is some other interaction.

vispy/plot/plotwidget.py
ktritz: Done!

Eric89GXL:
LGTM, will merge once the CIs come back happy

PR #1267, opened 2016-09-12
webgl backend fix for updated versions of ipython/jupyter notebook

itsagoodbrain:
The location of the widgets has changed in current versions of ipython/jupyter. Credit goes to @SamiAhola who posted this fix several months ago on the gitter vispy channel. However no PR for the fix has been made so I have made one here.

### DIFF
--- a/vispy/html/static/js/webgl-backend.js
+++ b/vispy/html/static/js/webgl-backend.js
@@ -18,7 +18,7 @@ define(function(require) {
     }
 
     var vispy = require("/nbextensions/vispy/vispy.min.js");
-    var widget = require("widgets/js/widget");
+    var widget = require("jupyter-js-widgets");
 
     var VispyView = widget.DOMWidgetView.extend({
 


rougier:
Great ! Thank you. Any  idea why the appveyor is failing ?

itsagoodbrain:
None so far. I was just looking at that, I'm quite surprised since it's a one line change that to my knowledge has worked for at least 4 people with different installs. I'll get back to you if I'm able to track it down.

itsagoodbrain:
Ah, it seems there's a test that is being run in this install method that wasn't run with the standard install method. I likely just need to modify the test. I'll update the PR soon with a fix hopefully.

itsagoodbrain:
Hm... I was really hoping that would have done the trick. I'm currently unable to replicate this locally. Any suggestions?

rougier:
I'm no specialist either. Can someone test it locally ?

astrofrog:
I'm fixing AppVeyor temporarily by pinning IPython to 4.2 instead of allowing it to use 5.x: https://github.com/vispy/vispy/pull/1269

astrofrog:
@itsagoodbrain - the failure is not due to the changes in this PR, and is happening for other PRs. It's due to changes in IPython 5.x that have yet to be tracked down.

astrofrog:
@itsagoodbrain - you will need to rebase this now - I think you should explicitly include the ipython version pinning in the CONDA_DEPENDENCIES as is now the case in master (when installing Jupyter, ipython gets installed anyway, and we want to avoid using IPython 5 for now)

itsagoodbrain:
@astrofrog Thanks for finding that issue and notifying me. Rebased and should be good now. I removed the extra changes since they were just attempting to address that issue. I actually hadn't had any issues with IPython 5 locally yet. But I'll certainly keep that in mind in case any issues do pop up.

astrofrog:
@itsagoodbrain - can you rebase again btw? There were some Travis failures due to continuum updating conda to include pyqt5 by default, which I fixed in https://github.com/vispy/vispy/pull/1272

itsagoodbrain:
Rebased again @astrofrog
@drafter250 I understand there's many other updates needed to make this current. But those are handled currently in a manner that still works but gives deprecation warnings. What I've submitted is only meant to address the issue several people had come across which was not handled with warnings. As stated above however, this isn't my fix, but it has worked for multiple people now and I think it should be included before someone takes the time to sort through all the deprecation warnings, etc.

Eric89GXL:
@itsagoodbrain good to go from your end?

itsagoodbrain:
Yep.

Eric89GXL:
Thanks @itsagoodbrain

PR #1271, opened 2016-09-17
MIN: fix several issues with ColorBarWidget/ColorBarVisual

### LINKED ISSUE #1251: Update label for ColorBarWidget (which relies on ColorBarVisual)
Hi,

I'll start with that I really like vispy! I wonder if the following would be worth a pull request. 

Using vispy 0.5.0.dev0, I was trying to update dynamically the colorbar label from ColorBarWidget. 

**(1)** To set the label from the colorbar widget (ColorBarWidget), I need to pass the label_str directly as the 'label' parameter isn't linked to anything. This works well. This means that the call to ColorBarVisual in ColorBarWidget could be changed to  

```
def __init__(self, cmap, orientation,
                 label="", clim=("", ""),
                 border_width=0.0, border_color="black", **kwargs):

        dummy_size = (1, 1)
        self._colorbar = ColorBarVisual(size=dummy_size, cmap=cmap,
                                        orientation=orientation,
                                        **label_str**=label, clim=clim,
                                        border_width=border_width,
                                        border_color=border_color, **kwargs)
```

**(2)** To update it I tried two things. First, I try to simply recreate (self.grid.remove_widget followed by self.grid.add_widget) the whole widget and add it back to the grid. E.g.
```
cbar = scene.ColorBarWidget(orientation=position,
                                    label_str=label,
                                    cmap=cmap,
                                    clim=clim,
                                    border_width=border_width,
                                    border_color=border_color,
                                    **kwargs)

self.grid.remove_widget(self.cbar)
self.cbar_left = self.grid.add_widget(self.cbar, row=0, col=0)
self.cbar_left.width_max = self.cbar_left.width_min = CBAR_LONG_DIM
```

Somehow, it doesn't refresh and I end up with the previous text being drawn underneath the new text -- which makes everything unreadable.

The second thing I tried is keep track of cbar and simply update its parameters. E.g.

```
self.cbar.clim = clim
self.cbar.label = label
self.cbar.cmap = cmap 
```
It works fine for clim and cmap with no change. However, the setter for label directly calls ColorBarVisual @label.setter which should not accept text directly as in __init__, it is set via label_str as follow:

```
self._label_str = label_str
self._label = TextVisual(text=self._label_str)
```

Adding a setter for label_str can be used to then update it via the label setter. So that in ColorBarWidget, the setter for label would be modifying self.label_str. This work fine. E.g. 

   ```
# In ColorBarWidget
@label.setter
def label(self, label):
        self._colorbar.label_str = label

# In ColorBarVisual
@property
def label(self):
        """ The vispy.visuals.TextVisual associated with the label
        """
        return self._label

@label.setter
def label(self, label):
        self._label = label
        self._update()

@property
def label_str(self):
        return self._label_str

@label_str.setter
def label_str(self, label_str):
        self._label_str = label_str
        self._update()
```

--- 
_As a side note_, I also added the following to ColorBarWidget so that I can modify the color and size of the ticks. I guess this could be added to the lot. 
```
@property
def ticks(self):
        return self._colorbar.ticks

@ticks.setter
def ticks(self, color, font_size):
        self._colorbar.ticks[0].font_size = font_size
        self._colorbar.ticks[1].font_size = font_size
        self._colorbar.ticks[0].color = color
        self._colorbar.ticks[1].color = color
```

### PR DESCRIPTION
**Pull request by kmuehlbauer**
closes #1251 

As written in the above issue, there are some glitches using ColorBarWidget/ColorBarVisual. I tried to address all  issues mentioned.

Main thing is that the labels and colors are propagated to the Textvisuals and they are exposed (somehow) to the upper layers for easy change.

### DIFF
--- a/examples/basics/scene/colorbar_widget.py
+++ b/examples/basics/scene/colorbar_widget.py
@@ -16,11 +16,11 @@ canvas.size = 800, 600
 canvas.show()
 grid = canvas.central_widget.add_grid(margin=10)
 
-
-cbar_widget = scene.ColorBarWidget(cmap="cool", orientation="right")
+cbar_widget = scene.ColorBarWidget(label="ColorBarWidget", clim=(0, 99),
+                                   cmap="cool", orientation="right",
+                                   border_width=1)
 grid.add_widget(cbar_widget)
 
-cbar_widget.pos = (300, 100)
 cbar_widget.border_color = "#212121"
 grid.bgcolor = "#ffffff"
 
--- a/vispy/plot/plotwidget.py
+++ b/vispy/plot/plotwidget.py
@@ -465,7 +465,7 @@ class PlotWidget(scene.Widget):
         self._configure_2d()
 
         cbar = scene.ColorBarWidget(orientation=position,
-                                    label_str=label,
+                                    label=label,
                                     cmap=cmap,
                                     clim=clim,
                                     border_width=border_width,
--- a/vispy/plot/plotwidget.py
+++ b/vispy/plot/plotwidget.py
@@ -73,6 +73,8 @@ class PlotWidget(scene.Widget):
         #     |         |                       | xlabel|         |         |
         #  r5 |         +-----------------------+-------+---------+         |
         #     |         |                       | cbar  |         |         |
+        #  r6 |---------+-----------------------+-------+---------+---------|
+        #     |                           padding                           |
         #     +---------+-----------------------+-------+---------+---------+
 
         # padding left
@@ -84,6 +86,11 @@ class PlotWidget(scene.Widget):
         padding_right = self.grid.add_widget(None, row=0, row_span=5, col=6)
         padding_right.width_min = 30
         padding_right.width_max = 60
+
+        # padding right
+        padding_bottom = self.grid.add_widget(None, row=6, col=0, col_span=6)
+        padding_bottom.height_min = 20
+        padding_bottom.height_max = 40
 
         # row 0
         # title - column 4 to 5
--- a/vispy/scene/widgets/colorbar.py
+++ b/vispy/scene/widgets/colorbar.py
@@ -70,7 +70,7 @@ class ColorBarWidget(Widget):
     minor_axis_ratio = 0.05
 
     def __init__(self, cmap, orientation,
-                 label="", clim=("", ""), label_color='black',
+                 label="", label_color='black', clim=("", ""),
                  border_width=0.0, border_color="black", **kwargs):
 
         dummy_size = (1, 1)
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by kmuehlbauer** [comment]
While trying to get the ColorBarWidget to display in a reasonable size, I added `padding` and `axis_ratio` as additional keyword arguments. Also reworked the `calc_size` function a bit to follow this.

**Comment by astrofrog** [comment]
@kmuehlbauer - can you rebase this? I fixed some Travis issues in https://github.com/vispy/vispy/pull/1272

**Comment by kmuehlbauer** [comment]
@astrofrog Hopefully my mediocre git-fu brought success :laughing:

**Comment by kmuehlbauer** [comment]
@astrofrog

I found that for positioning the colorbar always the center of the actual colorbar is referenced. This might work for most of the time, but actually the whole "colorbar" including the tick and label text should be taken into account. I'm a bit lost on how to get the `rect` include also the TextVisuals. Any thoughts?

The `vispy/examples/basics/plotting/colorbar.py` shows the effect (bottom colorbar).

**Comment by kmuehlbauer** [comment]
@astrofrog This might be related to #1268. So it should be no showstopper here.

**Comment by Eric89GXL** [comment]
Good to go from your end @kmuehlbauer?

**Comment by kmuehlbauer** [comment]
@Eric89GXL 

Yes, this is good to go from my end now.

The described problem 

> I found that for positioning the colorbar always the center of the actual colorbar is referenced. This might work for most of the time, but actually the whole "colorbar" including the tick and label text should be taken into account. I'm a bit lost on how to get the rect include also the TextVisuals. Any thoughts?

is still there. To fix `vispy/examples/basics/plotting/colorbar.py` (bottom colorbar) for now, I added a 6th row to `PlotWidget` for padding. This adds a bit more space to the bottom of the figure to show the colorbar with the Labels.

**Comment by Eric89GXL** [comment]
A bit of a hack, but it should work better at least :)

Thanks @kmuehlbauer

**Comment by kmuehlbauer** [comment]
> A bit of a hack, but it should work better at least :)
> 
> Thanks @kmuehlbauer

Yes, quite hacky. I'll open an issue to keep track of the underlying problem.

**Comment by macrocosme** [comment]
@kmuehlbauer I might be missing something here, but I think the commit did not fix all issues from [#1251](https://github.com/vispy/vispy/issues/1251). Say you want to update the colorbar label. I would do something like self.cbar.label = label; which would call the following:

In widgets/colorbar.py:
```
    @label.setter
    def label(self, label):
        self._colorbar.label = label
```
Which links to this in visuals/colorbar.py:
```
@label.setter
    def label(self, label):
        self._label = label
        self._update()
```

Hence, label.setter will change _label into a string, instead of TextVisual(). I would suggest to add the following without changing anything else: 

In widgets/colorbar.py:
```
@property
    def label_str(self):
        return self._colorbar.label_str

    @label_str.setter
    def label_str(self, label_str):
        self._colorbar.label_str = label_str
```

and in visuals/colorbar.py:

```
    @property
    def label_str(self):
        return self._label_str

    @label_str.setter
    def label_str(self, label_str):
        self._label_str = label_str
        self._update()
```

PR #1272, opened 2016-09-18
Fix Travis failures by pinning Qt/PyQt versions

astrofrog:
I think continuum recently added PyQt5 to conda and called it pyqt with version 5, so we need to pin the version to 4.


astrofrog:
@Eric89GXL - just for info, I'll merge this when it passes to make sure other builds don't fail.

astrofrog:
This passed but just had to re-word the commit message, so merging now.

PR #1280, opened 2016-09-27
Travis: make sure PyQt stays pinned to PyQt4

astrofrog:
Currently, one of the builds is failing because it is using PyQt5


Eric89GXL:
Thanks @astrofrog

Issue #1277, opened 2016-09-26
Invalid operation: GL.glDisable(GL.GL_LINE_SMOOTH)


astrofrog:
The following code:

```
from vispy import scene


canvas = scene.SceneCanvas(keys=None, size=(800, 600), show=True)
view = canvas.central_widget.add_view()
scene_transform = scene.STTransform()
view.camera = scene.cameras.TurntableCamera(parent=view.scene,
                                            fov=0., distance=4.0)

ax = scene.visuals.Axis(pos=[[-1.0, 0], [1.0, 0]],
                        tick_direction=(0, -1),
                        parent=view.scene, axis_label='X',
                        anchors=['center', 'middle'])

canvas.render()
```

produces the following error on AppVeyor (in another application):

```
Traceback (most recent call last):
  File "test.py", line 15, in <module>
    canvas.render()
  File "C:\conda\envs\test\lib\site-packages\vispy\scene\canvas.py", line 244, in render
    self._draw_scene(bgcolor=bgcolor)
  File "C:\conda\envs\test\lib\site-packages\vispy\scene\canvas.py", line 253, in _draw_scene
    self.draw_visual(self.scene)
  File "C:\conda\envs\test\lib\site-packages\vispy\scene\canvas.py", line 291, in draw_visual
    node.draw()
  File "C:\conda\envs\test\lib\site-packages\vispy\scene\visuals.py", line 98, in draw
    self._visual_superclass.draw(self)
  File "C:\conda\envs\test\lib\site-packages\vispy\visuals\visual.py", line 588, in draw
    v.draw()
  File "C:\conda\envs\test\lib\site-packages\vispy\visuals\visual.py", line 588, in draw
    v.draw()
  File "C:\conda\envs\test\lib\site-packages\vispy\visuals\visual.py", line 432, in draw
    if self._prepare_draw(view=self) is False:
  File "C:\conda\envs\test\lib\site-packages\vispy\visuals\line\line.py", line 353, in _prepare_draw
    GL.glDisable(GL.GL_LINE_SMOOTH)
  File "C:\conda\envs\test\lib\site-packages\OpenGL\platform\baseplatform.py", line 402, in __call__
    return self( *args, **named )
  File "C:\conda\envs\test\lib\site-packages\OpenGL\error.py", line 232, in glCheckError
    baseOperation = baseOperation,
OpenGL.error.GLError: GLError(
	err = 1282,
	description = b'invalid operation',
	baseOperation = glDisable,
[...]

Issue #1276, opened 2016-09-26
Error on Windows with loading freetype?


astrofrog:
I'm having issues running e.g. the `axes_plot.py` example:

```
PS C:\Users\Administrator\vispy\examples\basics\scene> python .\axes_plot.py > windows.log
WARNING: Traceback (most recent call last):
  File ".\axes_plot.py", line 58, in <module>
    app.run()
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\app\_default_app.py", line 62, in run
    return default_app.run()
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\app\application.py", line 137, in run
    return self._backend._vispy_run()
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\app\backends\_qt.py", line 229, in _vispy_run
    return app.exec_()
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\app\backends\_qt.py", line 431, in event
    out = super(QtBaseCanvasBackend, self).event(ev)
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\app\backends\_qt.py", line 431, in event
    out = super(QtBaseCanvasBackend, self).event(ev)
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\app\backends\_qt.py", line 703, in paintGL
    self._vispy_canvas.events.draw(region=None)
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\util\event.py", line 455, in __call__
    self._invoke_callback(cb, event)
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\util\event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  << caught exception here: >>
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\util\event.py", line 471, in _invoke_callback
    cb(event)
  File "C:\Anaconda3\lib\site-packages\vispy-0.5.0.dev0-py3.5.egg\vispy\scene\canvas.py", line 207, in on_draw
    self._draw_scene()
[...]

PR #1282, opened 2016-09-28
Enable Python 3.5 on AppVeyor

astrofrog:
I think this will fail currently, but this will show that there are issues in Python 3.5. I think that https://github.com/vispy/vispy/issues/1277 and https://github.com/vispy/vispy/issues/1276 are due to this.

### DIFF
--- a/.opengl_info.py
+++ b/.opengl_info.py
@@ -1,0 +1,2 @@
+import vispy
+print(vispy.sys_info())
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ environment:
       MINICONDA_VERSION: "3.5.5"
       PYTHON: "C:\\conda"
       CONDA_DEPENDENCIES: "numpy scipy pyqt=4 matplotlib setuptools nose pytest coverage ipython=4.2.0 ipython-notebook pytest-cov"
+      PIP_DEPENDENCIES: "pytest-faulthandler"
 
   matrix:
       - PYTHON_VERSION: "3.4"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -52,5 +52,5 @@ build: false  # Not a C# project, build stuff at the test step instead.
 
 test_script:
   # Run the project tests (don't do examples because they are too slow)
-  - "python make test nobackend"
-  - "python make test pyqt4"
+  - "python -X faulthandler make test nobackend"
+  - "python -X faulthandler make test pyqt4"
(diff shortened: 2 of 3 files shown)


astrofrog:
@rougier - the Python 3.5 build fails with a segmentation fault, whereas the 3.4 one does not. I made sure to print out the OpenGL version information. Do you have any idea what could be going on here?

astrofrog:
@Eric89GXL @rougier - any ideas?

Eric89GXL:
You could try enabling faulthandler to see what the problem is

astrofrog:
@Eric89GXL - what is the correct way to enable the faulthandler? I tried adding ``-X faulthandler`` to the ``python`` call, but this didn't seem to change anything.

Eric89GXL:
Actually thinking about more it I thought our running scripts actually enabled it in their calls. But in any case, if pytest has an automated way to do it like you just pushed, that would be good

astrofrog:
@Eric89GXL - unfortunately it doesn't look like there is any more output, though by enabling verbose mode we can see that the test that fails is:

```
vispy/scene/cameras/tests/test_perspective.py::test_perspective_render <- c:\projects\vispy\vispy\scene\cameras\tests\test_perspective.py Downloading data from http://github.com/vispy/demo-data/raw/master/orig/crate.npz (76 kB)
[........................................] 100.00000 / downloading   Failed: unit failure (3221225477)
Testing failed (['pyqt4'] failed, 0 skipped) in 28.376 seconds
FAILURE
Command exited with code 1
```

Eric89GXL:
Argh. Do you have a Windows machine you can test?

Eric89GXL:
... but really, the easy fix here is to do a check within that test for being on AppVeyor (look at env vars) and 3.5, and `raise SkipTest`. At some point it's not worth the effort to fix one small failing test, especially since it could easily be a bug with the Mesa lib that does the rendering

astrofrog:
@Eric89GXL - the issue is that I'm also running into segmentation faults for tests in glue in Python 3.5 on Windows, which I think may be related. I agree we can just skip the test here, but I think it would be worth trying to track down the issue.

I do have an EC2 instance I've been using to do some tests. I don't see the segfault there *but* I see a whole other bunch of failures which include errors like:

```
E           OpenGL.error.GLError: GLError(
E               err = 1282,
E               description = b'invalid operation',
E               baseOperation = glPixelStorei,
E               cArguments = (GL_PACK_SWAP_BYTES, 0)
E           )
```

If you would like access to the EC2 instance, I can provide you with the details.

Eric89GXL:
@astrofrog any chance to come back to this? We could even do 3.6 now.

larsoner:
Taken care of in #1353

Issue #1266, opened 2016-09-04
suggestion: bundle with glfw, freetype and OpenSans


yhql:
It is a nice feature that you allow any backend to be selected by the user, however I think it would be nice if at least one was provided so that VisPy runs as-is.
Also, on corporate networks it is not easy to get VisPy to download freetype and OpenSans typefont from the dedicated repository.

One of the developers here (@rossant I think ? :) ) mentioned 'glfw' was the most maintained, and it's also very small (about 50 Ko). Freetype and OpenSans aren't too large either, so I modified VisPy to install those 3 in a separate vispy/misc/ folder and look there first when importing.

If needed I could make a pull request with this. So far I'm just switching between 32 and 64 bits versions of Freetype and glfw, both using VC2013, I would need to switch for linux/OSX as well.

What do you think ? :)

drafter250:
I just ran into this myself as I was trying out one of the basic plotting examples and the plot window kept locking up because it was trying to download stuff on the fly from a secondary demo-data repo through the company firewall. If these libraries are crucial to core functionality then shouldn't they be included in the main project. I also really don't like the idea of an implicit download from GitHub for security purposes.

queenish001:
@djhoese

djhoese:
There has been a lot of work regarding downloads and including stuff in vispy or adding dependencies that include external dependencies inside vispy. VisPy still downloads demo data of examples on-the-fly, but at least those are examples. There may be cases still where vispy downloads ancillary files on the fly. If those are still issues please file a new github issues.

As for backends, there has also been a lot of work in backends since this issue was originally created. I, however, don't support the idea of having a hard requirement on one specific backend. We do depend on the python wrapper of glfw now so it may be easier to have that has a hard requirement, but seeing as most of these backends wrap some amount of C/C++ code there is a non-zero complexity with getting them installed on all supported platforms. One option would be the new experimental tk backend, but even that depends on an external library for OpenGL support in tk. So I think what I'm saying is that I'd rather keep with the current strategy of no hard requirements on a specific backend. Feel free to provide some counter argument and we can continue this discussion but given the age of this I'm going to close the issue.

almarklein:
People who want to "ship glfw with vispy" can install vispy using ;)
```
pip install -U vispy glfw
```

I also think since we use Rougier's freetype wrapper, the dll is included via that package.

That leaves the fonts. And for these I actually agree that packaging them along makes a lot of sense. Relying on a post-install download can be fragile ...

[...]

PR #1284, opened 2016-10-05
windows standalone

yhql:
Implements #1266
 On windows, install vispy with glfw3, freetype253, requiring MSVC2013, and also copy OpenSans-Regular. On other platforms, do nothing.
 This provides 'corporate' installation which does not require internet connection to vispy's github repository to make it usable.


Eric89GXL:
I am -1 on packaging more binaries, as maintaining them can add substantial maintenance burden. Does it work in 64 bit, 32 bit, across OSes, does it require a runtime, etc. are just a subset of issues that can come up. It would probably be better as a separate package, or as a `zip` file saved somewhere that people could optionally download when downloading VisPy

yhql:
Thanks for your comment ! I understand the problem, however I wanted to avoid the connections that VisPy does when Freetype and OpenSans are nowhere to be found on the target machine, and also provide a default backend.
As Vispy is not tied or dependent on a specific-backend, I thought it wouldn't be a bad thing to just use one of a specific version and for a specific OS.
My modifications were intended for win32 and x64 only, as this is the type of environment Vispy's connections will be blocked (corporate network). This would limit the maintaining.

Also, as freetype and OpenSans are downloaded and stored at different places, I could not find an easy way to provide those with the current architecture. Your suggestion of having a separate package means VisPy should be modified in order to look in an optional folder for those files it's missing, before looking anywhere else. Would that be a good modification to bring ?

(also while I'm here : thanks to you and your 'colleagues' for this module :) a lot)

Eric89GXL:
Point taken. This would also potentially take care of the issues in #1327. @stxlvt would bundling these things help?

stxlvt:
@Eric89GXL: yes, I think bundling is an acceptable solution. :-)

davidh-ssec:
@Eric89GXL @yhql I'm just now finding this issue. Is there a reason these libraries need to be bundled rather than requiring them to be installed separately? Don't libraries like matplotlib require they be installed separately? Is this too much to ask from vispy users?

Eric89GXL:
> Is there a reason these libraries need to be bundled rather than requiring them to be installed separately?

Because in general Windows (and OSX?) users who do `pip install vispy` will be very unhappy with the result, and we will get a lot of complaints.

yhql:
>  I'm just now finding this issue. Is there a reason these libraries need to be bundled rather than requiring them to be installed separately? Don't libraries like matplotlib require they be installed separately? Is this too much to ask from vispy users?

Actually the problem is that vispy tries to download libraries like freetype by connecting to some repository and managing the connection by itself. This can be a security issue as @stxlvt reported.
Also this means the connection will be blocked by any corporate firewall, and you'll end up trying to get freetype and opensans on your own. (I admit that bundling a backend is not necessary though)
The libraries _could be_ required to be installed separately, but that's not how vispy works right now. 

(also matplotlib is a python module, this is much simpler to install on windows than freetype for example)

davidh-ssec:
> (also matplotlib is a python module, this is much simpler to install on windows than freetype for example)

@yhql I was referring to matplotlib's installation instructions that talk about installing libpng and freetype yourself. https://matplotlib.org/users/installing.html#building-on-macos

kmuehlbauer:
IMHO this is a tradeoff between the sustainability of the detection/search-code, bundling efforts or to document how those libs are to be installed on a per user basis. I'm also wondering if completely unbundling `freetype` and adding as a dependency would be a more sensible way.

larsoner:
We should probably do whatever `matplotlib` and other standard libraries do, which is probably to let `conda` or other packaging managers / maintainers deal with dependencies. IIRC our `freetype` on Windows is just one I built a while ago, and we hope it continues to work, so this could be taken care of simultaneously with #1327.

kmuehlbauer:
@larsoner Yes, [freetype](https://github.com/conda-forge/freetype-feedstock) is already build on `conda-forge` for the three major OS. It would only need to get @rougier 's python-bindings into a conda-recipe https://github.com/rougier/freetype-py/.

larsoner:
We already bundle those:

https://github.com/vispy/vispy/blob/master/vispy/ext/_bundled/freetype.py

kmuehlbauer:
> We already bundle those:

I know. The question is, is this really necessary? I might lack some understanding, but why was this bundled in the first place?

larsoner:
Strictly speaking, no. At one extreme, you could bundle everything (pretty much this PR). At the other extreme, you could eliminate all of `ext` and make a really long `requirements.txt` and set build requirements.

To me getting rid of the downloaded binary blobs is good for maintainability. But getting rid of the bundled requirements doesn't buy us much in terms of maintainability, at least not enough to offset the increased user install burden.

kmuehlbauer:
> To me getting rid of the downloaded binary blobs is good for maintainability. But getting rid of the bundled requirements doesn't buy us much in terms of maintainability, at least not enough to offset the increased user install burden.

Yes, that was what I meant by **tradeoff** a few comments above. If I get this correctly, we now try to eliminate the `downloading` by either bundling or making a dependency whichever suits us best.

djhoese:
Coming back to this after a couple years: #1664 removed bundled freetype and we now depend on freetype-py, OpenSans has been bundled with vispy now (I don't know when), and glfw is not downloaded automatically and is an external dependency like all other backend dependencies. I still agree with @larsoner that bundling more binaries is a bad idea for maintenance and users should use their own package managers for installing any other dependencies. If you feel that this should still be considered feel free to comment, otherwise I'm closing this.

PR #1292, opened 2016-11-03
Adding setter for colorbar's label_str

macrocosme:
Minor fix related to issue #1251 & #1271

https://github.com/vispy/vispy/pull/1271


Eric89GXL:
This needs rebase

kmuehlbauer:
@macrocosme Can you please rebase this. If you do not have time, I can take over. Just let me know.

vispy/scene/widgets/colorbar.py
```diff
@@ -133,6 +133,15 @@ def label(self, label):
         self._colorbar.label = label
 
     @property
+    def label_str(self):
+        return self._label_str
```
kmuehlbauer: Please add a short docstring with the description.

vispy/visuals/colorbar.py
```diff
@@ -597,6 +597,15 @@ def label(self, label):
         self._update()
 
     @property
+    def label_str(self):
+        return self._label_str
```
kmuehlbauer: Same here, add short docstring with description.

davidh-ssec:
I know this is out of the scope for this PR most likely but as a style guideline we should probably avoid keyword arguments with datatypes in them `label_str`. IMO the visual should have a `self._label_visual` and store the string as `self._label`. Or at least name the keyword argument `label`.

Also I'm a little confused about the the colorbar widget with this property. I think it is supposed to be `self._colorbar.label_str`.

vispy/scene/widgets/colorbar.py
davidh-ssec: Please change this to `self._colorbar.label_str`.

vispy/scene/widgets/colorbar.py
```diff
@@ -133,6 +133,15 @@ def label(self, label):
...
+
+    @label_str.setter
+    def label_str(self, label_str):
+        self._label_str = label_str
```
davidh-ssec: Please change this also to `self._colorbar.label_str`.

kmuehlbauer:
> I know this is out of the scope for this PR most likely but as a style guideline we should probably avoid keyword arguments with datatypes in them label_str.

@davidh-ssec This change was introduced with #979, there was surely some logic behind to do this. @bollu could you chime in, when you find the time?

davidh-ssec:
Yes of course, but that's why I think in the future we need to be careful about what the user sees and what they don't. Having the keyword argument `label` is a lot cleaner than `label_str`. This means `self._label` becomes `self._label_visual`. However, I've been finding more and more that user's need to access these sub-visuals by name. This either means that the interface is not complete enough or we need to make less "private" attributes: `self.label` versus `self._label`

vispy/scene/widgets/colorbar.py
macrocosme: Will do as soon as I can.

macrocosme:
I have added the requested changes.

kmuehlbauer:
@macrocosme  Thanks. Are ready for rebasing this to resolve the conflicts?

vispy/visuals/colorbar.py
```diff
@@ -598,7 +598,7 @@ def label(self, label):
 
     @property
     def label_str(self):
-        return self._label_str
```
kmuehlbauer: @macrocosme This will not work. There is no self._colorbar in vispy/visuals/colorbar.py

I meant to just add some docstring here.

kmuehlbauer: Please revert this change.

vispy/scene/widgets/colorbar.py
```diff
@@ -134,7 +134,7 @@ def label(self, label):
 
     @property
     def label_str(self):
-        return self._label_str
+        return self._colorbar._label_str
 
     @label_str.setter
     def label_str(self, label_str):
```
kmuehlbauer: @macrocosme You also need to do this change in the `setter`.

vispy/visuals/colorbar.py
macrocosme: I'm at a conference. I'll get back to it when I have a bit more time to concentrate! : )
This should be soon.

macrocosme:
Okay. I have now changed both setter and getter in widgets/colorbar.py, and also added docstrings to both files (widgets/colorbar.py and visuals/colorbar.py).

djhoese:
@macrocosme Awesome. Looks like you'll have to rebase or merge the current master branch to resolve some conflicts and get tests to pass.

macrocosme:
@djhoese Done!

vispy/scene/widgets/colorbar.py
```diff
@@ -138,6 +138,22 @@ def label(self, label):
...
+        """Set the colorbar string
+
+        Parameters
+        ----------
+        label_str : string label to be displayed with the colorbar
+        """
+        self._colorbar._label_str = label_str
+        self._update()
+
     def ticks(self):
```
djhoese: You lost the `@property` on this ticks property which is causing the errors on travis.

macrocosme: Oh my... ! It's back in.

macrocosme: Thanks for walking me through the process.

vispy/visuals/colorbar.py
```diff
@@ -604,6 +604,29 @@ def label(self, label):
         self._label = label
         self._update()
```
djhoese: I'm not sure why it took me this wrong to realize this, but why can't all of this be done with an `if` statement for the regular `label` property? I know I've expressed my dislike of `label_str` as a property and I didn't see a way around it before, but why not do:

```
@label.setter
def label(self, label):
    if isinstance(label, str):
        self._label.text = label
    else:
        self._label = label
        self._update()
```

I'm not sure if the `_update` needs to be done in both cases since the TextVisual would call its own update. With this change the `label_str` property could be removed. Additionally, have you actually tested this? When does the `ColorbarVisual._label_str` get used? It doesn't seem to show up in the `_update` function.

Thoughts?

kmuehlbauer: @djhoese :+1:

macrocosme: I was using this code here https://github.com/macrocosme/shwirl/blob/master/shwirl/extern/vispy/scene/widgets/colorbar.py without a fuss. But I should indeed sit down and properly test this. I'm travelling so it may be a few days.

macrocosme: As to merging the two with an if, I like it! 👍

macrocosme:
I finally made the changes, and tested it. Note that I have renamed `label_str` to `label_text` as it was discussed that having str in the variable name isn't ideal.

djhoese:
I'm not sure what happened with this PR and I'm extremely sorry we lost track of it (I think your last comment was right before SciPy 2018 started). I see the tests passed and things are basically as we've described. I know you will hate me but now that the setter can handle a string is there a reason we can't have `label=` in the `__init__` method?

@kmuehlbauer Thoughts?

kmuehlbauer:
@djhoese I'm not sure I understand. You mean instead of `self._label=...` it should be `self.label=...`?

macrocosme:
@djhoese @kmuehlbauer  It's okay. These things happen. Nearly 4 years in, I will admit I thought this project had slowly stopped. I will let these final changes to you two if that is okay.

djhoese:
@kmuehlbauer The keyword argument in master is `label_str`. Is there a reason it couldn't (or shouldn't) be just `label`?

kmuehlbauer:
> The keyword argument in master is `label_str`. Is there a reason it couldn't (or shouldn't) be just `label`?

Yes, sure, users which use matplotlib would expect this. :+1:

kmuehlbauer:
@djhoese This seems impossible to merge master into this branch. Should we extract all needed commits and rebase on current master? The this can run on the recent CI?

djhoese:
It is impossible to merge master because it is @macrocosme 's master branch (instead of a separate "feature" branch). GitHub doesn't give maintainers implicit permission to fork's master branches. We'll need to take the fork's master branch and create our own branch on our own fork, merge vispy master, then make a new PR.

djhoese:
This has been replaced by #2057 and don't worry I kept you as the author of the commit for your work in this PR (had to copy some files around to get the commit history to not be garbage). Thanks @macrocosme!

PR #1304, opened 2017-02-19
Remove duplicate shader code. It is already included in vispy/glsl.

drufat:
Remove duplicate shader code. It is already included in vispy/glsl.


Eric89GXL:
Thanks @drufat

PR #1305, opened 2017-02-23
Cast values to native ints to avoid issues on Windows with Python 2.7

astrofrog:
This fixes an issue that I've seen Windows Python 2.7 users come across. Numpy appears to return long integers when accessing array shapes, which causes issues in places such as this.


Eric89GXL:
LGTM, +1 for merge

PR #1273, opened 2016-09-23
Fix function call for pyopengl2.

keunhong:
glGetIntegerv and glGetFloatv were being called by an un-imported name.
Fix this by correctly calling the function from the corresponding
PyOpenGL package (OpenGL.GL).

PR #1306, opened 2017-03-01
Fix codegen backend selection and fix PyOpenGL glGetParameter.

keunhong:
- Fix codegen so that opengl correctly parses annotations with backend
  set to 'es'.
- Actually fix commit 3dd6018d1cd79c1ca828526d377c5c880f168a0e so that
  it is applied in codegen and not by modifying an auto-generated file.
- Adds syntax "# /---" to stop backend selection in annotation parsing.
- This also properly fixes what was attempted with pull request #1273.

### DIFF
--- a/codegen/headerparser.py
+++ b/codegen/headerparser.py
@@ -348,7 +348,7 @@ class Argument:
 
 if __name__ == '__main__':
     THISDIR = os.path.abspath(os.path.dirname(__file__))
-    print(THISDIR)
+
     # Some tests ...
     gl2 = Parser(os.path.join(THISDIR, 'headers', 'gl2.h'))
     import OpenGL.GL
--- a/codegen/headerparser.py
+++ b/codegen/headerparser.py
@@ -79,7 +79,10 @@ class Parser:
     def _get_lines(self):
         # Easy iterator
         while True:
-            yield self._get_line()
+            try:
+                yield self._get_line()
+            except StopIteration:
+                return
 
     def parse(self):
         """ Parse the header file!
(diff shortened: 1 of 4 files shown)


Eric89GXL:
Can you add some simple tests of this functionality?

djhoese:
This PR is currently stuck due to some changes done in another PR. See https://github.com/vispy/vispy/pull/1569#issuecomment-684268665

Problem is some changes were made to the *output* of the generator but not the codegen itself. We (@kmuehlbauer has ideas) need to do that first before we can fully test and successfully update the generated code.

kmuehlbauer:
@djhoese I've reapplied the changes from #1569 to this branch. This time to the correct place in `codegen/annotations.py`. I've had to fix the headerparser to comply with PEP479 which was targeted for Python 3.5 (which was released almost 5 years ago). Fingers crossed.

codegen/headerparser.py
```diff
@@ -345,7 +348,7 @@ def __init__(self, argAsString, cinput=True):
 
 if __name__ == '__main__':
     THISDIR = os.path.abspath(os.path.dirname(__file__))
-
+    print(THISDIR)
```
djhoese, line 351: Extra print @kmuehlbauer ?

djhoese (approved):
Once that print is gone, this looks good to me. Azure and Appveyor are failing for the same reason which seems to be some upstream package (pytest probably). Travis had a failure that I'm hoping was a hiccup and I've restarted it. @kmuehlbauer feel free to merge if Travis passes.

codegen/headerparser.py
kmuehlbauer: @djhoese Yes, that's where I started to look, I'll remove it...

kmuehlbauer:
@djhoese The one Mac build fails with the mesh_io test. Not sure, if we can merge...

kmuehlbauer:
@keunhong Thanks for the PR! And please accept our apologies for the massive delay of more than three years.

PR #1307, opened 2017-03-01
WIP/RFC: Implement mipmaps

keunhong:
This implements mipmapping by adding a third argument to the
'INTERPOLATION' glir command.

It's not clear what the right way to support mipmapping is since setting
the number of mipmap levels is only supported in >= OpenGL ES 3.0 and
>= OpenGL 3.0. This discrepancy is handled by checking whether
'glTextureStorage2D' is available in the gl package.

Any comments are welcome.

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -982,7 +982,8 @@ class GlirIndexBuffer(GlirBuffer):
 
 class GlirTexture(GlirObject):
     _target = None
-    
+    _mipmap_levels = None
+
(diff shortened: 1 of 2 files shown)


Eric89GXL:
Can you add tests, and preferably some simple example that shows the functionality?

djhoese:
@keunhong Let me know what your availability is for working/updating this PR. It'd be nice to add this feature, but as Eric requested we'd need to get some tests added. Let me know if you have ideas.

PR #1310, opened 2017-03-06
Allow depth in Texture2D.

keunhong:
This is needed to render depth to a framebuffer texture.


Eric89GXL:
LGTM but needs some test

djhoese:
@keunhong It's been a couple years, but any chance you would be able to add some tests for this and merge with master so we can get this merged?

djhoese:
@keunhong Any chance you are still in a place where you could add tests for this?

djhoese:
Ok I think I've added a couple tests for the depth stuff. I think this is good enough for a merge now. If/when the tests all pass then we can assume nothing is getting broken, merge it, and include it in the next release.

PR #1311, opened 2017-03-06
Modify gloo.read_pixels to allow different formats.

keunhong:
Currently gloo.read_pixels only allows RGB and RGBA values to be read.
This commit changes the API to have an format argument instead of an
alpha argument allowing the addition of 'depth'. Other buffers such as
the stencil may be added later.

### DIFF
--- a/vispy/gloo/framebuffer.py
+++ b/vispy/gloo/framebuffer.py
@@ -250,5 +250,10 @@ class FrameBuffer(GLObject):
         
         # todo: this is ostensibly required, but not available in gloo.gl
         #gl.glReadBuffer(buffer._target)
-        
-        return read_pixels((0, 0, w, h), alpha=alpha)
+
+        if mode == 'color':
+            format = 'rgba' if alpha else 'rgb'
+        elif mode == 'depth':
+            format = 'depth'
+
+        return read_pixels((0, 0, w, h), format=format)
--- a/vispy/gloo/wrappers.py
+++ b/vispy/gloo/wrappers.py
@@ -603,8 +603,8 @@ for name in dir(global_gloo_functions):
 ## Functions that do not use the glir queue
 
 
-def read_pixels(viewport=None, alpha=True, out_type='unsigned_byte'):
-    """Read pixels from the currently selected buffer. 
+def read_pixels(viewport=None, format='rgba', out_type='unsigned_byte'):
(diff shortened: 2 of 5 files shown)


vispy/gloo/framebuffer.py
```diff
@@ -250,5 +250,10 @@ def read(self, mode='color', alpha=True):
         
         # todo: this is ostensibly required, but not available in gloo.gl
         #gl.glReadBuffer(buffer._target)
-        
-        return read_pixels((0, 0, w, h), alpha=alpha)
+
+        if mode == 'color':
+            format = 'rgba' if alpha else 'rgb'
+        elif mode == 'depth':
+            format = 'depth'
```
Eric89GXL: don't shadow builtin, use `fmt` or something else

vispy/gloo/wrappers.py
```diff
@@ -615,9 +615,9 @@ def read_pixels(viewport=None, alpha=True, out_type='unsigned_byte'):
     viewport : array-like | None
         4-element list of x, y, w, h parameters. If None (default),
         the current GL viewport will be queried and used.
-    alpha : bool
-        If True (default), the returned array has 4 elements (RGBA).
-        If False, it has 3 (RGB).
+    format : str
+        Can be 'rgba', 'rgb' or 'depth'. The returned array have 4 elements if
+        'rgba', 3 elements if 'rgb', and 1 element if 'depth'.
```
Eric89GXL: should be added as last argument to avoid API break

davidh-ssec:
@keunhong Do you think you'll have time in the next month or two to write tests for some of your PRs so they can be merged?

keunhong:
I'll see if I can find some time on the weekend. Sorry for leaving these unattended.

davidh-ssec:
The whole project was unattended for quite some time so don't feel too bad. Thanks for trying. You'll want to rebase or merge the current master branch to get all the CI fixes.

asnt:
@djhoese This feature was added in #1660, it seems. With the stencil buffer as well.

djhoese:
I don't think #1660 adds the `format` keyword argument as this PR was going to but instead somewhere else an `alpha` keyword argument was added to support RGB versus RGBA return values. #1660 is an improvement on that to not return that many bands for other types of buffers (as far as I can tell from skimming my work there). Good find @asnt. Thanks.

PR #1315, opened 2017-03-26
Correction in oscilloscope demo to handle numpy API change

campagnola:
Fixes an error in the oscilloscope demo that was introduced by a recent numpy API change:

```
TypeError: Cannot cast ufunc multiply output from dtype('float64') to dtype('uint8') with casting rule 'same_kind'
```


Eric89GXL:
Thanks @campagnola

PR #1316, opened 2017-04-01
WIP: Geometry shaders

campagnola:
Adds support for geometry shaders to `gloo`.


campagnola:
Note: changes to `gl/glplus.py` address an issue that is also fixed by #1306. It looks like their changes are a better solution, so we can revert glplus.py here if #1306 is merged.

campagnola:
Update: this PR is basically finished, but it causes some unit test failures that I have not had time to look into. If anyone else wants to pick this up, that'd be great.

davidh-ssec:
@campagnola I'll take a swing at resolving the conflicts and seeing what I can do. I'm going to make a PR to merge with this PR so I can test my merge without losing the easy-to-read history of this PR.

davidh-ssec:
Yeah @campagnola the tests fail...because they weren't updated to match your changes :)

PR #1320, opened 2017-04-19
Remove misplaced flags from re.sub() call

jwilk:
The 4th argument of `re.sub()` is maximum number of substitutions, not flags.
Moreover, `re.M` affects only semantics of `^` and `$`, so it wouldn't have any effect on this regular expression.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


Eric89GXL:
Thanks @jwilk

Issue #99, opened 2014-02-11
framebuffer size vs. window size


rougier:
On OSX, with retina display, the actual framebuffer size can be different from the window size. Some backends are aware of this difference (glfw), some others are not (qt,glut,pyglet).

On solution to cope with this situation is to set the size of GL viewport using the framebuffer size. We can add a framebuffer size for all backend and only glfw will report different size and makes OSX users happy. For qt,pyglet,glut, framebuffer size will return window size.

PR #1323, opened 2017-04-21
Fix for Retina displays

xoolive:
Fix for Retina displays (mentioned in #99)
Also adding a mouse_press callback!


Eric89GXL:
Thanks @xoolive

PR #1329, opened 2017-05-11
MRG: Better mesh options

larsoner:
Allows doing a vertex value->colormap mapping with meshes.

### DIFF
--- a/vispy/gloo/buffer.py
+++ b/vispy/gloo/buffer.py
@@ -37,13 +37,13 @@ class Buffer(GLObject):
     nbytes : int | None
         Buffer byte size.
     """
-    
+
     def __init__(self, data=None, nbytes=None):
         GLObject.__init__(self)
         self._views = []  # Views on this buffer (stored using weakrefs)
         self._valid = True  # To invalidate buffer views
         self._nbytes = 0  # Bytesize in bytes, set in resize_bytes()
-        
+
         # Set data
         if data is not None:
             if nbytes is not None:
@@ -51,7 +51,7 @@ class Buffer(GLObject):
             self.set_data(data, copy=False)
         elif nbytes is not None:
             self.resize_bytes(nbytes)
-        
+
     @property
     def nbytes(self):
         """ Buffer size in bytes """
@@ -109,13 +109,13 @@ class Buffer(GLObject):
         else:
             # Use SIZE to discard any previous data setting
             self._glir.command('SIZE', self._id, nbytes)
-        
+
         if nbytes:  # Only set data if there *is* data
             self._glir.command('DATA', self._id, 0, data)
-    
+
     def resize_bytes(self, size):
-        """ Resize this buffer (deferred operation). 
-        
+        """ Resize this buffer (deferred operation).
+
         Parameters
         ----------
         size : int
@@ -194,10 +194,7 @@ class DataBuffer(Buffer):
         # This works around some strange NumPy bug where a float32 array
         # of shape (155407, 1) was said to have strides
         # (4, 9223372036854775807), which is crazy
-        if data.ndim == 2 and data.shape[1] == 1:
-            self._stride = data.strides[0]
-        else:
-            self._stride = data.strides[-1]
+        self._stride = data.strides[-1]
         self._itemsize = self._dtype.itemsize
         Buffer.set_data(self, data=data, copy=copy)
 
@@ -272,7 +269,7 @@ class DataBuffer(Buffer):
         # storage. Note this case (key is string) only happen with base buffer
         if isinstance(key, string_types):
             raise ValueError("Cannot set non-contiguous data on buffer")
-        
+
         # Setting one or several elements
         elif isinstance(key, int):
             if key < 0:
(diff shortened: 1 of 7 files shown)


vispy/visuals/mesh.py
```diff
@@ -28,12 +29,10 @@
 varying vec4 v_light_color;
 varying vec4 v_base_color;
 
-
 void main() {
-
     v_ambientk = $ambientk;
     v_light_color = $light_color;
-    v_base_color = $base_color;
+    v_base_color = $color_transform($base_color);
```
Eric89GXL: so here before we used to just use the variable `$base_color`, but I needed to add a color transform on top of that to use the colormap mapping

vispy/visuals/mesh.py
```diff
@@ -136,6 +127,21 @@
...
+
+def _build_color_transform(data, cmap, clim=(0., 1.)):
+    if data.ndim == 2 and data.shape[1] == 1:
+        fun = Function(_clim)
+        fun['cmin'] = clim[0]
+        fun['cmax'] = clim[1]
+        fun = FunctionChain(None, [fun, Function(cmap.glsl_map)])
+    else:
+        fun = Function(_null_color_transform)
+    return fun
```
Eric89GXL: These are the bits of Python code necessary to build the `$color_transform` variable above. If it's not a colormap, the `_null_*` is used, which is just a pass-through. If it is a colormap, then we chain the colormap's `glsl_map` (which is GLSL code) after applying some simlpe GLSL code to set upper and lower limits to the data (`clim`).

vispy/visuals/mesh.py
```diff
@@ -357,16 +424,77 @@ def _update_data(self):
...
+        ----------
+        direction : ndarray, shape (3,)
+            The light direction.
+        """
+        direction = np.array(direction, float).ravel()
+        if direction.size != 3 or not np.isfinite(direction).all():
+            raise ValueError('Invalid direction %s' % direction)
+        self._light_dir = tuple(direction)
+        self.mesh_data_changed()
+
```
Eric89GXL: All of these other changes are merely to expose some parameters that were formerly hard-coded.

vispy/visuals/mesh.py
```diff
@@ -321,13 +386,17 @@ def _update_data(self):
...
+                colors = colors.astype(np.float32)
             elif md.has_face_color():
-                self._colors.set_data(md.get_face_colors(indexed='faces'),
-                                      convert=True)
+                colors = md.get_face_colors(indexed='faces')
+                colors = colors.astype(np.float32)
+            elif md.has_vertex_value():
+                colors = md.get_vertex_values(indexed='faces')
+                colors = colors.ravel()[:, np.newaxis]
+                colors = colors.astype(np.float32)
```
Eric89GXL: oh, and of course we need to add a conditional that causes it to actually set `colors` when the vertex-value mode is available

vispy/visuals/mesh.py
```diff
@@ -357,16 +424,77 @@ def _update_data(self):
                 normals = (1., 0., 0.)
 
             self.shared_program.vert['normal'] = normals
-            self.shared_program.vert['base_color'] = colors
+            self.shared_program.vert['color_transform'] = \
+                _build_color_transform(colors, self._cmap, self._clim_values)
+            if colors.ndim == 1:
+                self.shared_program.vert['base_color'] = colors
+            else:
+                self.shared_program.vert['base_color'] = VertexBuffer(colors)
```
Eric89GXL: and a conditional based on whether we're doing the value->color conversion with colormap, or the old/standard way (per-vertex colors actually passed in)

larsoner:
This is ready for review/merge from my end

vispy/gloo/buffer.py
```diff
@@ -191,7 +191,13 @@ def set_data(self, data, copy=False, **kwargs):
...
         data = self._prepare_data(data, **kwargs)
         self._dtype = data.dtype
-        self._stride = data.strides[-1]
+        # This works around some strange NumPy bug where a float32 array
+        # of shape (155407, 1) was said to have strides
+        # (4, 9223372036854775807), which is crazy
+        if data.ndim == 2 and data.shape[1] == 1:
+            self._stride = data.strides[0]
+        else:
+            self._stride = data.strides[-1]
```
davidh-ssec: Any idea if this bug still exists?

larsoner: Perhaps it was a bleeding-edge NumPy bug. I don't see it anymore so I've removed it

vispy/visuals/mesh.py
```diff
@@ -136,6 +131,21 @@
 """)
 
 
+_null_color_transform = 'vec4 pass(vec4 color) { return color; }'
+_clim = 'float cmap(float val) { return (val - $cmin) / ($cmax - $cmin); }'
+
+
+def _build_color_transform(data, cmap, clim=(0., 1.)):
```
davidh-ssec: Is it worth it to try to combine this `_build_color_transform` with the one in `image.py`? I know in one of my own projects that uses vispy I end up using clims with an ImageVisual subclass.

larsoner: It looks like `visuals/image.py` (line ~92) uses a transformation of `(color.r + color.g + color.b / 3.)` whereas this just takes and converts a `float`. So it's not quite so straightforward unfortunately. I'll add a comment that eventually they could be unified, though.

Issue #1331, opened 2017-05-19
viridis colormap.py fails with stride 2 but works with stride changed to 4


pocdn:
Problem: colorbar crashes when using viridis colormap. The problem seems to be
the size of the _viridis_data passed to Colormap(ColorArray(...)).

To reproduce the crash, the example colorbar.py file with a slight modification as follows:

replace "hsl" with "viridis" in 
https://github.com/vispy/vispy/blob/master/examples/basics/plotting/colorbar.py

line 36:
plot.image(img, cmap="viridis")

line 45: 
cmap="viridis"

Then run test.py (=colorbar.py with 'viridis' change) ... complains (see error message at appendix below) and does NOT show any color bars.

But if I manually change the system wide vispy installed file:
https://github.com/vispy/vispy/blob/master/vispy/color/colormap.py
by changing the stride from [::2] to [::**4**] (reducing sampled _viridis_data

 passed length to 64 samples) on

Line 956:
viridis=Colormap(ColorArray(_viridis_data[::**4**])),

The original colorbar.py test file works now works with (4 instead of 2 for stride).

Is there a better fix? Or can this simple one character change be merged into the colormap.py?

For those interested the complete error message is

> $ python test.py
> WARNING: Error drawing visual <Image at 0x7f0387ce9510>
> WARNING: Traceback (most recent call last):
>   File "test.py", line 49, in <module>
>     fig.show(run=True)
>   File "/scratch/tmp/vispy/vispy/app/canvas.py", line 431, in show
>     self.app.run()
>   File "/scratch/tmp/vispy/vispy/app/application.py", line 137, in run
>     return self._backend._vispy_run()
>   File "/scratch/tmp/vispy/vispy/app/backends/_qt.py", line 229, in _vispy_run
>     return app.exec_()
>   File "/scratch/tmp/vispy/vispy/app/backends/_qt.py", line 431, in event
>     out = super(QtBaseCanvasBackend, self).event(ev)
>   File "/scratch/tmp/vispy/vispy/app/backends/_qt.py", line 431, in event
>     out = super(QtBaseCanvasBackend, self).event(ev)
>   File "/scratch/tmp/vispy/vispy/app/backends/_qt.py", line 703, in paintGL
>     self._vispy_canvas.events.draw(region=None)
>   File "/scratch/tmp/vispy/vispy/util/event.py", line 455, in __call__
>     self._invoke_callback(cb, event)
[...]

PR #1332, opened 2017-05-19
viridis colormap change from 128 to 64 samples

pocdn:
Changed number of samples from 128 to 64 in creation of the viridis color map (changed sample stride from 2 to 4). Note the original viridis colormap is defined by 256 samples.

This allows a good (barely visible artifacts) interpolated approximation to the original viridis colormap with at most <1% error as elaborated in #1331. Without this change less GPUs are able to use the viridis colormap along with the ColorBar. This change expands the number of GPUs that are compatible.

### DIFF
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -677,6 +677,12 @@ class _Diverging(Colormap):
 # Taken from original Viridis colormap data in matplotlib implementation
 # Sampled 128 points from the raw data-set of 256 samples.
 # Sub sampled to 128 points since 256 points causes VisPy to freeze.
+#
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
@pocdn Can you add a comment in the sources pointing to the github-issue #1331. Just to keep track of why the changes happened.

pocdn:
@kmuehlbauer Yes, I added a comment pointing to the github-issue #1331 in the source file.

Eric89GXL:
@davidh-ssec @kmuehlbauer I'll let you pull the trigger, LGTM :)

pocdn:
Great, thanks for incorporating this change to the viridis colormap sample set :)

kmuehlbauer:
@pocdn Thanks for your contribution!

Issue #862, opened 2015-04-04
Shared context example is broken


campagnola:
The example in `basics/scene/shared_context.py` no longer works correctly for me as of commit 592c973  (ping @almarklein). The mesh in the 3D view appears briefly, then disappears.

davidh-ssec:
@kmuehlbauer It is probably important to note that #1333 was attempting to fix a different shared_context example `examples/tutorial/app/shared_context.py`. Either way it is probably best for everyone to determine what kind of functionality a shared context should provide in vispy.

PR #1333, opened 2017-05-19
Fix for shared_context.py  in /examples/tutorial/app/

davidruffner:
The Text object from vispy.scene.visuals did not accept a list of parents. I'm not sure if that is the whole point of this file. But I made it work by adding two Text objects each with one parent scene.

Being able to integrate Vispy into a PyQt or PySide app will be very useful.

### DIFF
--- a/examples/tutorial/app/shared_context.py
+++ b/examples/tutorial/app/shared_context.py
@@ -9,8 +9,6 @@ This is a very simple example that demonstrates using a shared context
 between two Qt widgets.
 """
 
-# XXX THIS IS CURRENTLY BROKEN
-
 from PyQt4 import QtGui, QtCore  # can also use pyside
 from functools import partial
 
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
@davidruffner It's not really clear to me what you want to achieve. Do you want to integrate a vispy (scene, canvas, plot) in a PyQt app? There are several examples showing this, like https://github.com/vispy/vispy/blob/master/examples/basics/scene/isocurve_for_trisurface_qt.py. 

Or are you interested in using **shared_context**? AFAIR, the example you are referring is currently broken, but I do not remember the details. It is most likely connected to multiparenting. @Eric89GXL do you have the details of shared context and  multiparenting at hand?

davidh-ssec:
I definitely want to here if Eric or other original maintainers know what happened with this example. I'm not sure I agree with this solution for fixing it since I think the original intention was to show how a Visual could be shared between two canvases with a shared context...I think. The proposed changes don't seem to take advantage of the shared context at all.

davidruffner:
@kmuehlbauer @davidh-ssec Thanks for following up. Yes, I'm interested in integrating vispy elements inside a PyQt app. Thanks for the example that you shared. That's a great starting point.

I think you are right that my proposed changes don't address the problem with sharing a Visual between two canvases with a shared context. I'm not sure yet how to make that work. Should I close this pull request?

kmuehlbauer:
@davidruffner Glad the example is useful for your application. If more questions arise, please open another issue. Let's wait for @davidh-ssec to close this one here.

@davidh-ssec For discussing shared context we might use #862.

Issue #1312, opened 2017-03-18
vispy.plot not working in Jupyter notebook


codypiersall:
I cannot get `vispy.plot` to work in a Jupyter notebook.  I'm using the same code as in issue #875:

```
import vispy.app
vispy.app.use_app('ipynb_webgl')
from vispy import plot as vp
fig = vp.Fig(show=False)
fig1 = fig[0, 0]
fig1.plot(range(10000),marker_size=0)
fig.show(run=True)
```

When I run this in a Jupyter notebook, I get a black canvas for a second, then nothing.  The Javascript console contains the following messages:

```
Widget backend and frontend versions are compatible
:8888/nbextensions/vispy/vispy.min.js?v=20170317210211:1 Event loop started.
:8888/nbextensions/vispy/vispy.min.js?v=20170317210211:1 Event loop stopped.

:8888/nbextensions/vispy/vispy.min.js?v=20170317210211:1 Uncaught TypeError: this.glir[n] is not a function
    at n.d.command (:8888/nbextensions/vispy/vispy.min.js?v=20170317210211:1)
    at n.d.execute_pending_commands (:8888/nbextensions/vispy/vispy.min.js?v=20170317210211:1)
    at n.g.event_tick (:8888/nbextensions/vispy/vispy.min.js?v=20170317210211:1)
    at t (:8888/nbextensions/vispy/vispy.min.js?v=20170317210211:1)
```

I'm not getting any Python errors, although I'm getting two issues related to traitlets moving to a toplevel package.  

Python version:

```
(vispy) C:\Users\cody\Dev\vispy>python
Python 3.5.2 (v3.5.2:4def2a2901a5, Jun 25 2016, 22:01:18) [MSC v.1900 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
```
Jupyter version:

```
(vispy) C:\Users\cody\Dev\vispy>jupyter notebook --version
4.4.1
```

Vispy is up to date with the GitHub repository.   I'm running this on a Windows 10 box.  Any ideas?

codypiersall:
Do you guys think the issue is on the Python side or the JS side?

davidh-ssec:
@codypiersall What browsers are you trying this on? Has this ever worked for you? Have you tried older versions of ipython notebook (4.3, 4.2, 4.1)?

kmuehlbauer:
@codypiersall @davidh-ssec I can reproduce this. 

[...]

PR #1338, opened 2017-06-27
Fix vispy in jupyter notebooks

davidh-ssec:
See #1312 for details. Also see the related vispy.js PR vispy/vispy.js#18.

Work In Progress. Waiting for vispy.js PR to work completely. See that PR for issues I'm still having.


kmuehlbauer:
@david-ssec Thanks for taking the lead here. I'll also test here on my systems (mostly linux) and report back.

One thing to mention. The `spatial-filter.frag` is normally created with the use of [build-spatial-filters.py](https://github.com/vispy/vispy/blob/master/vispy/glsl/build-spatial-filters.py) where at the same time the corresponding filter textures [spatial-filtes.npy](https://github.com/vispy/vispy/tree/master/vispy/io/_data) are created. To have these changes made permanent, they need to be incorporated into `build-spatial-filters.py`.

davidh-ssec:
@kmuehlbauer Thank you. You'd think I'd notice the "do not edit" at the top. I've updated it, but when running it with python 3.6 I get the following warnings when generating the npy file:

```
/Users/davidh/repos/git/vispy/vispy/glsl/build-spatial-filters.py:558: RuntimeWarning: overflow encountered in double_scalars
  c4 = c4 + 2 * c6
/Users/davidh/repos/git/vispy/vispy/glsl/build-spatial-filters.py:551: RuntimeWarning: overflow encountered in double_scalars
  c6 = 2 * (m2 - i) * c2 / x - c3
/Users/davidh/repos/git/vispy/vispy/glsl/build-spatial-filters.py:560: RuntimeWarning: invalid value encountered in double_scalars
  c6 = 2 * c2 / x - c3
/Users/davidh/repos/git/vispy/vispy/glsl/build-spatial-filters.py:551: RuntimeWarning: invalid value encountered in double_scalars
  c6 = 2 * (m2 - i) * c2 / x - c3
```

Any idea if these are ok? The only change I made to the kernel generation classes was forcing a `samples` value to an integer because it would be in python 2.

Eric89GXL:
Probably caused an integer overflow, adding float(something) so you work
with floats might fix it.

kmuehlbauer:
Need to run this locally. Can't be of much help atm.

davidh-ssec:
I think I'm on my way to fixing it. Looks like there are a couple spots where integer division in python 2 was relied on.

davidh-ssec:
Looks like my `#extension` that I added to the SDF shader doesn't compile on some systems (see travis failures). I'll have to come up with something better.

kmuehlbauer:
@davidh-ssec Bummer! This works somehow on my linux system (after increasing the iopub_data_rate_limit) in a recent `conda-forge` environment.

It's a bit glitchy, sometimes the canvas get's black, but then the text is displayed. And no messages on the browser console.

vispy_test_text_webgl

kmuehlbauer:
@davidh-ssec Great, the tests are passing!

davidh-ssec:
@kmuehlbauer Yep. The issue seems to be that some shader compilers are more strict about where preprocessors are placed. The OpenGL spec. seems to want this too, but my laptop doesn't seem to care.

Based on your description, I think I am having the same experience. When you say the screen goes black do you mean it stays black until you click again? Or does it go black and then eventually show the text?

kmuehlbauer:
@davidh-ssec Mysteriously I can't get it working at my linux-box at home. 

I think that sometimes the canvas remained black until I clicked again, but I'm not really sure. I'll check again tomorrow and report back more clear. Also, if you want me to test also other things, please let me know.

kmuehlbauer:
OK, here we go:

After running the notebook with the code in one cell the canvas is created and after 1-2 seconds, the `48.0 pt` appears.

Then, if I once click the left button, the font_size is increased to 50.4 pt, but the canvas remaines black. If I change the canvas size in the notebook via mouse grabbing the lower right corner triangle, the canvas turns instantly white and the `50.4 pt` is shown.

By issuing mouse clicks in rapid fashion the canvas is also shown correct in half or the cases (rough estimate). Changing the canvas size gives correct result every time. It seems, that there is some canvas update missing at the right place or time. Hope you can figure it out.

davidh-ssec:
@kmuehlbauer Thanks. I get the same thing. I ended up making a key press event to rerun update and that seems to "fix" it.

```
        if event.key == 'u':
            self.update()
```

Hopefully this will make it easier to debug. I'll try to get back to this the end of next week (US holiday and actual paid work). I'll figure it out sometime.

kmuehlbauer:
@davidh-ssec OK, so since this works, does that mean your fixes/additions here and in vispy.js solved the webgl-issue?

davidh-ssec:
@kmuehlbauer No. In general the issue is "not all vispy features work in jupyter notebooks". My changes so far seem to fix the issue of supporting all of the GLIR specification, but obviously something is not working as intended. The end goal (for this PR at least) is for an example vispy script (gloo, scene, plotting API) to work on the desktop and in jupyter notebook with only the obvious modifications (declaring the backend, not needing to 'run' the app, etc).

kmuehlbauer:
@davidh-ssec Thanks for clarifying.

davidh-ssec:
Most recent commit fixes the text example, but there are still some issues with the plotting example (invalid formats for render buffer internal formats, etc). The above commit does something necessary that was never handled before...at least I don't know of another way to do it.

As described in the complete commit message, WebGL automatically swaps drawing buffers between animation requests when there has been a draw command like clear or draw. There is no way with just GLIR to know when this should happen. I had previously been stopping at every DRAW command, however this doesn't work if there is more than one shader program being rendered or multiple DRAW/Clear commands are needed.

On desktop OpenGL vispy calls a `_vispy_swap_buffers` method on the backend outside of the typical GLIR command queue. The desktop backends run their required method for swapping methods. As seen in my commits my solution for the notebook webgl backend is to send a new/separate "SWAP" command that is handled in the `vispy.js` as a signal to execute previous commands and stop on the SWAP command. @Eric89GXL Is this something that should be considered for GLIR? IMO the GLIR isn't really complete as a server/client way to send GL commands with out a way to indicate when to swap buffers.

kmuehlbauer:
@davidh-ssec After your last changes the mouse interaction is smooth, no black canvas any more. 

There was also a test timeout on travis, A restart did the trick.

davidh-ssec:
@kmuehlbauer Good. I am able to get the basic plotting example to "show up" but it takes a good 15-20 seconds for the plot to show up. Plus all the mouse interactions basically break the plot and you start getting JS errors about it trying to access GLObjects that don't exist anymore. I was able to get in to a weird state however where now I can't get it to show up. If I click on the canvas area I get:
screen shot 2017-07-04 at 8 39 39 am

I haven't done any performance testing to figure out what makes it so slow, but the obvious guess is text rendering.

kmuehlbauer:
@davidh-ssec Looks and feels familiar...
screenshot_20170704_155508

I also got the axis displayed correctly only once, now everytime I restart I get a white canvas. After clicking the weird one.

davidh-ssec:
@kmuehlbauer Cleared my browser cache and it loaded again via Ctrl+Shift+R (Cmd+Shift+R on mac).

kmuehlbauer:
@davidh-ssec That easy, didn't thought about it. Thanks.

davidh-ssec:
@Eric89GXL @kmuehlbauer So I spent some time yesterday investigating why mouse events were so weird with notebooks. I created a small example to try to recreate the issue shown above with the plot api and the corrupt-looking buffer. I don't think it ended up doing that, but I did discover that `visual_at` returns different values in notebook from what it does on a desktop. This is obvious when you look at the code since on notebook it uses `_visual_bounds_at(pos)` since it doesn't have access to `read_pixels`. The render picking used on a desktop returns the ViewBox in my below example, the bounds picking returns None because a ViewBox doesn't have bounds apparently. This results in pan/zoom not doing anything without the modifications to the canvas I included:

```
from vispy import app
app.use_app('ipynb_webgl')
from vispy.scene import SceneCanvas, visuals
import numpy as np

# vertex positions of data to draw
N = 10
pos = np.zeros((N, 2), dtype=np.float32)
pos[:, 0] = np.linspace(10, 390, N)
pos[:, 1] = np.sin(np.linspace(-np.pi, np.pi, N)) * 70.

class Canvas(SceneCanvas):
    def visual_at(self, pos):
        vis = super(Canvas, self).visual_at(pos)
        if vis is None:
            if len(self.central_widget._widgets) > 0:
                return self.central_widget._widgets[0]
        
c = Canvas(title='Plot-like Line', keys='interactive')
view = c.central_widget.add_view(camera='panzoom')
c.show()
text = visuals.Line(pos, color='red', parent=view.scene)
view.camera.set_range()
```

However, unless you comment out the super call the response time is pretty bad and laggy. I can put a little more time in to trying to debug the weird black spot on the previous comments, but I'm starting to think we mark this as "good enough" and just put a couple notes in the documentation. Maybe something along the lines of "Support for Jupyter Notebooks (beta)" until we can figure out exactly what is going on. Or maybe even "Limited support" given that sending GLIR commands as fast as we need and processing them does not provide the greatest performance on the browser...especially if text is involved.

davidh-ssec:
I'm still investigating the black blob that happens during pan/zoom of a PlotWidget. During some debugging I discovered some possible reasons for performance issues besides the obvious (sending GLIR over a socket). If you remove all the text visuals (axis labels, title, tick labels, etc) you go from ~250 GLObjects being created to ~170. If you then change the `Widget` base class to not have the 2 MeshVisuals that it uses, you go to ~40-72 depending on how many text visuals you include. This is because almost every high-level area of the Plot is turned in to a Widget object. So that's the central widget of the canvas, the plot widget, each axis (2), each text label (3), and *each padding space* (~7). Each of those Widget objects creates 2 MeshVisuals which each create 4 Buffer objects. Just a lot of objects being created that aren't *technically* needed, but optimizing the code to not use these would be a pain so I'm just going to keep going ahead but leave this note here. Just another thing that isn't an issue on desktop, but not good on webgl/remote-GLIR.

So far, I don't see any black blob when I comment out the text objects. Hopefully that means I'm on the right track.

davidh-ssec:
@larsoner @kmuehlbauer What are your thoughts on merging this PR after I add information in the documentation about notebook support being in "beta" or something similar to that?

I keep thinking "maybe if I put a little more time into it I'll figure it out", but at some point it's just wasting time. I've been having trouble reproducing certain issues consistently. I'm sure there is something wrong with `vispy.js`. Even if this PR doesn't make notebook functionality work perfectly, it is still a useful PR that makes a large chunk of vispy compatible with notebook.

kmuehlbauer:
@davidh-ssec IMHO this would be the best possible way to go. There also a good chance, that others might pick up your work.

larsoner:
Update the docs to say it's not fully functional then +1 from me

davidh-ssec:
@larsoner For modifying documentation I need to make a PR on `vispy/vispy-website`, right?

From what I can tell there are only a few references to "notebook":

1. [Installation Instructions](http://vispy.org/installation.html?highlight=notebook)
2. [use_app](http://vispy.org/app.html?highlight=notebook)
3. List of [backends](http://vispy.org/vispy.html?highlight=notebook).

Do you think I should add any additional information in the documentation for "known issues" or any other information on this backend? Thoughts on renaming the backend from `ipynb_webgl` to `jpynb_webgl`? Perhaps keeping both for backwards compatibility (even for a not fully functional feature?).

Otherwise, I was thinking of just modifying the installation instructions to mention that the notebook backend is not fully functional.

larsoner:
I think renaming it is fine, it was experimental anyway. However I haven't really seen them called `jpynb` as much as `ipynb` (still), so it might not be worth it. Yes the docs are kept over there. And adding more saying it's experiment would be helpful so people don't get their hopes up too much.

davidh-ssec:
> However I haven't really seen them called jpynb as much as ipynb (still)

Is there a better acronym than "jpynb" for jupyter notebook? In my experience jupyter is definitely being used now...even if it is as a correct after saying "ipython".

davidh-ssec:
Yeah but there seems to be "ipynb_webgl", "ipynb_static", and "ipynb_vnc".

larsoner:
I don't remember which each of those are :(

Maybe `jupyter_*` then. Or if there is a "best" one (`ipynb_webgl`?) that could be `jupyter` and the others could be `jupyter_*`

davidh-ssec:
@larsoner Any idea why the one test is failing. It says:

```
0.49s$ python -c "import vispy; print(vispy.sys_info())"
Platform: Linux-4.11.6-041106-generic-x86_64-with-debian-jessie-sid
Python:   3.6.2 |Continuum Analytics, Inc.| (default, Jul 20 2017, 13:51:32)  [GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]
NumPy:    1.13.1
App info-gathering error:
  File "/home/travis/build/vispy/vispy/vispy/util/config.py", line 422, in sys_info
    app = use_app(call_reuse=False)  # suppress messages
  File "/home/travis/build/vispy/vispy/vispy/app/_default_app.py", line 47, in use_app
    default_app = Application(backend_name)
  File "/home/travis/build/vispy/vispy/vispy/app/application.py", line 49, in __init__
    self._use(backend_name)
  File "/home/travis/build/vispy/vispy/vispy/app/application.py", line 244, in _use
    'PyQt' % [b[0] for b in CORE_BACKENDS])
RuntimeError: Could not import any of the backends. You need to install any of ['PyQt4', 'PyQt5', 'PySide', 'Pyglet', 'Glfw', 'SDL2', 'wx', 'EGL', 'osmesa']. We recommend PyQt
```

Can you give me permissions to travis so I can restart the test?

larsoner:
> Can you give me permissions to travis so I can restart the test?

Done. But this is not why your build failed (after this line it says `The command "python -c "import vispy; print(vispy.sys_info())"" exited with 0.`), your failures are Flake8 in nature

https://travis-ci.org/vispy/vispy/jobs/279525594#L2413

davidh-ssec:
Wow I'm smart. Thanks.

larsoner:
Nothing like a little Travis-Flake8 shaming to round out a productive Monday :)

davidh-ssec:
@larsoner This is ready to merge. Can you do one more quick look over and merge it?

larsoner:
Looks good, it's nice to make some progress even if it's incomplete

PR #1341, opened 2017-07-07
Fix integer division for Python3

aepsil0n:
The GLSL build script fails to build on Python3. This fixes the division error.

Edit: to clarify, by division error, I do not mean that the division fails, but that the later call to `np.zeros` fails with a TypeError.

### DIFF
--- a/vispy/glsl/build-spatial-filters.py
+++ b/vispy/glsl/build-spatial-filters.py
@@ -97,7 +97,7 @@ class SpatialFilter(object):
     def kernel(self, size=4*512):
         radius = self.radius
         r = int(max(1.0, math.ceil(radius)))
-        samples = size / r
+        samples = size // r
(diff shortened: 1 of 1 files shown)


davidh-ssec:
Thanks @aepsil0n. I'm going to mark this as a duplicate of my changes in PR #1338, specifically ec84dfb4b01e52b6946374db774693519e26332f. My changes include modifications that also make spatial filters ES2 compatible.

I'm also going to justify my use of `int(size / r)` versus your use of `size // r` because if `size` is ever a float then the result of `//` would be a float which fails when used in slicing.

aepsil0n:
Ah, I didn't realize this was already in progress. And you are absolutely right about `int` vs `//`. I was going to report the infinite loop as well, but now that's taken care of, great :-)

PR #1360, opened 2017-09-24
wx: delete the deprecated call EventLoop

### LINKED ISSUE #1359: wxPython 4 deprecated call
```
Python 3.5.3 (default, Sep  7 2017, 16:23:57) 
[GCC 6.3.0 20170406] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import vispy
>>> vispy.use('wx')
>>> print(vispy.sys_info())
Platform: Linux-4.10.0-35-generic-x86_64-with-Ubuntu-17.04-zesty
Python:   3.5.3 (default, Sep  7 2017, 16:23:57)  [GCC 6.3.0 20170406]
NumPy:    1.13.1
Backend:  wx
pyqt4:    None
pyqt5:    None
pyside:   ('PySide', '1.2.2', '4.8.7')
pyglet:   None
glfw:     None
sdl2:     sdl2 0.9.3
wx:       wxPython 4.0.0b2
egl:      None
osmesa:   OSMesa
_test:    None
```
Warning:
```
/home/eldar/src/vispy/vispy/app/backends/_wx.py:131: wxPyDeprecationWarning: Using deprecated class EventLoop. Use GUIEventLoop instead.
  self._event_loop = wx.EventLoop()
```

eldarkg:
Fixes #1359

### DIFF
--- a/vispy/app/backends/_wx.py
+++ b/vispy/app/backends/_wx.py
@@ -128,7 +128,7 @@ class ApplicationBackend(BaseApplicationBackend):
 
     def __init__(self):
         BaseApplicationBackend.__init__(self)
-        self._event_loop = wx.EventLoop()
+        self._event_loop = wx.GUIEventLoop()
         wx.EventLoop.SetActive(self._event_loop)
 
     def _vispy_get_backend_name(self):


davidh-ssec:
What versions of wxpython does this start with? Is there a reason people wouldn't be using the older versions? Perhaps change this to a try/except with the newer version in the try and the older version in the except (include comments explaining).

eldarkg:
@davidh-ssec [wxPython 4.0.0b2](https://pypi.python.org/pypi/wxPython/4.0.0b2). See [this](http://www.wxwidgets.org/news/2017/04/new-major-wxpython-release/)

eldarkg:
@davidh-ssec 
```sh
grep  -Rn 'class EventLoop' `find -name *.py`
```
```Python
class EventLoop(GUIEventLoop):
    """Class using the old name for compatibility."""
    pass
```

davidh-ssec:
So from what I get from this the project maintainers for wxPython are essentially dropping support for the "Classic" (<4.0) version of the library. There is a new website for the Phoenix project, they have packages on PyPI now, and they will stop hosting on sourceforge it looks like. So based on that it sounds like we should also stop supporting "Classic" wxPython.

However, the sourceforge page still shows thousands of downloads a week for the Classic package: https://sourceforge.net/projects/wxpython/files/wxPython/

So given that the issue this is solving is just a warning for phoenix versions of wxPython, why don't we mark this as something for vispy 0.6.0 (the release after next). The wxPython documentation makes it sound like they will try to keep these warnings around for a couple releases. This should give people a good amount of time to update to the new wxPython 4.0 package and still have a working modern copy of vispy to use. We can push this fix to a later vispy release if needed.

larsoner:
A `try/except` would be forward- and backward-compatible and seems easy enough

eldarkg:
@davidh-ssec, @larsoner wxPython Classic support this GUIEventLoop and mark EventLoop as old

davidh-ssec:
@larsoner I was just going to say that, but was trying to get a feel for the state of the wxPython project.

eldarkg:
@davidh-ssec 
https://github.com/wxWidgets/wxPython/blob/8960e7ee0b6d11c5ff02f5e89665f0bdf50b7c62/src/_evtloop.i#L123

davidh-ssec:
@eldarkg Ok so GUIEventLoop should work for new (Phoenix) and old (Classic) versions of the library? Sounds good to me then.

eldarkg:
@davidh-ssec yes

davidh-ssec:
Looks like this was added back in 2007: https://github.com/wxWidgets/wxPython/commit/bb5e9f35af25e3818f274ab0c22b446ff702e870#diff-30364a95b336de975f85042f7a56f190

Ok looks good to me.

PR #1363, opened 2017-09-26
Fix spatial_filters example on mac

davidh-ssec:
Mac OSX complains about the program having unset variables.

```
Example basics/gloo/spatial_filters.py failed (0):
INFO: Program has unset variables: {'u_shape'}
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -115,11 +115,16 @@ install:
     - if [ "${DEPS}" == "full" ]; then
         conda install --yes pyopengl scipy numpy$NUMPY networkx;
         pip install -q numpydoc PySDL2;
-        if [ "${PYTHON}" == "3.6" ]; then
-          pip install -q freetype-py husl pypng cassowary pillow decorator six;
-          rm -rf ${SRC_DIR}/vispy/ext/_bundled;
+        if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
+          if [ "${PYTHON}" == "3.6" ]; then
+            pip install -q freetype-py husl pypng cassowary pillow decorator six;
+            rm -rf ${SRC_DIR}/vispy/ext/_bundled;
+          else
+            conda install --yes --quiet wxpython matplotlib jupyter pyqt=4;
+            pip install -q mock;
+          fi;
         else
-          conda install --yes --quiet wxpython matplotlib jupyter pyqt=4;
+          conda install --yes --quiet matplotlib jupyter pyqt=4;
           pip install -q mock;
         fi;
       fi;
@@ -151,7 +156,7 @@ before_script:
     # For OSX: https://github.com/travis-ci/travis-ci/issues/7313#issuecomment-279914149
     - if [ "${TEST}" != "osmesa" ]; then
         export DISPLAY=:99.0;
-        if [ "${TRAVIS_OS_NAME}" = "osx" ]; then ( sudo Xvfb :99 -ac -screen 0 1400x900x24 +extension GLX +render +iglx; echo ok )& fi;
+        if [ "${TRAVIS_OS_NAME}" = "osx" ]; then ( sudo Xvfb :99 -ac -screen 0 1400x900x24 +render +iglx; echo ok )& fi;
         if [ "${TRAVIS_OS_NAME}" = "linux" ]; then
           /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1400x900x24 -ac +extension GLX +render;
         fi;
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ before_install:
     - chmod +x miniconda.sh
     - ./miniconda.sh -b -p ~/anaconda
     - export PATH=~/anaconda/bin:$PATH
+    - conda config --add channels conda-forge
     - conda update --yes --quiet conda
 
     - SRC_DIR=$(pwd)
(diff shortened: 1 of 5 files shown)


larsoner:
@davidh-ssec do you also want to add an `OSX`-3.6 build for `.travis.yml` under:
```
matrix:
    allow_failures:
        ...
```
? It would allow us to see this one at least change, and also help the existing OSX issues.

davidh-ssec:
@larsoner I'm sure more work needs to be done to get the OSX environment working on travis, but we'll see what goes wrong with this commit. I tried to lint the yaml file with their tools but it complains about stuff that is apparently already working so...we'll see. I'll work on this more tomorrow hopefully.

larsoner:
This looks correct:

https://travis-ci.org/vispy/vispy/builds/280118923

Will merge if the CIs come back happy (they should)

larsoner:
... actually they shouldn't, now we have a different warning (linux complains about the uniform not being active):

https://travis-ci.org/vispy/vispy/jobs/280096881#L2810

One hacky option is to instead catch the warning in the example file so that it passes. Come to think of it, we should probably add an allowed-failures OSX example-runner, too.

davidh-ssec:
@larsoner Fun stuff. I tested my branch on an Ubuntu VM and get the same INFO message. However, when comparing the internal state of vispy/GL it looks like linux is returning something different for `handle = gl.glGetUniformLocation(self._handle, name)`. On OSX this returns the proper reference to `u_shape`. On Linux this returns -1. This should be fun to figure out.

davidh-ssec:
Best guess is that linux is optimizing out the `u_shape` uniform because it isn't used for the default "Nearest" filter so it doesn't need to be initialized. OSX is *not* optimizing it out so you get a warning/error when you don't initialize it.

davidh-ssec:
@larsoner Further complication: The part of the code that parses the code for uniforms is pre-GLIR. The part that has the optimized/compiled shader code is post-GLIR. So the options I see:

1. `glir.py` reparses the shader code for uniforms that exist or do `if uniform_name in code:` to not print out the log message.
2. Change the example script to special case the Nearest neighbor shader code to not include the `u_shape` at all. Making a "simple example" that much more complicated.
3. Other?

larsoner:
> glir.py reparses the shader code for uniforms that exist 

I worry about overhead with this

> Change the example script to special case the Nearest neighbor shader code

I think it's okay to use a context manager (or hopefully just `verbose=False` somewhere?) to get this log message to go away.

davidh-ssec:
@larsoner I'm guessing this doesn't need to be there:

https://github.com/vispy/vispy/blob/master/examples/basics/visuals/line_transform.py#L18

larsoner:
Yeah that looks like debugging cruft to me. Feel free to remove in this PR.

+1 for merge if the CIs come back happy (with OSX acceptably sad)

davidh-ssec:
@larsoner Looks like the basics for osx work now and I have the linux tests using the newest wxpython. I'm guessing the failing osx build is due to a window not being able to close or something like that.

larsoner:
Sounds good, we can fix that in a follow-up PR. We probably want to make the tester actually print the examples it's running so we can figure out which one is problematic. Does it pass locally for you?

davidh-ssec:
I just tested it and there was an encoding error. I tried SSH'ing in to my local machine just to see if that made a difference for the display, ran `make examples` and the `colorbar.py` had an encoding error on python 3.6 where it tried to read it as 'ascii' but there are some degree symbols in there. In python 3 you can do `open(..., encoding='utf8')`. Python 2 I'm not sure yet.

larsoner:
`open(..., 'rb')` and `whatever.decode('utf-8')` should fix it. Or just use `codecs.open(..., 'utf-8')`

larsoner:
If this works we can remove the `allow_failures` right?

davidh-ssec:
I suppose we could. And then conda updates some package and every dies, but maybe I'm just being pessimistic.

larsoner:
It can happen, yes, but it's important to keep up with those updates

davidh-ssec:
@larsoner So it looks like there must be something else going on with the `make examples` on OSX. I can't reproduce it on my own systems. I tried ssh'ing in to another mac here at work with a test environment installed, start Xvfb with the same options at travis, and try running the examples. While no one is logged on the the remote system it gets a screen size of 0 which results in an exception. When someone is logged in to the computer console it displays the GUI windows on that users client and displays warnings/errors on my console.

So leave as is and merge?

larsoner:
One other option -- in the test runner, have it check `os.getenv('TRAVIS', 'false') == 'true' and sys.platform == 'darwin'` and if so skip the failing example?

davidh-ssec:
I'm not sure it is just one example. I could make a commit to print out the names of the example scripts that its running (is there an existing option to do that?), then add the code to skip the failing test.

larsoner:
That's what I would do to get it so we run as many examples as possible (and we could remove the allow_failures), but it's fine by me if you'd rather merge as-is, too -- it's not worth getting too hung up on

davidh-ssec:
Interestingly it hangs on `colorbar.py` which is the same script that had the unicode issue. I'm 99% sure that is unrelated, but still.

larsoner:
Easy enough to test, remove the unicode :)

davidh-ssec:
The colorbar example takes long even on my machine. I'll write the logic to skip it later tonight or tomorrow hopefully.

davidh-ssec:
@larsoner Looks like the unicode examples were causing the hang up but the examples still run ridiculously slow. Merge with allowed failures?

larsoner:
We should probably just get rid of the examples run. No reason to wait 50 minutes for something that will be a failed build anyway. That can be done here or in a follow-up PR, so feel free to merge if you're happy

Issue #1178, opened 2016-02-23
Test failures on MacOS X 10.8


astrofrog:
I'm seeing test failures on MacOS X 10.8 with Python 3.5:

```
$ make test
Cleaning build, pyc, ctags, and cache
python make test full
----------------------------------------------------------------------
Running tests with pyqt4 backend:
_VISPY_TESTING_APP=pyqt4   -m vispy_app_test --cov vispy --no-cov-on-fail
======================================================= test session starts ========================================================
platform darwin -- Python 3.5.1, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: /Users/tom/miniconda3/envs/dev35/lib/python3.5/site-packages/vispy-0.5.0.dev0-py3.5.egg/vispy, inifile: 
plugins: cov-2.2.1, xdist-1.14
collected 266 items 

vispy/app/tests/test_app.py ......
vispy/app/tests/test_backends.py .
vispy/app/tests/test_context.py ..
vispy/app/tests/test_qt.py .
vispy/app/tests/test_simultaneous.py .
vispy/gloo/gl/tests/test_basics.py ...s
vispy/gloo/gl/tests/test_functionality.py FFFs
vispy/gloo/tests/test_glir.py ..
vispy/gloo/tests/test_program.py .
vispy/gloo/tests/test_use_gloo.py ....
vispy/gloo/tests/test_wrappers.py ..
vispy/mpl_plot/tests/test_show_vispy.py .
vispy/plot/tests/test_plot.py .
vispy/scene/cameras/tests/test_perspective.py F
vispy/scene/tests/test_node.py .
vispy/util/tests/test_config.py .
vispy/util/tests/test_vispy.py s
vispy/visuals/tests/test_arrows.py ....
vispy/visuals/tests/test_collections.py .
vispy/visuals/tests/test_colorbar.py ...
vispy/visuals/tests/test_ellipse.py ....
vispy/visuals/tests/test_histogram.py .
vispy/visuals/tests/test_image.py .
vispy/visuals/tests/test_markers.py .
vispy/visuals/tests/test_polygon.py ...
vispy/visuals/tests/test_rectangle.py ....
vispy/visuals/tests/test_regular_polygon.py ...
vispy/visuals/tests/test_sdf.py .
vispy/visuals/tests/test_spectrogram.py .
vispy/visuals/tests/test_text.py .
vispy/visuals/tests/test_volume.py .

============================================================= FAILURES =============================================================
[...]

PR #1365, opened 2017-10-02
Add OSX standard/unit tests to travis

davidh-ssec:
Certain tests fail when a backend is used on OSX (see #1178). Currently Travis includes an OSX standard set of tests but without using any backends. This PR adds a test environment for unit testing backends. Currently this is just PyQt4, but should include wxPython once it is available for OSX on conda-forge (currently takes to long to build on CI).

This PR includes removing OSX as an environment that is allowed to fail. I'm assuming that `test_functionality.py` will fail also which will require some special handling (again see #1178).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -111,7 +111,6 @@ install:
     # Also install PyQt4, imaging (PIL or pillow), scipy, mpl, egl
     # On conda, can't install pyside-pyzo b/c it conflicts with pyqt4,
     # which is required by matplotlib :(
-    # No wx on py27 or py34, wxpython doesn't work properly there
     # wxpython available from conda-forge but not for OSX:
     #     https://github.com/conda-forge/wxpython-feedstock/issues/2
     # If we only need a single backend (DEPS=backend), then use PyQT4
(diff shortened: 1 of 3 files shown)


larsoner:
GLFW fails to create windows:

https://travis-ci.org/vispy/vispy/jobs/282238707#L2110

Could be a CI-specific problem. If so let's not build it (and note it somewhere). Then we should be good to go

davidh-ssec:
Yeah that's what I was just going to do. I think it has something to do with the virtual display buffer on OSX.

davidh-ssec:
@larsoner Look ok?

.travis.yml
```diff
@@ -120,11 +119,11 @@ install:
         conda install --yes pyopengl scipy numpy$NUMPY networkx;
         pip install -q numpydoc PySDL2;
         if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
+          conda install --yes wxpython matplotlib jupyter pyqt=4 pillow decorator six;
           if [ "${PYTHON}" == "3.6" ]; then
-            pip install -q freetype-py husl pypng cassowary pillow decorator six;
+            pip install -q freetype-py husl pypng cassowary imageio;
```
larsoner: I'm surprised this works, some other dependencies must pull in `six`, `pillow`, and `decorator`, then?

davidh-ssec: Don't miss that I just moved those dependencies to the conda install.

larsoner: Oh yes I did, thanks

PR #1366, opened 2017-10-02
Fix collections for python 3

### LINKED ISSUE #1340: typeError in vispy collection module
hay,
I checked if the issue existed but couldn't find it, which is odd because here is the thing:
when running the collection examples from python35, I get several typeError which can be corrected (at least so that the example runs)

example : agg_fast_path_collection.py line 127, "itemcount" is defined by a division which returns a float whereas it should be an integer. the fix is simply to "int()" the division or "//".

I don't know the code and implications of such a change.

Regards,

davidh-ssec:
Fix #1340.

If you have a chance @ThomJea it would be great if you could test this too.

### DIFF
--- a/vispy/visuals/collections/agg_fast_path_collection.py
+++ b/vispy/visuals/collections/agg_fast_path_collection.py
@@ -123,7 +123,7 @@ class AggFastPathCollection(Collection):
         """
 
         itemsize = itemsize or len(P)
-        itemcount = len(P) / itemsize
+        itemcount = int(len(P) / itemsize)
 
         P = P.reshape(itemcount, itemsize, 3)
         if closed:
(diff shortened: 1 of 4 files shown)


vispy/visuals/collections/agg_fast_path_collection.py
```diff
@@ -123,7 +123,7 @@ def append(self, P, closed=False, itemsize=None, **kwargs):
         """
 
         itemsize = itemsize or len(P)
-        itemcount = len(P) / itemsize
+        itemcount = int(len(P) / itemsize)
```
larsoner: why not just `//` here, too?

davidh-ssec: Because `itemsize` could come from the user and could be a float. It's possible that `itemsize` being a float could break something else though. I'll check and make the changes if necessary.

davidh-ssec:
For some reason one of the python 2.7 tests randomly fails with a shader compilation error. I think this was fixed by some of my changes in #1365 so I'm going to merge it and fix anything that results.

PR #1369, opened 2017-10-16
Remove year from all source file copyright lines

davidh-ssec:
As discussed on gitter, the easiest to maintain solution that still seems legally correct is to remove the year from all copyright header lines in source code files. Additionally a dated range is used in the `LICENSE.txt` file as an umbrella copyright for the entire project.

Some links that made their way in to the conversation:

https://softwareengineering.stackexchange.com/a/210491/189892
https://softwareengineering.stackexchange.com/a/34221/189892
http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/
https://www.thoughtco.com/copyright-expiration-1991424
https://wiki.osgeo.org/wiki/Code_Provenance_Review

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -3,8 +3,7 @@ Vispy licensing terms
 
 Vispy is licensed under the terms of the (new) BSD license:
 
-Copyright (c) 2015, authors of Vispy
-All rights reserved.
+Copyright (c) 2013-2017, Vispy Development Team. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
--- a/codegen/annotations.py
+++ b/codegen/annotations.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (c) 2013, Vispy Development Team.
+# Copyright (c) Vispy Development Team. All Rights Reserved.
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 
 """ 
--- a/codegen/createglapi.py
+++ b/codegen/createglapi.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
-# Copyright (c) 2013, Vispy Development Team.
+# Copyright (c) Vispy Development Team. All Rights Reserved.
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 
 """
--- a/codegen/get-deprecated.py
+++ b/codegen/get-deprecated.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
-# Copyright (c) 2013, Vispy Development Team.
+# Copyright (c) Vispy Development Team. All Rights Reserved.
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 
 """ Get deprecated funcsions by parsing gl.spec.
--- a/codegen/headerparser.py
+++ b/codegen/headerparser.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (c) 2013, Vispy Development Team.
+# Copyright (c) Vispy Development Team. All Rights Reserved.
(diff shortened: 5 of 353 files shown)


davidh-ssec:
Also note that 2013-2017 was used as the date range since `git log --reverse` shows files moving to a `pyvis` and then to `vispy` repositories around this time.

kmuehlbauer:
👍 for the changes!

I can't immediately tell why travis is breaking. 

I looked into the last working linux build PYTHON=3.6 DEPS=full TEST=standard (https://travis-ci.org/vispy/vispy/jobs/283188638) and from this PR (https://travis-ci.org/vispy/vispy/jobs/288763946), which is breaking in the wx-backend test.

The only differences (I found so far) in that environment are:

| package  | last working | current |
| ------------- | ------------- | ---------- |
| python  | 3.6.2  | 3.6.3 |
| pytest  | 3.2.2  | 3.2.3 |
| jupyter_console  | 5.1.0  | 5.2.0 |
| pillow | 4.2.1 | 4.3.0 |

The PYTHON=2.7 DEPS=full TEST=standard is working for the wx-backend but breaking for qt4-backend. There too, new python version 2.7.14 instead of 2.7.13 and pytest 3.2.3 instead of 3.2.2.

I'll restart these builds to see if this is some transient issue.

kmuehlbauer:
Hmm, PYTHON=3.6 (https://travis-ci.org/vispy/vispy/jobs/288763946) is working now, but the PYTHON=2.7 still failing at the same place. Trying a third run on this...

kmuehlbauer:
Breaking in the same place... 😞

davidh-ssec:
Great...

So something got updated on travis that is more strict on the shader compilation it looks like for the 2.7 failure. The 3.6 failure looks like it is complaining about the changes being too large (too many files changed). @larsoner any idea why this git size check is done? Comments suggest build time:
`# file size checks run on minimal build for time`.

larsoner:
> any idea why this git size check is done? 

Yes to prevent repo bloat. But in cases like this where a big-sized change is warranted you can safely choose to ignore the warning.

larsoner:
As for the failing test, we already know the `arrows` stuff is flakey on Travis, just add another skip like this:

https://github.com/vispy/vispy/blob/master/vispy/visuals/tests/test_arrows.py#L43

davidh-ssec:
@larsoner but that issue was about frame buffers. This error is about shader compilation. I'm going to spend 10 minutes on it.

larsoner:
Yeah, the error message isn't necessarily correct. I seem to remember needing to add one of those in a recent (in the last few months) PR to get things passing.

davidh-ssec:
FYI I was able to replicate the travis environment using their docker images and get the test to fail. I then tried running the tests again both directly from the `testenv` conda environment and by rerunning the CI script and both passed. So...something is weird. I checked all of the apt packages and they seem to be the same versions. I spent too much time on this today so I'll probably add the skipped test, but it feels like I'm really close to figuring something out too.

davidh-ssec:
Update: I made it fail again! Stopping for today, but to get it to fail again I deleted a ton of cache-like directories in the home directory and everything in `/tmp` and the shader compilation failed again. Any guesses while I'm sleeping are appreciated.

larsoner:
It's possible that Mesa software rendering uses LLVM, which could do some caching. But I don't really know

davidh-ssec:
@larsoner Ok I'm going to skip the test, I give up. However, based on the line of code you linked to above it should already be skipped. Right?

larsoner:
Yeah it looks that way, strange. I'd modify the `pytest` runner to run with the command-line option `--showlocals`, and store the backend name and TRAVIS name in local variables. That way when it fails you can see why those conditionals don't pass.

Also the check should be done before the loop, not inside it.

larsoner:
(Also in your next commit if you add a `[size skip]` in the commit message, it will pass/ignore the size test)

davidh-ssec:
@larsoner the backend property was `PyQt4` not `pyqt4` that's why it wasn't being skipped.

larsoner:
> the backend property was PyQt4 not pyqt4 that's why it wasn't being skipped.

:( But at least you fixed it!

PR #1371, opened 2017-10-20
Fix gallery generation

davidh-ssec:
Includes fixes to `scrolling_plots.py`. See commit messages for details.

### DIFF
--- a/examples/basics/plotting/colorbar.py
+++ b/examples/basics/plotting/colorbar.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright (c) Vispy Development Team. All Rights Reserved.
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
-# vispy: gallery 1
+# vispy: gallery 2
 """
 Plot different styles of ColorBar using vispy.plot
 """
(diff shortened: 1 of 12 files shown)


davidh-ssec:
@larsoner I don't know if it'll complain about that warning. It complains when the vispy logger puts out a warning, but not so sure about numpy warnings. This warning is semi-unavoidable without copying arrays and doing extra NaN or 0 checks. It is only happening because the canvas isn't fully set up from what I can tell. It is better than the complete failure it was doing before.

larsoner:
True, feel free to merge once the CIs are happy

Are you planning to add all examples to the website somewhere/somehow?

davidh-ssec:
> Are you planning to add all examples to the website somewhere/somehow?

Not until the release is out unless you feel strongly otherwise. Making websites look good is not my strong suit and I'd rather not waste a ton of time putting up examples. Maybe I'll feel differently this weekend, but there is also no reason it can't be added to the website after the PyPI release is done.

larsoner:
Sounds good to me

davidh-ssec:
Current OSX backlog on travis: 540 jobs

We'll have to wait.

PR #1375, opened 2017-10-26
Update readme

almarklein:
Removed ref to `mpl_plot`. Changed to genesis section. I like the story as an example in OSS, but to me the bullet list puts a lot of emphasis on the 4 original devs, while there have been so many devs since then that deserve as much credit. (I'd be fine with just mentioning the projects.)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -95,7 +95,6 @@ Currently, the main subpackages are:
    with/without WebGL, and others). Relatively stable API.
 -  **gloo**: a Pythonic, object-oriented interface to OpenGL. Relatively
    stable API.
--  **mpl\_plot**: an OpenGL backend for matplotlib. Experimental.
 -  **scene**: this is the system underlying our upcoming high level
    visualization interfaces. Under heavy development and still
    experimental, it contains several modules.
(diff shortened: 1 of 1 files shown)


davidh-ssec:
I agree that it seems a little odd to list the developers, but don't mind it. Since you are one of the listed, I'm ok if you want to remove the bulleted list.

kmuehlbauer:
BTW, should we mention also the new vispy conda-forge packages in the `Installation` section? At least there is already a reference to conda.

davidh-ssec:
@kmuehlbauer Yes. Installation instructions should probably just use `conda-forge` and not mention pip unless there is a missing dependency on conda-forge.

davidh-ssec:
Yeah looking at the README instructions we should probably switch to suggesting pip and conda installation.

kmuehlbauer:
@almarklein @davidh-ssec 

Before I push this to this branch, are you OK with it?

```
Installation
------------

VisPy runs on Python 2.7+ and Python 3.3+ and depends on NumPy. You also
need a backend (PyQt4/PySide, glfw, pyglet, SDL, or wx).

VisPy can be installed either via `pip`:

`pip install vispy`

or within the `Anaconda <https://www.anaconda.com/download/>`_ Python 
distribution. Anaconda provides a convenient package management system. 
Installing VisPy can then easily be achieved by adding `conda-forge` to the 
channels with:

`conda config --add channels conda-forge`

Once the `conda-forge` channel has been enabled, `vispy` can be installed with:

`conda install vispy`

Development
-----------

As VisPy is under heavy development at this time, we highly recommend
developers to use the development version on Github (master branch). You need
to clone the repository and install VisPy with
`python setup.py install`.

As a one-liner, assuming `git` is installed ::

    git clone https://github.com/vispy/vispy.git && cd vispy && python setup.py install --user

This will automatically install the latest version of vispy.
```

davidh-ssec:
`Development Installation` versus `Development`?

Could do `conda install -c conda-forge vispy` and remove the add channel line...but it's probably best that stays in there.

Otherwise LGTM

davidh-ssec:
Also if you can you could update the `doc` submodule for this PR.

kmuehlbauer:
@davidh-ssec Can you have a look, if this doc-submodule update is correct. I'm not sure if I did it right.

davidh-ssec:
@kmuehlbauer Looks correct.

kmuehlbauer:
@davidh-ssec  Just being curious, How is this branch build handled within travis? I mean there is twice the builds, one for the PR the other for the branch.

davidh-ssec:
I don't know why github shows it for both in this PR, but since this branch is *in* the repository that is checked by Travis it will run every new branch. I could disable it for non-master branches. It is indeed running 2 sessions.

kmuehlbauer:
It would surely be problematic if there were some automated stuff in the CI's with some unwanted results. I'm not sure if there is such.

davidh-ssec:
You mean like submitting coverage or generating the docs?

kmuehlbauer:
Yes, some such.

PR #1377, opened 2017-10-29
Fix HiDPI issue in pyqt5

### LINKED ISSUE #1347: SceneCanvas central_widget display area is wrong (PyQt5 & SDL2)
Hi,

I tried this code, but plot area is wrong because SceneCanvas central_widget is not placed at correct area.
https://github.com/vispy/vispy/issues/1189#issuecomment-198597473

I think the bug is because of PyQt5/SDL2, but I'm not familiar with Qt...

I tried PyQt5, PyGlet, Glfw and SDL2 as backend app.
In PyQt5 and SDL2, display area is wrong.
In PyGlet and Glfw, display area is correct.
PyGlet is a little slow, so I'll use Glfw.

My environment is...
OS: macOS Sierra
vispy: 0.4.0
Python: 3.6.2
Qt: 5.9.1
PyQt: 5.9
SLD2: 2.0.5
PySDL2: 0.9.5
Glfw: 3.2.1
PyGlet: 1.2.4

vispy, PyGlet and PySDL2 are installed by pip.
Python, Qt, PyQt, SLD2 are installed by Homebrew.

davidh-ssec:
Fix #1347

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -287,7 +287,12 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
         # Qt supports OS double-click events, so we set this here to
         # avoid double events
         self._double_click_supported = True
-        self._physical_size = p.size
+        if hasattr(self, 'devicePixelRatio'):
+            # handle high DPI displays in PyQt5
+            ratio = self.devicePixelRatio()
+        else:
+            ratio = 1
+        self._physical_size = (p.size[0] * ratio, p.size[1] * ratio)
 
         # Activate touch and gesture.
         # NOTE: we only activate touch on OS X because there seems to be


vispy/app/backends/_qt.py
```diff
@@ -287,7 +287,12 @@ def __init__(self, *args, **kwargs):
         # Qt supports OS double-click events, so we set this here to
         # avoid double events
         self._double_click_supported = True
-        self._physical_size = p.size
+        if hasattr(self, 'devicePixelRatio'):
+            # handle high DPI displays in PyQt5
+            ratio = self.devicePixelRatio()
+        else:
+            ratio = 1
```
larsoner: This can probably just be `ratio = getattr(self, 'devicePixelRatio', lambda: 1.)()`

davidh-ssec: Or `ratio = self.devicePixelRatio() if hasattr(self, 'devicePixelRatio') else 1.`. Your way isn't really my style (making a lambda as a default attribute).

davidh-ssec:
@larsoner Wow I never noticed that translucent issue. That's a problem. Is there a github issue for that?

larsoner:
Not that I know of

PR #1381, opened 2017-10-31
Fix sphere passing rows/cols backwards

### LINKED ISSUE #1259: Sphere.py rows and cols wrong way round
Hi,

Might be wrong, but it looks like in vispy/visuals/Sphere.py, when passing rows and cols to create_sphere, they are passed back to front. create_sphere expects rows then cols, but SphereVisual passes cols then rows. As they're unnamed, I was producing different meshes than I expected.

Thanks

davidh-ssec:
Fix #1259 

Based on the above issue it seemed a little weird that no one would have noticed two parameters being backwards. What was actually wrong was that the code for the `cube` method was wrong twice so the issue wasn't showing. The `latitude` method was actually the method that showed the issue. If @kalexander92 and maybe @kmuehlbauer or @larsoner could test `examples/basics/scene/sphere.py` with this branch and try changing the rows and cols of the third sphere and see if the numbers make sense to them.

### DIFF
--- a/vispy/geometry/generation.py
+++ b/vispy/geometry/generation.py
@@ -403,7 +403,7 @@ def _ico(radius, subdivisions):
 
 def _cube(rows, cols, depth, radius):
     # vertices and faces of tessellated cube
-    verts, faces, _ = create_box(1, 1, 1, rows, cols, depth)
+    verts, faces, _ = create_box(1, 1, 1, cols, rows, depth)
     verts = verts['position']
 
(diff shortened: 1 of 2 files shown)


larsoner:
Sorry I didn't have time to test. But I trust your judgment here :)

davidh-ssec:
No problem. I didn't wait very long. I wanted to clear out some PRs to handle the `doc` fixes and forgot I asked for you guys to double check this.

PR #1383, opened 2017-10-31
Fix doc directory in source tarball

### LINKED ISSUE #1382: Wrong install location for docs
From `setup.py`:
```
setup(
   [...]
    package_data={
        'vispy': [op.join('io', '_data', '*'),
                  op.join('html', 'static', 'js', '*'),
                  op.join('app', 'tests', 'qt-designer.ui'),
                  op.join('..', 'doc', '*'),
                  ],
```
This line `op.join('..', 'doc', '*')` is wrong for a system-wide install. It leads to the documentation being install under `dist-packages` or `site-packages`, which is definitely non-standard. IMO, the best would be to just not install the docs yourself, and let the package build system (conda or Debian) handle it.

davidh-ssec:
Fix #1382 

The 2nd commit in this PR will need to be cherry picked in to a maintenance branch so that the 0.5.0 package can be repackaged and upload to PyPI. We may need to make a new build revision for conda-forge (@kmuehlbauer FYI) so this gets rebuilt too.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,7 @@
+# Add documentation to the source tarball but not in the installation
+include doc/Makefile doc/.nojekyll doc/conf.py
+include doc/*.rst
+recursive-include doc/ext *
+recursive-include doc/_static *
+recursive-include doc/_templates *
+
--- a/setup.py
+++ b/setup.py
@@ -109,7 +109,6 @@ setup(
         'vispy': [op.join('io', '_data', '*'),
                   op.join('html', 'static', 'js', '*'),
                   op.join('app', 'tests', 'qt-designer.ui'),
-                  op.join('..', 'doc', '*'),
                   ],
 
         'vispy.glsl': ['*.vert','*.frag', "*.glsl"],


davidh-ssec:
@ghisvail I'd like your opinion on this too if this is the right way to do this. I copied the idea from numpy so if their way works this should too, hopefully.

ghisvail:
@davidh-ssec I agree with your proposed PR.

> the 0.5.0 package can be repackaged and upload to PyPI

Don't. What is published is published. Consider making a hotfix release instead (0.5.1).

kmuehlbauer:
@davidh-ssec Just ping me when this gets around the corner, l'll have conda-forge updated then.

davidh-ssec:
@ghisvail I was hoping PyPI had something equivalent to conda's build revisions to say "it was built wrong before but the code is the same".

ghisvail:
PyPI is a source (and optionally binary) package provider, conda is a binary package manager. Different goals, different constrains.

FYI, we have got the same constrains in Debian: if one messes up the source import, the source version has to be bumped, if one messes up the package build, the binary version has to be bumped. That's the only way to ensure both smooth upgrades and reproducibility of builds.

kmuehlbauer:
@larsoner Are there any guidelines on how to handle hotfixes for VisPy? I can't remember any hotfixes since I've been contributing to VisPy.

larsoner:
I am in the published-is-published camp with @ghisvail so I'd just cut a 0.5.1.

davidh-ssec:
@larsoner Tag it and github release?

larsoner:
Yeah, with the annoying 50-step process involving changing of `__version__` and such. Actually having a separate `maint/0.5` branch makes this slightly less annoying (because you just cherry-pick commits and bump the version there and otherwise leave `master` alone) so maybe now is a good time to start that branching process. I can do it if you're unsure

davidh-ssec:
Sure go for it. So `maint/0.5` branch with this PR merged, then commits to change the version to 0.5.1 and upload the package. I can update the documentation site if you'd like...actually I'd be curious if you get all the warnings I was getting.

larsoner:
Okay I will try to do it tomorrow

PR #1387, opened 2017-11-06
Refactor ellipse and rectangle visuals

### LINKED ISSUE #1349: minor: border width bug in rectangleVisual
Hi,

the border width argument of the rectangle.py (visuals) is not passed to the polygon class.
The last lines of code in the file should be:

`if not self._border_color.is_blank:`
`   self.border.set_data(pos=vertices[1:, ..., :2],`
`   color=self._border_color.rgba,`
 **`width=self._border_width`**)

davidh-ssec:
Fix #1349

Issues fixed:

 - Border width was not used in RectangleVisual
 - Border loops were not properly being closed due to a 1 versus -1 index error.

This refactor tries to reduce the amount of duplicated code. There is something weird going on in the original code where the center point of the polygon's face (ellipse and rectangle) was being passed as a mesh vertex but as far as I can tell it is not needed.

Lastly, updating the rectangle/ellipse vertices could be done differently but I would argue no less ugly (`_regen_pos()`).

### DIFF
--- a/vispy/testing/image_tester.py
+++ b/vispy/testing/image_tester.py
@@ -375,7 +375,7 @@ def get_test_data_repo():
     # This tag marks the test-data commit that this version of vispy should
     # be tested against. When adding or changing test images, create
     # and push a new tag and update this variable.
-    test_data_tag = 'test-data-6'
+    test_data_tag = 'test-data-7'
 
     data_path = config['test_data_path']
     git_path = 'http://github.com/vispy/test-data'
--- a/vispy/testing/image_tester.py
+++ b/vispy/testing/image_tester.py
@@ -150,7 +150,7 @@ def assert_image_approved(image, standard_file, message=None, **kwargs):
                 raise Exception("Test standard %s does not exist." % std_file)
             else:
                 if os.getenv('TRAVIS') is not None or \
-                                os.getenv('APPVEYOR') is not None:
+                        os.getenv('APPVEYOR') is not None:
                     _save_failed_test(image, std_image, standard_file)
                 raise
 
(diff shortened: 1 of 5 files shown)


davidh-ssec:
@larsoner Any ideas on this one? Looks like GL is rendering slightly different on linux compared to OSX.

larsoner:
I assume you generated this on OSX, then? It also does not match Windows that well either:

https://ci.appveyor.com/project/larsoner/vispy/build/1.0.1896/job/x2n6sf11wgj16171

It's weird because it actually seems like a big difference -- more than a single pixel for the red / filled part at least. I wonder if the problem is that the border is getting drawn beneath the filled part on one platform, and above it on the other (i.e., depth testing messes things up). I don't know how to fix this problem. You could try disabling depth testing and ensuring the draw order is set properly, but I don't know if this would have unintended consequences (or if it would even work).

BTW we should update this to respect `AppVeyor` as well:

https://github.com/vispy/vispy/blob/master/vispy/testing/image_tester.py#L152

If you really need to, if you can verify that this problem also exists in `master`, then as long as this PR doesn't make things worse then you could just increase the tolerance for the test, and open a separate issue for border/fill-drawing issues.

davidh-ssec:
Yes, generated on OSX. I thought at first maybe it was a high DPI issue. I have a linux VM that I'm able to reproduce this on. I'll see what I can do with another 30 minutes of testing.

davidh-ssec:
@larsoner Looks like I need to read the vispy documentation closer. `glLineWidth` is not guaranteed for line widths other than 1.0. It works on my mac but not on my linux VM. I can force the line method to 'agg' where the results seem consistent between my linux VM and OSX, however it is definitely a hack and the results aren't great when zoomed in.

So basically although we allow the width to be specified I think it shouldn't be relied on by users. So I'm kind of thinking of removing my new test. If I leave it in the image_tester code will upload a diff of the image every time it fails.

davidh-ssec:
I added a `border_method` keyword argument locally to PolygonVisual which produces a rectangle like the below with 'agg':

image

Not sure what's up with the part in the middle of the line.

PR #1388, opened 2017-11-06
Fix zoom event position in Pyglet backend

mkubanski:
Mouse scroll event reported wrong position


davidh-ssec:
Strange that this got missed when all the other mouse events seem to do the same logic.

PR #1393, opened 2017-11-08
FIX: Fix Qt5 transparency

### LINKED ISSUE #1238: pyqt5 backend bug
When i use PyQt5 backend, canvas going to be transparent if i draw primitives with alpha channel.

This is in Python 3.4 PyQt5 (incorrect):
image

This is in Python 2.7 PyQt4 (correct):
image

Do you have some ideas about this?

``` python
import numpy as np
from vispy import app, gloo

vertex = """
attribute vec3 a_position;
void main (void)
{
    gl_Position = vec4(a_position, 1.0);
}
"""

fragment = """
void main()
{
    gl_FragColor = vec4(1.0, 0.0, 0.0, 0.1);
}
"""

c = app.Canvas()
program = gloo.Program(vertex, fragment)

# gloo.set_state('translucent')
gloo.set_state('opaque')

program['a_position'] = np.array([
    [-1.0, -1.0, 0.0],
    [-1.0, 1.0, 0.0],
    [1.0, -1.0, 0.0],
    [1.0, 1.0, 0.0]
], dtype=np.float32)

@c.connect
def on_resize(event):
    gloo.set_viewport(0, 0, *event.size)

@c.connect
def on_draw(event):
    gloo.clear((0.0, 0.0, 0.0, 1.0))
    program.draw('triangle_strip')

c.show()
app.run()
```

larsoner:
I updated one of the Qt widget examples to support PyQt5 so I could ensure I didn't break it (still works here).

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -21,6 +21,7 @@ and segfaults.
 
 from __future__ import division
 
+from copy import deepcopy
 from time import sleep, time
 import os
 import sys
@@ -62,6 +63,7 @@ def _check_imports(lib):
         if lib2 in sys.modules:
             raise RuntimeError("Refusing to import %s because %s is already "
                                "imported." % (lib, lib2))
+
 
 # Get what qt lib to try. This tells us wheter this module is imported
 # via _pyside or _pyqt4 or _pyqt5
@@ -158,6 +160,8 @@ def message_handler(*args):
     else:
         msg = msg.decode() if not isinstance(msg, string_types) else msg
         logger.warning(msg)
+
+
 try:
     QtCore.qInstallMsgHandler(message_handler)
 except AttributeError:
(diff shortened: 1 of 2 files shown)


vispy/app/backends/_qt.py
```diff
@@ -642,7 +647,10 @@ class CanvasBackendDesktop(QtBaseCanvasBackend, QGLWidget):
     def _init_specific(self, p, kwargs):
 
         # Deal with config
-        glformat = _set_config(p.context.config)
+        config = deepcopy(p.context.config)
+        # For non-see-through windows OSX and Windows (gh-2138), use default
```
davidh-ssec: Wrong issue number?

davidh-ssec:
I'll test this in a bit. A couple questions:

1. Why did this become a problem between Qt4 and Qt5?
2. You say depth buffer but are modifying the alpha buffer?
2. The Qt docs are a little confusing about this. You can have `setAlpha(True)` (http://pyqt.sourceforge.net/Docs/PyQt4/qglformat.html#setAlpha) which enables the alpha buffer and is not the default but you can also `setRgba(True)` which is the default. Any idea how these are different? Alpha is for the window as a whole and Rgba is for colors in the canvas?

larsoner:
> Why did this become a problem between Qt4 and Qt5?

No idea, maybe a Qt bug

> You say depth buffer but are modifying the alpha buffer?

Comment error

Argh, doing `setAlphaBufferSize` implicitly enables alpha (so no need for `setAlpha`), but once I do `setAlpha(True)` it breaks again. So actually it's using an alpha buffer at all (which we need) that causes the problem. So this solution is incorrect actually :(

davidh-ssec:
If it makes you feel better it works for me. ;)

larsoner:
Okay I give up for now. I'll just leave the changes that update the example for now.

djhoese:
I can't be sure but I think this PR is unnecessary now with all of the other changes that went in to vispy 0.6 and making the PyQt5 backend work. I'm closing this since I don't think it is relevant any more.

PR #1394, opened 2017-11-08
[WIP] ENH: Migrate PyQt5 backend to QOpenGL API

### LINKED ISSUE #1238: pyqt5 backend bug

larsoner:
Rebase / potential takeover for #960.
 
Closes #1238.
Closes #960.

Let's see if it works!


QuLogic:
A few examples were still randomly failing last I checked, but if you want to look into that...

larsoner:
Well the CIs seem happy, I even restarted the `examples` build after succeeding once to see if it would succeed again and it did. Maybe we're good to go, then -- @QuLogic WDYT?

davidh-ssec:
@larsoner None of the CIs use PyQt5 over PyQt4, right?

larsoner:
> None of the CIs use PyQt5 over PyQt4, right?

Argh yes you're right, trying now

davidh-ssec:
It might be worth it to replicate one of the environments and use PyQt5 for all the tests. Maybe as an allowed failure.

larsoner:
Windows and Linux pass (including examples), just 4 failures on OSX:

https://travis-ci.org/vispy/vispy/jobs/299438698

I'll see if I can make some sense of them hopefully tomorrow.

larsoner:
Oh wait I didn't run the examples with PyQt5, let me fix that...

larsoner:
Examples pass, Linux crashes, OSX has 4 errors. I'm going to see if I can fix the OSX errors, maybe those will fix the Linux crash.

davidh-ssec:
Should we maybe wait until after 0.5.2 or maybe 0.6.0 to merge this?

larsoner:
Who knows how long it will take to get it working. I if the CIs are happy it's fine for 0.6 assuming it's not right before release.

larsoner:
On my OSX machine I get errors because the windows go into some tabbed mode, so the rendering sizes get messed up.

On my Linux machine I get a bunch of errors, seemingly anytime a `FrameBuffer` is used.

This one will be tough :(

QuLogic:
IIRC, the problem with the examples is that something somewhere assumes the current framebuffer, but with Qt5, it manages the framebuffer and the default is not the one we want to use. I added the commands to switch them, but there's still something not doing it correctly and I don't really know where to look right now.

djhoese:
I played around with this after merging current master back in to it. There are still some errors, but most things seem to work locally for me on OSX Python 3.6. Weird errors include shader compilation error in `test_app.py` that passes on my local travis linux docker image when I pass `-s` to pytest (to turn off stdout capture).

Another weird one is the example `examples/demo/scene/force_directed_graph.py` which fails because a warning message shows up talking about invalid values in a sqrt operation, but as far as I can tell there shouldn't be any NaNs or negative numbers in the calculation.

Lastly, the `sys_info()` when using pyqt5 doesn't show any GPU/GL information. If you use a straight pyopengl/pyqt5 example (https://github.com/baoboa/pyqt5/blob/master/examples/opengl/hellogl.py) then information can be retrieved.

I'll spend a little longer on this, but I'm kind of stuck on where to go from here.

larsoner:
If it helps, with `python make test pyqt5 -s` (or without the `-s`) on Linux I get:
```
         - vispy/app/tests/test_context.py:12 test_context_properties
         - vispy/gloo/tests/test_program.py:69 ProgramTest.test_error
         - vispy/gloo/tests/test_wrappers.py:152 test_wrappers
         - vispy/visuals/tests/test_text.py:9 test_text
```

djhoese:
@larsoner What version of numpy? It seems like numpy 1.14.3 was causing a lot of issues on my travis docker container.

djhoese:
@larsoner Examples pass! Other stuff doesn't though.

larsoner:
I was on bleeding edge 1.15.0.dev, if I build and use 1.14.x maintenance branch, I get two of the same and three new ones:
```
         - vispy/app/tests/test_context.py:12 test_context_properties
         - vispy/gloo/tests/test_use_gloo.py:26 test_use_framebuffer
         - vispy/gloo/tests/test_wrappers.py:152 test_wrappers
         - vispy/visuals/tests/test_infinite_line.py:20 test_set_data
         - vispy/visuals/tests/test_rectangle.py:67 test_rectpolygon_draw
```

larsoner:
With my latest commit I get just intermittent failures like:
```
         - vispy/visuals/tests/test_ellipse.py:44 test_ellipse_draw
```
I think it's related to canvases not properly cleaning up. In the old API we could do `self.context().reset()` in `_vispy_close` which invalidated the old context. I haven't found the equivalent for PyQt5 but if we can, it might solve the problem.

I stuck a couple `process_events` in there to try to fix the problem but I'm not sure if it did.

djhoese:
@larsoner Isn't `reset()` still available: https://doc.qt.io/qt-5/qglcontext.html#reset

djhoese:
Nevermind that's the old API.

djhoese:
@larsoner Looking at one of the remaining issues makes me think we need to figure out why we can't get any GL parameters (sys_info being empty). This is the only error I get on my travis vm:

```
        if viewport is None:
            viewport = gl.glGetParameter(gl.GL_VIEWPORT)
        viewport = np.array(viewport, int)
        if viewport.ndim != 1 or viewport.size != 4:
            raise ValueError('viewport should be 1D 4-element array-like, not %s'
>                            % (viewport,))
E           ValueError: viewport should be 1D 4-element array-like, not []
```

larsoner:
@djhoese I get that error (in `test_wrappers.py`) but only on some runs. This seems to indicate some problem with context switching / proper closing of contexts, thus why I worked a bit on tweaking them. But my solutions still leave it only intermittently working.

djhoese:
I mentioned this on gitter, but will mention it here for historical purposes. I was able to get `sys_info` to work by manually creating my app and canvas (the same way sys_info does) then showing the canvas which causes `QOpenGLWidget` to create the QOpenGLContext. Without this the context is not set up and things like `glGetParameter` won't return anything.

I agree that there seem to be some race-like conditions @larsoner and that it seems like PyQt5 is not treating the context as we expect.

larsoner:
FWIW my traceback is:
```
vispy/gloo/tests/test_wrappers.py:173: in test_wrappers
    y = gloo.read_pixels(alpha=False, out_type=np.ubyte)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

viewport = array([], dtype=int64), alpha = False, out_type = <class 'numpy.uint8'>

    def read_pixels(viewport=None, alpha=True, out_type='unsigned_byte'):
        """Read pixels from the currently selected buffer.
...
        if viewport.ndim != 1 or viewport.size != 4:
            raise ValueError('viewport should be 1D 4-element array-like, not %s'
>                            % (viewport,))
E           ValueError: viewport should be 1D 4-element array-like, not []
```
And if I do `pytest vispy/gloo` I don't get the error.

djhoese:
@EtienneCmb Thanks for putting your experience here. This follows what I've been seeing. I need to work on figuring out the best way for VisPy to create a Qt5 GL context when a widget isn't opened first. Essentially, I think we need to create our own context/surface that we render to and communicate with if we haven't show a widget and use the default Qt5 one when a widget has been show (a requirement in Qt5 for creating a GL context).

kmuehlbauer:
This did not have any effect on travis. Think we need to strip the build matrix to debug this.

djhoese:
@kmuehlbauer These tests never passed. Only recently did I get some of the OSX tests to pass. This is what I've been working on. Trying to figure out what about PyQt5 is so terrible.

kmuehlbauer:
This is a neverending story, I believe.

kmuehlbauer:
@djhoese I'm wondering if we should not just approaching `pyside2` (https://wiki.qt.io/Qt_for_Python).

djhoese:
We can try using PySide 2 as part of this PR and if we get that to work with travis then 👍 . However I would still like to have PyQt5 unless vispy 0.6 also drops python 2 support.

hmaarrfk:
drop it like its hot.

kmuehlbauer:
@hmaarrfk You referring to py2.7?

hmaarrfk:
Yes, unless vispy is going to get fixed in the next few weeks, I feel like holding on to the python 2 dream will be wasted effort.

hmaarrfk:
that said, the failures seem unrelated to python 2.

djhoese:
Agreed. Python 2 support doesn't seem to fully exist for PySide 2 so having working tests with PyQt5 (which don't seem to be failing for python 2 only) should be possible and be done first or at the same time.

hmaarrfk:
i think you need xenial for qt 5.12, not sure about other stuff.

djhoese:
@hmaarrfk @kmuehlbauer I got rid of pyside2 from the travis environment. Seems like it segfaults and may have been fixed in later versions of pyside2 that are not yet available on conda/conda-forge: https://bugreports.qt.io/browse/PYSIDE-585

I also upped the travis linux distribution to xenial. Only issue seems to be `libts-dev` not being available for xenial.

djhoese:
For OSX I added an expected fail for `test_perspective.py`. There isn't much I can do to fix this test and I'm not sure I'm willing to overwrite the "valid" image. Here is what `--vispy-audit-tests` shows:

image

Also going to bump testing to use python 3.7 just for "fun" and see what changes.

vispy/gloo/util.py
```diff
@@ -101,7 +101,7 @@ def check_enum(enum, name=None, valid=None):
 
 void main(void) {
     v_uv = a_texcoord.xy;
-    gl_Position = vec4(a_position, 0, 1);
+    gl_Position = vec4(a_position, 0., 1.);
```
hmaarrfk: crazy, how do you find these things?

djhoese: A long time ago when fixing new/random travis failures, I started a travis docker container on my local machine to debug some shader compilation failures. On the local container it either showed a better error message or I figured it out, maybe both. Some GL compilers don't like integers being mixed with floats when creating a vec.

djhoese:
@larsoner @kmuehlbauer @hmaarrfk Here is a summary of the current status of this PR and where I think we should go from here:

1. Windows tests on appveyor were failing when matplotlib and pyqt5 were installed. I noticed this after realizing that py27 was failing but py37 was passing. It was because nothing "important" was being installed in the py37 environment. There seemed to be some segmentation faults and other more regular failures. From what I can tell this is something wrong with the OpenGL version or C builds running on appveyor. @larsoner if you feel like updating the mesa libs for appveyor, maybe we could test more. For now I have appveyor running the nobackend/basic tests and they're passing.
2. While looking in to the remaining OSX failures I realized that doing `print(vispy.sys_info())` was causing a segmentation fault on my local OSX 10.14 machine even though the tests were passing. This caused me to re-enable the OffscreenSurface stuff I previously added to `_qt.py` which fixes the segmentation fault. The tests *all* pass on my python 3.7 environment with newest conda-forge packages on my OSX 10.14 laptop with dedicated GPU. However, these changes did not fix the OSX tests which as far as I can tell seem to be tied to the GL drivers on the OSX VM on travis (seems like things aren't getting drawn/resized) which is a OSX 10.13 system with APPLE GL drivers.
3. Given the info about OSX in number 2 above, I'd like to set the OSX "full" tests as an allowed failure. We can watch and see if things change over time.

Actually...I'm going to try upping the version of OSX on travis and see what breaks.

djhoese:
Cool, failing even worse now with a segmentation fault. I blame the OSX images on travis. Works for me. Good enough for the rest of you to put OSX "full" tests on allowed failures?

larsoner:
Skipping the *entire run* seems a bit drastic -- people will (reasonably) completely ignore it this way. It would be better to skip in as specific a way as possible the problematic tests, for example tests X, Y, and Z only when on OSX and Travis. That way we still run as many tests as possible.

djhoese:
"Skipping" is a strong word, but I understand. Now that I found out that travis has a debug mode I'll try to debug this further either today or tomorrow.

djhoese:
Passing tests! All the failing or seg faulting ones seemed to be related to multiple canvases being created in the test in different ways. I skipped segfaults for osx on travis, xfail for failed ones.

@larsoner @kmuehlbauer Merge?

larsoner:
Testing issues aside, did you work out the remaining PyQt5 issues such that VisPy works properly with modern PyQt5 commands now?

djhoese:
> did you work out the remaining PyQt5 issues such that VisPy works properly with modern PyQt5 commands now?

What do you mean by that? All tests pass on my local system. I can try using this branch in my application and see if everything works, but that's not a super heavy test.

vispy/app/backends/_qt.py
```diff
@@ -183,7 +198,7 @@ def message_handler(*args):
     resizable=True,
     decorate=True,
     fullscreen=True,
-    context=True,
+    context=not QT5_NEW_API or os.getenv(PYQT5_SHARE_ENV, False),
```
larsoner: `getenv` always returns a `str` so you could have `export VISPY_PYQT5_SHARE_CONTEXT=false` and this conditional will eval to `context=True`, you need to do a `str` comp like:
```
os.getenv(PYQT5_SHARE_ENV, 'false').lower() == 'true'
```

vispy/app/backends/_qt.py
```diff
@@ -199,15 +214,21 @@ def _set_config(c):
     glformat.setGreenBufferSize(c['green_size'])
     glformat.setBlueBufferSize(c['blue_size'])
     glformat.setAlphaBufferSize(c['alpha_size'])
-    glformat.setAccum(False)
-    glformat.setRgba(True)
-    glformat.setDoubleBuffer(True if c['double_buffer'] else False)
-    glformat.setDepth(True if c['depth_size'] else False)
+    if QT5_NEW_API:
+        # Qt5 >= 5.4.0 - below options automatically enabled if nonzero.
+        glformat.setSwapBehavior(glformat.DoubleBuffer if c['double_buffer']
```
larsoner: This is the "modern" sort of stuff I was talking about -- the new API introduced in #960

vispy/app/backends/_qt.py
```diff
@@ -219,6 +240,10 @@ class ApplicationBackend(BaseApplicationBackend):
 
     def __init__(self):
         BaseApplicationBackend.__init__(self)
+        # sharing is currently buggy and causes segmentation faults for tests with PyQt 5.6
+        if QT5_NEW_API and os.getenv(PYQT5_SHARE_ENV, False):
```
larsoner: would be good to deduplicate this with a tiny helper function (seems overkill, but it's wrong in currently two places and having a little helper for it will make it easier on the next dev who gets shared contexts working to change one thing to change the default)

vispy/gloo/wrappers.py
```diff
@@ -677,7 +676,12 @@ def get_gl_configuration():
     # XXX eventually maybe we can ask `gl` whether or not we can access these
     gl.check_error('pre-config check')
     config = dict()
-    gl.glBindFramebuffer(gl.GL_FRAMEBUFFER, 0)
+    canvas = get_current_canvas()
+    if canvas and hasattr(canvas, '_backend'):
```
larsoner: when does `canvas` not have `_backend`?

vispy/gloo/wrappers.py
```diff
@@ -694,15 +698,35 @@ def get_gl_configuration():
...
+
+    for key, enum in (('stereo', GL_STEREO),
+                      ('double_buffer', GL_DOUBLEBUFFER)):
+        val = gl.glGetParameter(enum)
+        try:
+            gl.check_error('post-config check')
+        except Exception as exp:
+            logger.warning('Failed to get %s: %s' % (key, exp))
+        else:
+            config[key] = bool(val)
```
larsoner: Do we still need these changes? Might be good to add a comment why they're needed

vispy/app/backends/_qt.py
djhoese: Yes, of course. This was meant to be more of a hack, but I agree the more flexible thing would be to actually interpret the truthiness of the value.

vispy/app/backends/_qt.py
djhoese: I still a little confused. This PR was started by you @larsoner. I simply took what you had and modified a few things to get the tests to pass. The main thing being QOffscreenSurface use for rendering and getting a GL context without explicitly showing a Widget first.

This PR should be a replacement for #960 based on previous conversations we've had.

vispy/app/backends/_qt.py
djhoese: Agreed.

vispy/gloo/wrappers.py
djhoese: I don't know. This was added by you.

vispy/gloo/wrappers.py
djhoese: Again, this was you. I was hoping you have a better idea why they're needed.

vispy/gloo/wrappers.py
larsoner: Maybe it is unnecessary now. If you feel up to removing them to see if it breaks anything go for it. If not, we can try removing it later if someone is motivated to

vispy/gloo/wrappers.py
larsoner: No recollection :(

vispy/gloo/wrappers.py
djhoese: I took out the `hasattr` check. I could see this having problems with some backends, but right now they all (including ipynb) are based on the `BaseCanvasBackend` which sets this `_backend`.

vispy/gloo/wrappers.py
djhoese: I changed the `except Exception` to `except RuntimeError` in an effort to have more explicit exception handling. I don't have a huge issue with more explicit warnings and error messages. We'll see if they still pass.

larsoner:
Thanks @djhoese !

PR #1395, opened 2017-11-11
FIX: fix order of alignment values for proper processing

### LINKED ISSUE #641: Alignment wrong computation when aligned on 8 bytes
Hello,

While reviewing code in vispy/glir.py I found this comment that looks erroneous:

```
# we know the alignment is appropriate
# if we can divide the width by the
# alignment cleanly
# valid alignments are 1,2,4 and 8
# put 4 first, since it's the default
alignments = [4, 8, 2, 1]
for alignment in alignments:
    if width % alignment == 0:
        return alignment
```

the issue is that if width is multiple of 8 then the alignment returned will always be 4... since it is always multiple of 4 too.

So either either remove the "8" case for speed or define alignment to [8,4,2,1].

Laurent

kmuehlbauer:
closes #641

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -1041,8 +1041,8 @@ class GlirTexture(GlirObject):
         # if we can divide the width by the
         # alignment cleanly
         # valid alignments are 1,2,4 and 8
-        # put 4 first, since it's the default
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
Nothing breaking, seems this is good for merge.

PR #1404, opened 2017-12-16
FIX: correct rows/cols in IsocurveVisual 

### LINKED ISSUE #1403: Possible dimensions error in Isocurve
Hi,
working by Vispy, that is great! I think I found a possible problem in
` vispy/vispy/visuals/isocurve.py.`
Isocurve works only if the data first dimension is equal to the second dimension. Taking a look at the code it seems that there is a wrong order of the arguments for the function np.meshgrid. In particular, if data has a shape, for instance (n,m), the result of the meshgrid function are two matrix (m,n).

https://github.com/vispy/vispy/blob/9d46fac28eac985fbe5d032d1581710fb74999ea/vispy/visuals/isocurve.py#L107-L109

my solution:

            if self._X is None or self._X.T.shape != data.shape:
                self._X, self._Y = np.meshgrid(np.arange(data.shape[1]), **_<- changed_**
                                               np.arange(data.shape[0])) **_<- change_**

Am I wrong? I hope it helps you.
Kind regards,
Luigi

kmuehlbauer:
* correct rows/cols in IsocurveVisualfor matplotlib isoline tracer
* fix examples to have (200, 100) dimension

closes #1403

### DIFF
--- a/examples/basics/scene/isocurve.py
+++ b/examples/basics/scene/isocurve.py
@@ -21,8 +21,8 @@ canvas.show()
 view = canvas.central_widget.add_view()
 
 # Create the image
-img_data = np.empty((100, 100, 3), dtype=np.ubyte)
-noise = np.random.normal(size=(100, 100), loc=50, scale=150)
+img_data = np.empty((200, 100, 3), dtype=np.ubyte)
+noise = np.random.normal(size=(200, 100), loc=50, scale=150)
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
Wow, after all that travis issues, I really thought this would break somehow. 

This is good to go from my end.

PR #1406, opened 2017-12-17
Don't draw zero-size markers

EtienneCmb:
This PR add a if test to the vertex shader of markers.py to avoid drawing markers that have a 0. size

### DIFF
--- a/vispy/visuals/markers.py
+++ b/vispy/visuals/markers.py
@@ -32,7 +32,7 @@ varying float v_edgewidth;
 varying float v_antialias;
 
 void main (void) {
-    if (a_size != 0.)
+    if (a_size > 0.)
     {
         $v_size = a_size * u_px_scale * u_scale;
         v_edgewidth = a_edgewidth * float(u_px_scale);
--- a/vispy/visuals/markers.py
+++ b/vispy/visuals/markers.py
@@ -32,14 +32,17 @@ varying float v_edgewidth;
 varying float v_antialias;
 
 void main (void) {
-    $v_size = a_size * u_px_scale * u_scale;
-    v_edgewidth = a_edgewidth * float(u_px_scale);
-    v_antialias = u_antialias;
-    v_fg_color  = a_fg_color;
-    v_bg_color  = a_bg_color;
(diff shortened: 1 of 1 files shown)


EtienneCmb:
The following code draw markers with half of them having a size of 0. but I think edges are still drawn :

```python
import numpy as np

from vispy import app
from vispy.scene.cameras import TurntableCamera
from vispy.scene.visuals import Markers

cam = TurntableCamera()
sc = SceneCanvas(bgcolor='black')

xyz = np.random.rand(100, 3)
size = [0.] * 50 + [10.] * 50
m = Markers(pos=xyz, face_color='white', parent=wc.scene, edge_color='white', size=size)

sc.show()
app.run()
```

vispy canvas_012

larsoner:
> The following code draw markers with half of them having a size of 0. but I think edges are still drawn :

Yes probably. If you duplicate the size-checking logic in the fragment shader with `discard` it will probably fix this

EtienneCmb:
Thanks @larsoner  for the suggestion. This new PR use `discard` in the frag shader. Why it's better to use `discard` instead of the if test in the vert shader?

larsoner:
Did it work? I didn't check but I think it should.

You could use both to save some GPU cycles, but it's probably fast enough just to use `discard` in the fragment shader. In this case I assumed (without looking at the code, but seeing the described problem) that the you'd need the `discard` in the fragment shader, otherwise the edge-drawing logic would kick in and draw regardless of what you did with the vertices.

davidh-ssec:
@EtienneCmb Did you test this? I don't think `v_size` exists in the fragment shader.

EtienneCmb:
@davidh-ssec  Yes, I tested it and both solutions works. `v_size` is also used 6 lines bellow `float size = $v_size + 4.*(edgewidth + 1.5*v_antialias);`

@larsoner oh I see, thx :)

image

larsoner:
@davidh-ssec `v_size` comes from the `Varible` magic during `Markers` instantiation

https://github.com/EtienneCmb/vispy/blob/51dac632cf9d7aca2ef45a67fe75264af3b9ed2c/vispy/visuals/markers.py#L498

PR #1409, opened 2017-12-20
MRG: Use CPU for TextVisual rendering

### LINKED ISSUE #1408: BUG: TextVisual SDF calculation on GPU is too slow
... perhaps we should just use bitmapped fonts (how it was originally implemented!).

Alternatively we could use Cython to compute the SDF (it was also this way at one point...).

@rougier which method did you end up going with for fonts in Glumpy?

larsoner:
Closes #1408.

@davidh-ssec can you give this a try with your WebGL setup to see if it helps?


davidh-ssec:
I will...when I get the WebGL stuff working. I currently don't have an environment on my laptop that has vispy working on a jupyter notebook. I'll see what I can do.

davidh-ssec:
We'll definitely need to update the setup.py and the travis yaml. Cython should not be needed as a dependency by a normal user. It is only needed for regenerating the C files if something changes.

larsoner:
>  Cython should not be needed as a dependency by a normal user. It is only needed for regenerating the C files if something changes.

I'd rather not check in C code unless necessary. AFAIK it's standard practice not to do so (SciPy, scikit-learn, pandas, etc.). Doing so is a bit ugly because it creates unnecessary redundancies.

I don't see Cython as a particularly heavy dependency, at least not compared to the other requirement for using compiled code: getting a usable C++ compiler.

davidh-ssec:
Maybe I'm missing something huge, but if you include cython you are just recompiling the cython code to C at installation and then compiling the C source to a shared library object.

davidh-ssec:
Scipy does not generate the cython if it is being installed from a source tarball: https://github.com/scipy/scipy/blob/master/setup.py#L399

larsoner:
> you are just recompiling the cython code to C at installation and then compiling the C source to a shared library object.

Yes this is how it works. But if you check in both, now any time a change is made to the `pyx` file you have to check in changes to it *and* a `.c` file. In general the Cythonizing is quite fast (relative to the building), too. FWIW Cython+build on my system currently is too fast for me to even notice.

> Scipy does not generate the cython if it is being installed from a source tarball: 

Yes, but the generated files are not checked into the repo. Cython gets run (presumably) by whomever is doing the source tarball release.

I'm not sure how much we gain much by adding these additional methods. I'd rather wait until someone motivated to do it (a wheel builder probably?) has the time and expertise, unless it we know it will break current release methods. Maybe using `Cython` used to be more difficult than it is now? At least with the CIs, it was as easy as adding `conda` to the dependency list, and I see that there are wheels on PyPi.

larsoner:
... at least it has worked for Linux and Windows so far. We might have to wait some hours to see what OSX does.

davidh-ssec:
From running the `text_visual.py` example with a click event to change the size and value of the text and clicking 5 times:

```
$ wc -l glir_text_*
    2266 glir_text_master.txt
     283 glir_text_sdf.txt
    2549 total
```

Give or take a few 100 for any osx driver errors.

larsoner:
Roughly an order of magnitude fewer GLIR commands is nice!

davidh-ssec:
While trying to run this through notebook and display some text I now get:

```
WebGL: INVALID_OPERATION: texImage2D: ArrayBufferView not big enough for request
```

In the browser console. This should be fun to debug.

davidh-ssec:
These conflicts may be part of me upgrading webpack locally and it defaulting to a dev build so it didn't minify the JS. I also probably updated ipywidgets so that could be placing different stuff in the final JS files. I'll look at this later. After that I think this is ready for merge for 0.6. I wanted to make another 0.5.x release with the fixes for new ipywidgets, but I'm not seeing an easy way to cherry picky that functionality out.

davidh-ssec:
@larsoner I had to do a merge with master to resolve the conflicts that this branch had. Didn't want to merge to master by pushing because I want to make sure the CIs pass as expected (hopefully no JS errors/warnings).

PR #1420, opened 2018-02-05
First attempt at PySide2 support

hmaarrfk:
You can build pyside2 yourself or use
https://github.com/fredrikaverpil/pyside2-linux

I mostly searched for PySide and PyQt5 and kinda merged them together.

it ran the `examples/basics/scene/grid.py`, `magnify.py` and `cube.py`. Good enough for a first day. Haven't tested other things yet. But I think it is OK :D.

I had to add a `qWait` function.

I have no idea what `has_uic` is

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -5,18 +5,18 @@
 """
 Base code for the Qt backends. Note that this is *not* (anymore) a
 backend by itself! One has to explicitly use either PySide, PyQt4 or
-PyQt5. Note that the automatic backend selection prefers a GUI toolkit
-that is already imported.
+PySide2, PyQt5. Note that the automatic backend selection prefers
+a GUI toolkit that is already imported.
 
-The _pyside, _pyqt4 and _pyqt5 modules will import * from this module,
-and also keep a ref to the module object. Note that if two of the
-backends are used, this module is actually reloaded. This is a sorts
-of poor mans "subclassing" to get a working version for both backends
-using the same code.
+The _pyside, _pyqt4, _pyside2 and _pyqt5 modules will import * from
+this module, and also keep a ref to the module object. Note that if
+two of the backends are used, this module is actually reloaded. This
+is a sorts of poor mans "subclassing" to get a working version for
+both backends using the same code.
 
-Note that it is strongly discouraged to use the PySide/PyQt4/PyQt5
-backends simultaneously. It is known to cause unpredictable behavior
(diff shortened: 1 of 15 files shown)


vispy/util/tests/test_import.py
```diff
@@ -80,7 +80,9 @@ def test_import_vispy_app2():
     """ Importing vispy.app should not pull in any backend toolkit. """
     allmodnames = loaded_vispy_modules('vispy.app', 2, True)
     assert_not_in('PySide', allmodnames)
+    assert_not_in('PySide2', allmodnames)
     assert_not_in('PyQt4', allmodnames)
+    # assert_not_in('PyQt5', allmodnames) # should PyQt5 be tested against too
```
davidh-ssec: Yes probably.

vispy/app/qt.py
```diff
@@ -17,11 +17,16 @@
 
 
 if qt_lib in ('pyqt4', 'pyside'):
+    # Why is this not a seperate option for pyside.
+    # PySide and PyQt4 are different aren't they?
```
davidh-ssec: Yes this does seem strange doesn't it. If @larsoner or @kmuehlbauer don't have any ideas why this is like this, it seems like it should be split in to a separate if branch.

kmuehlbauer: No idea, seems odd to me too.

larsoner: Agreed this looks like a bug

vispy/app/backends/_qt.py
```diff
@@ -143,7 +149,7 @@ def message_handler(*args):
 
     if qt_lib in ("pyqt4", "pyside"):
         msg_type, msg = args
-    elif qt_lib == "pyqt5":
+    elif qt_lib == ("pyqt5", "pyside2"):  # Is this correct for pyside2?
```
csachs: Don't know more details, but `==` definetly doesn't work here.

`elif qt_lib in ("pyqt5", "pyside2"):`

davidh-ssec: Wow, good catch. @hmaarrfk would you mind changing that?

davidh-ssec:
@hmaarrfk If you don't mind, could you rebase your branch on to the current master and then force push to your branch to start the tests over again and see if everything passes (I committed some fixes to master today)?

hmaarrfk:
@davidh-ssec, cool. I think i just rebased and pushed to my version. Travis should do his work now.

PR #1421, opened 2018-02-07
Utilize the cmap.setter

liuyenting:
When trying to modify `vispy.visuals.volume.Volume` to integrate with multi-color volume, I figured that `cmap.setter` might be better utilized - instead of setting it independently in `__init__` and `method.setter`, which can also potentially alleviate `self._cmap` and `frag['cmap']` being mistmatch.

### DIFF
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -407,7 +407,7 @@ class VolumeVisual(Visual):
         self._need_vertex_update = True
 
         # Set the colormap
-        self._cmap = get_colormap(cmap)
+        self.cmap = cmap
 
         # Create gloo objects
         self._vertices = VertexBuffer()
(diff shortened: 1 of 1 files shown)


liuyenting:
Seems like test failures are not caused by my modifications...

kmuehlbauer:
@liuyenting I think I don't get the idea what you are trying to do. But this won't work this way.

At the time of initialisation the fragment shader isn't yet initialised, so the call ` self.shared_program.frag['cmap'] = Function(self._cmap.glsl_map)` in `cmap.setter` will fail.

I also don't see why `self._cmap` and `frag['cmap']` should mismatch if you use the standard methods. Cam you explain a bit in more detail and maybe setup an example showing your problem?

davidh-ssec:
Additionally, and I could be wrong, when the method changes, the fragment shader changes, which I *think* makes a new shader object so all of the parameters have to be reassigned. Removing the cmap assignment would cause it to never be assigned. Again, I could be wrong.

davidh-ssec:
@liubenyuan If you have no further comments or arguments for why your changes don't break things in the way we've described then I'm going to close this. We can reopen this in the future if needed.

PR #1424, opened 2018-02-11
use passed in variable P instead of gl_PointCoord in demo/cloud.py

rgkirch:
The cloud.py demo can render a few different shapes. The functions are using the global variable gl_PointCoord and ignoring the gl_PointCoord that is passed into the function. This does not affect behavior.

### DIFF
--- a/examples/demo/gloo/cloud.py
+++ b/examples/demo/gloo/cloud.py
@@ -219,7 +219,7 @@ class Canvas(app.Canvas):
         ps = self.pixel_scale
 
         # Create vertices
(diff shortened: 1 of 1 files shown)


davidh-ssec:
Thanks. Our tests are currently failing from problems unrelated to your PR. Once I fix the tests I'll merge this.

rgkirch:
Ok, thanks! By the way, thank you guys so much for this awesome project and the awesome examples. I just found it yesterday and I was able to get what I needed by adapting the cloud.py example.

PR #1431, opened 2018-02-15
A few cleanups for triangulation.py

suzil:


davidh-ssec:
Thanks. In the future I don't think you need to spend time removing debug messages from the vispy code base, even if they are commented out. They may come in handy later on when we need to debug things again. Now that you've tackled so many style issues, how would you feel about trying some of the 0.6 milestone issue: https://github.com/vispy/vispy/milestone/6 ;)

Your help is greatly appreciated. Thank you.

vispy/geometry/triangulation.py
```diff
@@ -719,28 +640,18 @@ def _merge_duplicate_points(self):
...
             dups_arr[dups_arr > j] -= 1
-            #assert not np.any(self.edges[:,0] == self.edges[:,1])
 
         self.pts = self.pts[pt_mask]
 
         # remove zero-length edges
         mask = self.edges[:, 0] != self.edges[:, 1]
         self.edges = self.edges[mask]
 
-    def _distance(self, A, B):
```
davidh-ssec: It seems that some tests still use this method. Could you maybe remove the `_` from the name and patch any tests that use it?

larsoner: if it's currently private let's not make it public without good reason. It's likely it was private because it's not bulletproof / ready for public use

suzil:
> They may come in handy later on when we need to debug things again.

The commented-out code continues to exist in version control.

If the messages are still useful, having them in debug-level logging would be a good alternative though. The author comments on why they chose to use a custom `debug` function instead of using logger:

> Note: using custom #debug instead of logging because there are MANY messages and logger might be too expensive

But I wouldn't think some extra log messages would be that expensive, especially when they'd only be emitted in debug-mode.

davidh-ssec:
@suzil Putting debug messages in for loops in python can be a performance hit. Even when you call `logger.debug` it still has to figure out that you don't want to log that message and it has to format the log message. With the way python logging works it would have to send the message (record) to each handler and parent logger. I agree that this shouldn't take too much time, but it could be significant when worrying about performance. Not sure I would have done it that way, but at least you fixed it now.

I'm not sure I approve of completely removing the distance function and the test that was using it. I was thinking you would add the function back in so the test would pass... @larsoner?

larsoner:
My preference would be to deprecate `triangulation` altogether and rely on SciPy for Delaunay support. But that's perhaps too drastic :)

I guess if `_distance` isn't used internally for anything then it can be removed. We can get it back from the history if we need it. I'm +0 on removing it, so either way is fine with me.

Issue #1437, opened 2018-03-06
VisPy isocurve incompatible with matplotlib 2.2.0


davidh-ssec:
Matplotlib 2.2 is now out and the private contour api `from matplotlib import _cntr` has been removed. We knew this day would come. @QuLogic pointed this thread out on gitter: https://mail.python.org/pipermail/matplotlib-users/2018-March/001313.html

The TL;DR of that is that we can either copy the source from old mpl releases or we could try using skimage. I'm not a huge fan of adding such a huge (optional) dependency on to vispy, but as `isocurve.py` sits right now it uses the `_cntr` classes as an optional dependency so it shouldn't be a huge deal.

http://scikit-image.org/docs/dev/auto_examples/edges/plot_contours.html

As a quick hack I will try to make a PR in the next couple days to switch the import of `_cntr` to a try/except and use the slow contouring method if it isn't found. That should at least let our tests pass and give github users working code.

rougier:
Note that there is GPU only solution but I don't know to what extent "real" contour lines are needed.

kmuehlbauer:
Yes, we also have the IsolineFilter which can be attached to eg. Image Visual.

rougier:
Ah, ok. Sorry for the noise.

kmuehlbauer:
@rougier Did you mean with GPU-method something like in IsolineFilter?

rougier:
I'm not sure actually. I've made them for glumpy (https://github.com/glumpy/glumpy/blob/master/examples/isocurves.py) but I don't remember if I've ported them to vispy...

![isocurves](https://user-images.githubusercontent.com/327203/37111152-48d41e4c-223f-11e8-8211-456ccffca40d.png)

kmuehlbauer:
@rougier Actually we, means you, @larsoner and I discussed this some years ago. This was implemented when we did the spatial-filters using packed floats stuff. Here is the example https://github.com/vispy/vispy/blob/master/examples/basics/scene/contour.py
You surely remember the eye of Mona Lisa.

davidh-ssec:
I'm currently working on replacing the mpl cntr usage with skimage. Is everyone (@larsoner, @kmuehlbauer, @rougier) ok with me completely deprecating the mpl usage?

My thinking is if someone wants new vispy features then they should be ok making updates to all packages including MPL. Then again, they may not have skimage installed.

larsoner:
[...]

PR #1438, opened 2018-03-06
Fix appveyor tests and other deprecation warnings

davidh-ssec:
This PR:

1. Fixes Appveyor tests not being able to download Windows GL libraries by allowing older SSL protocols. I'm not sure how this happens between github and appveyor, but it works now.
2. Removes a warning where unit tests were passing a string of parameters to pytest instead of a list of arguments.
3. Comments out coveralls upload from appveyor. @larsoner recently moved appveyor tests to the shared `vispy` organization instead of running them from his account. The token used for uploading coverage stats was only on his account. He will work on fixing this in the near future.
4. Adds workaround for #1437 but does *not* fix it. It simply uses the slower contour methods that already existed if you didn't have mpl installed.
5. Updates JS dependencies for the notebook widget so that it can properly build on Windows and it should remove the security vulnerability that github was showing to maintainers.
6. Adds `pytest.ini` file to ignore the `js` directory since this was causing some kind of infinite recursion (javascript dependencies, yay) ending in too long of a path for Windows; or at least that's my best guess.


davidh-ssec:
Travis is failing because of the size of the changes (JS updates updated a lot). Going to merge anyway.

The notebook widget also seems to fail to build on travis, but I'll handle that in the future in another PR.

PR #1445, opened 2018-03-11
GLSL shader for iso-filles value in colorbar and meshVisual

sylm21:
For scientific visualisation it's importnat to have the possibility to change continous colormap with vertices data to iso-filled colormap

step 1 :  update shader in colorbar, add PR #1292 and modify an add an example
step 2 : update MeshVisual shader
step 3 : add an example
step 4 : correct flake8 and travis error

Step 5 :  Try to modify MeshVisual vert and frag shader and add iso-filled colormap capability


sylm21:
The example using PyQt5 causes an error because PyQt5 isn't an imported module in Travis.

I don't understand the other mistake in Travis

In my program I have buffers that aren't always used is this serious ?

davidh-ssec:
@sylm21 Are you doing anything that requires PyQt5? PyQt5 is on our list for support for version 0.6, but right now it is not completely supported so it is not used for testing examples. Could you remove the PyQt5 requirement and import PyQt4 for now?

For the buffers, it would be best if you could structure your code so that there aren't buffers that are only used some times?

vim.exe.stackdump
```diff
@@ -0,0 +1,14 @@
+Stack trace:
```
davidh-ssec: This file should be removed.

vispy/visuals/colorbar.py
```diff
@@ -204,6 +216,35 @@ def _prepare_draw(self, view):
...
+    def nband(self):
+        return self._nband
+
+    @nband.setter
+    def nband(self, nband):
+        self._nband = nband
+        self.shared_program.frag['color_transform'] = self.define_fun()
+        self._update()
+
+    def define_fun(self):
```
davidh-ssec: Can you rename this to something private and make it more clear what it is doing? Something like `_get_color_transform()`? Open to other names. Mainly just thinking that this doesn't need to be public and `define_fun` is not very clear.

vispy/visuals/mesh.py
```diff
@@ -171,12 +187,17 @@ class MeshVisual(Visual):
...
+    nband : int
+        the number of band in the glsl banded shader
     **kwargs : dict
         Keyword arguments to pass to `Visual`.
     """
     def __init__(self, vertices=None, faces=None, vertex_colors=None,
                  face_colors=None, color=(0.5, 0.5, 1, 1), vertex_values=None,
-                 meshdata=None, shading=None, mode='triangles', **kwargs):
+                 meshdata=None, shading=None, mode='triangles', banded=False,
+                 nband=10, **kwargs):
```
davidh-ssec: I'm not a huge fan of all this stuff getting added to Visuals that uses a colormap. I think it would better if users accessed stuff like `nband` and `banded` directly from the colormap object. Something like `my_mesh.cmap.nband = 6`.

davidh-ssec: The above assignment assumes that changing a colormap property automatically triggers an update to the visual using it.

vim.exe.stackdump
sylm21: Yes it's an error ...

vispy/visuals/colorbar.py
sylm21: Ok i change this name

vispy/visuals/mesh.py
sylm21: Good idea i try to make it and i change example to PyQt4

kmuehlbauer:
@sylm21 This looks really nice. Good work so far.

The problem with this implementation is, as @davidh-ssec already pointed out, the many changes in the visuals. How about trying to implement this via `filters`.

I'm not totally sure how the `filter` should look like, but just add this to your example file for an impression:

Add somewhere in your example Canvas `__init__`:
```python
from vispy.visuals.filters import IsolineFilter
level = 16
self.iso = IsolineFilter(level=level, width=1., color='white')
```
Add first in `on_resize`-method (this is a bit of a hack, but works):
```python
self.meshes[0].attach(self.iso)
```
This adds GPU isolines to the mesh-sphere. With a little adaption we could enforce contours. Your function should also be possible to implement as `filter` with a bit more ambitious coding than the IsolineFilter, but it should be possible nevertheless. Using this approach we can keep the `mesh-visual` unchanged and also the colorbar visual might be treated similar via `filter`.

Makes sense?

Xref: #1444 as there might be synergies (using 'filters`)

vispy/visuals/mesh.py
sylm21: @davidh-ssec  If in MeshVisual tou change the original _clim function in ''float cmap(float val) { return floor((val - $cmin) / ($cmax - $cmin)*10)/10; }' you don't  have an iso-filled contour because the data is in vert shader.  For the MeshVisual in this case the data must be in the frag shader via v_vert_data = a_vert_data in the vert shader.

sylm21: But put nband and the banded flag in the colormap is a good idea.

sylm21:
I don't know filter, i look at this this ...

sylm21:
@kmuehlbauer, 
I look at IsoLineFilter but my problem is that it converts the color into a float and then convert again in a color is generalizable to all colormap ?

kmuehlbauer:
@sylm21 You are right, `IsolineFilter` works only on the fragment shader. It takes the (perceptual) luminance of a fragment and mixes with the (line) color if around some distinct value.

If we switch the clim-glsl code into Isoline-Filter colortransform then we will get no isolines but filled contours, which is something you want to achieve. But Isoline and also this contour approach will only work on the fragments (ImageVisual) etc..

And now we have to look, how we can use the `filter`-approach to attach your code at the right places (Vertex and Fragment-Shader) for `MeshVisual`. For `ColorBarVisual` you only need the bands(levels) and can work just in the Fragment-Shader (with a filled Contour-filter). Makes sense?

@davidh-ssec I think we can implement @sylm21's needs into a `filter`. WDYT?

@larsoner Is there somewhere a writeup about `filters` (hooks and the like) to get us on the right track here. Also your comments are welcome about implementing filled-contours via `filters`.

larsoner:
> Is there somewhere a writeup about filters (hooks and the like) to get us on the right track here

Not in the code. You can look back (probably via the GitHub `git blame` interface) for the PR that added them to see the discussion. But I would rather look at an example to figure it out. You can look at how picking is done in `scene` (via filters) to see one.

sylm21:
@larsoner in the filter approach, What do you think about the problem of using the color to find a float then recalculate a color (like in IsolineFilter). It's possible and/or it's necessary to inverse colormap ? For me it's OK only for linear colormap via a gray translation. Because I need vertices data information in the frag shader to iso-filled shader

kmuehlbauer:
@sylm21 If I understand the 'filter' stuff, you can also change the vertex shader. So what you are doing should be working using filters. It won't work changing only fragment shader, that's correct. Unfortunately I do not have the cycles atm to dive into this.

sylm21:
@kmuehlbauer , @davidh-ssec  and @larsoner I think that the problem is more important : 

In MeshVisual : the colormap interpol vertex data in the vert shader and in the frag shader you are a linear interpolation in the polygon (triangle) of the vertex color that not link with the  colormap's interpolation. I think but i'm not sure that the good way is interpolate the vertices data in the frag shader to respect the colormap.

larsoner:
That sounds reasonable to me. Not sure if it will make much difference for most use cases but it could for some I suppose

sylm21:
@larsoner no change for linear colormap but for nonlinear or stepped colormap it's very different. I try to modify MeshVisual frag and vert shader.

djhoese:
@sylm21 What other work needs to be done on this PR? Let me know if you're ready for me to re-review this.

PR #1449, opened 2018-03-15
Changed vispy.plot.Fig.__init__() to allow passing 'keys' argument

jimofthecorn:
(previously forced to 'interactive' in call to super().__init__()).

I couldn't figure out how else to pass key events up to a PyQt5 QMainWindow.

### DIFF
--- a/vispy/plot/fig.py
+++ b/vispy/plot/fig.py
@@ -36,10 +36,11 @@ class Fig(SceneCanvas):
     PlotWidget : the axis widget for plotting
     SceneCanvas : the super class
     """
-    def __init__(self, bgcolor='w', size=(800, 600), show=True, **kwargs):
+    def __init__(self, bgcolor='w', size=(800, 600), show=True,
+                 keys='interactive', **kwargs):
         self._plot_widgets = []
         self._grid = None  # initialize before the freeze occurs
-        super(Fig, self).__init__(bgcolor=bgcolor, keys='interactive',
+        super(Fig, self).__init__(bgcolor=bgcolor, keys=keys,
                                   show=show, size=size, **kwargs)
         self._grid = self.central_widget.add_grid()
         self._grid._default_class = PlotWidget


davidh-ssec:
There are event handlers on Canvases and Canvas subclasses. The easiest way to use them is to do either:
```
@canvas.connect
def key_press(event):
    pass
```
Or maybe it's `on_key_press`. You can also subclass the canvas and define your own but that makes more sense for the Scene or gloo level APIs and not so much for the plotting API. There are other ways to define event handlers but they all revolve around using `canvas.events`. Your solution seems like a good fix regardless.

kmuehlbauer:
For this you can use `EventEmitter`. This is how I do it most of the time:

```python
from vispy.util.event import EventEmitter
```
 in the Canvas:

```python
self.key_pressed = EventEmitter(source=self, type="key_pressed")

    self.key_pressed(event)
```

In the main thread:

```python
self.canvas.key_pressed.connect(self.keyPressEvent)

def keyPressEvent(self, event):
    do_something()
```

jimofthecorn:
It appears that using @canvas.connect _adds_ an event handler, rather than replacing it, right? That's the behavior I'm observing. So with the current code base,  I need to create a Fig sub-class in order to replace the existing behavior?I

djhoese:
@jimofthecorn I think you are right. The connect method **adds** callbacks to the event. You can disconnect all callbacks for an event by doing `events.key_pressed.disconnect()` if I'm reading the code correctly. That said I'm ok merging this PR if @kmuehlbauer is ok with it. It doesn't seem to remove any functionality and only gives users more control.

You will need to rebase the current master branch to make the tests pass.

rougier:
@jimofthecorn rebase needed.

jimofthecorn:
Oops. Sorry for the delay...

djhoese:
Thanks!

PR #1457, opened 2018-03-27
Creating	texture-based colormaps for vispy

### LINKED ISSUE #1397: Adopting matplotlib colormaps
As mpl is a de-facto standard, some library developers (I) have written the code converting some foreign (from a file of a proprietary format, which was reverse engineered) colormaps into matplotlib format. I guess that vispy colormaps either should be compatible with matplotlib ones using duck typing, or receive matplotlib colormaps into the colormap constructor to give vispy colormaps.

EDIT: There are a few other colormaps listed in #648 we should add.

jarno-ssec:
Closes #1397.

There is one texture per colormap. Colormap texture implementation has been tested on vispy's examples, 2 simultaneous canvases with dynamically changing colormaps on both canvases and 4 different colormaps on a canvas. 

The implementation uses 2d-texture maps (t-dim,1) instead of 1d-texture maps (t-dim) as 1d-texture map implementation had some visual artifacts in a smooth color gradient near zero.


vispy/color/colormap.py
```diff
@@ -114,17 +116,78 @@ def step(colors, x, controls=None):
...
+                if i == 0:
+                    ifs = 'if (t < %.6f)' % (controls[i+1])
+                elif i == (ncolors-2):
+                    ifs = 'else'
+                else:
+                    ifs = 'else if (t < %.6f)' % (controls[i+1])
+                adj_t = '(t - %s) / %s' % (controls[i],
+                                       controls[i+1] - controls[i])
+                s += ("%s {\n    return mix($color_%d, $color_%d, %s);\n} " %
+                      (ifs, i, i+1, adj_t))
```
larsoner: Let's just remove this old conditional code. It was meant to be a temporary placeholder for the proper texture-based access

vispy/color/colormap.py
```diff
@@ -700,18 +765,13 @@ def __init__(self, limits=(0.33, 0.66, 1.0)):
...
-# colormaps. To get 64 samples
-# the original Viridis colormap data is sampled with a stride of 4 ie [::4].
-#
-# HACK: Ideally, all 256 points should be included, with VisPy generating
-# a 1D texture lookup for ColorMap, rather than branching code.
+# 128 viridis sample size fails on some GPUs 
+# but lowering to 64 samples allows more GPUs to use viridis. 
+# 
+# VisPy has beem updated to use a texture map lookup.
+# Thus, sampling of the Viridis colormap data is no longer necessary.
```
larsoner: Can you update the `viridis` definition (and any others), too, then?

vispy/color/colormap.py
```diff
@@ -986,7 +1046,7 @@ def __init__(self, limits=(0.33, 0.66, 1.0)):
     winter=_Winter(),
     light_blues=_SingleHue(),
     orange=_SingleHue(hue=35),
-    viridis=Colormap(ColorArray(_viridis_data[::4])),
+    viridis=Colormap(ColorArray(_viridis_data[::1])),
```
larsoner: ahh never mind, I see you have. This can just be `(_viridis_data)`, then, no need to slice.

vispy/color/colormap.py
```diff
@@ -114,17 +116,78 @@ def step(colors, x, controls=None):
...
+    ----------
+    colors : list of lists, tuples, ndarrays or ColorArrays
+        The control colors used by the colormap (shape = (ncolors, 4)).
+
+    controls : The list of control points for the given colors. It should be
+        an increasing list of floating-point number between 0.0 and 1.0.
+        The first control point must be 0.0. The last control point must be
+        1.0. The number of control points depends on the interpolation scheme.
+
+    texture_map_data : Numpy array of size of 1D texture lookup data for
```
larsoner: this is not formatted properly

vispy/color/colormap.py
```diff
@@ -138,13 +141,13 @@ def _glsl_mix(controls=None, colors=None, texture_map_data=None):
         The first control point must be 0.0. The last control point must be
         1.0. The number of control points depends on the interpolation scheme.
 
-    texture_map_data : Numpy array of size of 1D texture lookup data for 
-                       luminance to RGBA conversion.
-                       The size of np.float32 array is (LUT_len, 4).
-                       If texture_map_data is not None,
-                       the RGBA data for the array is computed.
+    texture_map_data : Numpy array of size of 1D texture lookup data for
```
larsoner: `: ndarray, shape (n_colors,)`

Also the `controls` should just be `controls : list` and the rest should be description, following more or less [the numpy docstring standard](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)

vispy/color/colormap.py
```diff
@@ -115,16 +118,82 @@ def step(colors, x, controls=None):
...
+    ----------
+    colors : list of lists, tuples, ndarrays or ColorArrays
+        The control colors used by the colormap (shape = (ncolors, 4)).
+
+    controls : The list of control points for the given colors. It should be
+        an increasing list of floating-point number between 0.0 and 1.0.
+        The first control point must be 0.0. The last control point must be
+        1.0. The number of control points depends on the interpolation scheme.
+
+    texture_map_data : Numpy array of size of 1D texture lookup data for
```
larsoner: `: ndarray, shape (n_colors,)`

Also the `controls` should just be `controls : list` and `colors : array-like, shape (n_colors, 4)` and the rest should be description, following more or less [the numpy docstring standard](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)

I'm also not sure why you need `texture_map_data`, this seems redundant with `colors`. There should be a way just to use `colors` and properly use the first `n_colors` values from the `1024` possible ones.

jarno-ssec: For uniformly distributed control points a texture map color interpolation would work, but for non-uniformly distributed control-points you want to compute texture_map_data.

larsoner: Why don't we leave this to the user, and they can do e.g. `colors = ['k', 'k', 'k', 'r', 'g', 'b']` if they want half black then RGB?

davidh-ssec:
@jarno-ssec It looks like there are tests failing. You can see some of the failures on travis here: https://travis-ci.org/vispy/vispy/jobs/359029146#L3486

You may need to update the tests to match your changes. If you haven't already, it would be nice to have tests for your specific changes too.

Lastly, you may want to do a `git rebase` with the current master branch if you haven't recently since there were some fixes for broken tests applied in the last couple months.

vispy/color/colormap.py
```diff
@@ -115,45 +118,101 @@ def step(colors, x, controls=None):
...
     assert (controls[0], controls[-1]) == (0., 1.)
     ncolors = len(controls)
     assert ncolors >= 2
-    if ncolors == 2:
-        s = "    return mix($color_0, $color_1, t);\n"
+    if texture_map_data is None:
+        s2 = ""
+        if ncolors == 2:
+            s = "    return mix($color_0, $color_1, t);\n"
+        else:
```
davidh-ssec: Rather than putting the conditional code back in which I know @larsoner requested be removed, how about you update the tests to not need them? Or wouldn't it be possible to construct the `texture_map_data` from the provided colors and controls?

davidh-ssec:
@jarno-ssec Is there a reason you added the cmap to the LineVisual? I think I would prefer that in a separate pull request in the future. I'm not sure it makes sense to have a colormap in a LineVisual since it doesn't have values associated with it.

davidh-ssec:
Additionally, how does the colormap get used in the LineVisual with the way you have it? When is the cmap used in a shader?

jarno-ssec:
Example `examples/basics/scene/colored_line.py` uses colormap with line visual.

davidh-ssec:
Thanks for the clarification, you are right. Sorry.

jarno-ssec:
My implementation of texturebased-color map seem to have broken basic line examples in `vispy/visuals/tests/test_arrows.py`. So, I reverted back to an older commit without colormap code in `vispy/visuals/line/line.py`. Now, only `examples/basics/scene/colored_line.py` should fail in the Travis build.

davidh-ssec:
@jarno-ssec Could you do some more testing with this? I just tried using this branch for my application (SIFT) and none of the previously working colormaps are working at all.

davidh-ssec:
Hold on @jarno-ssec. I forgot I'm using a custom subclass of ImageVisual. That may be the issue.

vispy/color/colormap.py
```diff
@@ -115,45 +118,79 @@ def step(colors, x, controls=None):
...
+    LUT = texture_map_data
+    LUT_len = texture_map_data.shape[0]
+    LUT_tex_idx = np.linspace(0.0, 1.0, LUT_len)
+
+    for i in range(LUT_len):
+        t = LUT_tex_idx[i]
+        j = find_color_index(controls, t, ncolors)
+        adj_t = (t - controls[j]) / (controls[j+1] - controls[j])
+        LUT[i, 0, :] = _mix_simple(Color(colors[j]).rgba,
+                                   Color(colors[j+1]).rgba, adj_t)
```
davidh-ssec: I may be misunderstanding what this chunk of code is doing, but couldn't this be replaced by `np.interp` in some way (https://docs.scipy.org/doc/numpy/reference/generated/numpy.interp.html)?

At the least, we want to avoid doing all of the `adj_t` and `_mix_simple` stuff individually for each LUT value. With `find_color_index` and the rest of this code the way it is, it's a lot of complexity in python. If we can move it to C via numpy the performance should be much better as things grow.

Lastly, @larsoner @kmuehlbauer your thoughts are appreciated on this too, all of the uses of `linspace` in the colormap code for texture indexes, should they include the end point of 1.0? Right now they do not. Is 1.0 a valid texture coordinate?

larsoner: IIRC yes they span 0 - 1 inclusive

And I agree that `np.interp` should be used if possible

kmuehlbauer: @larsoner Beat me about some seconds :+1:

davidh-ssec: Oops, looks like `endpoint=True` is the default. No changes needed regarding linspace. https://docs.scipy.org/doc/numpy/reference/generated/numpy.linspace.html

kmuehlbauer: I have to look this up again and again too, I always mix this up with np.arange which has a half open interval.

jarno-ssec:
Beore commit 49b1a6d850d00932522039ace03c61c3f64b2d17, the code had tests for multiple different types of incoming colors
```
if(isinstance(colors, ColorArray)):
-                LUT[i,0,:] = _mix_simple(colors[j].rgba, colors[j+1].rgba, adj_t)
-            elif(isinstance(colors[j], str)):
-                LUT[i,0,:3] = _mix_simple(ColorArray(colors[j]).rgb, ColorArray(colors[j+1]).rgb, adj_t)
-            else:
-                LUT[i,0,:len(colors[j])] = _mix_simple(np.array(colors[j]), np.array(colors[j+1]), adj_t)
```
Even that code ran into some issues in some cage. So, I figured that this was more of a procedural way of thinking and changed the code into
`LUT[i,0,:] = _mix_simple(Color(colors[j]).rgba, Color(colors[j+1]).rgba, adj_t)` 
which should work even if Color class is changed to accept new ways to specify the color.

vispy/color/colormap.py
jarno-ssec: Before commit 49b1a6d850d00932522039ace03c61c3f64b2d17, the code had tests for multiple different types of incoming colors
```
if(isinstance(colors, ColorArray)):
                LUT[i,0,:] = _mix_simple(colors[j].rgba, colors[j+1].rgba, adj_t)
            elif(isinstance(colors[j], str)):
                LUT[i,0,:3] = _mix_simple(ColorArray(colors[j]).rgb, ColorArray(colors[j+1]).rgb, adj_t)
            else:
                LUT[i,0,:len(colors[j])] = _mix_simple(np.array(colors[j]), np.array(colors[j+1]), adj_t)
```
Even that code ran into some issues in some cases. So, changed the code into
`LUT[i,0,:] = _mix_simple(Color(colors[j]).rgba, Color(colors[j+1]).rgba, adj_t)` 
which should work even if Color class is changed to accept new ways to specify the color.

davidh-ssec:
You could pass the original `colors` parameter to `ColorArray(colors)` which would do the same thing, but allow you to have a numpy array of all the colors at once. We want to try as much as possible to move for loops in to C land by using numpy for calculations. By using a `ColorArray` you could do stuff like:

```
colors = ColorArray(colors).rgba
# vectorize all the other code
_mix_simple(colors[array_of_js], colors[array_of_js + 1])
```

Or something like that. I'm still not a fan that `.rgba` copies the input data, but I'm more worried about what problems it would cause if we removed the copy.

larsoner:
> I'm still not a fan that .rgba copies the input data, but I'm more worried about what problems it would cause if we removed the copy.

These operations are being done in private code just use the private attribute `ColorArray(colors)._rgba`

jarno-ssec:
I converted `find_color_index` into Matlab-style element wise matrix operation to move the loop in to C land. Although, it made the intent of the code more difficult to understand.

vispy/color/colormap.py
```diff
@@ -116,44 +119,78 @@ def step(colors, x, controls=None):
...
-            s += ("%s {\n    return mix($color_%d, $color_%d, %s);\n} " %
-                  (ifs, i, i+1, adj_t))
-    return "vec4 colormap(float t) {\n%s\n}" % s
-
-
-def _glsl_step(controls=None):
+    assert (texture_map_data is not None)
+
+    LUT = texture_map_data
+    LUT_len = texture_map_data.shape[0]
```
larsoner: this shadows the global variable, one of them should have a different name for code clarity

vispy/color/colormap.py
```diff
@@ -116,44 +119,78 @@ def step(colors, x, controls=None):
...
-        s += """%s {\n    return $color_%d;\n} """ % (ifs, i)
-    return """vec4 colormap(float t) {\n%s\n}""" % s
+    assert (texture_map_data is not None)
+
+    LUT = texture_map_data
+    LUT_len = texture_map_data.shape[0]
+    LUT_tex_idx = np.linspace(0.0, 1.0, LUT_len)
+
+# Replicate indices to colormap texture and color control points.
+# The resulting matrices have size of (LUT_len,len(controls)).
```
larsoner: these comments should be at the same indentation as the code they accompany (what is done here makes it hard to follow the scope, and is inconsistent with other VisPy comments)

vispy/color/colormap.py
```diff
@@ -116,44 +119,78 @@ def step(colors, x, controls=None):
...
+    t2 = np.repeat(LUT_tex_idx[:, np.newaxis], len(controls), 1)
+    controls2 = np.repeat(controls[:, np.newaxis], LUT_len, 1).transpose()
+
+# Perform element wise comparison to find a control point for all LUT colors.
+    bn = np.sum(controls2 >= t2, axis=1)
+    j = np.clip(bn-1, 0, ncolors-1).astype(int).tolist()
```
larsoner: why `tolist` (you can iterate below on `ndarray`)? why `int` (it should already be `int`)?

vispy/color/colormap.py
```diff
@@ -116,44 +119,78 @@ def step(colors, x, controls=None):
...
+    assert (texture_map_data is not None)
+
+    LUT = texture_map_data
+    LUT_len = texture_map_data.shape[0]
+    LUT_tex_idx = np.linspace(0.0, 1.0, LUT_len)
+
+# Replicate indices to colormap texture and color control points.
+# The resulting matrices have size of (LUT_len,len(controls)).
+    t2 = np.repeat(LUT_tex_idx[:, np.newaxis], len(controls), 1)
+    controls2 = np.repeat(controls[:, np.newaxis], LUT_len, 1).transpose()
```
larsoner: Assuming I understand this code, it would be better to avoid `np.repeat` and use broadcasting to implicitly do it for you with something like:
```
bn = np.sum(LUT_tex_idx[:, np.newaxis] >= t2[np.newaxis], axis=1)
```

vispy/color/colormap.py
```diff
@@ -116,44 +119,78 @@ def step(colors, x, controls=None):
...
+
+# Perform element wise comparison to find a control point for all LUT colors.
+    bn = np.sum(controls2 >= t2, axis=1)
+    j = np.clip(bn-1, 0, ncolors-1).astype(int).tolist()
+
+# Copying color data from ColorArray to array-like makes LUT assignment faster.
+    colors_rgba = ColorArray(colors[:])._rgba
+    LUT[:, 0, :] = [colors_rgba[i] for i in j]
```
larsoner: `LUT[:, 0, :] = colors_rgba[j]`

vispy/color/tests/test_color.py
```diff
@@ -212,8 +212,9 @@ def test_colormap_interpolation():
     import vispy.color.colormap as c
     assert_raises(AssertionError, c._glsl_step, [0., 1.],)
 
-    c._glsl_mix(controls=[0., 1.])
-    c._glsl_mix(controls=[0., .25, 1.])
+# dont' call _glsl_mix() directly
+#    c._glsl_mix(controls=[0., 1.])
+#    c._glsl_mix(controls=[0., .25, 1.])
```
larsoner: remove if no longer needed

kmuehlbauer:
@davidh-ssec I did some tests  using examples and some of my programs and did not find suspicious behaviour so far. I'm with @larsoner when it comes to tests. :+1:

jarno-ssec:
Travis runs quite a few tests involving colormaps for pyqt4 backend.

https://travis-ci.org/vispy/vispy/jobs/368896594
examples/basics/plotting/colorbar.py 
examples/basics/plotting/plot_colorbars.py
examples/basics/plotting/volume.py
examples/basics/scene/colorbar_widget.py
examples/basics/scene/colored_line.py
examples/basics/scene/isocurve_updates.py
examples/basics/scene/volume.py
examples/demo/gloo/galaxy.py
examples/demo/gloo/imshow_cuts.py
examples/demo/gloo/imshow.py
examples/demo/scene/picking.py
examples/demo/scene/force_directed_graph.py
examples/demo/scene/picking.py

djhoese:
If I recall correctly the examples aren't tested for accurate output. What do the existing colormap tests test for? Do they include rendering to a buffer and checking the output colors? If so then I am ok with that being enough tests assuming that tests will be/have been added to completely test the new colormap code (with texture data, without texture data, with control points, etc).

jarno-ssec:
Yes, those colormap tests didn't test for the rendering output. They only tested whether the code runs successfully or not.

However, vispy/visuals/tests/test_colorbar.py
https://github.com/vispy/vispy/blob/master/vispy/visuals/tests/test_colorbar.py
tests a colormap as a part of a colorbar by rendering to a buffer and checking the output colors.

Following that test, I started adding new colormap tests into vispy/visuals/tests/test_colormap.py
https://github.com/vispy/vispy/pull/1457/commits/3c144726e74e609446bee09de0cd8852587cd348
So far, I have only been able to run it locally as the image data is stored in a separate repository.

djhoese:
Two options:
1. Modify your tests to not need "expected" images and just check a couple values inside the array or only make a small ImageVisual/Canvas so that the array you are comparing against is small.
2. Add the images to the test data repository in a pull request and we can merge that so that these tests pass.

djhoese:
@jarno-ssec Let me know when you think this is ready for re-review. Looks like there is one example failing in the tests. I restarted two of the other test environments since it seemed they failed from HTTP timeout downloading some packages.

jarno-ssec:
I think that a fail is due to travis acting up.

At least, the difference between an earlier commit, which passed the Travis test,
https://github.com/vispy/vispy/pull/1457/commits/75100f262c6835a157c8f8a3b809d1350025b8c8
and the latest failed commit is only the existence of the vispy/visuals/tests/test_colormap.py file in the latter.

Something similar happened to me earlier.  At that time, I triggered travis to run again without changing the code for a few times and eventually Travis build was a success.

I'll let you know once all the test are successful.

jarno-ssec:
@djhoese I haven't really made any changes into the code lately, but the code is passing Travis tests now.
So, I think this is ready for re-review.

djhoese:
Awesome @jarno-ssec. Thanks. I'll try to look at it early next week. @larsoner @kmuehlbauer if you have any free time to re-review this it would be appreciated. I'd like to merge it next week if possible.

vispy/visuals/tests/test_colormap.py
```diff
@@ -0,0 +1,879 @@
...
+                      [[230, 228, 255, 255],
+                       [233, 232, 255, 255],
+                       [236, 234, 255, 255],
+                       [239, 238, 255, 255],
+                       [242, 241, 255, 255],
+                       [244, 243, 255, 255],
+                       [247, 247, 255, 255],
+                       [250, 249, 255, 255],
+                       [253, 253, 255, 255],
+                       [255, 255, 255, 255]]], dtype=np.uint8)
```
larsoner: these would be better hosted as actual images in `vispy-testing-data`

larsoner: https://github.com/vispy/test-data

vispy/visuals/tests/test_colormap.py
```diff
@@ -0,0 +1,879 @@
...
+def test_colormap_gray():
+    """Test grayscale colormap"""
+    with TestingCanvas(size=size, bgcolor='w') as c:
+        image = Image(cmap='grays', clim=[0, 1], parent=c.scene)
+        shape = (size[0], size[1])
+        np.random.seed(379823)
+        data = np.random.rand(*shape)
+        image.set_data(data)
+        assert_image_match(c.render(), img)
+#        print repr(c.render())
```
larsoner: remove

djhoese:
@larsoner Sorry about all the arrays in the code. I think this was a misunderstanding between me and Jarno. I had hoped that if there were only a couple tests and the images were small enough that adding them as arrays inside the code would just be simpler, but looking at it now there is quite a bit.

@jarno-ssec If you make a PR for the test data repository as @larsoner has suggested I'll merge it right away so that we can get these updated tests working.

vispy/visuals/mesh.py
```diff
@@ -405,6 +408,22 @@ def _update_data(self):
                 colors = self._color.rgba
         self.shared_program.vert['position'] = self._vertices
 
+        if self._cmap.texture_map_data is not None:
```
kmuehlbauer: @djhoese @jarno-ssec 
Just thinking if moving this block (L411-L426) into a dedicated function would make things bit more easy with respect to code duplication (used 4 times in different Visuals). 

Since all code in this block depends only on `self._cmap` we could put the function (eg. get_cmap_texture_lut or similar) into `vispy.color.colormap` and return the `Texture2D` object:

``` python
self._texture_LUT = get_cmap_texture_lut(self._cmap)
self.shared_program['texture2D_LUT'] = self._texture_LUT
```
Maybe the underscore variable isn't necessary at all.

Thoughts?

djhoese: Sounds good to me.

djhoese: Is it more PEP8 friendly to do `texture_lut`? Not a big deal either way.

djhoese:
@kmuehlbauer @jarno-ssec Would it make more sense for this new cmap texture function to be a method on the Colormap class?

kmuehlbauer:
Yes, why not. I did not think that far, but it seems possible.

jarno-ssec:
@djhoese @kmuehlbauer My initial try to convert the new cmap function into a method on the Colormap class didn't work
https://github.com/vispy/vispy/pull/1457/commits/507b784be538322236adcfc1998a2ecbec51bc97

I could try again after the new colormap test images work
https://github.com/vispy/test-data/pull/6

kmuehlbauer:
@djhoese @jarno-ssec  See https://github.com/vispy/test-data/pull/6#issuecomment-394588567, commit tag is missing, so travis fails.

kmuehlbauer:
@djhoese After restarting two jobs, which were breaking spurious, green lights now.

jarno-ssec:
@djhoese I made the new cmap texture function to be a method on the Colormap class. So, both code duplication and PNG test data issues should be addressed now.

vispy/color/colormap.py
```diff
@@ -116,44 +120,82 @@ def step(colors, x, controls=None):
 
 
 # GLSL interpolation functions.
-def _glsl_mix(controls=None):
+def _glsl_mix(controls=None, colors=None, texture_map_data=None):
```
djhoese: The function signature for `_glsl_mix` and `_glsl_step` seem a little odd to me since they have assertion checks on some of the arguments not being None or being a certain way, but they are still keyword arguments. Seems like they should be positional arguments, but I suppose having them the way they are probably matches up better with other user methods and there is no point in breaking things for older user code that may be depending on this...I think.

PR #1461, opened 2018-03-29
Fix #2 for line_draw example to work with current SceneCanvas API

### LINKED ISSUE #2: API Specification: event system

### LINKED ISSUE #1455: examples/basics/visuals/line_draw.py
Hi,
I can't run the example line_draw.py. Someone can try if it is only my problem or if there is an issue?

```

  File "/home/........../basics/visuals/line_draw.py", line 178, in <module>
    win = Canvas()
  File "/home/........../basics/visuals/line_draw.py", line 154, in __init__
    self.pos = np.zeros((n, 3), dtype=np.float32)
  File "/home/........../lib/python3.6/site-packages/vispy/util/frozen.py", line 16, in __setattr__
    'attributes' % (key, self))
AttributeError: 'pos' is not an attribute of class <Canvas (PyQt5) at 0x7f4334088f98>. Call "unfreeze()" to allow addition of new attributes

Process finished with exit code 1

```

fschill:
Fixes #1455 

(maybe compare with @davidh-ssec version to select a fix)


davidh-ssec:
I had never heard of `SceneCanvas.node_transform`. Any idea if that is the recommended way?

Also, if we're being picky I think you could limit the `pos_scene[:3]` before passing it to the lower functions. Maybe.

I guess the only other important difference/decision is what the best way to handle mouse events is. If the canvas has to do something complicated to connect them to the line visual then maybe yours is the best. If not, then maybe it would be best for the event to be passed to the handler methods on the line visual. Either way, yours is nice because it does the event transform in one place.

davidh-ssec:
Looks like you have some lines that are too long (see travis).

fschill:
oh great, that again... I honestly don't know much about Travis. Is there a simple way to run syntax/formatting checks locally, instead of going through github and waiting 15 minutes? The Auto-formatting in PyCharm usually takes care of things, but it seems Travis doesn't like visually aligned indenting (which I'd prefer for clarity...). How are continuation lines supposed to be indented?

fschill:
thanks, that was helpful. flake8 is at least happy locally now, let's hope it passes. For the future I guess I'll try to find out how to change the rules in PyCharm to the flake8 expectations.

PR #1463, opened 2018-03-31
Add ShadingFilter for meshes by separating it from MeshVisual

asnt:
Implement mesh shading as a filter

Relies on #1462 for ensuring regular mesh indexing mode.

The flat shading is implemented without duplication of data (as was previously the case) as described there: https://stackoverflow.com/a/40101324/2528098


kmuehlbauer:
@asnt Sorry that this still sits around, as well as #1462 and #1444. 

Do you think you can rebase/combine your PRs on top of latest master (I wasn't successful so far). Efforts are there to release vispy 0.6 in near time. Would be a nice addition.

asnt:
@kmuehlbauer Great! I'll have a look in the coming days.

examples/basics/scene/mesh_shading.py
```diff
@@ -0,0 +1,63 @@
...
+vertices, faces, normals, texcoords = read_mesh(args.mesh)
+
+canvas = scene.SceneCanvas(keys='interactive', bgcolor='white')
+view = canvas.central_widget.add_view()
+
+view.camera = 'arcball'
+view.camera.depth_value = 1e3
+
+mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1))
+mesh.set_gl_state('translucent', depth_test=True, cull_face=True)
```
djhoese, line 24: Is this needed? The only difference from what the MeshVisual does is `cull_face`, but in the case of this example it doesn't seem to change the resulting visual, at least on my machine.

djhoese:
I mentioned this in other places, but this PR should rewrite the added Filter to use the base Filter class that was added after this PR was first created. After the other PRs are merged (mainly #1462) then I can rewrite the filter (hopefully).

examples/basics/scene/mesh_shading.py
asnt: I think it makes a difference with transparency (`alpha < 1` in the color), which I was playing with while testing. So, indeed, this seems redundant.

djhoese: That makes sense. i tried setting the alpha for the mesh to 0.5 and `cull_face=True` mostly did what it should have. I actually like the effect of `cull_face=False` with transparency :wink: 

But anyway, yeah it isn't needed for this. Whoever updates this PR next can remove it.

vispy/visuals/isosurface.py
```diff
@@ -42,6 +43,9 @@ def __init__(self, data=None, level=None, vertex_colors=None,
         self._recompute = True
         self._update_meshvisual = True
 
+        shading = kwargs.pop('shading', None)
```
djhoese, line 46: Can't we add this kwarg to the `__init__` declaration?

vispy/visuals/mesh.py
```diff
@@ -166,42 +88,20 @@ class MeshVisual(Visual):
...
-
-        else:
-            Visual.__init__(self, vcode=vertex_template,
-                            fcode=fragment_template,
-                            **kwargs)
-
-        self.set_gl_state('translucent', depth_test=True,
-                          cull_face=False)
+                 face_colors=None, color=(1, 1, 1, 1), vertex_values=None,
+                 meshdata=None, mode='triangles', **kwargs):
```
djhoese, line 98: I didn't see this outline anywhere in the PR, but why did the default color change? Pure white seems risky to work for all examples that we have. And `shading` was removed, are users expected to add the filter themselves? Would it be possible to add the filter here if `shading` is specified? Or are there some other side effects to this?

asnt: Indeed, good point. The removed `shading` param breaks backward compatibility and is maybe not what we want.
I initially assumed users would add the filters manually so I removed these options from the MeshVisual.
I don't see any reason why it wouldn't work to add the shading filter here depending on `shading`.

vispy/visuals/isosurface.py
asnt: Given the comment below, this change wouldn't be needed anymore. The shading filter would be part of the `MeshVisual` the `shading` param forwarded to it in the `kwargs`. (As before the changes, if I understand well.)

vispy/visuals/mesh.py
djhoese: We could keep the shading keyword in the MeshVisual but print a deprecation warning and point to the example to show how to attach the ShadingFilter to a created MeshVisual. Then in the future we could remove it.

Either way it is probably a good idea if we add to the MeshVisual docstring the various filters that apply to the MeshVisual. Thoughts? @kmuehlbauer @larsoner

larsoner: Indeed the behavior should be documented. I'm fine with leaving the option in the `mesh` constructor, both for backward compat and going forward. It's not too hard to leave it in there (it should only be a one or two-liner, right?) so little maintenance cost. And in the docs for Mesh you can always tell people that for advanced/other options, they should use a filter instead.

kmuehlbauer: @djhoese I would second @larsoner.

vispy/visuals/filters/mesh.py
```diff
@@ -105,3 +107,232 @@ def _update_texcoords_buffer(self, texcoords):
...
+        self._normals.set_data(normals, convert=True)
+
+    def _attach(self, visual):
+        # vertex shader
+        vert_post = visual._get_hook('vert', 'post')
+        vert_post.add(self.vertex_program)
+        render2scene = visual.transforms.get_transform('render', 'scene')
+        visual2scene = visual.transforms.get_transform('visual', 'scene')
+        scene2doc = visual.transforms.get_transform('scene', 'document')
+        doc2scene = visual.transforms.get_transform('document', 'scene')
```
djhoese, line 322: This is a big problem when this is used outside of the `SceneCanvas`. The `transforms` system only exists with the `SceneCanvas`. Except now that I'm looking at the code...this is the way it was in the original Visual so...this never work in a regular Canvas? Or am I missing something?

asnt: I am not sure about this one. I've only worked with the SceneCanvas for meshes.

vispy/visuals/mesh.py
asnt: @djhoese I'd also go for @larsoner's proposal. Keeping the `shading` keyword in the constructor for backward compatibility and ease of use, while allowing customisation with `shading=None`.

vispy/visuals/gridmesh.py
```diff
@@ -49,9 +50,13 @@ def __init__(self, xs, ys, zs, colors=None, shading='smooth',
         self.__vertices = None
         self.__meshdata = MeshData()
 
-        MeshVisual.__init__(self, shading=shading, **kwargs)
+        self._shading_filter = ShadingFilter(shading=shading)
+
+        MeshVisual.__init__(self, **kwargs)
         self.set_data(xs, ys, zs, colors)
```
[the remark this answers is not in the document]
asnt: Tentative implementation in https://github.com/asnt/vispy/tree/mesh-shading-filter-dynamic
that seems to work.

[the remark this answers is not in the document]
djhoese: Yes that would great @asnt. Besides that I think the only remaining thing (besides fixing any other failing tests) is to rewrite the filter to use the base filter classes. I can do that after you make the updates you mentioned.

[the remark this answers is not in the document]
asnt: @djhoese I did both.

asnt:
Some issues I've spotted:
- In the mesh_shading.py example, the shading fails if the camera is too close to the object. The backside is illuminated instead of the front side (object appears dark). Maybe a shift in the camera/light position to account for.
- In MeshVisual, properties related to the shading should probably be removed as they are exposed in the filter. Should they be accessed like this `mesh.shading_filter.shininess`? (The mesh_texture.py example has to be updated accordingly.)

djhoese:
Wow, lots of work done today. Nice job.

For the removal of the properties in the mesh visual related to shading, I'm ok with this. It isn't great for backwards compatibility but I think we have to draw the line somewhere on what to support and what not to support. This is another chance for @kmuehlbauer @larsoner to disagree though and I don't feel too strongly about it anymore. I'm not sure how widely used the shading was used since I think (?) it is a relatively new feature in vispy anyway.

larsoner:
I haven't looked in depth but IIRC shading never worked properly or flexibly, if this solves that problem then it's worth the backward compatibility break

djhoese:
Sounds about right.

@asnt the failing travis is not your fault and is caused by a new version of flake8. I'll fix it in master this weekend hopefully. Your first bullet point is for fixing the issue when the camera is close to the mesh and shading breaks down. I thought you had this checked, as in it was fixed, but it isn't anymore. Is this fixed? What do you think it would take to fix it?

asnt:
> Your first bullet point is for fixing the issue when the camera is close to the mesh and shading breaks down. I thought you had this checked, as in it was fixed, but it isn't anymore. Is this fixed? What do you think it would take to fix it?

@djhoese It isn't fixed yet, I mistakenly checked it. Not sure about the cause of the issue. I am going to have a closer look.

asnt:
@djhoese The example is fixed. (See bullet point for details.)
I haven't notice any further obvious issues.

examples/basics/scene/mesh_texture.py
```diff
@@ -28,7 +28,7 @@
 
 shading = None if args.shading == 'none' else args.shading
 mesh = Mesh(vertices, faces, shading=shading, color='white')
-mesh.shininess = 1e-3
+mesh.shading_filter.shininess = 1e+1
```
djhoese, line 31: Why the change in shininess?

examples/basics/visuals/tube.py
```diff
@@ -56,7 +56,7 @@
 
 l4 = scene.visuals.Tube(points4,
                         color=colors,
-                        shading='flat',
+                        shading='smooth',
```
djhoese, line 59: I thought I asked this before, but why the change in shading here?

examples/basics/scene/mesh_texture.py
asnt: Good point. In short, `shininess` breaks backwards compatibility. `shininess_new = 1 / shininess_previous`.

I adapted to the Phong shading implementation to follow the [book formulas](https://en.wikipedia.org/wiki/Phong_reflection_model#Description) more closely. The `shininess` is now the Phong specular exponent `alpha`. Before, `alpha=1/shininess`. This way, I found that no hard-coded factor is needed anymore, [as before](https://github.com/vispy/vispy/blob/ad1f3ffbde688499d4922597c3b639beba6eea70/vispy/visuals/mesh.py#L92). Subjectively, the shading effect now seems better behaved to me.
So in the proposed changes, `shininess` takes values in `[0, infinity]`, with higher values being more mirror-like surfaces (smaller specular highlight).

examples/basics/visuals/tube.py
asnt: `flat` doesn't work with the mesh in `mode=lines`.
See the bottom-right tube in this image:
tube_lines_shading
Left: `shading=smooth`. Right: `shading=flat`.

I think this is because flat shading is implemented directly in [this way](https://github.com/vispy/vispy/blob/5002cb117810a476f8ce29240de849746d88bdc4/vispy/visuals/filters/mesh.py#L64). That is, the normals are computed in the shader, assuming that we are drawing faces (I suspect `dFdx` and `dFdy` do not make sense in `mode=lines`).
This was initially introduced to allow flat shading with indexed buffers (and limitations of the GLSL version, I think). However, at the moment, this is not needed since we assume expanded buffers of faces (where we could pass a single normal direction for each face instead). It might still be useful in the future but there are also other approaches, e.g. the `flat` [type qualifier](https://www.khronos.org/opengl/wiki/Type_Qualifier_(GLSL)) (search for `flat` in the link) but not in earlier GLSL, I think.
I'd have to double-check these claims in practice. If they hold, we could change the content of the normals buffer instead of the current proposed approach for `flat` shading to work in `mode=lines`. Do we absolutely need `flat` shading in `lines` mode if smooth works?

djhoese: It depends what worked before, but in the end I think whatever works is good. In the long term if there is an approach to all this that makes every shading style work with every Mesh then great. If that approach causes performance problems or months of development time then it isn't worth it (not for free at least).

I can't say what users will want to use this for so until someone requests it or adds the functionality I think we stick with what you've done.

djhoese:
Once #1983 is merged, I can merge master back into this PR to fix the existing merge conflicts and then we can get going on merging this.

vispy/visuals/mesh.py
```diff
@@ -165,48 +88,47 @@ class MeshVisual(Visual):
     meshdata : instance of MeshData | None
         The meshdata.
     shading : str | None
-        Shading to use.
+        Shading to use. This uses the
+        :class:`ShadingFilter <vispy.visuals.filters.mesh.ShadingFilter>`
```
djhoese, line 92:
```suggestion
        :class:`~vispy.visuals.filters.mesh.ShadingFilter`
```

djhoese:
I just added a relatively simple test that does some basic checks of running this code by using the `shading` keyword argument to the MeshVisual. Copying the test to a separate script I get the following images (just for sanity sake):

#### None

image

#### flat

image

#### smooth

image

djhoese:
Don't ask me how long it took me to write this test. If no one has any objections, I'll merge this PR this afternoon.

djhoese:
@kmuehlbauer Just noticed coveralls doesn't report on PRs. Did it ever? Do we want it to (comment and/or "check" status)?

kmuehlbauer:
@djhoese Also not sure, if it ever did. It would be nice. We can just create a dummy pr and see if it works.

djhoese:
Thanks @asnt!

PR #1464, opened 2018-04-03
Enable EGL backend for headless rendering

landersson:
Hi,

Here's a few changes that re-enables the EGL backend and makes it usable for accelerated off-screen rendering. This makes my application, as well as vispy/examples/demo/gloo/offscreen.py able to utilize truly headless rendering using EGL, using both standard OpenGL and OpenGL ES. Only tested with recent NVIDIA drivers.

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -497,7 +497,14 @@ class GlirParser(BaseGlirParser):
             self.capabilities['gl_version'] = gl.glGetParameter(gl.GL_VERSION)
             self.capabilities['max_texture_size'] = \
                 gl.glGetParameter(gl.GL_MAX_TEXTURE_SIZE)
-            this_version = self.capabilities['gl_version'].split(' ')[0]
+            this_version = self.capabilities['gl_version'].split(' ')
+            if this_version[0] == "OpenGL":
+                # For OpenGL ES, the version string has the format:
+                # "OpenGL ES <version number> <vendor-specific information>"
+                this_version = this_version[2]
+            else:
+                this_version = this_version[0]
+
             this_version = LooseVersion(this_version)
             if this_version < '2.1':
                 if os.getenv('VISPY_IGNORE_OLD_VERSION', '').lower() != 'true':
--- a/vispy/app/backends/_egl.py
+++ b/vispy/app/backends/_egl.py
@@ -128,12 +128,6 @@ class ApplicationBackend(BaseApplicationBackend):
 
 # ------------------------------------------------------------------ canvas ---
 
-COMMON_ATTRIB_LIST = [egl.EGL_RED_SIZE, 8,
-                      egl.EGL_BLUE_SIZE, 8,
-                      egl.EGL_GREEN_SIZE, 8,
-                      egl.EGL_ALPHA_SIZE, 8,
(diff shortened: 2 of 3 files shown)


vispy/gloo/glir.py
```diff
@@ -497,7 +497,14 @@ def _gl_initialize(self):
             self.capabilities['gl_version'] = gl.glGetParameter(gl.GL_VERSION)
             self.capabilities['max_texture_size'] = \
                 gl.glGetParameter(gl.GL_MAX_TEXTURE_SIZE)
-            this_version = self.capabilities['gl_version'].split(' ')[0]
+            this_version = self.capabilities['gl_version'].split(' ')
+            if this_version[0] == "OpenGL":
+                # For OpenGL ES, the version string has the format:
+                # "OpenGL ES <version number> <vendor-specific information>"
+                this_version = this_version[2]
```
davidh-ssec: Is this difference documented in the official standards? Do you have a link?

landersson: Yes, the ES version format string is specified in the description for GL_VERSION in the ES2 reference:

https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glGetString.xml 

OpenGL<space>ES<space><version number><space><vendor-specific information>

davidh-ssec:
@larsoner Could you review this if this is something you are familiar with? I don't really have any experience with this.

larsoner:
I don't remember much about it. I would say that it's important to get (or at least try to get) Travis tests working for this backend, since that should provide an excellent headless-rendering test case.

davidh-ssec:
@larsoner And how should @landersson go about enabling the tests for this backend?

larsoner:
Take a look at the testing framework for how each backend is listed to be tested, I can't remember how offhand

landersson:
Ok, I will try to look into this when I get a spare moment. Just enabling tests may be as simple returning True for "testable" in _egl.py, but the fact that it's fully off-screen may introduce other complications.

djhoese:
@landersson Since all of the tests are drawn to a fake display it might not be terrible.

bhack:
Any news?

landersson:
After making the changes in this PR, accelerated rendering into an offscreen buffer worked fine for me with an NVIDIA GPU. I never got around to adding the tests etc that was requested though, so I guess it hasn't been merged.

bhack:
Too bad that this change is not in. Also glumpy has this problem unsolved https://github.com/glumpy/glumpy/issues/100

djhoese:
@landersson There have been some changes to the testing files (.travis.yml) since this was first created. If you could merge the master branch in to your master branch, make the tests "testable" as you said in an earlier comment, and then push your changes to your master branch again, this should be mergeable.

landersson:
@djhoese I wish I had some spare time to spend on this but I'm insanely busy at the moment (and for the foreseeable future), so I most likely wont be able to.

@bhack, if you need this merged, feel free to adopt my PR, enable the tests and go from there.

bhack:
Ok. Cause there are some server/cloud/containers contexts where GLX is not available.
See https://github.com/nvidia/nvidia-docker/wiki/Frequently-Asked-Questions#is-opengl-supported

PR #1480, opened 2018-05-07
Add new example for mouse editing/drawing of shapes

fschill:
Following up on the line_draw example, this is a slightly more advanced example which lets the user draw rectangle and ellipse visuals interactively with the mouse, resize them using control points, move and delete them. It demonstrates the use of visual_at, adding and modifying visuals in a scene, mouse handling and transformations. It also has shapes at fixed positions (not zooming/scrolling with the scene) which are used as "buttons" to select the edit mode. It is meant as a starting point for more complex interactive mouse-driven vector drawing applications.

I have some questions for the maintainers: 
- currently this example is in examples/basics/visuals, because that's where the line_draw example was, and where I found the line_plot and polygon_visual examples that served as inspiration. However, thinking about it, it might be more at home in examples/basics/scene, as it is heavily using the scene infrastructure. Should I move it there?

- I found the performance a bit slow once I add 20-30 shapes. Dragging/zooming, etc. feels much more sluggish than with fewer objects, which should be purely a vispy/scene thing (nothing to do with my example code). I'm not quite sure why that is - given the usual speed of vispy it shouldn't be an issue at all with such few objects. I noticed there is a still-open pull request that seems to address this issue but haven't had time to look into it. I did some profiling (graph attached), but I don't know enough about vispy internals to figure out where the time goes. A lot seems to be going on in the event handling. Maybe it is useful to you? I'm running on a reasonably powerful i5 with a GeForce GTX660 on ubuntu, and vispy generally performs really well on this machine.
[shape_draw_profiling.pdf](https://github.com/vispy/vispy/files/1981227/shape_draw_profiling.pdf)

### DIFF
--- a/examples/basics/scene/shape_draw.py
+++ b/examples/basics/scene/shape_draw.py
@@ -1,0 +1,350 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018, Felix Schill.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+
+"""
+Simple demonstration of drawing and editing shapes with the mouse
+This demo implements mouse picking on visuals and markers using the
+vispy scene and "visual_at" mechanism.
+Left mouse button on empty space creates new objects. Objects can be
+selected by clicking, and moved by dragging. Dragging control points
+changes the size of the object.
+Vispy takes care of coordinate transforms from screen to ViewBox - the
+demo works on different zoom levels.
+Lastly, additional objects are added to the view in a fixed position as
+"buttons" to select which type of object is being created. Selecting
+the arrow symbol will switch into select/pan mode where the left drag
+moves the workplane or moves objects/controlpoints.
+"""
+
+import numpy as np
+
+from vispy import app, scene
+from vispy.color import Color
+
+
+class ControlPoints(scene.visuals.Compound):
+    def __init__(self, parent):
+        scene.visuals.Compound.__init__(self, [])
+        self.unfreeze()
+        self.parent = parent
+        self._center = [0, 0]
+        self._width = 0.0
+        self._height = 0.0
+        self.selected_cp = None
+        self.opposed_cp = None
+
+        self.control_points = [scene.visuals.Markers(parent=self)
+                               for i in range(0, 4)]
+        for c in self.control_points:
+            c.set_data(pos=np.array([[0, 0]],
+                                    dtype=np.float32),
+                       symbol="s",
+                       edge_color="red",
+                       size=6)
+            c.interactive = True
+        self.freeze()
+
+    def update_bounds(self):
+        self._center = [0.5 * (self.parent.bounds(0)[1] +
+                               self.parent.bounds(0)[0]),
+                        0.5 * (self.parent.bounds(1)[1] +
+                               self.parent.bounds(1)[0])]
+        self._width = self.parent.bounds(0)[1] - self.parent.bounds(0)[0]
+        self._height = self.parent.bounds(1)[1] - self.parent.bounds(1)[0]
+        self.update_points()
+
+    def update_points(self):
+        self.control_points[0].set_data(
+            pos=np.array([[self._center[0] - 0.5 * self._width,
+                           self._center[1] + 0.5 * self._height]]))
+        self.control_points[1].set_data(
+            pos=np.array([[self._center[0] + 0.5 * self._width,
+                           self._center[1] + 0.5 * self._height]]))
+        self.control_points[2].set_data(
+            pos=np.array([[self._center[0] + 0.5 * self._width,
+                           self._center[1] - 0.5 * self._height]]))
+        self.control_points[3].set_data(
+            pos=np.array([[self._center[0] - 0.5 * self._width,
+                           self._center[1] - 0.5 * self._height]]))
+
+    def select(self, val, obj=None):
+        self.visible(val)
+        self.selected_cp = None
+        self.opposed_cp = None
+
+        if obj is not None:
+            n_cp = len(self.control_points)
+            for i in range(0, n_cp):
+                c = self.control_points[i]
+                if c == obj:
+                    self.selected_cp = c
+                    self.opposed_cp = \
+                        self.control_points[(i + n_cp / 2) % n_cp]
+
+    def start_move(self, start):
+        None
+
+    def move(self, end):
+        if not self.parent.editable:
+            return
+        if self.selected_cp is not None:
+            self._width = 2 * (end[0] - self._center[0])
+            self._height = 2 * (end[1] - self._center[1])
+            self.update_points()
+            self.parent.update_from_controlpoints()
+
+    def visible(self, v):
+        for c in self.control_points:
+            c.visible = v
+
+    def get_center(self):
+        return self._center
+
+    def set_center(self, val):
+        self._center = val
+        self.update_points()
+
+
+class EditVisual(scene.visuals.Compound):
+    def __init__(self, editable=True, selectable=True, on_select_callback=None,
+                 callback_argument=None, *args, **kwargs):
+        scene.visuals.Compound.__init__(self, [], *args, **kwargs)
+        self.unfreeze()
+        self.editable = editable
+        self._selectable = selectable
+        self._on_select_callback = on_select_callback
+        self._callback_argument = callback_argument
+        self.control_points = ControlPoints(parent=self)
+        self.drag_reference = [0, 0]
+        self.freeze()
+
+    def add_subvisual(self, visual):
+        scene.visuals.Compound.add_subvisual(self, visual)
+        visual.interactive = True
+        self.control_points.update_bounds()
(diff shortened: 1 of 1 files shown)


djhoese:
Awesome work. Thanks for the example.

1. Yes please move it to the basic scene examples. I could also see it going in the `demo/scene` examples. I'll let others comment on that if they feel like it.
2. This is a known issue. Scene degrades quickly when a lot of visuals are added. The Scene is nice for automatically handling cameras, transforms, and a hierarchy of updates, however performance is not it's strong suit. The best solution for these cases where adding a lot of visuals is needed is to try to combine multiple in to one Visual object. This is what `collections` are for which were originally added by @rougier. However, right now there is no Visual wrapper around them so they can't be used out of the box in the Scene.

fschill:
1. ok - I'll wait a bit for other opinions. Actually, the demo directory seems a bit empty, and there are currently imho many "cooler" examples in "basics". Coming from the outside, I might expect the most impressive examples in "demo" (e.g. the ones featured on the website gallery). I don't really mind if my example goes in demo or basics, but it definitely would make more sense in "scene" than in "visuals". 

2. thanks for letting me know - it was a bit unexpected. Do you have any idea why it is and how to fix it? How likely is it that this will improve? The mouse drawing is a side project for me, but something I've wanted for a long time, and I'm planning to build a bigger program around this. If there are severe performance issues with scene, I should address this early on in case it requires a fundamental rethinking how to handle the low-level stuff. Just browsing through other pull-requests, I noticed that this came up before, will do a bit of reading there as well.

djhoese:
In my opinion, besides newer OpenGL features that I want for my own work (tessellation shaders, etc), the performance of the Scene is my (our?) number one priority. Maybe add documentation and consistency of interfaces just below that. But that priority list comes after resolving bugs and getting the number of issues/PRs to a minimum (I'm still a 'new' maintainer in my opinion).

The main performance problem with the Scene is that each Visual is a separate GL Program. These can only be drawn one at a time (that's how GL works) so the more you add, the more time it takes to completely draw the Scene. However, if you draw thousands of shapes in one shader program, that's only one program.

djhoese:
That is pretty much what collections are supposed to be. I haven't looked at the internals of the collections as they are now, but it is very possible (if done correctly) that they would replace the individual shape visuals (Polygon, Rectangle, etc) but it depends. The issues involved in all of this are a pretty common GL issue. For example, you can draw 100 separate triangles, specifying the 3 vertex coordinates for each one. Or you can use a [triangle strip](https://en.wikipedia.org/wiki/Triangle_strip) and specify a lot less vertices. It's not the same issue, but similar, at least that's how I like to think of this issue. 😉

fschill:
thanks - I'll have a closer look at collections, and if I can already use them somehow or not. I was considering if it is better for a drawing app to take care of the scene handling itself, and only use vispy for the rendering backend (as I might also want to render to other backends, like svg, etc.). Then I could bypass scene - but the views, cameras, visual_at, etc. just worked so well and made life rather easy... :)
Re. OpenGL performance - I do remember in the "old days" to use triangle strips, and more importantly, precompiled display lists, which made quite a difference. However, I feel there is something else afoot here, as I'd expect this sort of thing to make the difference between drawing 100k triangles or 1 million triangles, but not 3 rectangles vs. 30 rectangles, with one CPU core at 100% while panning. I don't know how it can possibly take this long with such a small number of shapes, and I'd suspect some hidden loops, or lots of redundant calls or memory allocation/deallocation somewhere. Just to be clear, I'm not criticizing vispy, it's an amazing project, just trying to get to the bottom of this.

djhoese:
The two performance cases aren't exactly the same. In the triangle case you are poorly describing the problem to the GPU in one program versus intelligently describing it in one program. In the shapes case (visuals) you are telling the GPU to draw one shape, then telling it to draw another, then another, BUT each shape requires communication between the CPU and the GPU (individual GL programs) instead of one "description" going to the GPU once.

While I would guess that the number of programs is the main issue, having multiple visuals complicates this as well since there are now more nodes to work through. Some bugs in vispy may also hurt performance but *should* be minimized by now (shaders being recompiled, arrays being needlessly copied, etc).

fschill:
At the moment I'm really not sure where the problem is. If I find time, I'll try to do some more benchmarking, but it's quite the steep task for me to fully understand the underlying system without being too familiar with vispy. Probably a good learning experience :) It's just my gut feeling right now that drawing only 10-20 simple shapes just cannot be that slow on modern hardware, especially considering how fast everything else is. I'll try to compare a few different ways of drawing these shapes (without scene, with scene, etc.) to see if that brings some clarity. It could indeed be the individual programs - I remember once running CUDA code on Windows, and each function call apparently came with a 1ms penalty due to the driver architecture, wiping out all the GPU advantage (however, none of that happened under Linux, so it's not that, but it could be something similar). 
In any case, I think this little demo is good as is - the performance is another topic, and doesn't affect the demo itself very much, as nobody will use it for massive diagrams.

fschill:
I've done some testing of the frame rate in different scenarios, and the results are a bit surprising. For testing, I cut down my example, and simply filled the canvas with 100 rectangles (10x10 grid). I then tested mouse panning of the canvas, as well as dragging/resizing one rectangle, and have vispy print the frame rate. Here are the results:

border_method="gl", radius=0:
Fill="orange"+border="white":  panning: 3.4 fps, dragging/resizing:  6.0 fps
Fill="orange",border=None:   panning: 5.3 fps, dragging/resizing: 16.2 fps
Fill="None", border="white":  panning: 4.0 fps, dragging/resizing: 8.6 fps

border_method="agg", radius=0:
Fill="orange"+border="white":  panning: 2.7 fps, dragging/resizing:  4.2 fps
Fill="orange",border=None:   panning: 5.3 fps, dragging/resizing: 16.2 fps
Fill="None", border="white":  panning: 3.1 fps, dragging/resizing: 5.5 fps

border_method="gl", radius=3:
Fill="orange"+border="white":  panning: 3.3 fps, dragging/resizing:  6.0 fps
Fill="orange",border=None:   panning: 5.3 fps, dragging/resizing: 16.1 fps
Fill="None", border="white":  panning: 3.9 fps, dragging/resizing: 8.4 fps

Fill=None+border=None: panning: 6.6 fps  (dragging not possible, objects not visible)
and again without the GridLines, exactly the same:
Fill=None+border=None: panning: 6.6 fps 

What I find baffling here: I thought panning should be faster, as none of the object positions change, and only the scene transform is different. That dragging/resizing is so much faster surprises me - as far as I can tell, all objects are redrawn anyway, and I made sure to resize the object to cover all other objects, etc. which gives the same performance. 
Second surprising thing is that line drawing is so much slower than the polygon fill. I verified that the fast "gl" method is used (and indeed the "agg" is a bit slower, but not much.
Lastly, I tested a corner radius, as I would expect the re-triangulation at resize to take more time, and it would presumably also result in more triangles and line segments. However, this doesn't seem to affect the performance at all (below measurement noise). It appears that what I thought would be "heavy lifting" (re-triangulating, updating mesh, etc.) comes almost for free, which makes it all the more surprising that rendering those 100 rectangles takes so long (particularly when panning, where really nothing should change other than the scene offset - even when all objects are invisible, and should not even contain a mesh or line).

I have currently no explanation that would make sense for these observations. I hope however that it may be useful for others who are more familiar with the render pipeline, and maybe guide the search for performance improvements.

djhoese:
Regarding the name of the examples, maybe we should start an "interactive" section of examples compared to ones that are just animations. Not sure how many examples would be in each group. @larsoner @kmuehlbauer thoughts on this?

@fschill Another thing to look in to is how many times the canvas' `update` method is called and how many mouse events are being sent. If panning shows up as 100 mouse events then that's a lot of wasted processing even with the small amount of debouncing we try to accomplish. If zooming with the scroll wheel shows up as fewer events then that means less updating.

For the filled versus bordered performance, this makes sense to me if vispy is being smart about it and only running one GL program versus two GL programs (line and mesh).

fschill:
re. events: That is a possibility. However, from what I could see, zooming with the scroll wheel results in even lower update rate (2.6 fps). Also, I did spot somewhere that vispy sets an "update_pending" flag that is meant to prevent multiple redraws if it's already doing one.

re. filled vs. bordered: yes, that is correct - if color is None or alpha==0, it will not generate the triangles or line segments for that object. It makes sense that just drawing fill is faster than drawing fill+border, but I was surprised that just the border is much slower than just the fill. Intuitively I thought triangles would be more work (more pixels, more complex, etc). But in any case, it is almost certainly not the actual GL performance, as the GPU can draw a thousand times more triangles quite easily than in this example (as is indicated by no performance hit with border radius, which would be a lot more triangles/lines than a sharp corner).

djhoese:
I'd be curious if the shader(s) are getting compiled multiple times too. It is something that *shouldn't* be happening but it isn't something we don't specifically test.

rougier:
Can we merge this PR? Looks good to me.

djhoese:
Thank you!

PR #1481, opened 2018-05-15
More flexible input for PolygonVisual

campagnola:
* PolygonVisual can optionally receive PolygonData as its input rather than just an array of positions. This allows more complex edge specification.
* PolygonData makes use of its `edges` property (previously it was ignored).

### DIFF
--- a/vispy/geometry/polygon.py
+++ b/vispy/geometry/polygon.py
@@ -85,6 +85,20 @@ class PolygonData(object):
 
         If no vertices or faces are specified, the function returns None.
         """
+        if self._edges is None:
+            npts = self._vertices.shape[0]
+            if np.any(self._vertices[0] != self._vertices[1]):
+                # start != end, so edges must wrap around to beginning.
+                edges = np.empty((npts, 2), dtype=np.uint32)
+                edges[:, 0] = np.arange(npts)
+                edges[:, 1] = edges[:, 0] + 1
+                edges[-1, 1] = 0
+            else:
+                # start == end; no wrapping required.
+                edges = np.empty((npts-1, 2), dtype=np.uint32)
+                edges[:, 0] = np.arange(npts)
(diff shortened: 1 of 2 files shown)


campagnola:
Open question: should the PolygonData.faces property be read-only? I can't think of a situation where we want the user to be able to set it; the whole purpose of PolygonData seems to be triangulation.

vispy/geometry/polygon.py
```diff
@@ -85,6 +85,20 @@ def edges(self):
...
+                edges = np.empty((npts, 2), dtype=np.uint32)
+                edges[:, 0] = np.arange(npts)
+                edges[:, 1] = edges[:, 0] + 1
+                edges[-1, 1] = 0
+            else:
+                # start == end; no wrapping required.
+                edges = np.empty((npts-1, 2), dtype=np.uint32)
+                edges[:, 0] = np.arange(npts)
+                edges[:, 1] = edges[:, 0] + 1
+            self._edges = edges
```
djhoese: Would it be better to move this chunk of code to the setter and then in init use `self.edges = edges` (instead of the private)?

djhoese: I might be missing something, but `self._edges` is never read directly, it is always `self.edges`. Additionally `self._edges` is never written to directly except for init. This block of code is just a validation step of the edges being applied to the PolygonData object so shouldn't that validation happen at set (one time) rather than at read (possibly many times).

djhoese: Ok it makes more sense now because I realize that they are just autogenerated from the vertices. However, when you set `vertices` don't you need to set `self._edges` to `None` so they are regenerated?

djhoese:
@campagnola Any chance you can fix the CI failures?

djhoese:
@campagnola You around? Think you could merge master in to this and see what the state of the tests are?

Issue #1407, opened 2017-12-19
Memory leak when using LinePlot.set_data


UUSim:
I've modified the example [plot.py](https://github.com/vispy/vispy/blob/master/examples/basics/plotting/plot.py) to continuously update the plot data.
To achieve this, I'm using the `set_data` function on the `line` object (`LinePlot`), as follows:

```
N_SAMPLES = 1000000
x = np.linspace(0, 1, N_SAMPLES)
y = np.random.rand(N_SAMPLES)-.5
line.set_data((x,y), symbol=None)
```

From a functional perspective this works, i.e. the new data is being displayed, however, this results in a huge memory leak (approx. 57 MiB leaked per call to `set_data`).

I'm using vispy `0.5.2` and `python3.5.2` on linux.

I am aware of the [realtime_signals.py](https://github.com/vispy/vispy/blob/master/examples/demo/gloo/realtime_signals.py) example, which has updating data. However, this example doesn't have axes, etc.

How can I achieve a plot with updating data, without memory leaks, and with plot labels, axes, etc?

davidh-ssec:
> How can I achieve a plot with updating data, without memory leaks, and with plot labels, axes, etc?

The answer is simple: pull requests are welcome! 😉 

Seriously though, we'll need a little more information to track this down. If you can find the problem and fix it yourself we welcome the contribution. How do you know the memory is leaking? How many iterations of updates have you tried and is the leak continuously leaking? What is `line` in your example above? The `LinePlot` object? The `LineVisual` underneath?

UUSim:
To elabolate, here is the complete code:
```
#!/usr/bin/env python3
"""
Plot data with different styles
"""
import numpy as np
import time
from vispy import plot as vp
  
fig = vp.Fig(size=(800, 400))

N_SAMPLES = 1000000
x = np.linspace(0, 1, N_SAMPLES)
y = np.random.rand(N_SAMPLES)-.5
line = fig[0, 0].plot((x, y))
 
print("line: {}".format(line)) # Shows that line is a vispy.scene.visuals.LinePlot

if __name__ == '__main__':
    fig.show()#run=True) # disabled to allow an update loop below
    
    lastupdate = time.time()
    while True:
        delta = time.time() - lastupdate
        if delta > 2: # Update the data every few seconds
            lastupdate = time.time()
            line.set_data((x,y), symbol=None) # if symbol=None is omitted, symbols will appear after the set_data call.
[...]

PR #1488, opened 2018-06-27
FIX: Memory leak in LinePlot.set_data (#1407)

mkubanski:
Fix memory leak in LinePlot (#1407) when no markers are to be rendered.

### DIFF
--- a/vispy/visuals/markers.py
+++ b/vispy/visuals/markers.py
@@ -590,16 +590,16 @@ class MarkersVisual(Visual):
     def symbol(self, symbol):
         if symbol == self._symbol:
             return
-        if symbol is not None:
+        if (symbol is not None and self._symbol is None and
+                self._data is not None):
             # Allow user to configure symbol after a set_data call with
             # symbol=None. This can break down if the user does a consecutive
             # marker.symbol = 'disc'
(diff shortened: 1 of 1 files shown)


mkubanski:
Skipping setting of the actual _data causes a crash in compute_bounds so we cant just bail out of set_data early. Regarding your second point - it does break. I have added an extra check for this - the user can break it (memory leak occurs again) with consecutive symbol changes (one with a valid symbol, the next with None) without drawing but that is a strange case. The other option would be to force a GLIR flush, but that seems a poor way to go about it I think?

vispy/visuals/markers.py
```diff
@@ -584,6 +590,16 @@ def symbol(self):
...
         if symbol == self._symbol:
             return
+        if symbol is not None:
+            # Allow user to configure symbol after a set_data call with
+            # symbol=None. This can break down if the user does a consecutive
+            # marker.symbol = 'disc'
+            # marker.symbol = None
+            # without drawing. At this point the memory leaking ensues
+            # but this case is unlikely/makes no sense.
+            if self._symbol is None and self._data is not None:
```
djhoese: Could you combine this if statement with the one above it?

PR #1489, opened 2018-06-28
Use matplotlib colormaps (if installed)

EtienneCmb:
Note that in [line.py](https://github.com/vispy/vispy/blob/master/vispy/visuals/line/line.py#L231) there's a `try/except` which is used to switch between colormaps and unique color. So I add a `KeyError` inside the `MatplotlibColormap` class but there's probably a better way to handle `KeyError`.

Small script to test it :
```python
import numpy as np

from vispy import scene
from vispy import app

cam = PanZoomCamera(rect=(0, 0, 20, 10))
sc = scene.SceneCanvas(bgcolor='black', show=True)

data = np.arange(200).reshape(10, 20).astype(float)
im_red = scene.visuals.Image(data, parent=wc.scene, cmap='tab20b')

app.run()

```
image

### DIFF
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -452,7 +452,10 @@ class MatplotlibColormap(Colormap):
 
     def __init__(self, name):
         assert has_matplotlib()
-        from matplotlib.cm import ScalarMappable
+        from matplotlib.cm import ScalarMappable, cmap_d
+        if name not in sorted(cmap_d):
+            raise KeyError("Possible values for Matplotlib colormaps are"
+                           " %s" % ', '.join(sorted(cmap_d)))
 
         vec = ScalarMappable(cmap=name).to_rgba(np.arange(LUT_len))
         Colormap.__init__(self, vec)
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -1113,10 +1113,10 @@ def get_colormap(name, *args, **kwargs):
             raise TypeError('colormap must be a Colormap or string name')
         if name in _colormaps:  # vispy cmap
             cmap = _colormaps[name]
-        else:
-            if has_matplotlib:  # matplotlib colormap
+        else:  # matplotlib colormap
+            try:
                 cmap = MatplotlibColormap(name)
-            else:
+            except:
                 raise KeyError('colormap name %s not found' % name)
 
         if inspect.isclass(cmap):
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -451,9 +451,7 @@ class MatplotlibColormap(Colormap):
     """
 
     def __init__(self, name):
-        from matplotlib.cm import ScalarMappable, cmap_d
-        assert has_matplotlib()
-        assert name in cmap_d
+        from matplotlib.cm import ScalarMappable
 
         vec = ScalarMappable(cmap=name).to_rgba(np.arange(LUT_len))
         Colormap.__init__(self, vec)
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -451,11 +451,9 @@ class MatplotlibColormap(Colormap):
     """
 
     def __init__(self, name):
-        assert has_matplotlib()
         from matplotlib.cm import ScalarMappable, cmap_d
-        if name not in sorted(cmap_d):
-            raise KeyError("Possible values for Matplotlib colormaps are"
-                           " %s" % ', '.join(sorted(cmap_d)))
+        assert has_matplotlib()
+        assert name in cmap_d
 
         vec = ScalarMappable(cmap=name).to_rgba(np.arange(LUT_len))
         Colormap.__init__(self, vec)
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -11,6 +11,7 @@ from .color_array import ColorArray
 from ..ext.six import string_types
 from ..ext.cubehelix import cubehelix
 from ..ext.husl import husl_to_rgb
+from ..testing import has_matplotlib
 import vispy.gloo
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


vispy/color/colormap.py
```diff
@@ -440,6 +441,26 @@ def texture_lut(self):
...
+    Parameters
+    ----------
+    name : string
+        Name of the colormap.
+    """
+
+    def __init__(self, name):
+        assert has_matplotlib()
+        from matplotlib.cm import ScalarMappable, cmap_d
+        if name not in sorted(cmap_d):
```
djhoese: Is `cmap_d` a dictionary? Is that an advertised/public/documented dictionary that won't be removed any time soon? Wouldn't `if name in cmap_d` work while avoiding the sorting?

EtienneCmb: Hello @djhoese :)

Yes, [cmap_d](https://github.com/matplotlib/matplotlib/blob/3448de363d66d448e853218421f2763bb3508730/lib/matplotlib/cm.py#L20) is a dict with the key the name of the colormap and values are colormap object. I don't think it will be removed soon. I removed the `sorted`.

Vispy also have a [get_colormaps](https://github.com/vispy/vispy/blob/master/vispy/color/colormap.py#L1105) method. If matplotlib is installed, it should probably return the union of vispy and mpl colormaps.

vispy/color/colormap.py
```diff
@@ -440,6 +441,24 @@ def texture_lut(self):
...
+
+    Parameters
+    ----------
+    name : string
+        Name of the colormap.
+    """
+
+    def __init__(self, name):
+        from matplotlib.cm import ScalarMappable, cmap_d
+        assert has_matplotlib()
```
djhoese: This assertion is happening after the import. You could move it up a line or would it be cleaner to change the if statement below to be `if name in _colormaps; elif has_matplotlib(); else: KeyError(...)`

larsoner: also `assert` statements are skipped when running Python in optimized code so it shouldn't be used as an actual check that you want to be guaranteed to run

EtienneCmb: Oh yes, sorry for the mistake. @larsoner what's Python optimized code? So finally I should remove `assert` in the `MatplotlibColormap`?

larsoner: running with `python -O`

I would change the code to:
```
if not has_matplotlib():
    raise ImportError('informative message')
from matplotlib.cm ...
```

EtienneCmb: In `get_colormap` @djhoese suggest something like `if name in _colormaps: ... elif has_matplotlib():...` so the `if not has_matplotlib(): raise ImportError('informative message')` inside `MatplotlibColormap` isn't really useful?

EtienneCmb: The only clean way I see right now to avoid matplotlib error message is that  `get_colormaps` return colormap name of both vispy and mpl if mpl is installed

It should looks like :
```py
if name not in get_colormaps():
    raise KeyError('colormap name %s not found' % name)
if name in _colormaps:  # vispy cmap
    cmap = _colormaps[name]
elif has_matplotlib():  # matplotlib colormap
    cmap = MatplotlibColormap(name)
```

djhoese: Are you calling `get_colormaps` inside `get_colormaps`? I think I prefer my original suggestion with the `if name in _colormaps; elif has_matplotlib(); else: raise KeyError`.

larsoner: That seems fine to me. And you probably don't really even need the check I suggested because the ImportError that will be raised just by the nested mpl import is probably sufficient

EtienneCmb: No, I'm not calling `get_colormaps`

EtienneCmb: ok for the `if name in _colormaps; elif has_matplotlib(); else: raise KeyError`. In case of a typo, the default Matplotlib error message is going to be raised, but not the one of vispy

djhoese: I'm ok with that because I think any other way would be putting mpl logic outside of the colormap class. I didn't really understand your code above where you put `if name not in get_colormaps()`.

EtienneCmb:
Hello @djhoese @larsoner 

There's an issue with Matplotlib colormap integration for LineVisual. The LineVisual can either have a uniform color or use a colormap. There's a [try/except](https://github.com/vispy/vispy/blob/master/vispy/visuals/line/line.py#L231) to switch between uniform / colormap colors but the `KeyError` is raised. A quick fix would be to remove the `KeyError` and just have `except:`?

djhoese:
@EtienneCmb I'm confused. Isn't the current KeyError what we want to catch? What exception is occurring that isn't a KeyError? Either they provide a colormap name (vispy or matplotlib) or they provide a color.

Regardless, I'm not in favor of a catch-all except block.

EtienneCmb:
See for example [travis test](https://travis-ci.org/vispy/vispy/jobs/414884115#L3560) for arrows. The default matplotlib error is raised.

djhoese:
I guess the MatplotlibColormap class should catch the ValueError and raise a KeyError or maybe even better (but uglier) catch the ValueError in vispy's `get_colormap` and raise a KeyError instead.

EtienneCmb:
I think this is not an error. The try/except in LineVisual is only used as a switch to detect if the user want a uniform color or a colormap. So, IMO, no error should be raised

djhoese:
We aren't raising an error in the LineVisual or rather that's not what I'm suggesting. I'm suggesting that we override the ValueError raised by matplotlib and raise a KeyError inside vispy's get_colormap instead. The LineVisual, as you said, is using the try/except to figure out if the user provided a colormap or a basic color. This is a valid use of a try/except and I don't see another logical way to do it It also goes along with python's ask forgiveness preference (even if I don't like this): https://docs.python.org/2/glossary.html#term-eafp

EtienneCmb:
Ok I understand now. So, should I wrap the `cmap = MatplotlibColormap(name)` in [colormap.py](https://github.com/vispy/vispy/pull/1489/files#diff-b0cdeb61a9230fe7192cfc12d6c2b804R1115) with a `try/except ValueError` and raise a KeyError instead?

djhoese:
Yes or some equivalent. Overall we are correcting for the inconsistency between vispy and matplotlib where vispy's developers chose a KeyError for "this colormap doesn't exist" and matplotlib's developers chose a ValueError for "the colormap you asked for doesn't make sense".

djhoese:
@larsoner Any last comments? Otherwise I'll probably merge this this week.

Issue #1033, opened 2015-07-31
Post-update-scenegraph issues


almarklein:
cc @Eric89GXL, @campagnola 
- [ ] ViewNode and a way to automatically create a view on a scenegraph subtree
- [ ] using fly-camera keys in volume.py example, while fly camera is not yet active results in errors
- [ ] moving mouse with CTRL+RMB gives errors in fly-camera `AttributeError: 'SceneMouseEvent' object has no attribute 'map_to_canvas'`
- [ ] document limitations of `_visual_bounds_at` (e.g. does not work in 3D now)
- [ ] Fix basisc/scene/stereo.py example
- [ ] Fallback for picking on IPython
- [ ] Make the `picking` property of visuals private (or a property on the draw event instead?)
- [ ] Ticks and gridlines should be 1px width or positioned with subpixel precision.

Eric89GXL:
@almarklein I'm going to eliminate the issues from the list that already have their own place for discussion (#926, #914, #628)

almarklein:
sure

KelSolaar:
I'm adding my share of issues (as described on Gitter):

- [ ] `Viewbox.canvas` returns `None` and not the current `SceneCanvas` instance.
- [ ] `SceneCanvas` `resize` signal is not emitted anymore, @Eric89GXL suggesting using `transform_change` instead but this does not exist:
```python
Traceback (most recent call last):
  File "./../colour_analysis/commands.py", line 168, in system_exit_wrapper
    if object(*args, **kwargs):
  File "./../colour_analysis/commands.py", line 353, in main
    arguments.layout)
  File "./../colour_analysis/analysis.py", line 252, in __init__
    self.__create_views()
  File "./../colour_analysis/analysis.py", line 859, in __create_views
    border_color=border_colour)
  File "./../colour_analysis/views/gamut_view.py", line 219, in __init__
    self.__scene_canvas.events.transform_change.connect(
AttributeError: 'EmitterGroup' object has no attribute 'transform_change'
```
Ultimately I was using that to put a `TextVisual` as a title on my views and make it react to `SceneCanvas` geometry changes. If there is a cleaner alternative please let me know (I suppose there should be one).
- [ ] The nasty one: Setting a `ViewBox` camera seems to prevent proper hotkeys processing, and as a result the hotkey pressed event is repeatedly happening forever. Here is my Gitter report doctored:

> Allright found why the application hangs, it is quite weird.
[...]

PR #1491, opened 2018-07-10
Make events only propagate to callbacks that were present when the event was emitted

campagnola:
Fixes the "nasty" bug described in #1033, which was caused by an event callback changing the list of callbacks.


larsoner:
Thanks @campagnola

PR #1501, opened 2018-08-07
Replace Grid Widget cassowary solver with kiwisolver

MatthieuDartiailh:
on the whole I believe that looking at the enaml grid helper may provide some inspiration if this was to be further refactored.

Notes:
- stay constraints are gone in kiwisolver (I still have to find an old discussion about this, and how to emulate them but I remember @sccolbert explaining there were not that useful)
- the solver needs to be explicitly called to update variables (suggestValue)
- value is a method not an attribute
- one can use suggestValue to update an edit variable (but I need to check if this is the right way to do this)


djhoese:
@MatthieuDartiailh I'd like to get back in to this in the next 6 months or so. Do you have any updates or ideas that you think we should be aware of with any recent kiwisolver changes?

MatthieuDartiailh:
No recent changes in kiwisolver no. However I know that matplotlib updated their use of kiwisolver to avoid under constrained systems so it may be worth having a look at what they are doing now.

djhoese:
Can you point us to where that work is happening? Pull requests?

MatthieuDartiailh:
Here is the PR https://github.com/matplotlib/matplotlib/pull/17494

djhoese:
@almarklein Do you have any experience with this "gridding stuff"? If not, maybe I'll have to look at this. It'd be great if we could get rid of cassowary.

almarklein:
> Do you have any experience with this "gridding stuff"?

Nope.

djhoese:
Ok, I'd like to start working on this as I said I'd like to get rid of cassowary. I just know about nothing about the grid solving stuff we're doing here. To start, I've rebased this PR on the current main branch.

djhoese:
@larsoner @almarklein @kmuehlbauer How do you want to handle this dependence on kiwisolver? List it as a hard requirement even though it is only used in the `Grid` Widget which is only used for plotting? Or list it as an extra requires for plotting?

I should note that although `Grid` is only ever used with `Widget.add_grid` which is only used behind-the-scenes for plotting, it is used in a few examples (mostly the ones specifically showing the gridding behavior) with just a plane SceneCanvas.

almarklein:
kiwisolver is c++. They provide a ton of pre-build binaries o pypi, but it is a liability, e.g. if a new Python version comes out, we rely on their devs to push a new release before we can update to that Python version. Also thinking of people on more exotic systems for which there are no pre-build binaries available. My vote would I think be to make it an optional dependency.

djhoese:
`kiwisolver` is a hard dep of matplotlib so we definitely wouldn't be the biggest project depending on it: https://github.com/matplotlib/matplotlib/blob/36dd5b6a6ba874a93446f8cfd15b6f3289e635a1/setup.py#L328

djhoese:
So the only test failing is the `test_perspective_render` test. If I run:

```bash
python vispy/scene/cameras/tests/test_perspective.py --vispy-audit-tests
```

I get:

image

So that's the test image in the left-most pane, the expected image in the middle, and the difference in the right. The test does `grid.padding = 20`. If I make `padding = 0` then the test passes. I would assume that the units of `padding` are pixels? So in that case even though my local screen ruler (KRuler) says the space is about 70 pixels, that makes more sense than the expected result which I would expect with zero padding. But maybe I don't know what I'm talking about.

@MatthieuDartiailh Any guesses?

djhoese:
If I set padding to 5 (25% of the default) then I get this:

image

This makes me think that the old solver is wrong and ignoring padding.

djhoese:
So I just tested this on `main` and it looks like padding and margin are completely ignored in the `test_perspective_render`. I can comment padding out, set it to 100, it doesn't change the output of the test.

@almarklein Any feelings about this? It just seems like the current solver is completely not working as expected (or this test is using it wrong).

almarklein:
I have no idea, TBH. But if this is the case for the old solver and not the new one, I suppose the old solver was wrong.

djhoese:
Yeah, I think that's what I figured too. I think the other work @MatthieuDartiailh brings up at the top are optimizations and better corrections that could be made, but given that neither of us know how this stuff is supposed to work...we might just have to leave it as is. Tests are passing and I did look through the examples that use `add_grid` and they all looked fine. Merge it?

almarklein (approved):
LGTM!

Issue #1508, opened 2018-08-20
Mesh not showing correctly


PaulDebus:
Hey, 
I am having a weird issue while displaying a 2D mesh using a `panzoom` view. I tried three different approaches and got varying results.

The base for all approaches is a (valid) mesh consisting of vertex positions (numpy array `vertices`) and face indices into the vertices (numpy array `faces`). 

## Polygons
First I created one [Polygon](https://vispy.readthedocs.io/en/latest/scene.html#vispy.scene.visuals.Polygon) for every triangle in my mesh. While this showed the desired result, the performance was abysmal. Zooming and panning did not really work... or at least took longer then I was prepared to wait

## Mesh
```python
mesh = scene.visuals.Mesh(parent=self.view.scene, color="red", mode="lines",
                              vertices=vertices, faces=faces)
```
This shows some points and lines between them with great performance, though the lines are not what I expected and not all points where connected/visible.

## Lines
I finally tried painting lines connecting all the vertices with the following code:
```python
border = np.empty((border_pos.shape[0]*3,2))
for i in range(border_pos.shape[0]):
    l = border_pos[i]
    border[3*i+0] = [l[0], l[1]]
    border[3*i+1] = [l[1], l[2]]
    border[3*i+2] = [l[2], l[0]]
uv = scene.visuals.Line(pos=vertices, connect=border, parent=self.view.scene, color="black")
```
This provides the desired result with good performance, but I cannot use the features of Polygons, like colored shapes and so on. I am mostly confused by the wrong results of the mesh. Below you find a screenshot of the mesh results and the line results side by side. The bottom black figure is the desired result via lines, the top red figure is the wrong results via the mesh.

![image of the results](https://user-images.githubusercontent.com/4587428/44349694-e4306e00-a49d-11e8-8a0e-12ba7071b252.png)

Do you have any pointer what I am doing wrong? Tell me if you need further information/data, I tried to provide a MWE.

Thanks in advance!

djhoese:
So there are a couple things I can think of for you to try and some questions I have:

1. In your mesh example why are you using `mode='lines'` if you want triangles?
2. How are you viewing the results and how are you saving the screenshots? Are you using a direct call to `render`? This has been shown to not draw things as expected, please use `vispy.gloo.util._screenshot` or take a screenshot of an open GUI window using another application or your OS's tools.
[...]

PR #1510, opened 2018-08-21
Support drawing only edges of the mesh

PaulDebus:
To only draw the edges of a mesh, or to produce a compound visual that
displays the edges, the "lines" mode of the mesh can be used. This
requires passing in the faces not as triangles but as edge list, so not
as an (n_faces, 3) array, but (n_edges, 2) array. This commit also
allows for passing in the face index matrix, that will be converted into
an edge index matrix. Should work for all meshes with constant type
polygons, so triangle or quad meshes.

### DIFF
--- a/vispy/visuals/mesh.py
+++ b/vispy/visuals/mesh.py
@@ -173,6 +173,7 @@ class MeshVisual(Visual):
     **kwargs : dict
         Keyword arguments to pass to `Visual`.
     """
+
     def __init__(self, vertices=None, faces=None, vertex_colors=None,
                  face_colors=None, color=(0.5, 0.5, 1, 1), vertex_values=None,
                  meshdata=None, shading=None, mode='triangles', **kwargs):
@@ -212,6 +213,17 @@ class MeshVisual(Visual):
 
         # Init
         self._bounds = None
+
+        # Convert from face triangle indexing to face edge
+        # indexing for line mode
+        if mode == "lines" and faces.ndim == 2 and faces.shape[1] > 2:
+            poly_type = faces.shape[1]  # number of vertices per face
+            new_faces = np.empty(
+                (faces.shape[0]*poly_type, 2), dtype=np.uint32)
+            for i in range(poly_type):
+                new_faces[i::poly_type] = np.roll(faces, i, axis=1)[:, :2]
+            faces = new_faces
+
         # Note we do not call subclass set_data -- often the signatures
         # do no match.
         MeshVisual.set_data(


djhoese:
Ah setting `_draw_mode` configures the base visual drawing behavior to send to GL.

PaulDebus:
I created this pull request as a result of issue #1508, where I did not get the correct display of a mesh. The mesh already exists and gets rendered correctly if using the supported mode `triangles`.

I am aware, that `lines` is neither supported nor documented, but it is used in several places to render the edges/borders of meshes, for example in the [mesh test](https://github.com/vispy/vispy/blob/master/vispy/visuals/tests/test_mesh.py)

I do not know if this addition is wanted, I added it because vispy was not behaving as I expected it to. Maybe this change is the wrong approach to tackling this behavior. Maybe it is more suitable to change `scene.visuals.Mesh` to accept also a `border_color` parameter as does [the box](https://github.com/vispy/vispy/blob/982bed13646100b4c557b65916cb492510e07078/vispy/visuals/box.py#L43)

PaulDebus:
Thanks for the feedback, glad to contribute!

My reasoning for the roll was to allow for any mesh with regular polygon type (= vertices per face). Say we have faces consisting of the vertices 
```python
[1, 2, 3]
[2, 3, 4]
```
for the edges we need the following lines:
```python
[1, 2]
[2, 3]
[3, 1]
[2, 3]   # not really necessary
[3, 4]
[4, 2]
```
So the inner loop in this case would get executed 3 times. In the first execution, the array is rolled by 0 and the first to columns are inserted into every third row of the final index matrix. In the next step, the array is rolled by 1 and the first to columns are inserted into every third row, starting from the second (index `1`), and so on. Line 224 could be rewritten as
```python
new_faces[i::poly_type] = faces[:, [i, (i+1)%poly_type]]
```
if I am not mistaken. In this case, no temporary arrays should be created at all. Do you think this is better?

PaulDebus:
And apart from that: the failing travis-ci is not caused by my changes, is it?

EtienneCmb:
Note that `'line_strip'` and ` 'line_loop'` are other possible [drawing modes](https://github.com/vispy/vispy/blob/master/vispy/gloo/program.py#L410)

PaulDebus:
Yes, that is right, though I understand the [OpenGL doc](https://www.khronos.org/opengl/wiki/Primitive#Line_primitives) that those are not really useful in this case.

djhoese:
@PaulDebus Any chance you have time to merge current vispy master with your branch to update this and see if the tests pass now?

os-gabe:
For what its worth, I used the following code to calculate the array of point pairs needed for only drawing mesh edges using a line visual:
```
nf = len(faces)
mask = np.tile([0, 1, 1, 2, 2, 0], nf) + np.repeat(np.arange(0, nf * 3, 3), 6)
edges = faces.reshape(-1)[mask]
edge_visual = scene.visuals.Line(pos=verts[edges], connect="segments", color="white")
```

PaulDebus:
@djhoese thanks for bringing the pull request to my attention again. I tried running the tests, but even on a clean copy of the master branch, they do not run successfully for me. 
```
  File "/usr/lib/python3.7/unittest/loader.py", line 205, in loadTestsFromName
    test = obj()
TypeError: circular() missing 1 required positional argument: 'adjacency_mat'
```
Did I miss something there?

I will gladly test my code again and look into what @os-gabe  provided as a fix. I will try to figure out, which computation is the fastest.

djhoese:
@PaulDebus How are you running the tests? If you do `make unit` in the root of the repository it should use pytest to find and run the tests in vispy.

PaulDebus:
Ah, I used `python setup.py test` and it did not work. Running `make test` results in one failing test `test_font_glyph` because the Noto Emoji Font is not scalable, which is not connected to my change, I guess.

djhoese:
Can you push your merged version to your master branch on github? The PR will be automatically updated and will start running the tests.

djhoese:
Ah great, github doesn't notify me of forced pushes...

Just saw this, restarting the one failed test.

djhoese:
@PaulDebus It looks like one of the examples is failing. Here's the whole traceback from travis:

```
Example /basics/visuals/tube.py failed (0):
  File "<string>", line 12, in <module>
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/travis/build/vispy/vispy/examples/basics/visuals/tube.py", line 82, in <module>
    canvas.show()
  File "/home/travis/build/vispy/vispy/vispy/app/canvas.py", line 429, in show
  File "/home/travis/build/vispy/vispy/vispy/app/backends/_qt.py", line 405, in _vispy_set_visible
    self.showNormal()
  File "/home/travis/build/vispy/vispy/vispy/app/backends/_qt.py", line 501, in event
  File "/home/travis/build/vispy/vispy/vispy/app/backends/_qt.py", line 501, in event
  File "/home/travis/build/vispy/vispy/vispy/app/backends/_qt.py", line 818, in paintGL
  File "/home/travis/build/vispy/vispy/vispy/util/event.py", line 455, in __call__
  File "/home/travis/build/vispy/vispy/vispy/util/event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  << caught exception here: >>
  File "/home/travis/build/vispy/vispy/vispy/util/event.py", line 471, in _invoke_callback
    cb(event)
  File "/home/travis/build/vispy/vispy/vispy/scene/canvas.py", line 217, in on_draw
    self._draw_scene()
  File "/home/travis/build/vispy/vispy/vispy/scene/canvas.py", line 266, in _draw_scene
    self.draw_visual(self.scene)
  File "/home/travis/build/vispy/vispy/vispy/scene/canvas.py", line 304, in draw_visual
    node.draw()
  File "/home/travis/build/vispy/vispy/vispy/scene/visuals.py", line 99, in draw
  File "/home/travis/build/vispy/vispy/vispy/visuals/mesh.py", line 531, in draw
    Visual.draw(self, *args, **kwds)
  File "/home/travis/build/vispy/vispy/vispy/visuals/visual.py", line 435, in draw
  File "/home/travis/build/vispy/vispy/vispy/visuals/mesh.py", line 526, in _prepare_draw
    if self._update_data() is False:
  File "/home/travis/build/vispy/vispy/vispy/visuals/mesh.py", line 400, in _update_data
    normals = md.get_face_normals(indexed='faces')
  File "/home/travis/build/vispy/vispy/vispy/geometry/meshdata.py", line 301, in get_face_normals
    v[:, 2] - v[:, 0])
IndexError: index 2 is out of bounds for axis 1 with size 2
```

Could you have a look at this?

PaulDebus:
Hm... this error seems to be caused by my change, as it does not occur on the previous commit. I will look into it.

PaulDebus:
Ok, as far as I understand it, the problem with my change is, that the mesh is not represented as triangles anymore, but as edges. The example - and I guess all tubes and maybe more types - use the face normal information for the rendering. As there are no more faces, computing the normal is not possible, which causes this error.

As I am not really sure how to handle that, I would like to get your opinion on it. You have a better overview over the internal workings of vispy, so maybe you have a better idea then me. One way to "fix" this issue is to maintain a separate list of face based vertex indices, like there originally was. Then all the computations would have to access this list and not the list, that has been converted to edge based. This would require many changes in many places and I am pretty sure that there would be many side effects breaking stuff on all ends.

I did introduce the change in the first place i found, right in the `__init__` method of the mesh visual. Maybe this is not the correct place and the change should be further down the line, only where the actual rendering happens. I will have another look into this, when I find some time. 

Sorry for the mess, I hope we can find a solution

djhoese:
Let me start by saying faces/edges and meshes really aren't my strong suit. Obviously, backwards compatibility is a priority. I was ok with your changes because they seemed to be specific to someone providing face information with 2 coordinates (shape[1] == 2) instead of 3. It seems based on the tests that the changes don't work in any case, right?

@os-gabe In your code, is `verts` the 3-element faces array? Is that something that could be dropped in here?

@PaulDebus It looks like there is an unused `edges` keyword argument in `MeshData` from `vispy/geometry/meshdata.py`. There are also the changes by @asnt in #1462 and in #1444 which will likely change a lot of what's possible regarding meshes.

PaulDebus:
Yeah, the tests show, that the change does break some stuff. I looked deeper in what happens, especially as the failing test (`/basics/visuals/tube.py`) works without my change and does display correctly only the edges. Which really confuses me.

I also looked at the `MeshData` and it does correctly compute the edges, if they are not provided in the constructor. So in theory, the required logic is already there. From my now different understanding, I would guess that the better change would be, to pass `MeshData.get_edges()` to the renderer instead of `MeshData.get_faces()` if the drawing mode is `lines`. But I got lost while searching for a good place for this change. The further I go, the deeper I get into the internals of vispy and the less I know what is happening. I will continue there, but I do not know what I will find or when I will find something.

asnt:
@PaulDebus @djhoese Maybe a new visual is needed in this case that draws a different geometry than the mesh on top of it. (I am not sure a filter as in #1444 is adapted unless we are willing/able to hook the MeshVisual.draw())

os-gabe:
Per @djhoese's earlier question, verts is the array of mesh vertices. Faces is the array of indices into verts to form the triangular faces.

djhoese:
@asnt Based on the recent merges of your PRs, what do you think of this PR?

asnt:
> @asnt Based on the recent merges of your PRs, what do you think of this PR?
> 
> Edit: By that I mean, do your PRs implement this PR in an easier way or make this PR unnecessary?

@djhoese As I tried to highlight in #1665, I see this as an alternative to #1689. A drawback of #1689 is that it's not possible to set a specific colour per edge (the colour is defined by neighbouring vertices, thus two adjacent edges always share the same colour around the common vertex). I think #1665 could allow per-edge colour. I haven't looked into, though.

Otherwise, implementing #1665 as a filter should not be an issue.
Not sure about the higher-level API, but the `wireframe` keyword in the MeshVisual constructor could maybe be used to select between variants.

rougier:
@PaulDebus Probabyl a stupid question but can't you use polygon mode (`glPolygonMode(GL_FRONT_AND_BACK,GL_LINE)`)?

PaulDebus:
Uh, I haven't looked at it in quite some time now to be honest :sweat_smile: 

This would probably be possible, though I am not very good with OpenGL.

Also, I think a lot of stuff has happened since I opened this PR, and maybe some progress has been made somewhere else? I am not up to date anymore, so maybe somebody else can help?

PR #1511, opened 2018-08-21
ImageVisual and masked array

### LINKED ISSUE #1509: ImageVisual doesn't support masked array
One nice feature when plotting images using Matplotlib is that it support [masked array](https://docs.scipy.org/doc/numpy-1.13.0/reference/maskedarray.html) which mean that array values with a `True` set to the mask are not displayed (and not take into account in the colormap)

```py
import numpy as np

import matplotlib.pyplot as plt

from vispy import scene

n = 1000

data = np.ones((n, n))
y, x = np.ogrid[-n / 2:n / 2, -n / 2:n / 2]
mask = x**2 + y**2 >= (n / 2)**2
data[mask] = 0.
image = np.ma.masked_array(data, mask=mask)

# Matplotlib :
plt.imshow(image)
plt.show()

# Vispy :
cam = PanZoomCamera(rect=(0, 0, n, n))
sc = scene.SceneCanvas(bgcolor='black', show=True)
im_vispy = scene.visuals.Image(image, parent=wc.scene)
sc.app.run()

```

image

EtienneCmb:
Hi @djhoese 

This is not a fully working pull request. Right now :
1. It fix `ImageVisual` if there's a `np.nan` in the 2D image (beacause of the `clim = np.min(data), np.max(data)`)
2. If a masked_array is passed, masked values are set to `np.nan`

IMO the intended behavior would to to discard nan pixels (as discussed in #1509 ). For now, nan pixels are set to 0. (somewhere... I can't figure it out). Do you have an idea how to discard those pixels?

(Note : OpenGL seems to have a [isnan](https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/isnan.xhtml) function but it's not working for me. [Here](https://github.com/metsci/glimpse/blob/ec72fae680de6334bdf64aea5cc950f3c245e304/core/src/main/resources/shaders/colormap/tagged_colorscale_shader.fs#L68) they said that `x != x` could be used instead)

```py
import numpy as np

from vispy import scene

n = 10

# data = np.random.rand(10, 10)
data = np.arange(100).reshape(10, 10).astype(np.float32)
y, x = np.ogrid[-n / 2:n / 2, -n / 2:n / 2]
mask = x**2 + y**2 >= (n / 2)**2
data = np.ma.masked_array(data, mask=mask)

# Vispy :
cam = PanZoomCamera(rect=(0, 0, n, n))
sc = scene.SceneCanvas(bgcolor='black', show=True)
im_vispy = scene.visuals.Image(data, parent=wc.scene)
sc.app.run()
```
image

closes #1509

### DIFF
--- a/vispy/visuals/image.py
+++ b/vispy/visuals/image.py
@@ -233,7 +233,7 @@ class ImageVisual(Visual):
         image : array-like
             The image data.
         """
-        data = np.asarray(image)
+        data = image if np.ma.is_masked(image) else np.asarray(image)
         if self._data is None or self._data.shape != data.shape:
             self._need_vertex_update = True
         self._data = data
@@ -400,6 +400,11 @@ class ImageVisual(Visual):
             # deal with clim on CPU b/c of texture depth limits :(
             # can eventually do this by simulating 32-bit float... maybe
             clim = self._clim
+            # deal with masked_array
+            if np.ma.is_masked(data):
+                mask = data.mask
+                data = np.array(data)
+                data[mask] = np.nan
             is_num = np.isfinite(data)
             if isinstance(clim, string_types) and clim == 'auto':
                 clim = np.min(data[is_num]), np.max(data[is_num])
--- a/vispy/visuals/image.py
+++ b/vispy/visuals/image.py
@@ -400,8 +400,9 @@ class ImageVisual(Visual):
             # deal with clim on CPU b/c of texture depth limits :(
             # can eventually do this by simulating 32-bit float... maybe
             clim = self._clim
+            is_num = np.isfinite(data)
             if isinstance(clim, string_types) and clim == 'auto':
-                clim = np.min(data), np.max(data)
+                clim = np.min(data[is_num]), np.max(data[is_num])
             clim = np.asarray(clim, dtype=np.float32)
             data = data - clim[0]  # not inplace so we don't modify orig data
             if clim[1] - clim[0] > 0:


vispy/visuals/image.py
```diff
@@ -400,8 +400,14 @@ def _build_texture(self):
...
             clim = self._clim
+            # deal with masked_array
+            if np.ma.is_masked(data):
+                mask = data.mask
+                data = np.array(data)
+                data[mask] = np.nan
+            is_num = np.isfinite(data)
             if isinstance(clim, string_types) and clim == 'auto':
-                clim = np.min(data), np.max(data)
+                clim = np.min(data[is_num]), np.max(data[is_num])
```
QuLogic: `np.nanmin` / `np.nanmax`

EtienneCmb: Yes but the array can contains `np.inf`. Actually, using `plt.imshow`, non-numerical pixels are not displayed.

vispy/visuals/image.py
```diff
@@ -400,8 +400,14 @@ def _build_texture(self):
             # deal with clim on CPU b/c of texture depth limits :(
             # can eventually do this by simulating 32-bit float... maybe
             clim = self._clim
+            # deal with masked_array
+            if np.ma.is_masked(data):
+                mask = data.mask
+                data = np.array(data)
```
djhoese: Does this make a copy of the array?

vispy/visuals/image.py
```diff
@@ -400,8 +400,14 @@ def _build_texture(self):
...
+                data[mask] = np.nan
+            is_num = np.isfinite(data)
```
djhoese: Can we move this inside the `if` statement below so that it is only computed if it is needed?

vispy/visuals/image.py
djhoese: @EtienneCmb So you are saying in matplotlib `inf` values are ignored just like `nan`?

EtienneCmb: Yes, `inf` and `nan` are ignored when plotting and the automatic `clim` is based on numerical values only.

vispy/visuals/image.py
EtienneCmb: Are you talking about the `np.ma.is_masked(data)`? If yes, no, it only return a bool indicating if the array has a `mask` attribute.

vispy/visuals/image.py
EtienneCmb: I would say no because non-numerical values can happen in non-masked arrays

vispy/visuals/image.py
djhoese: No, I meant the `np.array` on a masked array.

vispy/visuals/image.py
djhoese: Sure, but I'm saying you create `is_num` above the if statement and then don't use it until you are *inside* the if statement. It should be created inside the if statement to avoid needlessly creating it (when clim is not `'auto'`).

EtienneCmb:
@djhoese In it's current form, this PR doesn't add the support for masked-array plotting. My original idea was to turn masked values into `nan` in the `data` array. Then, in the shader, add an `if` statement in order to discard nan pixels. But I couldn't do it because, for a reason that I don't understand, non-finite values are turned into `0` **before** entering in the shader.

But, at least, this PR fix the issue of automatic clim when there's non-finite values. To me two possibilities : 
* I make an other small PR to only fix the issue of automatic clim
* Someone has an idea why nan values are converted before coming inside the shader and then I try to add the support for masked-array

djhoese:
```python
In [1]: import numpy as np                                                                                                                                                                              

In [2]: np.array([np.nan]).astype(np.uint8)                                                                                                                                                             
Out[2]: array([0], dtype=uint8)
```

vispy/visuals/image.py
EtienneCmb: Oh yes, sorry, I didn't read your message correctly. Yes, indeed, I would be better inside the `if isinstance(clim, string_types) and clim == 'auto':`

vispy/visuals/image.py
EtienneCmb: Yes, it returned a copy of the array but not `data.data`

djhoese: So it is a view of the array and does **not** copy the data to another memory location?

EtienneCmb:
but there's no `int` conversion, buffer use `np.float32` arrays?

vispy/visuals/image.py
EtienneCmb: Yes, it's a [view of the array](https://docs.scipy.org/doc/numpy-1.13.0/reference/maskedarray.generic.html#accessing-the-data)

djhoese:
Hm I suppose you're right. I was confusing this with number of bits. Even though we may be providing 32-bit floats, they get stored in the GPU as 8-bit floats (8 bits for R, 8 bits for G, etc) from 0 to 1. See table 2 in: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexImage2D.xhtml

However, the conversion from NaN to 0 with integer conversion seems to be accurate and I swear I tested this with my own application. I was trying to put float data with NaNs in to the texture and was not seeing any NaNs. I ended up building a custom texture with R32F internal format for full floats. I'm not sure we can put NaNs in using the ES2 compatible/default RGB/A internal formats. @rougier does this ring a bell at all?

kmuehlbauer:
IIRC the data is aligned between 0 and 1. Everything which does not fit will either be 0 or 1.

rougier:
@djhoese I think you're right. I remember having tried to have NaN on GPU without much success.

mkubanski:
I think I did something similar to what you are attempting to do with nans at work for a project - I did this years ago with older versions of the color module, but I think the concept is still workable and will allow you to do the masking with nans.

The gist of it is simply set a reserved normalized color value for nan - I used 1.0. I had a subclass of ImageVisual called NanImageVisual which allowed nans in, set true color values of 1.0 to 0.9999, and then set nans to 1.0.

In the color module I had _glsl_mix updated to handle the case of the normalized color being 1.0 with an additional else if statement and in this case return vec4(0.0, 0.0, 0.0, 0.0). I realize the module has gone through some changes and uses texture lookup now but I assume it is still feasible to do this with the texture coordinate and an if statement?

All my colormaps were for image visuals so this was not an ideal implementation for cases where the colormap could possibly be used outside of the NanImageVisual subclass I built. But I guess it could be possible to implement all normalization in vispy between 0 and 0.9999 and reserve 1.0 for nan in a generic manner? Or some variation of this?

EtienneCmb:
That's a very interesting idea. I did something similar, but for meshes. I defined an additional buffer full of {0., 1.}. Inside the vertex shader, I had an if statement to discard vertices. That's may be a solution, introduce a new buffer, specifically for the mask, to discard or not pixels

djhoese:
@EtienneCmb What is the current state of this PR? Do your changes do what you expect for all of the cases discussed? It looks like a `np.min/max` call has to be changed to `np.nanmin/nanmax` at least, right?

EtienneCmb:
Hello @djhoese ,

No, this PR doesn't includes everything I planned to have. At least if I remember correctly it fixes a small issue to plot images that might contains NaN / Inf values. In that case, automatic color limits (`clim`) are not correct.

But the main idea was to add the support of masked array at least for images and right now it doesn't. Reading again the idea of @mkubanski there's probably a workaround, but not as clean as I wanted originally (I thought I was going to be able to "simply" add a `discard` inside shaders to skip non-float values but it never works).

EtienneCmb:
It might be better to close without merging and I can make a first new one starting from the latest master (because I don't know if it changes). And then, if I've time, try to work a bit on my side on masked arrays.

As you want ;)

djhoese:
Yeah let's close this since it seems like it got off topic. If you make a new PR make sure to mention this one to connect the conversations.

> but it never works

Right, now I remember. The data floats are converted to integers for the default internal texture format and that is converting NaNs to 0s. We'd have to store floats in the texture to get this to work out of the box. Keep in mind this isn't a vispy limitation, this is OpenGL.

PR #1516, opened 2018-09-08
multi-radii tubes

schlegelp:
In the current master TubeVisuals can only have a single radius. This PR introduces the option to pass a list of radii to `vispy.scene.visuals.Tube` - one for each point of the tube.

It also forces coordinates into `floats` - otherwise `tangents /= mags[:, np.newaxis]` in [tube.py](https://github.com/vispy/vispy/blob/a6ec64bb7b56739b463e3c16245a31940ca4f5f7/vispy/visuals/tube.py#L123) throws a TypeError.

A minimal example:
```Python
import vispy
import numpy as np

# Create canvas
canvas = vispy.scene.SceneCanvas()
view.camera = vispy.scene.ArcballCamera()
canvas.show()

# Create tube with 5 points and 5 individual radii
coords = np.array([[0, 0, 0], [1, 1, 1], [2, 2, 2], [3, 3, 3], [4, 4, 4]])
tu = vispy.scene.visuals.Tube(coords, radius=[1, 4, 2, 3, 1])

# Add tube to canvas and set range
view.add(tu)
view.camera.set_range()
```

tube_w_radii

### DIFF
--- a/examples/basics/visuals/tube.py
+++ b/examples/basics/visuals/tube.py
@@ -26,6 +26,10 @@ points3[:, 2] += 30
 points4 = points1.copy()
 points4[:, 0] += 41.
 
+points5 = points1.copy()
+points5[:, 0] += 20.4
+points5[:, 2] += 15
+
 colors = np.linspace(0, 1, len(points1))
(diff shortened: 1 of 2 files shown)


djhoese:
Supersceded by #1658

PR #1518, opened 2018-09-24
BUGFIX: Properly update opacity in `VisualNode`s

kne42:
# Description
Fixes a bug in `VisualNode`'s `_update_opacity` function wherein opacity was being updated by setting a `ColorFilter`'s `color` property when the correct property to set was `filter`. Additionally calls the visual's `update` method upon filter change.

Since opacity only affects the `alpha` channel, additionally opts to replace the overly explicit `ColorFilter` with an `Alpha` filter. This prevents tampering with the RGB channels in the private filter.

### DIFF
--- a/vispy/scene/visuals.py
+++ b/vispy/scene/visuals.py
@@ -17,7 +17,7 @@ import weakref
 
 from .. import visuals
 from .node import Node
-from ..visuals.filters import ColorFilter, PickingFilter
+from ..visuals.filters import Alpha, PickingFilter
 
 
 class VisualNode(Node):
@@ -28,7 +28,7 @@ class VisualNode(Node):
         Node.__init__(self, parent=parent, name=name,
                       transforms=self.transforms)
         self.interactive = False
-        self._opacity_filter = ColorFilter()
+        self._opacity_filter = Alpha()
         self.attach(self._opacity_filter)
 
         self._id = VisualNode._next_id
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
@kne42 Well spotted. Looking at the code history, the `ColorFilter` was used since the big scenegraph overhaul. @campagnola If you are available, could you have a look, if we are on the right track here?

@djhoese I can't think of any additional functionality. It might just have slipped in the wrong place. PR #928 was huge.

One thing is that the `ColorFilter` doesn't have a `color` property at all. To my understanding that adds `color` to the `ColorFilter` instance but does just nothing.

The tests fail IMHO, because the image to check against in the [vispy-test repo](https://github.com/vispy/test-data/blob/master/visuals/colormap_nu.png) breaks. We just might need to change this particular image. Pinging @jarno-ssec as creator of this particular image.

djhoese:
@kmuehlbauer I'm not sure I agree with the failing image. I can test it locally to double check, but the exception seems to be raised from trying to render the image to test. The test should compare the images and if there is a failure it should upload the difference to the vispy website. It doesn't seem to be doing that.

kmuehlbauer:
@djhoese Sorry, obviously I made the wrong assumptions. 

Here for easy access the lines where it breakes:

https://travis-ci.org/vispy/vispy/jobs/432691173#L3814-L3821

djhoese:
There were some fixes to some conda-forge packages so I'm going to try to rerun the travis tests for this and see if they pass now.

PR #1520, opened 2018-09-25
 Adding missing app.sleep function 

thomasdeneux:
This should be quite straightforward.


vispy/app/_default_app.py
```diff
@@ -77,3 +77,19 @@ def process_events():
...
+    CPU stress when VisPy is run in interactive mode.
+    see inputhook.py for details
+
+    Parameters
+    ----------
+    duration_sec: float
+        Time to sleep in seconds
+    """
+    use_app(call_reuse=False)
+    return default_app.sleep(duration_sec)
```
kmuehlbauer: @thomasdeneux Please add a newline here.

djhoese:
I'm ok merging this, but are there a lot of cases where a user doesn't have access to an `app` already by the time they need the `sleep` function?

thomasdeneux:
sorry i am not sure about the question; of course the line `use_app(call_reuse=False)` will probably never be useful, but i have put it for the sake of sanity

besides, the reason altogether for this pull request is that I needed a sleep that, contrary for example to time.sleep, would let the events in the queue be processed (in my case, regular display updates triggered by a timer). I was surprised that app.sleep did not exist, whereas it was mentioned in the doc (http://vispy.org/app.html?highlight=sleep#vispy.app.Application.sleep), so i have added it.

djhoese:
The `app.sleep` *does* exist as a method on the `Application` *instance* not the `app` module within vispy. Does that clear things up?

```python
from vispy.app import use_app
app = use_app('pyqt4')
app.sleep(5)
```

thomasdeneux:
oh i see! i never used the app instance, but always the app module, in particular to finish scripts with app.run().
So, well, it is up to the vispy team to decide whether to provide the two syntaxes or rather enforce one over the other...

djhoese:
I'm going to close this since I think it is best for users to use the `app` object directly when it comes to things like `sleep`. That said, I am ok if @larsoner or @kmuehlbauer want to overrule me, reopen it, and merge it anyway.

PR #1504, opened 2018-08-10
 switched examples from PyQt4 to PyQt5 

thomasdeneux:

PR #1521, opened 2018-09-25
Switch examples from PyQt4 to PyQt5

thomasdeneux:
Hello, 
I did some mess with the previous pull request with the same name. So here i re-created the same pull request, and it contains only the really related commits.
Best,

see also, discussion in #1504


kmuehlbauer:
@djhoese Recreated the PR setup (assignees and label) from #1504. No offence meant :grinning:

djhoese:
Thanks @kmuehlbauer and @thomasdeneux. FYI the reason #1504 wasn't merged yet is I'm still working on getting the Qt5 migration working all the way (#1394).

djhoese:
Thanks @thomasdeneux, I adapted some of what you did based on what @larsoner did in his original PyQt5 PR: getting PyQt5 if possible, falling back to PyQt4, or letting the `app` find the best default.

PR #1523, opened 2018-09-26
Re-implement vispy-provided filters

### LINKED ISSUE #1434: Can attach but not detach alpha filter
Hi,

I'm writing some sort of STL viewer with PyQt5 + vispy. Here is my Canvas widget:

```python
#!/usr/bin/python
# coding: utf-8

import os
from vispy import app, geometry
import vispy.scene
from vispy.app import use_app
from typing import Tuple, Union
from vispy.visuals.filters import Alpha


# TODO: remove trimesh, vispy should soon be able to load STL files
import trimesh

use_app('PyQt5')


class MyCanvas(vispy.scene.SceneCanvas):

    """"
    Class to view a STL file. Camera is embedded, so rotations/translations are
    possible
    https://github.com/vispy/vispy/issues/1425#issuecomment-365369727
    """

    def __init__(self, size: Tuple[int, int]=(800, 500)):

        """
        size is the size in pixels (x, y). watch_dir is the path to watch
        for new STL files
        """

        vispy.scene.SceneCanvas.__init__(self, keys='interactive',
                                         size=size, bgcolor='white')

        # Unfreeze the class, so adding attributes is possible
        self.unfreeze()

        # Create a dict to store several meshes
        # Meshes are stored with their names so they can be reloaded/reviewed
        # namely
        self.meshes = dict()

        # Add an 'arcball' camera
        self.view = self.central_widget.add_view()
        self.view.camera = 'arcball'

        self.translucent_filter = Alpha(0.3)

        self.viewMesh("test")

        # Freeze the class again. Might not be necessary
        self.freeze()

        self.setTranslucent(True)
        self.setTranslucent(False)

        self.show()


    def viewMesh(self, file_name: str):

        """
        Will load the file file_name in the STL viewer. If the mesh didn't
        exist before, it will be created. Else, it will be updated
        """

        # file_name is passed wo extension, so add .stl
        file_name = file_name + ".stl"

        file_to_view = file_name

        # Load the stl file. Works with binary or ascii stl files
        mesh = trimesh.load(file_to_view)
        mdata = geometry.MeshData(mesh.vertices, mesh.faces)

        # Create a Mesh visual, with the view as parent
        self.meshes[file_name] = visuals.Mesh(meshdata=mdata,
                                              shading='flat',
                                              parent=self.view.scene,)


    def setTranslucent(self, translucent: bool):

        for mesh in self.meshes.values():
            if translucent:
                mesh.attach(self.translucent_filter)
            else:
                mesh.detach(self.translucent_filter)


if __name__ == '__main__':
    win = MyCanvas()
    win.viewMesh("test")

    import sys
        app.run()
        print(app)
```

It should be a working example. Here is a file ```test.stl``` if you want to run the code:

[test.stl.txt](https://github.com/vispy/vispy/files/1748729/test.stl.txt)

If you run this code with ```setTranslucent(False)``` commented, it will work perfectly and the mesh should be translucent. If ```setTranslucent(False)``` is executed, the script crashes with:

```
  File "canvas.py", line 94, in <module>
    win = MyCanvas()
  File "canvas.py", line 57, in __init__
    self.setTranslucent(False)
  File "canvas.py", line 90, in setTranslucent
    mesh.detach(self.translucent_filter)
  File "/home/djipey/.local/share/virtualenvs/cad/lib/python3.6/site-packages/vispy/visuals/visual.py", line 499, in detach
    filt._detach(view)
AttributeError: 'Alpha' object has no attribute '_detach'
```

So it seems I can attach an Alpha filter to a mesh, but not detach it. Am I doing something wrong ?

### LINKED ISSUE #1522: Most provided filters cannot be detached.
# Description
Most filters provided by `vispy` itself cannot be detached due to a missing `_detach` method. This issue is meant to collect references to all such filters and document related issues.

# References
## [color.py](https://github.com/vispy/vispy/blob/master/vispy/visuals/filters/color.py)
[IsolineFilter](https://github.com/vispy/vispy/blob/master/vispy/visuals/filters/color.py#L11)
[Alpha](https://github.com/vispy/vispy/blob/master/vispy/visuals/filters/color.py#L113) (see #1434)
[ColorFilter](https://github.com/vispy/vispy/blob/master/vispy/visuals/filters/color.py#L137)
[ZColormapFilter](https://github.com/vispy/vispy/blob/master/vispy/visuals/filters/color.py#L161)

## [material.py](https://github.com/vispy/vispy/blob/master/vispy/visuals/filters/material.py)
What is even going on here? On [line 15](https://github.com/vispy/vispy/blob/master/vispy/visuals/filters/material.py#L15), it attempts to import from a non-existent module.

## [picking.py](https://github.com/vispy/vispy/blob/master/vispy/visuals/filters/picking.py)
[PickingFilter](https://github.com/vispy/vispy/blob/master/vispy/visuals/filters/picking.py#L10)

kne42:
# Description
Provides a new filter base-class and re-implements most filters.

These changes include the following:
~~- freezes all filter attributes to prevent incorrect assignment which can occur when switching filters (see #1518)~~
- implements the `_detach` method for proper filter detachment (see #1522 and #1434)

# References
closes #1522 
closes #1434

### DIFF
--- a/vispy/visuals/filters/color.py
+++ b/vispy/visuals/filters/color.py
@@ -8,8 +8,6 @@ from ...color import colormap, Color
 
 
 class IsolineFilter(Filter):
-    __slots__ = ('_level', '_width', '_color', '_antialias')
-
     FRAG_SHADER = """
         void isoline() {
             if ($isolevel <= 0 || $isowidth <= 0) {
@@ -109,8 +107,6 @@ class IsolineFilter(Filter):
 
 
 class Alpha(Filter):
-    __slots__ = ('_alpha',)
-
     FRAG_SHADER = """
         void apply_alpha() {
             gl_FragColor.a = gl_FragColor.a * $alpha;
@@ -132,8 +128,6 @@ class Alpha(Filter):
 
 
 class ColorFilter(Filter):
-    __slots__ = ('_filter',)
-
     FRAG_SHADER = """
         void apply_color_filter() {
             gl_FragColor = gl_FragColor * $filter;
@@ -156,8 +150,6 @@ class ColorFilter(Filter):
 
 
 class ZColormapFilter(Filter):
-    __slots__ = ('cmap',)
-
     FRAG_SHADER = """
         void z_colormap_support() {
             $zval = $position.z;
@@ -185,5 +177,5 @@ class ZColormapFilter(Filter):
         self.fshader['zval'] = self.vshader['zval']
 
     def _attach(self, visual):
-        super(ZColorMapFilter, self)._attach(visual)
+        super(ZColormapFilter, self)._attach(visual)
         self.vshader['position'] = visual.shared_program.vert['position']
--- a/vispy/visuals/filters/color.py
+++ b/vispy/visuals/filters/color.py
@@ -2,70 +2,72 @@
 # Copyright (c) Vispy Development Team. All Rights Reserved.
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 
-import weakref
-
+from .base_filter import Filter
 from ..shaders import Function, Varying
 from ...color import colormap, Color
 
 
-class IsolineFilter(object):
-    def __init__(self, level=2., width=2.0, antialias=1.0, color='black'):
-        self.fshader = Function("""
-            void isoline() {
-                if ($isolevel <= 0 || $isowidth <= 0) {
-                    return;
-                }
+class IsolineFilter(Filter):
+    __slots__ = ('_level', '_width', '_color', '_antialias')
 
-                // function taken from glumpy/examples/isocurves.py
-                // and extended to have level, width, color and antialiasing
-                // as parameters
-
-                // Extract data value
-                // this accounts for perception,
-                // have to decide, which one to use or make this a uniform
-                const vec3 w = vec3(0.299, 0.587, 0.114);
-                //const vec3 w = vec3(0.2126, 0.7152, 0.0722);
-                float value = dot(gl_FragColor.rgb, w);
-
-                // setup lw, aa
-                float linewidth = $isowidth + $antialias;
-
-                // "middle" contour(s) dividing upper and lower half
-                // but only if isolevel is even
-                if( mod($isolevel,2.0) == 0.0 ) {
-                    if( length(value - 0.5) < 0.5 / $isolevel)
-                        linewidth = linewidth * 2;
-                }
-
-                // Trace contour isoline
-                float v  = $isolevel * value - 0.5;
-                float dv = linewidth/2.0 * fwidth(v);
-                float f = abs(fract(v) - 0.5);
-                float d = smoothstep(-dv, +dv, f);
-                float t = linewidth/2.0 - $antialias;
-                d = abs(d)*linewidth/2.0 - t;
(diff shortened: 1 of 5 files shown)


kne42:
I'm having difficulty running tests on my machine and I'm a bit tired, so I'll go ahead and let Travis handle this one for now. I'll manually investigate it and patch any fixes/cleanup documentation tomorrow, finishing with a rebase.

vispy/visuals/filters/color.py
```diff
@@ -180,10 +185,5 @@ def __init__(self, cmap, zrange=(0, 1)):
         self.fshader['zval'] = self.vshader['zval']
 
     def _attach(self, visual):
-        self._visual = visual
-        vhook = visual._get_hook('vert', 'post')
-        vhook.add(self.vshader(), position=9)
-        fhook = visual._get_hook('frag', 'post')
-        fhook.add(self.fshader(), position=3)
-
+        super(ZColorMapFilter, self)._attach(visual)
```
kmuehlbauer: `ZColormapFilter` was used above.

kmuehlbauer:
@kne42  Great work!

I'm having a hard time running tests locally, too. Unit tests error out, examples tests give mysterious error (although running in a clean cloned vispy repo):
```bash
Cleaning build, pyc, ctags, and cache
python make test examples
Skipping example test: Cannot test examples unless in vispy git directory
```

@djhoese Any idea, what I might be doing wrong? Unfortunately I'm can't further investigate today.

vispy/visuals/filters/color.py
djhoese: Just because it took me more than 5 seconds to figure out what you were saying: the `m` in the class name should be lowercase.

kne42:
I've decided to remove the `__slots__` attribute for now.

I have also encountered a bug with this system when attaching, detaching, and then attaching a filter to the same visual. It does not have this problem if I attach it to two different visuals, however.

I'd also like to hear your thoughts on the actual class design. Do you think it is unintuitive? Is there any other way you would prefer to see it implemented?

kne42:
Travis appears to be failing the Python 3.6 examples test due to an external package comparing `e != None` instead of `e is not None`. Not sure what lead to the chain of events that made that happen, but you can view them here: https://travis-ci.org/vispy/vispy/jobs/433797088#L3413-L3464

djhoese:
Interesting I think cassowary is supposed to come from the internal/bundled version of cassowary within the vispy package. I'm getting really sick of the failing examples on travis.

kmuehlbauer:
@djhoese `cassowary` is installed via pip: https://travis-ci.org/vispy/vispy/jobs/433797088#L1147

kmuehlbauer:
> Any chance you have vispy installed in your environment and are then trying to run the examples from the repository? My guess is that the make script is picking up the vispy that is installed and therefore thinks it isn't in a git repository.

Something along those lines... :man_facepalming: Anyway I'm back at testing. `PyQt5` backend tests (`unit`) work, `osmesa` and `nobackend` many fails with some of thos `'NoneType' object has no attribute 'decode'` errors. 

There are also some examples which depend on PyQt4/PySide, they break. Some spurious breaks (mostly `gloo` stuff).

PR #1534, opened 2018-10-04
passing title as encode('utf-8') in glfwSetWindowTitle

### LINKED ISSUE #1533: GLFW glfw.glfwSetWindowTitle issue
Hello there, 
I am using a library https://github.com/p5py/p5 which is dependent on vispy. When passing a new title as str to glfw.glfwSetWindowTitle only the first letter of the passed string is coming up as the changed title. I was able to find a workaround to this by passing the title by encoding it to utf-8 in
https://github.com/vispy/vispy/blob/0ff2bacef0ec941c039708935a4cbd694ecf02ab/vispy/app/backends/_glfw.py#L327
 like the implementation in
https://github.com/vispy/vispy/blob/0ff2bacef0ec941c039708935a4cbd694ecf02ab/vispy/ext/glfw.py#L483

Now I am confused wether to make a pull request to p5py or vispy about this issue as p5py is accepting title as str. 
Thank You

Gotham13121997:
fixes #1533 .

### DIFF
--- a/vispy/app/backends/_glfw.py
+++ b/vispy/app/backends/_glfw.py
@@ -324,7 +324,7 @@ class CanvasBackend(BaseCanvasBackend):
         if self._id is None:
             return
         # Set the window title. Has no effect for widgets
-        glfw.glfwSetWindowTitle(self._id, title)
+        glfw.glfwSetWindowTitle(self._id, title.encode('utf-8'))
 
     def _vispy_set_size(self, w, h):
         if self._id is None:


djhoese:
@Gotham13121997 Does this work on python 2?

djhoese:
Nevermind, this should be fine on python 2. I'm still confused at why this is needed (why it wasn't seen before), but with an ok from @larsoner I'm fine merging this. You may need to rebase/merge with master to get travis to pass its tests.

larsoner:
>  I'm still confused at why this is needed (why it wasn't seen before)

Sometimes encoding issues only manifest if you have non-ASCII chars. Looks reasonable to me

Gotham13121997:
.encode('utf-8') converts it into bytes. I think we were meant to pass bytes like https://github.com/vispy/vispy/blob/0ff2bacef0ec941c039708935a4cbd694ecf02ab/vispy/ext/glfw.py#L483

I think its related to ctypes.
https://stackoverflow.com/questions/27127413/converting-python-string-object-to-c-char-using-ctypes
Maybe this link can help

Gotham13121997:
@djhoese what is the status on this?

PR #1536, opened 2018-10-12
Fix a typo

JoeXinfa:
As the docstring said "diagonal line".

### DIFF
--- a/vispy/plot/fig.py
+++ b/vispy/plot/fig.py
@@ -27,7 +27,7 @@ class Fig(SceneCanvas):
         >>> from vispy.plot import Fig
         >>> fig = Fig()
         >>> ax = fig[0, 0]  # this creates a PlotWidget
-        >>> ax.plot([[0, 1], [0, 1]])
+        >>> ax.plot([[0, 1], [1, 0]])
 
     See the gallery for many other examples.
 


djhoese:
I'm going to be honest when I first got the email about this PR I thought "this person is trying really hard for hacktoberfest PRs", but now that I've investigated I am pleasantly surprised how bad this example was. For future reference, `[[x1, y1], [x2, y2]]` are two points (because it is a list/array), but `([x1, x2], [y1, y2])` are a tuple of x-coordinates and a tuple of y-coordinates. So the original version of this example should have been a tuple to work, but your version works just fine too. The line is just in a different direction.

djhoese:
I was just about to merge this. What did you do? Now there are two plot lines?

JoeXinfa:
Thanks @djhoese for the quick reply. The original example didn't work for me so I simply thought it's just a typo, now I understand better. I added another line so it has both examples (tuple and list).

djhoese:
Please go back to your original change or fix the original code. I'd prefer if there was only one of these lines (and no comment after) to keep the example simple. The `.plot` documentation and gallery are there for other examples. Otherwise you could split it in to two complete example code blocks...but just stay with one.

JoeXinfa:
Done! Simple is good, and I agree with you. I'm new to making PR (to be honest, this is my 2nd one and the 1st was to Luke's PyQtGraph). Thank you very much for teaching me.

JoeXinfa:
@djhoese Would you merge this? I will delete my branch after merge. Thanks.

Issue #1505, opened 2018-08-10
Updating collections


thomasdeneux:
Hello, I don't find how to update basic values of collections.

For example, i can create a points collection:
```
points = PointCollection("agg", color="shared", size="local")
n = 1000
self.points.append(np.random.normal(0.0, 0.5, (self.n, 3))
    color=np.column_stack((np.random.random((self.n, 3)), np.ones(self.n))),
    size=8)
```

But trying to change its vertices coordinates:
```
points['position'] = np.random.normal(0.0, 0.5, (self.n, 3))
```

Gives me the error:
```
File "/AlphaI/test.py", line 36, in change
  self.points['position'] = pos
File "/vispy/vispy/visuals/collections/collection.py", line 192, in __setitem__
  BaseCollection.__setitem__(self, key, value)
File "/vispy/vispy/visuals/collections/base_collection.py", line 432, in __setitem__
  V[key] = data
File "/vispy/vispy/gloo/buffer.py", line 271, in __setitem__
  raise ValueError("Cannot set non-contiguous data on buffer")
```

Besides, I managed to update the colors:
```
points['color'] = np.column_stack((np.random.random((self.n, 3)), np.ones(self.n)))
canvas.update()  # this involves redrawing the points
```

This was possible only with `color="shared"` when constructing the PointCollection object, while `color="local"` resulted in the same error as above when trying to update the color; but i could not guess appropriately what these keywords mean. I tried a similar `position="shared"` but this was not accepted. In general i am suffering from the lack of documentation!

djhoese:
@rougier Any advice?

thomasdeneux:
> but i could not guess appropriately what these keywords mean

Of course i understand this is about different types of declarations in the shaders. I am not very familiar with those yet.

thomasdeneux:
@rougier First sorry for the delay, i just was en vacances.

So, no it clearly does not work! I show more code below, this is the latest version of the vispy master branch.

It appears that in the `BaseCollection.__setitem__` method there is a call for setting the vertices buffer using a string key (which in my case is `'position'`):

```
403    def __setitem__(self, key, data):
           (...)

420        V = self._vertices_buffer
           (...)

428        # Setting a whole field
429        if isinstance(key, str):
430            # Setting a named field in vertices
431            if key in self.vtype.names:
432                V[key] = data
               (...)
```

But writing in a `DataBuffer` with a string key is actually not handled:

```
265    def __setitem__(self, key, data):
[...]

PR #1538, opened 2018-10-15
Cpu buffer

thomasdeneux:
Following issue [#1505](https://github.com/vispy/vispy/issues/1505), i implemented an optional local CPU buffer in the Buffer class. By default it is not used, but it is used by Collections so that one can set individual fields of the VertexBuffer.

### DIFF
--- a/vispy/gloo/CPUData.py
+++ b/vispy/gloo/CPUData.py
@@ -1,0 +1,176 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2009-2016 Nicolas P. Rougier. All rights reserved.
+# Distributed under the (new) BSD License.
+# -----------------------------------------------------------------------------
+"""
+CPU data is the base class for any data that needs to co-exist on both CPU and
+GPU memory. It keeps track of the smallest contiguous area that needs to be
+uploaded to GPU to keep the CPU and GPU data synced. This allows to update the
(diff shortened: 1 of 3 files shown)


thomasdeneux:
Hello, this pull request was not accepted, was it? Note that after implementing this "local CPU buffer" feature, i encountered more problems with collections, so i did not continue in that direction and now i am only writing my own shaders. So even myself i am not using this feature. Should we drop it or make the effort to go until acceptance of the PR because this might be useful in the future?

djhoese:
What kind of issues did you run in to with collections? If this gets us closer to more glumpy-like collections that are more flexible then I'm all for getting it merged.

Maybe you could merge this with current master and we'll see if the tests pass now? I have other changes (like `CPUData` -> `cpu_data`) that I'd like, but getting tests added and getting them passing is more important.

thomasdeneux:
If i try to do a new pull request i get a button "view pull request" that sends me back here, and no button to create a new pull request. So i assume that i must first close this current PR. I try it.

djhoese:
You can merge master in to your branch and push it to this same existing branch and the PR will update.

thomasdeneux:
oh too late, i have already issued the new pull request...

thomasdeneux:
did not pass the tests again... and i have difficulty reading the test reports to determine where is the problem

PR #1539, opened 2018-10-16
Add connect to line plot

JoeXinfa:
Add kwarg `connect` to lines at plot level.

### DIFF
--- a/vispy/plot/plotwidget.py
+++ b/vispy/plot/plotwidget.py
@@ -279,7 +279,7 @@ class PlotWidget(scene.Widget):
 
     def plot(self, data, color='k', symbol=None, line_kind='-', width=1.,
              marker_size=10., edge_color='k', face_color='b', edge_width=1.,
-             title=None, xlabel=None, ylabel=None):
+             title=None, xlabel=None, ylabel=None, connect='strip'):
         """Plot a series of data using lines and markers
 
         Parameters
@@ -310,6 +310,8 @@ class PlotWidget(scene.Widget):
             The label to display along the bottom axis
         ylabel : str | None
             The label to display along the left axis.
+        connect : str | array
+            Determines which vertices are connected by lines.
 
         Returns
         -------
@@ -321,7 +323,7 @@ class PlotWidget(scene.Widget):
         marker_types, LinePlot
         """
         self._configure_2d()
-        line = scene.LinePlot(data, connect='strip', color=color,
+        line = scene.LinePlot(data, connect=connect, color=color,
                               symbol=symbol, line_kind=line_kind,
                               width=width, marker_size=marker_size,
                               edge_color=edge_color,


djhoese:
Doesn't this perform the same purpose as `line_kind` in the existing interface? I think I would prefer if these two options could be combined, if possible. Also the docstring for that class will need to be updated.

What is your use case for needing the `connect` keyword argument?

JoeXinfa:
Thanks for the quick response. I've been busy at work and just get time reply you. No it's not the same; `connect` defines which vertices are connected by lines while `line_kind` is for the style of lines e.g. solid, dashed or dotted. I updated the docstring accordingly. An example is below.
```
import numpy as np
import vispy.plot as vp

x = np.array([0,1,2,3,4,5])
y = np.array([0,1,2,3,4,5])
data = (x, y)
connect = np.array([[0,1], [2,3], [4,5]])

fig = vp.Fig(show=False)
ax = fig[0, 0]
line = ax.plot(data, symbol='o', connect=connect)

if __name__ == '__main__':
    fig.show(run=True)
```
My use case is plotting the block lines in oil & gas exploration (Gulf of Mexico, for example). The data are vertices (XY coordinates) and segments (which vertices are connected).

djhoese:
Just curious, how would you plot this with matplotlib? Is there an interface that vispy should be adding for plotting lines like yours? Is a plot actually what you want or do you just need the lines on the screen (plotting versus SceneCanvas/Canvas with Visuals)?

JoeXinfa:
I don't know if matplotlib can plot this or not (connect is the key). With the change in this PR, vispy is enough for my need. We usually don't just plot the block lines, but overlay them with lots of other things such as coastlines, wells, seismic surveys, leases, etc. Therefore, I need the scene/canvas that I can add/remove items and pan/zoon interactively. Thanks.

djhoese:
Ok but you want the plot axes, ticks, axes labels, etc? I'm not completely against merging this PR I just want to make sure things are being used as intended and if not figure out what use cases aren't being handled by existing functionality.

JoeXinfa:
Yes I do. This PR is quite simple, just remove the hard-coded `connect` in `scene.LinePlot` and expose to `def plot()` so user can set at that level. To my knowledge, it is only better than before. Is there any reason that people would be against it?

djhoese:
@JoeXinfa You are right, this only adds more functionality for the user. My hesitation is only because this reveals a low-level interface in a high-level interface. So far users haven't needed this and if this functionality isn't available in matplotlib then I'm wondering if it requires other changes. Now that I think about it more I suppose this is possible in matplotlib as separate line segments which in vispy would require separate LineVisuals and would have terrible performance.

JoeXinfa:
That's true. Before I found this `connect`, I tried plot the thousands of segments as separate LineVisuals, which choked on memory. Thank you.

PR #1544, opened 2018-11-05
Use proper exception NotImplementedError

asnt:
I think NotImplementedError was intended.

See:
https://docs.python.org/3.7/library/exceptions.html#NotImplementedError
https://docs.python.org/3.7/library/constants.html#NotImplemented


djhoese:
Sorry about this. I somehow didn't see the email for this. This fix has now been included in my own PR in #1546.

PR #1546, opened 2018-11-22
Fix flake errors

djhoese:
This PR fixes flake errors that have been causing travis to fail. As part of this I moved the exclusions and ignored rules to setup.cfg so they are a little less hidden (or at least I think so). I also added W504 to the list of ignored rules which I think is a default ignore in the recent versions of flake8.

This PR also changes a `raise NotImplemented` to a `raise NotImplementedError` to satisfy flake8.

@mkkb There are still errors in `text_visual.py` that you semi-recently made changes to. The main issues are a `\e` and a `\?`. Any idea if those are necessary?

### DIFF
--- a/examples/basics/visuals/text_visual.py
+++ b/examples/basics/visuals/text_visual.py
@@ -55,7 +55,7 @@ class Canvas(app.Canvas):
         big_test_string += '\t-\t\\n\n'
         big_test_string += '\t-\t\\v\n'
         big_test_string += '\t-\t\\t\n'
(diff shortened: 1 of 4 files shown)


mkkb:
@djhoese I don't understand why those would give errors. But anyways I agree that in this context the intent of the escape sequences is only to format the text, thus \n, \t, and \v (and maybe \" and \'?) is the only ones that are necessary. So the others could surely be removed from the example.

mkkb:
Yep, you are right.

PR #1547, opened 2018-11-24
Add element-wise rotation to TextVisual

EtienneCmb:
For now, only a unique rotation can be set to `TextVisual`. This PR allows to set an individual rotation to each text element.

```python
import numpy as np

from vispy import scene
from vispy import app

# Camera and scene canvas :
cam = scene.cameras.PanZoomCamera(rect=(-2, -1.5, 4, 3))
sc = scene.SceneCanvas(bgcolor='white')


n_s = 20
theta = np.linspace(0, 2 * np.pi, n_s, endpoint=False)
text = ['Text_%i' % i for i in range(n_s)]
xy = np.c_[np.cos(theta), np.sin(theta)]

text_v = scene.visuals.Text(text=text, pos=xy, parent=wc.scene,
                            rotation=np.rad2deg(theta[::-1]))

sc.show()
app.run()
```

image

### DIFF
--- a/vispy/visuals/tests/test_text.py
+++ b/vispy/visuals/tests/test_text.py
@@ -19,6 +19,7 @@ def test_text():
         
         text.text = ['foo', 'bar']
         text.pos = [10, 10]  # should auto-replicate
+        text.rotation = [180, 270]
         try:
             text.pos = [10]
         except Exception:
--- a/vispy/visuals/text/text.py
+++ b/vispy/visuals/text/text.py
@@ -482,7 +482,7 @@ class TextVisual(Visual):
 
     @rotation.setter
     def rotation(self, rotation):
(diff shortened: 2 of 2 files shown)


EtienneCmb:
About performances, can you tell me how to measure it?

djhoese:
I think you can add `--vispy-fps` to the command line when running a script. Show a text example with a camera that doesn't use rotation before and after your changes and see if there is any difference in the FPS. I doubt there would be any noticeable difference. There could be a different in some edge cases like having a ton of text elements, but I don't think we have to worry about that.

EtienneCmb:
Ok, I tested on the master branch and on my PR and I don't see a difference in the fps (values between 1.9 and 34 FPS in both cases)

Issue #1549, opened 2018-12-03
Visuals cube and box duplicate?


JoeXinfa:
Looks to me visuals/cube.py and visuals/box.py are very similar. Is there opportunity to merge them and avoid confusion? Thanks.

larsoner:
No idea

kmuehlbauer:
resolved by #1550

PR #1550, opened 2018-12-03
Remove cube visual in favor of box visual

JoeXinfa:
Check issue https://github.com/vispy/vispy/issues/1549

### DIFF
--- a/vispy/visuals/cube.py
+++ b/vispy/visuals/cube.py
@@ -4,12 +4,11 @@
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 # -----------------------------------------------------------------------------
 
-from ..geometry import create_cube
-from .mesh import MeshVisual
-from .visual import CompoundVisual
+import warnings
+from .box import BoxVisual
 
 
-class CubeVisual(CompoundVisual):
+class CubeVisual(BoxVisual):
     """Visual that displays a cube or cuboid
 
     Parameters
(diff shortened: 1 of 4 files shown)


djhoese:
Thanks for the quick pull request on this. I could be wrong, but github shows that you deleted both the cube and the box modules. We need at least one of them 😉 

Additionally, while you may not need both of them some people may have been using Box and expect the keyword arguments as they are. I see a couple of options that seem like good workarounds:

1. The keyword arguments currently available in Box (being deprecated) get copied to Cube. The Box class/visual would then just point to Cube (`Box = Cube`). The copied kwargs could either become the new way of creating the Cube or they could be accepted and have deprecation warnings associated with them:

```python
import warnings
warnings.warn("The 'size' parameter is no longer supported, use 'width', 'height', and 'depth' instead.", DeprecationWarning)
```

2. Keep Box as it is, but add a deprecation warning inside the `__init__`.
3. Take out the inner logic of Box and make it a subclass of Cube. Add a DeprecationWarning and then call Cube's `__init__`.

If `Box` provides any functionality or its kwargs make it easier to create a cuboid because of how they are structured then we should consider copying them over to `Cube`. The key point overall though is we don't want to cause peoples code to stop working from these changes if possible. DeprecationWarnings and migration paths are preferred.

kmuehlbauer:
@djhoese I just had a quick look at the diffs. Seems that only `CubeVisual` is removed. No changes to `BoxVisual.

kmuehlbauer:
@JoeXinfa example basics/scene/cube.py needs adaption too. This is why Travis is failing.

djhoese:
Thanks @kmuehlbauer I realize now I saw the deleted example and thought it was the other visual module.

JoeXinfa:
@djhoese @kmuehlbauer Thank you both! I edited the commit per your suggestions. The `visuals/cube.py` is kept by using subclass and deprecation, which allows the examples unchanged.

vispy/visuals/cube.py
```diff
@@ -29,41 +28,13 @@ class CubeVisual(CompoundVisual):
...
-        return self._border
-
-    @border.setter
-    def border(self, border):
-        self._border = border
+            width = height = depth = size
+        BoxVisual.__init__(self, width=width, height=height, depth=depth,
+                           vertex_colors=vertex_colors,
+                           face_colors=face_colors, color=color,
+                           edge_color=edge_color, **kwargs)
```
kmuehlbauer: Please add empty line at end of file, to fix flake error.

JoeXinfa: how about now?

JoeXinfa:
@djhoese has this been merged?

djhoese:
No it has not. I will merge it later today after I re-review it. Were there any other uses of `CubeVisual` in the examples or other code in vispy?

I'm mostly worried about the properties not being the same between Cube and Box visuals.

djhoese:
@JoeXinfa I am ok with the changes that have been done with the visual but I agree with @kmuehlbauer that the examples should be updated to use the Box instead of the Cube. From what I can find these include:

- [ ] `examples/basics/visuals/cube.py` (rename to `rotating_box.py`?)
- [ ] `examples/basics/visuals/arcball.py`
- [ ] `examples/basics/scene/cube.py` (rename to `box.py` or maybe `turntable_box.py`?)

JoeXinfa:
@djhoese Thanks for the suggestions. I've updated the three examples. Please review and merge.

Issue #1554, opened 2018-12-13
Link two flycameras


SuyiWang:
The mouse rotation for two linked FlyCameras like the following is not working properly.
```
    vb1 = scene.widgets.ViewBox(border_color='yellow', parent=canvas.scene)
    vb2 = scene.widgets.ViewBox(border_color='blue', parent=canvas.scene)
    # Creates flycamera for two side-by-side scene. Setup steps omitted.
    camfly1, camfly2 = FlyCamera(), FlyCamera()
    vb1.camera = camfly1
    vb2.camera = camfly2
    vb1.camera.link(vb2.camera)
```
Namely, the current mouse rotation for linked FlyCameras has an inertia effect. For example, let's say we have two movements with left button down. Move 1: Dragging the screen from `mouse_down_pos` to `mouse_pos_A`. Move 2: Without releasing left mouse button, dragging the screen from `mouse_pos_A` back to `mouse_down_pos`. Without linked camera, the FlyCamera always looks at where mouse points at, which means it starts to look back to `mouse_down_pos` in Move 2. However, with FlyCameras linked, those cameras never look back, unless one keeps Move 2 and moves mouse pass `mouse_down_pos`.

I also noticed that at Line 55-56
```
    # Linking this camera likely not to work very well
    _state_props = PerspectiveCamera._state_props + ('rotation', )
```

I have found that this issue is caused by how set_state is implemented (BaseCamera.py, Line 322 ). It copies all properties from one camera to another. 
```
        for key, val in D.items():
            if key not in self._state_props:
                raise KeyError('Not a valid camera state property %r' % key)
            setattr(self, key, val)
```
However in FlyCamera.py the `rotation` property is: 
```
    @property
    def rotation(self):
        """ Get the full rotation. This rotation is composed of the
        normal rotation plus the extra rotation due to the current
        interaction of the user.
        """
        rotation = self._rotation2 * self._rotation1
        return rotation.normalize()

    @rotation.setter
    def rotation(self, value):
        assert isinstance(value, Quaternion)
        self._rotation1 = value
```
So what happened for linked camera is that it reads rotation2*rotation1 as rotation from one camera and writes to rotation1 in another. This rotation1 from the other camera will later affect rotation1 in current camera.
I have a quick fix in FlyCamera class, which asks set_state to copy _rotation1 and _rotation2 instead of rotation property.
```
    # In Fly.py, class FlyCamera()
[...]

PR #1557, opened 2018-12-20
Fixed a bug in linking two flycameras

SuyiWang:
See issue #1554. The bug is fixed as discussed.
- Replaced actual `rotation.setter` function by a warning of using `rotation1.setter instead`
- Added property `rotation1` and `rotation1.setter`
- Added property `rotation2` and `rotation2.setter`
- Normalizes Quaternion before assigning.

### DIFF
--- a/vispy/scene/cameras/fly.py
+++ b/vispy/scene/cameras/fly.py
@@ -52,8 +52,8 @@ class FlyCamera(PerspectiveCamera):
 
     """
 
-    # Linking this camera likely not to work very well
-    _state_props = PerspectiveCamera._state_props + ('rotation', )
+    # Using _rotation1 and _rotation2 for camera states instead of _rotation
+    _state_props = PerspectiveCamera._state_props + ('rotation1', 'rotation2')
 
     def __init__(self, fov=60, rotation=None, **kwargs):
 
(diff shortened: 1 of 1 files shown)


djhoese:
@SuyiWang Sorry about losing track of this. I restarted the failing test and we'll see if it passes now. Is there anything else you think needs to be done with this?

I'm not sure I fully understand every change here or if there is a better way to do it, so I'm ok merging this as long as @kmuehlbauer and @larsoner don't have any reservations.

kmuehlbauer:
@djhoese @SuyiWang I've briefly looked over the code and did not spot any problems so far. But using the example provided in #1554 I do not see any inertia effect whatsoever. I also can't see any difference using the new implementation. Good chance I do not fully understand where to look.

djhoese:
@SuyiWang Do you have any free time to look at this more? Could you better describe what you're seeing (I *think* I was able to reproduce this on my machine) to @kmuehlbauer so we can test this further.

SuyiWang:
Using the example in the comment dated on "Dec 13, 2018" would reproduce the bug. It happens when you drag the screen using left mouse key.
An example is also provided in the following video.
https://youtu.be/ivV99JuNleI

djhoese:
Coming back to this after waaaay too long. Tested it locally and seems to work from the example in #1554. Thanks @SuyiWang!

SuyiWang:
Great! Thank you for testing!

SuyiWang:
Great! Thank you for testing and confirming the change!

PR #1558, opened 2018-12-21
Add element-wise colors to TextVisual

cimbi:
Ability to pass different colors for different parts of text in a list similar to pos argument


vispy/visuals/text/text.py
```diff
@@ -474,7 +477,17 @@ def color(self):
 
     @color.setter
     def color(self, color):
-        self._color = Color(color)
+        if isinstance(color, str):
+            self._color = ColorArray(Color(color))
+        else:
+            color = np.atleast_2d(color).astype(np.float32)
+            if color.shape[1] == 3:
+                color = np.concatenate((color, np.ones((color.shape[0], 1),
```
djhoese: Doesn't `ColorArray` handle this case? If not, should it?

djhoese: Judging by this https://github.com/vispy/vispy/blob/master/vispy/color/color_array.py#L91-L97, this chunk of code shouldn't be needed. Or am I missing what this is doing?

cimbi: no, you are correct, edited in f372f1e8848dd1eae44cf1d6e189b26c8caf1d26

vispy/visuals/text/text.py
```diff
@@ -474,7 +477,14 @@ def color(self):
 
     @color.setter
     def color(self, color):
-        self._color = Color(color)
+        if isinstance(color, str):
+            self._color = ColorArray(Color(color))
+        else:
+            color = np.atleast_2d(color).astype(np.float32)
+            if color.shape[1] not in [3, 4]:
```
djhoese: Doesn't ColorArray handle all of these cases? `str` and not having elements with length `3` or `4`?

vispy/visuals/text/text.py
djhoese: Thanks. I added another question about this handling in a separate comment.

vispy/visuals/text/text.py
cimbi: indeed it does! thanks. fixed in d7057b6a4490eacb7d8d2b04e890b73a47b8b3f3

vispy/visuals/text/text.py
```diff
@@ -559,6 +563,26 @@ def _prepare_draw(self, view):
...
+                text = ''.join(text)
+            else:
+                repeats = [4 * len(text)]
+            n_text = len(repeats)
+            color = self.color.rgba
+            if color.shape[0] < n_text:
+                color = np.repeat(color,
+                                  [1]*(len(color)-1) + [n_text-len(color)+1],
+                                  axis=0)
+            color = np.repeat(color[:n_text], repeats, axis=0)
```
djhoese: Would it be a bad idea to try to re-cache these new color arrays back in to `self._color`? Maybe not the one repeated for each RGB color but the one that is made the same length as the text object? Or maybe that would cause confusion for the user that `.color` is not the same as what they set it to? Thoughts?

cimbi: I personally have a bad experience with this when I used clim='auto' in the Image visual and when I tried to read it back I got an array. This caused problems in my PyQt application. It is not a big issue but I would keep the value as it was set. It is just my personal preference though.
From the other point of view - what would be the benefit of re-caching the new color arrays?

djhoese: I suppose I was concerned that some of these color calculations would be recalculated every time the text was changed and perhaps they didn't need to be. Just worried about performance.

cimbi: Right. It depends on the length of the text I guess. I could test this and we can decide based on the test results?

djhoese: If you have the time, that would be great. Reducing number of operations that are required pre-draw or on-draw is a benefit.

cimbi: I ran the test as follows - I created a long random text updated it fast with and without re-caching the color looking for drop in FPS but did not see any. So I don't think there is any significant performance problem if we recalculate color every time the text changes. Although I have to say the FPS drop pretty fast with larger text. But the bottleneck seems to be somewhere else.

Let me know if you think of another way to test this.

djhoese: That is good enough for me I think. It is worth merging this and getting more eyes on it. I'll try merging this with current master and see if that fixes the test failures.

PR #1565, opened 2019-01-28
Fix argument validation in PanZoomCamera.zoom method

mthrok:
The docstring of `PanZoomCamera.zoom` method says when `center` argument is not provided, it uses the current center, but argument validation is performed before such substitution thus the current code cannot work without explicitly passed `center` argument.

This PR fixes this by flipping the order of default value substitution and parameter validation.


mthrok:
Thank you for the update.

Issue #1566, opened 2019-01-31
GLSL - code duplication


kmuehlbauer:
While digging into #1563 I found that several glsl-functions have duplicates in the code base.

Here are some examples

1. signed segment distance [here](https://github.com/vispy/vispy/blob/master/vispy/glsl/arrows/common.glsl) and [here](https://github.com/vispy/vispy/blob/master/vispy/glsl/math/signed-segment-distance.glsl)
2. arrows/arrowheads [here](https://github.com/vispy/vispy/blob/master/vispy/glsl/arrows/stealth.glsl) and [here](https://github.com/vispy/vispy/blob/master/vispy/glsl/arrowheads/stealth.glsl)
3. definition of markers [here](https://github.com/vispy/vispy/tree/master/vispy/glsl/markers) and [here](https://github.com/vispy/vispy/tree/master/vispy/glsl/markers)

Some definitions are completely equivalent, others have some differences regarding eg. `size`. There are sure reasons why things are done that way, but nevertheless is would be a good idea to have all GLSL functions in the glsl-subfolder. Thoughts?

djhoese:
I'd like to hear @rougier's opinion on this. I wouldn't be surprised if these have already been cleaned up in glumpy. I'm assuming most of these came from glumpy originally.

rougier:
I think there's no such duplicate in glumpy (need to check) and I'm not sure what it origin of such duplicatino in vispy.

kmuehlbauer:
@rougier I meant, that there are eg. this file `vispy/glsl/arrows/common.glsl` and this file `vispy/glsl/math/signed-segment-distance.glsl`. Both include the same formulation of the signed segment distance. And this example is also in glumpy, I just checked. Just curious.

Issue #1563, opened 2019-01-22
Plotting wind barbs


sjoro:
a feature request:
i'm using vispy to plot meteorological products. it would be great to have have a marker or a way to plot wind barbs. this type of visualization is available in [matplotlib](https://matplotlib.org/examples/pylab_examples/barb_demo.html)

djhoese:
Hi @sjoro. I think we had a user a year or so ago bring up wind barbs on [gitter](https://gitter.im/vispy/vispy) or in an issue. I can't find the discussion on github anywhere, but I think someone has done something like this. I feel like @kmuehlbauer was in on the conversation too. I feel like I remember seeing some shader code for it too.

The "best" idea would be to have a custom Markers visual where direction and magnitude could be used inside the shader to draw the barbs. I'm not sure how possible this is. But it would be better than having to generate the individual line segments for each point.

kmuehlbauer:
@sjoro I think this was discussed before, unfortunately I wasn't able to dig this up.

I have the same impression as @djhoese that a custom MarkerVisual is the best way to go. And this would indeed be a nice addition. How many different windbarbs are out there?

sjoro:
thanks for the info!  shame that the old discussion could not be found.
regarding the barbs, the representation depends on the wind speed. so, as individual markers there's many different ones. [https://blog.mapbox.com/mapping-wind-barbs-to-show-speed-and-direction-4d3078add03d](Here) is a good example. would each of these barbs need it's own Marker or could it be build on-fly?

kmuehlbauer:
@sjoro Currently `Marker` just get an array of positions where to plot. You could create different markers for the different wind speed ranges. But this won't work here as you also have wind direction as parameter to take into account.

Nevertheless, it should be possible to also provide wind speed and direction to calculate the barb from those values within the shader. This would need some tweaking in `MarkerVisual`-class but should be possible. We also might consider subclassing MarkerVisual for that purpose. WDYT @djhoese @larsoner? Up for an adventure, @sjoro? PR welcome.

sjoro:
i'm happy to assist in anyway i can, but please keep in mind that i have no openGL programming skills :)

kmuehlbauer:
Means, for a range up to 200 knots we need only 45 different windbarb polygons. Depending on the direction of wind the windbarb polygon just need to be rotated accordingly. For use on the southern hemisphere, we just need to flip on x-axis. Could all be assembled in one dedicated function.

[...]

PR #1567, opened 2019-01-31
Add WindbarbVisual

kmuehlbauer:
This is a follow up PR from #1563. 

This adds a special WindbarbVisual as well as an example. The Visual might need some tweaking here and there. Would be nice if someone can test and review. There is also shader code which can be moved to the gsls subfolder (see also #1566).

Ping @sjoro and @rougier.

@larsoner @djhoese How and where should I add tests for this? Is the example tested anyway?

### DIFF
--- a/vispy/visuals/__init__.py
+++ b/vispy/visuals/__init__.py
@@ -27,6 +27,7 @@ from .line import LineVisual, ArrowVisual  # noqa
 from .linear_region import LinearRegionVisual  # noqa
 from .line_plot import LinePlotVisual  # noqa
 from .markers import MarkersVisual, marker_types  # noqa
+from .windbarb import WindbarbVisual # noqa
 from .mesh import MeshVisual  # noqa
 from .plane import PlaneVisual  # noqa
 from .polygon import PolygonVisual  # noqa
--- a/vispy/visuals/__init__.py
+++ b/vispy/visuals/__init__.py
@@ -45,3 +45,4 @@ from .xyz_axis import XYZAxisVisual  # noqa
 from .border import _BorderVisual  # noqa
 from .colorbar import ColorBarVisual  # noqa
 from .graphs import GraphVisual  # noqa
+from .windbarb import WindbarbVisual  # noqa
--- a/vispy/visuals/__init__.py
+++ b/vispy/visuals/__init__.py
@@ -27,7 +27,6 @@ from .line import LineVisual, ArrowVisual  # noqa
 from .linear_region import LinearRegionVisual  # noqa
 from .line_plot import LinePlotVisual  # noqa
 from .markers import MarkersVisual, marker_types  # noqa
-from .windbarb import WindbarbVisual # noqa
 from .mesh import MeshVisual  # noqa
 from .plane import PlaneVisual  # noqa
 from .polygon import PolygonVisual  # noqa
--- a/examples/basics/visuals/windbarb_quiver.py
+++ b/examples/basics/visuals/windbarb_quiver.py
@@ -14,6 +14,7 @@ from vispy import app
 from vispy.visuals import WindbarbVisual
 from vispy.visuals.transforms import NullTransform
 
+
 class Canvas(app.Canvas):
     def __init__(self):
         app.Canvas.__init__(self, title="Windbarb plot", keys="interactive",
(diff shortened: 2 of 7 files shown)


kmuehlbauer:
@sjoro @djhoese This is now ready for review.

sjoro:
i'll try and test this hopefully next week.

djhoese:
@sjoro Did you ever get a chance to test this?

kmuehlbauer:
@djhoese Where should I add tests here? Via image comparison? Thoughts?

djhoese:
I'd be ok with image comparison. If you can make the image/buffer small enough maybe do raw array comparison without having to add images to vispy-data?

kmuehlbauer:
@djhoese  Yes, if I only consider one barb this might work. Let me try this.

sjoro:
sorry guys, i haven't tested this yet. hope to have some time soon.

kmuehlbauer:
@djhoese Raw Image comparison isn't feasible, but I think utilizing `numpy.histogram` should do the trick. I rebased on current master.

kmuehlbauer:
So the histogram differs from my local histogram, but in a consistent way. I just might use those values then.

kmuehlbauer:
Some OT:
@djhoese @larsoner One thing which I'm thinking about is using something like `phash` (eg. https://github.com/JohannesBuchner/imagehash) to test creation of images. That way it would be possible to get rid of the `test-data` repo while still test for correctness of created images via hashes. If there is interest, I would check further and try to implement that alongside the current implementation.

larsoner:
Not sure it's worth it. We wouldn't have to download things on CIs but is this a problem? Is there some other problem it would fix?

A few concerns that occur to me off the top of my head:

- The image hashing would need to be sufficiently flexible in the ways that our current checkers are
- When there are failures hopefully it should be at least as informative as to why
- If we switch, we are stuck with whatever ways the image hasher is flexible with no ways to add new types of flexibility
- When there are failures, currently you can get a nice side-by-side of the desired and actual plus diff, this would not be possible without having `test-data`

kmuehlbauer:
@larsoner Thanks, this all makes sense. 

Then I'll need to add the image for this visual here to the vispy-test repo. I can't make the image(s) small enough to create raw array tests and the `histogram` approach is obviously not robust enough.

djhoese:
I agree with @larsoner. As much as it feels wrong to me to have a separate repository for storing test/compare images, having `--vispy-audit` functionality (I think that's the flag) for comparing images when tests fail is pretty nice.

vispy/visuals/__init__.py
```diff
@@ -44,3 +45,4 @@
 from .border import _BorderVisual  # noqa
 from .colorbar import ColorBarVisual  # noqa
 from .graphs import GraphVisual  # noqa
+from .windbarb import WindbarbVisual  # noqa
```
djhoese: Why is the visual imported twice?

kmuehlbauer: Not sure, two are better than one? I'll clean-up.

kmuehlbauer:
windbarb

kmuehlbauer:
@djhoese Is there something wrong with travis? I rebased on latest master, but travis seems not to start.

larsoner:
I thought that closing and reopening might fix it but it didn't. Let's just make sure to check the Travis status before merge

kmuehlbauer:
@larsoner It worked! Let's see, if the image test works correctly, hopefully.

kmuehlbauer:
@larsoner Do you have any idea, why the image test might fail. It worked locally...

larsoner:
Travis:
```E                   Exception: Test standard /Users/travis/Library/Application Support/vispy/test_data/visuals/windbarb.png does not exist.
```

Did you follow all steps in https://github.com/vispy/vispy/blob/master/vispy/testing/image_tester.py#L4 ?

kmuehlbauer:
Did not increment the tag...

larsoner:
Do you have permissions to do it, or do you need me to?

kmuehlbauer:
AFAIKT I would also need to increase that number in https://github.com/vispy/vispy/blob/6894b3fa81c043cab545c4e5350ce99df86bf89c/vispy/testing/image_tester.py#L367

I'll try to add the tag and do the other change on top of this PR.

kmuehlbauer:
Created the tag in test-data and changed it accordingly in `get_test_data_repo`. :crossed_fingers:

kmuehlbauer:
Nice, now the green travis build is linked correctly in the checks above. :green_heart:

kmuehlbauer:
@larsoner Thanks for the help with the test-data. @djhoese This is ready for final review/merge.

djhoese:
Looks good to me. Great work!

sjoro:
just tried your windbarb example @kmuehlbauer, looks great! thank you for your great work!

kmuehlbauer:
@sjoro Glad you find it useful!

PR #1569, opened 2019-02-01
FIX: fix pyopengl and osmesa

### LINKED ISSUE #1568: pyopengl issues
See for further discussion:

https://github.com/mcfletch/pyopengl/issues/19
https://github.com/conda-forge/pyopengl-feedstock/issues/16

First I installed pyopengl and pyopengl-accelerate from latest github master (https://github.com/mcfletch/pyopengl). Then I applied the following patch to vispy:
[FIX__pyopengl.patch.txt](https://github.com/vispy/vispy/files/2817944/FIX__pyopengl.patch.txt)

That fixed the pyopengl-tests. Did not had the time to dig further.

kmuehlbauer:
This fixes #1568 alongside with https://github.com/mcfletch/pyopengl/pull/20
- [x] fix glGetActiveAttrib in _pyopengl2.py
- [x] fix travis conda-forge dependencies 
- [x] fix backend issues (osmesa, glfw)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,9 @@ matrix:
     # to avoid having the linker load the wrong libglapi.so which would cause
     # OSMesa to crash
     - env: PYTHON=3.6 DEPS=osmesa TEST=osmesa
+      addons:
+        apt:
+          packages:
 
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,11 +51,7 @@ matrix:
     # has (on-screen) OpenGL installed, we need to setup environment variable
     # to avoid having the linker load the wrong libglapi.so which would cause
     # OSMesa to crash
-    - env: PYTHON=2.7 DEPS=full TEST=osmesa
-      addons:
-        apt:
-          packages:
-            - *full_apt
+    - env: PYTHON=3.6 DEPS=osmesa TEST=osmesa
 
 
 before_install:
@@ -70,6 +66,9 @@ before_install:
     - export PATH=~/anaconda/bin:$PATH
     - conda config --add channels conda-forge
     - conda update --yes --quiet conda
+    # remove defaults, depend only on conda-forge
+    - conda config --remove channels defaults
+
 
     - SRC_DIR=$(pwd)
     # file size checks run on minimal build for time
@@ -100,11 +99,13 @@ before_install:
 
 
 install:
-    # Install numpy, flake
+    # create basic conda-forge environment
     - conda create -n testenv --yes --quiet pip python=$PYTHON;
     - source activate testenv;
-    - conda install --yes --quiet numpy$NUMPY nose pytest cython;
-    - pip install -q coveralls pytest-cov pytest-sugar flake8
+    # add needed packages common to all backends/tests
+    - conda install --yes --quiet numpy scipy nose pytest cython coveralls pytest-cov flake8;
+    # install pytest-sugar via pip, until 0.9.2 is available at conda-forge
+    - pip install -q pytest-sugar;
 
     # On Python3, install system-wide copies of bundled libraries instead
     # Also install PyQt4, imaging (PIL or pillow), scipy, mpl, egl
@@ -118,21 +119,27 @@ install:
     # WX requires OSMesa (mesa on conda) which has typically been an
     # additional test environment. With llvm=3.3 the combination of
     # EGL and mesa causes segmentation faults. See issue #1401.
+    # conda-forge does not provide qt4, so we try defaults on darwin
     - if [ "${DEPS}" == "full" ]; then
-        conda install --yes pyopengl scipy numpy$NUMPY networkx;
-        pip install -q numpydoc PySDL2;
+        conda install --yes pyopengl networkx numpydoc pysdl2;
         if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
-          conda install --yes matplotlib jupyter pyqt=4 pillow decorator six scikit-image;
+          conda install --yes matplotlib jupyter pyqt=4 pillow decorator six scikit-image glfw;
           if [ "${PYTHON}" == "3.6" ]; then
-            pip install -q freetype-py husl pypng cassowary imageio;
+            conda install --yes freetype-py imageio;
+            pip install -q husl pypng cassowary;
             rm -rf ${SRC_DIR}/vispy/ext/_bundled;
           else
-            pip install -q mock;
+            conda install --yes mock;
           fi;
         else
-          conda install --yes matplotlib jupyter pyqt=4 scikit-image;
-          pip install -q mock;
+          conda install --yes -c defaults pyqt=4;
+          conda install --yes matplotlib jupyter scikit-image mock;
         fi;
+      fi;
+
+     # Install OSMesa
+    - if [ "${TRAVIS_OS_NAME}" == "linux" ] && [ "${DEPS}" == "osmesa" ]; then
+        conda install --yes mesalib libglu;
       fi;
 
     # Install vispy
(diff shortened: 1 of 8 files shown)


kmuehlbauer:
needs https://github.com/conda-forge/pyopengl-feedstock/pull/17

kmuehlbauer:
So, the pyopengl errors are gone. Only the flake8 errors and one example error.

kmuehlbauer:
I cherry-picked the flake8-fixes from #1567 to see how this works. Finally only the osmesa-test is breaking, which might also relate to the new pyopengl-release. @djhoese @larsoner Any thoughts?

djhoese:
@kmuehlbauer Based on [this line](https://travis-ci.org/vispy/vispy/jobs/487809966#L3426) it looks like maybe the original build of osmesa that we are using (which @larsoner built iirc) is no longer compatible with travis' OS images...maybe?

```
OSError: /usr/lib/x86_64-linux-gnu/mesa/libGL.so.1: undefined symbol: _glapi_tls_Dispatch
```

djhoese:
Also, unless I missed it, shouldn't be force the pyopengl version for these changes? I'm guessing the GL function changes are backwards compatible?

kmuehlbauer:
@djhoese Yes, osmesa might be outdated, but this line is also present in former green builds. But much has changed in Travis and conda-forge stack.

I did not get your last comment. Do you mean we should pin to the latest pyopengl conda-forge version?

djhoese:
I meant we should force the version in setup.py since not all users are using conda-forge. I would be ok having vispy 5.0.4+ require this newer version of pyopengl. However, I'm realizing now that pyopengl isn't even a requirement in setup.py.

kmuehlbauer:
Ah, OK, for this we have to wait for the PR at pyopengl being pulled in and a new version released on PyPI. For the moment I patched the current version in conda-forge. @almarklein merged it, so we are able to use it, but only within conda-forge.

Yes, I would follow your suggestion.

kmuehlbauer:
On conda-forge there is `mesalib` which is actually built as offscreen. Could we use this instead?

I tried locally but got lost in the vispy machinery. Also running tests is getting hard and harder (we have had one issue I recall). Sigh.

djhoese:
That could be worth a shot. When you say you got lost in the vispy machinery do you mean the tests were failing and you couldn't figure it out because of vispy internals or you couldn't figure out how to modify travis to use the mesalib package? If the former, would you be willing to make a separate branch (in the main repository even) and/or PR that swaps out our custom osmesa with the `mesalib` package?

kmuehlbauer:
For osmesa there are several env variables which need to be set (indicated in the examples/offscreen). There are some hooks inside vispy which automate that passing of correct osmesa-lib. I do not understand atm why there (in the current osmesa) is also liblgu provided and where we should get it otherwise.

The testing-machinery is broken afaics, using the make-way it collects all tests found in my conda-env. Not sure why this happens. Other tests fail with some texts sdf import error, which I know comes from starting within vispy root folder. Maybe I'm doing something wrong. 

What I really need is an indication, that the osmesa-lib was used in the tests. I did run the offscreen-example and the osmesa.png was created.

Anyway, it is a good idea to make a plain simple branch using only osmesa as you suggested. This will take a while, I'll not have time before next week.

kmuehlbauer:
@larsoner Is it a difference between libOSMesa.so and libOSMesa32.so? The latter is built with 

```bash
./configure \
    --prefix=$PREFIX \
    --disable-gles1 \
    --disable-gles2 \
    --disable-va \
    --disable-gbm \
    --disable-xvmc \
    --disable-vdpau \
    --enable-shared-glapi \
    --enable-texture-float \
    --disable-dri \
    --with-dri-drivers= \
    --with-gallium-drivers=swrast \
    --disable-egl \
    --with-egl-platforms= \
    --enable-gallium-osmesa \
    --disable-glx \
    --enable-llvm \
    --disable-llvm-shared-libs \
    --with-osmesa-bits=32
```
Do you remember how the vispy-osmesa was built?

kmuehlbauer:
I reverted the osmesa changes and will create another testing PR for this .

kmuehlbauer:
Please follow osmesa here: #1570

kmuehlbauer:
@djhoese How do we proceed with `pyopengl`? 

* Merging and not waiting for new `pyopengl`-release might break workflows based on vispy github-master.  
* Not merging might break workflows using conda-forge and vispy github-master.
* conda-forge vispy is currently broken (although we could patch it there)

For a real solution we would need a merge and a new vispy version and an updated `pyopengl`.

djhoese:
If I recall correctly, since pyopengl is not a hard dependency for vispy (setup.py or conda) then most users probably won't have an issue with the newer pyopengl. What if we merge this in to master and cherry pick it on to 0.5 for a 0.5.4 release? Maybe pyopengl's version should be listed as an extra_requires in setup.py so we at least know the version number that is required...except I still think most users won't need the pyopengl GL backend so maybe it would be best to not force it (even in an extra_requires).

kmuehlbauer:
@djhoese I'm pro merge, since this gets our ci-workflow back on track.

kmuehlbauer:
@djhoese Green lights everywhere :green_heart:

I changed travis a bit to use conda-forge wherever possible. Also fixed osmesa and glfw backend installation. Not sure why glfw isn't picked up in py3.6 tests, but for 2.7 it works. We can track this down later, I think.

If you approve, we can merge and get along with the pending PR's, finally. Might also test the pyqt5 stuff with these changes.

kmuehlbauer:
Have to put this on hold for now. Somehow the system-libraries made it into the osmesa-build. :angry:

.travis.yml
```diff
@@ -51,11 +51,10 @@ matrix:
     # has (on-screen) OpenGL installed, we need to setup environment variable
     # to avoid having the linker load the wrong libglapi.so which would cause
     # OSMesa to crash
-    - env: PYTHON=2.7 DEPS=full TEST=osmesa
+    - env: PYTHON=3.6 DEPS=osmesa TEST=osmesa
```
djhoese: Do the osmesa-based tests not need the "full" dependencies?

.travis.yml
```diff
@@ -118,49 +122,35 @@ install:
...
         else
-          conda install --yes matplotlib jupyter pyqt=4 scikit-image;
-          pip install -q mock;
+          conda install --yes -c defaults pyqt=4;
+          conda install --yes matplotlib jupyter scikit-image mock;
         fi;
       fi;
 
+     # Install OSMesa
+    - if [ "${TRAVIS_OS_NAME}" == "linux" ] && [ "${DEPS}" == "osmesa" ]; then
```
djhoese: Could/should this be changed to just checking for osmesa?

.travis.yml
kmuehlbauer: @djhoese I removed the system (apt) libs to not give chance for clashes, because I use `mesalib` and `libglu` from conda-forge.

.travis.yml
kmuehlbauer: Yes, would make sense, since osmesa is just on linux atm.

.travis.yml
djhoese: I see you removed `full_apt` (can the whole `addons` section be removed?), but changing `DEPS=full` to `DEPS=osmesa` means that pyqt, matplotlib, imageio, and other dependencies won't be installed, right? Is that expected?

kmuehlbauer: These are not needed for osmesa, I think.

kmuehlbauer: No, the whole section can't be removed. I did it, then the main apt-section takes over.

djhoese: Ok I just want to make sure that tests aren't being skipped due to missing dependencies so they look like they are passing. Similar thing happened on the PyQt5 PR where the tests were still running with PyQt4... 😭

kmuehlbauer: Yes, this need to be checked, but I'm quite confident, that the main dependencies are installed with the first conda install batch. Will check anyway.

kmuehlbauer: I would suggest, to check the dependencies for each backend and create the conda-forge env accordingly in travis. Do you think that can be postponed to another PR? I checked the builds from 4 weeks ago, they seem all the same when it comes to collected, deselected and selected.

djhoese: I checked the total number of tests run for a previous success of the osmesa environment and for this new one and they have the same number passed, skipped, and deselected. So I'm good to go. Do you want to fix that if statement on line 144 or should be just merge this?

kmuehlbauer: Let me just push the fix for line 144.

kmuehlbauer:
@djhoese I've checked locally, we can install `pyside`, `pyside2`, `wxpython`, `pyqt4`, pyqt5`, `sdl2`, `pyglet`, `glfw` and `osmesa` backends from conda-forge. Not naturally all in one backend. I'll give this a try and also check for backend dependencies. Seems it is taking up momentum.

djhoese:
@kmuehlbauer We have a problem. Unless I'm missing something, this PR edits `_pyopengl2.py` manually without editing the necessary parts of the `codegen/` code which *generates* it. Now I'm trying to #1306 and it loses the changes from this PR. Any ideas what might need to be done in codegen to handle this?

kmuehlbauer:
@djhoese Yes, [I know](https://github.com/vispy/vispy/issues/1568#issuecomment-461331378), unfortunately I lost track of it. We need to add these changes in the codegen sources and recreate then manually.

Issue #819, opened 2015-03-27
Port Klein bottle example from Glumpy


rossant:
https://github.com/glumpy/glumpy/blob/master/examples/geometry-parametric.py
#817 implements some geometry facilities required by this example.

Need to fix some bugs in `vp.mesh()` first.

ericdill:
Is this issue complete? There is a referenced PR that has been merged

Eric89GXL:
Nope, it hasn't been ported even though some of the necessary sub-functions have been

GuillaumeFavelier:
I would like to take care of this task, what are the constraints of this example? I could reference a prototype using `SceneCanvas`.

kmuehlbauer:
resolved by #1576, thanks @GuillaumeFavelier

PR #1576, opened 2019-02-08
Klein bottle example

GuillaumeFavelier:
This PR try to adress #819

The code so far is really basic, feel free to tell me what you think, I'm ready to work on it.
And I'm not even sure where to put the example?

### DIFF
--- a/examples/demo/scene/klein.py
+++ b/examples/demo/scene/klein.py
@@ -1,0 +1,36 @@
+from vispy import app, scene
+from vispy.geometry.parametric import surface
+
+
+def klein(u, v):
+    from math import pi, cos, sin
+    if u < pi:
+        x = 3 * cos(u) * (1 + sin(u)) + \
+                (2 * (1 - cos(u) / 2)) * cos(u) * cos(v)
+        z = -8 * sin(u) - 2 * (1 - cos(u) / 2) * sin(u) * cos(v)
+    else:
+        x = 3 * cos(u) * (1 + sin(u)) + (2 * (1 - cos(u) / 2)) * cos(v + pi)
+        z = -8 * sin(u)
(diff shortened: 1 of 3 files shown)


djhoese:
Nice catch on the division issues. I think putting it in the demo directory is fine. I'm not sure what that was originally for, but it doesn't seem right for any of the other examples directories. Any opinion @kmuehlbauer @larsoner? I have not tested this.

kmuehlbauer:
@GuillaumeFavelier Very nice!

kmuehlbauer:
The abysses of different OS, the hanging indent was catched by macOSX flake8 but not by the linux based install.

kmuehlbauer:
@GuillaumeFavelier The CI waits for your interaction :grin: I'll stop the older running builds.

GuillaumeFavelier:
Ok, I'll keep an eye on it to make it as green as possible.

GuillaumeFavelier:
Apparently, Travis is not happy because of an error in the `basics/plotting/colorbar.py` example.

kmuehlbauer:
Might be spurious. Restarted that test.

larsoner:
I see a seam. Is this actually a mesh definition problem rather than a rendering error (there are also rendering problems, possibly as expected for non-OIT rendering, which is fine):

screenshot from 2019-02-08 13-51-11

GuillaumeFavelier:
I saw nothing wrong when I tried to inspect the geometry but it was less informative than I thought because of the clutter.

Pseudo-wireframe (line_strip)

klein_wireframe

but apparently near the seam, the properties of the normals change:
`gl_FragColor = vec4(v_normal_vec, 1.0);`

klein_normal

larsoner:
Okay so it's just some rendering bug probably in the shading code. Thanks @GuillaumeFavelier

PR #1577, opened 2019-02-09
fix size bug for offscreen.py example

jiangwei221:
In the example/demo/gloo/offscreen.py example, if we change the size to (1200, 600), e.g. the width is not equal to the height, we will get:
screenshot from 2019-02-09 11-02-40
Where the right half of the image is empty.
I think it is a bug related to the size argument.
In this pull request, I change the order of width/height for Framebuffer and Texture2D, and get the result looks like:
screenshot from 2019-02-09 11-05-42
I'm not sure it is the best solution.

### DIFF
--- a/examples/demo/gloo/offscreen.py
+++ b/examples/demo/gloo/offscreen.py
@@ -103,10 +103,10 @@ class Canvas(app.Canvas):
         app.Canvas.__init__(self, show=False, size=size)
         self._t0 = time()
         # Texture where we render the scene.
-        self._rendertex = gloo.Texture2D(shape=self.size + (4,))
+        self._rendertex = gloo.Texture2D(shape=self.size[::-1] + (4,))
         # FBO.
         self._fbo = gloo.FrameBuffer(self._rendertex,
-                                     gloo.RenderBuffer(self.size))
+                                     gloo.RenderBuffer(self.size[::-1]))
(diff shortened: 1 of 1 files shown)


djhoese:
Any idea why this works? I don't have time to look in to the low-level gloo code at the moment. What made you realize that flipping the shape (`::-1`) would work?

What happens if you make the size taller than wider?

jiangwei221:
in the texture.py, https://github.com/vispy/vispy/blob/d96c26ec3093b9ea576bed95072a9c3ef6be1e27/vispy/gloo/texture.py#L489-L497
height is shape[0], and width is shape[1].
and in the framebuffer.py,
https://github.com/vispy/vispy/blob/d96c26ec3093b9ea576bed95072a9c3ef6be1e27/vispy/gloo/framebuffer.py#L207-L216
says ```New buffer shape (h, w)```
But, in the canvas.py, https://github.com/vispy/vispy/blob/d96c26ec3093b9ea576bed95072a9c3ef6be1e27/vispy/app/canvas.py#L26-L34
it says ```size : (width, height)```, which is not consistent with the other two.

jiangwei221:
and for height > width test:
with original code:
screenshot from 2019-02-09 12-07-05
with fix:
screenshot from 2019-02-09 12-07-24

PR #1578, opened 2019-02-13
Add shading to SphereVisual

GuillaumeFavelier:
#### What does this implement/fix?
The goal of this PR is to enable the `shading` capabilities of the `vispy.visuals.SphereVisual`.

#### Additional information
Some tweaking might be needed, tell me what do you think?

### DIFF
--- a/vispy/visuals/sphere.py
+++ b/vispy/visuals/sphere.py
@@ -42,10 +42,12 @@ class SphereVisual(CompoundVisual):
     edge_color : tuple or Color
         The `Color` to use when drawing the sphere edges. If `None`, then no
         sphere edges are drawn.
+    shading : str | None
+        Shading to use.
     """
     def __init__(self, radius=1.0, cols=30, rows=30, depth=30, subdivisions=3,
                  method='latitude', vertex_colors=None, face_colors=None,
-                 color=(0.5, 0.5, 1, 1), edge_color=None, **kwargs):
+                 color=(0.5, 0.5, 1, 1), edge_color=None, shading=None, **kwargs):
 
         mesh = create_sphere(rows, cols, depth, radius=radius,
                              subdivisions=subdivisions, method=method)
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
@GuillaumeFavelier  LGTM so far. I'm a bit unsure with respect to the `gl_state`.

In `MeshVisual` it is set to ```self.set_gl_state('translucent', depth_test=True, cull_face=False)``` and in SphereVisual it is set to ```self.mesh.set_gl_state(polygon_offset_fill=True,                         polygon_offset=(1, 1), depth_test=True)```. @rougier Does the second call override the first settings? Or does this make sense?

I checked with a slightly altered `example/basics/scene/sphere.py` and it works quite well (flat and smooth shading).

rougier:
@kmuehlbauer I'm not sure why the `polygon_offset_fill` and  `polygon_offset` are set. Those are used when you want to draw something over something else. Typically, this is used to draw the border of a cube (first you draw the filled cube, then the outline)

kmuehlbauer:
@rougier The SphereVisual draws a meshed sphere. I think it is to see the mesh structure. So if you do not see any problems using that way of setting gl_state, we can move forward.

kmuehlbauer:
As @rougier seems fine with it, we could move forward. @djhoese Did we have this windows error before? Seems related to sdf_text.

djhoese:
I'm starting the appveyor tests over again for sanity sake. No I have not seen this error before.

djhoese:
@kmuehlbauer is this branch up to date with your recent changes to `master` and the CI environments?

kmuehlbauer:
It's everything in besides the Klein bottle example.

I did not change appveyor setup.

djhoese:
@GuillaumeFavelier do you think you could merge current master in to your branch? I think I've fixed the Windows failures.

GuillaumeFavelier:
yeah, sure. I'll do it right now.

PR #1584, opened 2019-03-01
Add CONDA_CHANNELS environment variable to appveyor

djhoese:
Windows builds have started failing recently. Let's see if we can fix them.


djhoese:
Hhhmm probably didn't need to add conda-forge...but not a bad idea given that vispy is available on conda-forge.

Issue #818, opened 2015-03-27
Port skybox example from Glumpy


rossant:
https://github.com/glumpy/glumpy/blob/master/examples/skybox.py

GuillaumeFavelier:
I tried to investigate this issue and I think the port of the skybox as it is requires the implementation of a `TextureCube` class in `vispy.gloo.texture.py` with eventually some modifications in `vispy.gloo.program.py` too.
Plus, I didn't find any alternative for the `Trackball` interaction used in the example.

GuillaumeFavelier:
As #1589 and [#9](https://github.com/vispy/demo-data/pull/9) has been merged, I think this can be closed right?

djhoese:
Yep, thanks again.

PR #1589, opened 2019-03-05
[MRG] Add TextureCube object and port skybox example from glumpy

GuillaumeFavelier:
**What does this implement/fix?**
Port skybox example from Glumpy #818 

**Additional information**
1) I fixed the `offset` tuple in the `set_data()` method of the `GlirTextureCube` class in `vispy/gloo/glir.py` because I wasn't sure about what it exactly does.

2) I did not find any alternatives for `trackball` so I used another way to interact from:
`examples/demo/gloo/terrain.py`. Most likely there is a more practical way.

3) I stored the images on my machine and tested locally but I guess they can be uploaded in `vispy/demo-data`.

4) I am not sure about the location of `skybox.py`, I put it in `examples/demo/gloo` by default.

It is still a work in progress but here is a rendering:
skybox

### DIFF
--- a/examples/demo/gloo/skybox.py
+++ b/examples/demo/gloo/skybox.py
@@ -7,7 +7,7 @@
 Demonstration of cube mapping with trackball interaction.
 """
 import numpy as np
-import math
+from math import cos, sin, pi
 from vispy import app, gloo
 from vispy.gloo import gl
 from vispy.io import read_png, load_data_file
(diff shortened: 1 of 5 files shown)


GuillaumeFavelier:
Travis cannot test the example yet because the images can't be loaded.

GuillaumeFavelier:
Honestly, I don't know why Travis is not happy this time. 
Except from that, I still wonder how to handle the `offset` parameter of the `set_data()` method in the `GlirTextureCube` class.
Feel free to tell me what you think.

GuillaumeFavelier:
I just realise this is a duplicate of #1303.

djhoese:
@GuillaumeFavelier Am I correct that this now has all of the functionality of #1303 included? What were the differences between your PR and that one?

GuillaumeFavelier:
Both PRs added the *cube mapping* feature needed for #818.

@keunhong created `TextureCubeMap` and `GlirTextureCubeMap` classes, I called them `TextureCube` and `GlirTextureCube` respectively in this version. I wasn't sure of my implementation and since #1303 had already been reviewed, I decided to use his (merged in this [commit](https://github.com/vispy/vispy/pull/1589/commits/ee17f5c968a97737c2c5879d2f62d6997c9adc03)).

I think most of the differences are in the example script itself. @keunhong decided to demonstrate the cube mapping with generated textures and automatic camera interaction if I understood correctly whereas I tried to reproduce the *trackball* camera interaction, loaded the sky images and kept the example script fairly similar to the [source](https://github.com/glumpy/glumpy/blob/master/examples/skybox.py) as I considered it as a port.

PR #1590, opened 2019-03-05
Fix event loop detection triggering on blocked events

kne42:
Delay checking if stuck in an event loop until after checking if the event emitter is being blocked. This allows callback cycling. 

Toy example:
```python
e = EventEmitter(source=None, type='test')

def foo(x):
    print('foo')
    with e.blocker():
        e()

e.connect(foo)
e()
```

However, this check can be erroneously triggered during asynchronous operations. IMO if there is no real good reason for the check (since the loop will automatically trigger a `RecursionError` anyways), I'd like to just flat-out remove it.

### DIFF
--- a/vispy/util/tests/test_event_emitter.py
+++ b/vispy/util/tests/test_event_emitter.py
@@ -663,4 +663,81 @@ def test_emitter_block():
     assert_state(True, True)
 
 
+def test_emitter_reentrance_allowed_when_blocked1():
+    # Minimal re-entrance example
+
+    e = EventEmitter(source=None, type='test')
+    count = 0
+
+    @e.connect
+    def foo(x):
+        nonlocal count
+        count += 1
+        with e.blocker():
+            e()
+
+    e()
+    assert count == 1
+
+
+def test_emitter_reentrance_allowed_when_blocked2():
+    # More realistic re-entrance example
+    # The real world case for this might be a longer chain of events
+    # where event1's callback knows that what it's doing will trigger
+    # the same event so it blocks it. So event1 -> foo -> event2
+    # -> bar -> event1 -> (ignored bc blocked).
+
+    e1 = EventEmitter(source=None, type='test1')
+    e2 = EventEmitter(source=None, type='test2')
+    count = 0
+
+    @e1.connect
+    def foo(x):
+        nonlocal count
+        count += 1
+        with e1.blocker():
+            e2()
+
+    @e2.connect
+    def bar(x):
+        nonlocal count
+        count += 10
+        e1()
+
+    e1()
+    assert count == 11
+
+
+def test_emitter_reentrance_allowed_when_blocked3():
+    # Like the previous, but blocking callbacks instead of the event itself.
+    # Allows more fine-grained control. To some extent anyway - all callbacks
+    # of the event must be blocked to prevent raising the emitter loop error.
+
+    e1 = EventEmitter(source=None, type='test1')
+    e2 = EventEmitter(source=None, type='test2')
+    count = 0
+
+    @e1.connect
+    def foo(x):
+        nonlocal count
+        count += 1
+        with e1.blocker(foo):
+            e2()
+
+    @e2.connect
+    def bar(x):
+        nonlocal count
+        count += 10
+        e1()
+
+    @e2.connect
+    def eggs(x):
+        nonlocal count
+        count += 100
+        e1()
+
+    e1()
+    assert count == 2111
+
+
 run_tests_if_main()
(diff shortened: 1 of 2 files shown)


djhoese:
I know it's been a while, but could you explain this to me with more detail? What is the use case here and why would you want to do this? I don't understand the toy example.

kne42:
Yes, apologies, I explained it so poorly that even I was getting confused re-reading it.

So basically, the EventEmitter automatically checks if it is being triggered in an infinite cycle and breaks that cycle by erroring. However, this check is done before it checks if the event has been blocked, which doesn’t make sense because if the event is being blocked, then it shouldn’t trigger anyways and wouldn’t have created that infinite cycle.

This PR makes it so that if the event is being blocked by all listeners, it will not claim that there is an infinite loop.

djhoese:
Thanks. Ok, so in the toy example you are doing `event -> callback -> blocked(event)` in which case the callback is not called again because it is being blocked, but you/we also don't want it to raise an exception. The real world case for this might be a longer chain of events where event1's callback *knows* that what it's doing will trigger the same event so it blocks it. So `event1 -> callback -> event2 -> callback -> event1 -> Exception`. With your changes the `Exception` wouldn't happen and instead it would be a no-op.

Right? Darn, this is hard to explain.

kne42:
@djhoese Precisely!

djhoese:
I feel like this is you telling me how "my" library works :wink: 

So last question is, right now your implementation checks per-callback. Logically this looks correct, but should we consider a check before the callback for loop?

Lastly, do you think you could add a test for this? You'll likely want to rebase given how old this PR is now (sorry).

djhoese:
@kne42 Any updates on my last set of questions?

djhoese:
@kne42 Thoughts on my last question?

almarklein:
> right now your implementation checks per-callback. Logically this looks correct, but should we consider a check before the callback for loop?

if the check is before the loop, things will work if the event itself has been blocked. If put in the loop, things will work also if not the event but all its registered callbacks are blocked.

I can try adding a test tomorrow.

vispy/util/tests/test_event_emitter.py
```diff
@@ -663,4 +663,81 @@ def assert_state(a, b):
...
+    e()
+    assert count == 1
+
+
+def test_emitter_reentrance_allowed_when_blocked2():
+    # More realistic re-entrance example
+    # The real world case for this might be a longer chain of events
+    # where event1's callback knows that what it's doing will trigger
+    # the same event so it blocks it. So event1 -> foo -> event2
+    # -> bar -> event1 -> (ignored bc blocked).
```
almarklein, line 688: @djhoese I included your explanation because that helped me figure out the use-case :)

vispy/util/tests/test_event_emitter.py
```diff
@@ -663,4 +663,81 @@ def assert_state(a, b):
...
+        e1()
+
+    e1()
+    assert count == 11
+
+
+def test_emitter_reentrance_allowed_when_blocked3():
+    # Like the previous, but blocking callbacks instead of the event itself.
+    # Allows more fine-grained control. To some extent anyway - all callbacks
+    # of the event must be blocked to prevent raising the emitter loop error.
```
almarklein, line 714: I don't fully understand the use-case for this, TBH. but I also don't see any harm.

almarklein:
I this is ready to go.

vispy/util/tests/test_event_emitter.py
djhoese: I'm sure a use case for this would be something like resize event that triggers some transform updates that do some other kind of resize-like event...I don't know.

Anyway, I don't understand the final assertion value here.

Count starts at 0.

```
e1() -> foo() (+ 1) -> e2() ->
    bar() (+ 10) -> e1() -> blocked callback
    eggs() (+ 100) -> e1() -> blocked callback
```

So 1 + 10 + 100 => 111, right? Where does the 2000 come from?

djhoese: :man_facepalming: Nevermind, I was looking at the original unfixed commit.

djhoese:
Thanks @kne42 and @almarklein!

I've added the `bug` label as the more I thought about it this isn't really just a performance/user-experience kind of change, this is really fixing a bug that should allow for this type of behavior and handling. Another 2+ year old PR merged!

kne42:
Thanks for cleaning up my mess!

vispy/util/tests/test_event_emitter.py
almarklein: Yeah sorry, I added something to the test which I removed later, but forgot to update the check.

Issue #1588, opened 2019-03-05
Slowdown with Grid and AxisWidget + possible solution?


felixwoolford:
Hi!

I'm new to all this so sorry if this isn't the right way to go about things. I've encountered a slowdown issue when using Grid and Axis widgets. After a few minutes running a simple animation I begin to see a steady drop in fps.

Here's a fairly minimal example which produces the issue.

```
from vispy import app, scene
import numpy as np

class Test:
    def __init__(self):

        self.canvas = scene.SceneCanvas(keys = "interactive", show=True, bgcolor = 'b')
        self.canvas.measure_fps()
        self.grid = self.canvas.central_widget.add_grid(margin=10)

        self.yaxis = scene.AxisWidget(orientation='left',
                                 axis_label='Y Axis',
                                 axis_font_size=12,
                                 axis_label_margin=50,
                                 tick_label_margin=5)
        self.yaxis.width_max = 80
        self.grid.add_widget(self.yaxis, row=0, col=0)

        self.xaxis = scene.AxisWidget(orientation='bottom',
                                 axis_label='X Axis',
                                 axis_font_size=12,
                                 axis_label_margin=50,
                                 tick_label_margin=5)

        self.xaxis.height_max = 80
        self.grid.add_widget(self.xaxis, row=1, col=1)


        right_padding = self.grid.add_widget(row=1, col=2, row_span=1)
        right_padding.width_max = 50

        self.view = self.grid.add_view(row=0,col=1, border_color='white')
        self.view.camera = 'panzoom'
        self.view.camera.set_range(x=(-1,1),y=(-1,1))

        self.xaxis.link_view(self.view)
        self.yaxis.link_view(self.view)

        self.pos = np.zeros(2).reshape(1,2)
        self.m = scene.Markers(pos=self.pos,parent=self.view.scene)

        self.t = 0


    def update(self, ev):
        self.t = (self.t + 0.01) % (2*np.pi)
        self.pos[0,0] = np.sin(self.t*2)
        self.pos[0,1] = np.cos(self.t)
        self.m.set_data(pos = self.pos)


if __name__ == '__main__':
    test = Test()
    timer = app.Timer(interval=0, connect=test.update, start=True)
    app.run()
```

[...]

PR #1591, opened 2019-03-07
Fix FPS drop with grid and axis (from vispy 0.5.3)

felixwoolford:
As requested in #1588, this is how I addressed a slowdown issue in 0.5.3. In that version this reduces calls to _view_changed in axis.py, and may it still be useful?

### DIFF
--- a/vispy/scene/widgets/grid.py
+++ b/vispy/scene/widgets/grid.py
@@ -457,10 +457,12 @@ class Grid(Widget):
         if self._need_solver_recreate:
             self._need_solver_recreate = False
             self._recreate_solver()
-
+            
         # we only need to remove and add the height and width constraints of
         # the solver if they are not the same as the current value
+        not_same = False
         if rect.height != self._var_h.value:
+            not_same = True
             if self._height_stay:
                 self._solver.remove_constraint(self._height_stay)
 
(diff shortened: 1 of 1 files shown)


djhoese:
So I took the day off from work today and decided to review some of the bug/performance PRs that have been sitting around in vispy (what was I thinking lol). I'm rereading #1588 and the various issues and PRs that sprung from that and although it sounds like this isn't as big of a deal I think it would be foolish to lose these changes which at a quick glance make sense.

One concern I have is if the `if not_same` needs to be entered if the solver was recreated (earlier in the method). If the solver is new and may have changed things then maybe the dims of the widgets need to be updated too? @felixwoolford I know its been a while but do you have any idea if we should worry about this?

felixwoolford:
You could be right about that concern, I'm sorry I can't offer much insight - as I mentioned in #1588 it really was just a hack which seemed to work in the circumstances. I haven't been using this change because whatever update that was made between 0.5.3 and 0.6.0 to reduce calls to `_view_changed` resolved the performance issues that I was experiencing. I have used applications like the one in #1588 quite extensively since then (currently using 0.6.2) with no performance issues whatsoever!

djhoese:
Ok good to know you haven't had any issues with performance. I guess we'll close this and depend on the linking/references between the issues if we need it in the future. I'm just a little too nervous about that solver rebuild and how these changes could break something. If we had tests for all the various parts of this it might be a different story. Thanks again for all your help.

PR #1592, opened 2019-03-07
Fix memory leak in DataBuffer.

codypiersall:
Weak references were being kept around in the list of Buffer._views,
which was causing unbounded memory growth.

### DIFF
--- a/vispy/gloo/buffer.py
+++ b/vispy/gloo/buffer.py
@@ -260,7 +260,17 @@ class DataBuffer(Buffer):
 
         view = DataBufferView(self, key)
         self._views.append(weakref.ref(view))
+        self._remove_stale_views()
         return view
+
+    def _remove_stale_views(self):
+        # remove stale weak references to the view os our _views list doesn't
+        # grow without bound
+        # iterate backwards so we can remove while iterating
+        for i in reversed(range(len(self._views))):
+            view = self._views[i]
+            if view() is None:
+                del self._views[i]
 
     def __setitem__(self, key, data):
         """ Set data (deferred operation) """
--- a/vispy/gloo/buffer.py
+++ b/vispy/gloo/buffer.py
@@ -40,7 +40,7 @@ class Buffer(GLObject):
 
     def __init__(self, data=None, nbytes=None):
         GLObject.__init__(self)
-        self._views = []  # Views on this buffer (stored using weakrefs)
+        self._views = weakref.WeakSet()  # Views on this buffer
         self._valid = True  # To invalidate buffer views
         self._nbytes = 0  # Bytesize in bytes, set in resize_bytes()
 
@@ -125,9 +125,8 @@ class Buffer(GLObject):
         self._glir.command('SIZE', self._id, size)
         # Invalidate any view on this buffer
         for view in self._views:
-            if view() is not None:
-                view()._valid = False
-        self._views = []
+            view._valid = False
+        self._views = weakref.WeakSet()
 
 
 # -------------------------------------------------------- DataBuffer class ---
@@ -259,18 +258,8 @@ class DataBuffer(Buffer):
         """ Create a view on this buffer. """
 
         view = DataBufferView(self, key)
-        self._views.append(weakref.ref(view))
-        self._remove_stale_views()
+        self._views.add(view)
         return view
-
-    def _remove_stale_views(self):
-        # remove stale weak references to the view os our _views list doesn't
-        # grow without bound
-        # iterate backwards so we can remove while iterating
-        for i in reversed(range(len(self._views))):
-            view = self._views[i]
-            if view() is None:
-                del self._views[i]
 
     def __setitem__(self, key, data):
         """ Set data (deferred operation) """


djhoese:
I wonder if this could be done with something like a `WeakSet`: https://stackoverflow.com/a/6272503/433202

or also described in that SO answer, a custom weak list (I'd prefer WeakSet if a list isn't needed).

djhoese:
Looks like https://github.com/vispy/vispy/blob/e6142bd3d4f04eb18e6319183c2bea5ee8aedf14/vispy/gloo/buffer.py#L127-L130 is the only thing that needs to iterate over the existing views, should work with a weakset, right?

codypiersall:
This is a PR based on some stuff I wrote in #1588.  Just to reiterate what I said there: with the sample provided by @foloxweilf, a couple of memory leaks were exposed in Vispy.  This addresses them.  The code below is from that issue, but adapted to print the length of the `_views` attribute of the buffer.

```python
from vispy import app, scene
import numpy as np

class Test:
    def __init__(self):
        self.i = 0
        self.canvas = scene.SceneCanvas(keys = "interactive", show=True, bgcolor = 'b')
        self.grid = self.canvas.central_widget.add_grid(margin=10)

        self.yaxis = scene.AxisWidget(orientation='left',
                                 axis_label='Y Axis',
                                 axis_font_size=12,
                                 axis_label_margin=50,
                                 tick_label_margin=5)
        self.yaxis.width_max = 80
        self.grid.add_widget(self.yaxis, row=0, col=0)

        self.xaxis = scene.AxisWidget(orientation='bottom',
                                 axis_label='X Axis',
                                 axis_font_size=12,
                                 axis_label_margin=50,
                                 tick_label_margin=5)

        self.xaxis.height_max = 80
        self.grid.add_widget(self.xaxis, row=1, col=1)


        right_padding = self.grid.add_widget(row=1, col=2, row_span=1)
        right_padding.width_max = 50

        self.view = self.grid.add_view(row=0,col=1, border_color='white')
        self.view.camera = 'panzoom'
        self.view.camera.set_range(x=(-1,1),y=(-1,1))

        self.xaxis.link_view(self.view)
        self.yaxis.link_view(self.view)

        self.pos = np.zeros(2).reshape(1,2)
        self.m = scene.Markers(pos=self.pos,parent=self.view.scene)

        self.t = 0


    def update(self, ev):
        self.t = (self.t + 0.01) % (2*np.pi)
        self.pos[0,0] = np.sin(self.t*2)
        self.pos[0,1] = np.cos(self.t)
        self.m.set_data(pos=self.pos)

        if self.i % 100 == 0:
            print('len _views:', len(self.m._vbo._views))

        self.i += 1


if __name__ == '__main__':
    test = Test()
    timer = app.Timer(interval=0, connect=test.update, start=True)
    try:
        app.run()
    except KeyboardInterrupt:
        pass
```

Without this PR, this prints out
```
len _views: 10
len _views: 510
len _views: 1010
len _views: 1510
len _views: 2010
len _views: 2510
len _views: 3010
len _views: 3510
len _views: 4010
len _views: 4510
len _views: 5010
len _views: 5510
len _views: 6010
len _views: 6510
len _views: 7010

```

With this PR, it prints out
```
len _views: 6
len _views: 6
len _views: 6
len _views: 6
len _views: 6
len _views: 6
len _views: 6
len _views: 6
len _views: 6
len _views: 6

```

I found this leak with the excellent [objgraph](https://github.com/mgedmin/objgraph) package.

djhoese:
Just curious, do you have any numbers for how bad this memory leak is? I would hope/assume that weakrefs being kept around isn't actually that bad...eh on second thought I guess it could be (i.e. `weak_ref` object != `weak_ref()` which would be None).

codypiersall:
Hmmm, I didn't consider a weakset.  Do the `_views` need order to be kept?  And I also don't know what would happen if (somehow) a weakref vanished while the `_views` was being iterated over.

codypiersall:
Our conversation is slightly out of sync with each other :-)

Yeah, I posted the numbers in the previous post.  The leak is worse than I thought it would be; it's growing the list by roughly 100 items/second.

djhoese:
The python GIL should prevent that. Same argument could be made for "what if I try to use a weakref'd item that has been deleted in another thread. See https://docs.python.org/3/library/weakref.html#weak-reference-objects
Also found this: https://github.com/python/cpython/blob/master/Lib/_weakrefset.py#L14

codypiersall:
Oh heck yeah, good find. For other people following along, @djhoese linked to an 
class _IterationGuard class, which prevents removal while the weakrefset is being iterated.  So the only issue I can think of is whether the order of the `_views` matters; if not, then a weakrefset is for sure a better way to go.

djhoese:
I linked to some code further up (while we were out of sync). There are only two places (I think) in the base class where you'd have to make sure it was a WeakSet instead of list. Otherwise it should just be iterating over things to invalidate them. If your test code works and the tests pass then I think it'd be good to go.

codypiersall:
Good thought on `WeakSet`.  It at least works locally; let's see if it passes CI.  I've got another PR in the pipeline for another memory leak, but `WeakSet`'s won't work there... thanks a bunch for your help @djhoese!

codypiersall:
Just wanted to ping about potentially getting this merged.  What do you all think?

vispy/gloo/buffer.py
```diff
@@ -125,9 +125,8 @@ def resize_bytes(self, size):
         self._glir.command('SIZE', self._id, size)
         # Invalidate any view on this buffer
         for view in self._views:
-            if view() is not None:
-                view()._valid = False
-        self._views = []
+            view._valid = False
```
djhoese: So iterating over the `WeakSet` automatically dereferences them? (no need for the parentheses)

codypiersall: Yep!  That's exactly correct. And you're guaranteed to never get a bad reference, since you always get the dereferenced object.

codypiersall:
> Please be patient as I sort through all of the long standing PRs that I've delayed due to that PyQt5 PR. It's been 9 days and I do not get paid to maintain this project.

I really appreciate your maintenance of this project!  Sorry if I gave the impression of being impatient, I just wanted to make sure this didn't get lost.

PR #1593, opened 2019-03-07
Fix memory leak caused by keeping weak references to callbacks around.

codypiersall:
The list of callbacks gets iterated over very often, so it's important
to make sure there aren't useless entries laying around.  Plus the
memory can grow without bound without correcting this.

### DIFF
--- a/vispy/visuals/transforms/base_transform.py
+++ b/vispy/visuals/transforms/base_transform.py
@@ -194,6 +194,10 @@ class BaseTransform(object):
     def __repr__(self):
         return "<%s at 0x%x>" % (self.__class__.__name__, id(self))
 
+    def __del__(self):
+        # we can remove ourselves from *all* events in this situation.
+        self.changed.disconnect()
+
 
 class InverseTransform(BaseTransform):
     def __init__(self, transform):
--- a/vispy/visuals/transforms/chain.py
+++ b/vispy/visuals/transforms/chain.py
@@ -240,6 +240,14 @@ class ChainTransform(BaseTransform):
         tr = ",\n                 ".join(map(repr, self.transforms))
         return "<ChainTransform [%s] at 0x%x>" % (tr, id(self))
 
+    def __del__(self):
+        # remove all the children transforms from our callback, since we are
+        # being deleted.  (But we do *not* want to remove children from other
+        # callbacks)
+        for t in self._transforms:
+            t.changed.disconnect(self._subtr_changed)
+        self.changed.disconnect()
+
 
 class SimplifiedChainTransform(ChainTransform):
     def __init__(self, chain):


codypiersall:
This is similar in spirit to #1592.  The same strategy won't work with WeakRefs, since callbacks need to be called in order.  I'm *definitely* open to suggestions for improvements though! Here's a similar reproducing script:

```python
from vispy import app, scene
import numpy as np

class Test:
    def __init__(self):
        self.i = 0
        self.canvas = scene.SceneCanvas(keys = "interactive", show=True, bgcolor = 'b')
        self.grid = self.canvas.central_widget.add_grid(margin=10)

        self.yaxis = scene.AxisWidget(orientation='left',
                                 axis_label='Y Axis',
                                 axis_font_size=12,
                                 axis_label_margin=50,
                                 tick_label_margin=5)
        self.yaxis.width_max = 80
        self.grid.add_widget(self.yaxis, row=0, col=0)

        self.xaxis = scene.AxisWidget(orientation='bottom',
                                 axis_label='X Axis',
                                 axis_font_size=12,
                                 axis_label_margin=50,
                                 tick_label_margin=5)

        self.xaxis.height_max = 80
        self.grid.add_widget(self.xaxis, row=1, col=1)


        right_padding = self.grid.add_widget(row=1, col=2, row_span=1)
        right_padding.width_max = 50

        self.view = self.grid.add_view(row=0,col=1, border_color='white')
        self.view.camera = 'panzoom'
        self.view.camera.set_range(x=(-1,1),y=(-1,1))

        self.xaxis.link_view(self.view)
        self.yaxis.link_view(self.view)

        self.pos = np.zeros(2).reshape(1,2)
        self.m = scene.Markers(pos=self.pos,parent=self.view.scene)

        self.t = 0


    def update(self, ev):
        self.t = (self.t + 0.01) % (2*np.pi)
        self.pos[0,0] = np.sin(self.t*2)
        self.pos[0,1] = np.cos(self.t)
        self.m.set_data(pos=self.pos)

        if self.i % 100 == 0:
            print(len(self.view._transform._inverse.changed._callbacks))

        self.i += 1


if __name__ == '__main__':
    test = Test()
    timer = app.Timer(interval=0, connect=test.update, start=True)
    try:
        app.run()
    except KeyboardInterrupt:
        pass
```

This list grows really quickly as long as mouse events are happening.

codypiersall:
The reason we can't use `WeakSet` like in #1592 is twofold: 1. order of callbacks is important, and 2. we are keeping two lists in sync: `EventEmitter._callbacks` and `EventEmitter._callback_refs`.

djhoese:
Given your explanation and what I see, your solution makes sense. However, doesn't this chunk of code already handle this? https://github.com/vispy/vispy/blob/e3d2038dac890273e0f82e531c61c608b74bfcbb/vispy/util/event.py#L453-L473

If not, could your solution be applied here where we are already doing the check for a valid callback (if weakref no longer exists, don't add it to the list, after looping through add the new list to the callbacks attribute)?

codypiersall:
Hmmm, yeah, that does seem to do the right thing, so it must not be getting called.  So either the `__call__` method isn't getting entered (but isn't this how callbacks are fired?) or `blocked.get(None, 0) > 0` is getting entered, so the function is returning before it should: https://github.com/vispy/vispy/blob/e3d2038dac890273e0f82e531c61c608b74bfcbb/vispy/util/event.py#L450-L451

I'll try and run some tests tonight or in the next couple days to see what's going on there.

codypiersall:
This is very fun and exciting.

So you're *absolutely right* that the snippet that you linked to in `__call__ ` should work, and it does work in most circumstances.  For this specific list of callbacks: `self.view._transform._inverse.changed._callbacks`, the `__call__` method is not getting called.  I haven't looked in to why at all, any ideas?

Taking a bit of a step back, though, might it make sense for the check to be in `connect` instead of in `__call__`?  `connect` gets called less often, and the list of callbacks can only grow whenever connect is called.

I *think* this is this issue which is causing our GUIs at work to eventually slow to a crawl or consume all our system memory.  One can hope anyway...  I think this is only an issue with the higher level plotting API?

djhoese:
I think we need to take it even one more step back. Why are all of these callbacks getting added to an event? These things shouldn't be growing over time unless I'm missing something. That, or things are being done poorly.

If these callbacks do need to be added for some reason, why aren't they being removed by the thing *adding* them?

I'm not completely against the cleaning being done during the connect, but it feels like putting a bandaid in the wrong place. Besides the fact that the callbacks are already being iterated over during `__call__`, doing things during connect that affect other callbacks/events (removing them) feels like a bad side effect. An event occurring effecting things sounds cleaner to me. However, I now realize that this cleaning should only be cleaning invalid/old callbacks so maybe that is ok. My first questions are still the most important IMO.

codypiersall:
> Why are all of these callbacks getting added to an event? These things shouldn't be growing over time unless I'm missing something.

That's something I'm really curious about.  I'm not very clear on how events work on vispy at all, so I probably need to do some more reading through the docs.  I think every event that happens calls `connect`, but I would have thought that each event that fires would call `__call__`, so I'm missing something.  I should check why `__connect__` is getting called so frequently and why `__call__` never gets called for this emitter.

> However, I now realize that this cleaning should only be cleaning invalid/old callbacks so maybe that is ok. 

Yeah, since they're *only* removing the invalid callbacks it's fine.  The references are already dead.  As far as correctness goes I think it's good.  If the `EventEmitter` was the one actually creating the weakrefs, we could use the callback parameter to mark the objects as dirty, but alas the `EventEmitter` does not control them.

I think we've seen this problem at work when using the [plotwidget](https://github.com/vispy/vispy/blob/master/vispy/plot/plotwidget.py) interface.  I'll see if I can reproduce this using the plotwidget too.

codypiersall:
Okay, here's a simpler reproducer that shows the leaky behavior, using the high-level plot interface.  Running this example requires objgraph to be installed.  It will print out the new references coming alive:

```python
import vispy.plot as vp
from vispy import app
import objgraph

plotwidget = fig[0, 0]

fig.title = "beep boop"
data = [(x, x**2) for x in range(0, 100)]
line = plotwidget.plot(data, title="y = x^2")

i = 0
def update(ev):
    global i
    i += 1
    if i % 100 == 0:
        objgraph.show_growth()


if __name__ == '__main__':
    timer = app.Timer(interval=0, connect=update, start=True)
    fig.show(run=True)
```

I think the thing that is causing this is linking the views of the axes with the main plot.  If those lines are removed from the original example, this issue goes away (because `self.view._transform._inverse is None == True`).  I'll try to track down what's going on in more detail tonight or in the next few days.

codypiersall:
Okay... so I printed a traceback inside `EventEmitter.connect()` to see what was exercising this, and got the following traceback:

```
  File "derp.py", line 68, in <module>
    app.run()
  File "C:\Users\cody\Dev\vispy\vispy\app\_default_app.py", line 62, in run
    return default_app.run()
  File "C:\Users\cody\Dev\vispy\vispy\app\application.py", line 152, in run
  File "C:\Users\cody\Dev\vispy\vispy\app\backends\_qt.py", line 237, in _vispy_run
    return app.exec_()
  File "C:\Users\cody\Dev\vispy\vispy\app\backends\_qt.py", line 448, in event
  File "C:\Users\cody\Dev\vispy\vispy\app\backends\_qt.py", line 416, in mouseMoveEvent
    modifiers=self._modifiers(ev),
  File "C:\Users\cody\Dev\vispy\vispy\app\base.py", line 208, in _vispy_mouse_move
    ev = self._vispy_canvas.events.mouse_move(**kwargs)
  File "C:\Users\cody\Dev\vispy\vispy\util\event.py", line 461, in __call__
  File "C:\Users\cody\Dev\vispy\vispy\util\event.py", line 477, in _invoke_callback
    cb(event)
  File "C:\Users\cody\Dev\vispy\vispy\scene\canvas.py", line 359, in _process_mouse_event
    getattr(picked.events, event.type)(scene_event)
  File "C:\Users\cody\Dev\vispy\vispy\util\event.py", line 461, in __call__
  File "C:\Users\cody\Dev\vispy\vispy\util\event.py", line 477, in _invoke_callback
    cb(event)
  File "C:\Users\cody\Dev\vispy\vispy\scene\cameras\panzoom.py", line 226, in viewbox_mouse_event
    self.pan(p1s-p2s)
  File "C:\Users\cody\Dev\vispy\vispy\scene\cameras\panzoom.py", line 127, in pan
    self.rect = self.rect + pan
  File "C:\Users\cody\Dev\vispy\vispy\scene\cameras\panzoom.py", line 149, in rect
    self.view_changed()
  File "C:\Users\cody\Dev\vispy\vispy\scene\cameras\base_camera.py", line 380, in view_changed
    self._update_transform()
  File "C:\Users\cody\Dev\vispy\vispy\scene\cameras\panzoom.py", line 276, in _update_transform
    self.transform.zoom((1, 1, 1/d))
  File "C:\Users\cody\Dev\vispy\vispy\visuals\transforms\linear.py", line 214, in zoom
    self._set_st(scale=scale, translate=trans)
  File "C:\Users\cody\Dev\vispy\vispy\visuals\transforms\linear.py", line 173, in _set_st
    self.update()   # inform listeners there has been a change
  File "C:\Users\cody\Dev\vispy\vispy\visuals\transforms\base_transform.py", line 153, in update
    self.changed(*args)
  File "C:\Users\cody\Dev\vispy\vispy\util\event.py", line 461, in __call__
  File "C:\Users\cody\Dev\vispy\vispy\util\event.py", line 477, in _invoke_callback
    cb(event)
  File "C:\Users\cody\Dev\vispy\vispy\scene\widgets\axis.py", line 88, in _view_changed
    tr = self.node_transform(self._linked_view.scene)
  File "C:\Users\cody\Dev\vispy\vispy\scene\node.py", line 611, in node_transform
    return ChainTransform(self.node_path_transforms(node))
  File "C:\Users\cody\Dev\vispy\vispy\visuals\transforms\chain.py", line 49, in __init__
    self.transforms = trs
  File "C:\Users\cody\Dev\vispy\vispy\visuals\transforms\chain.py", line 94, in transforms
    t.changed.connect(self._subtr_changed)
  File "C:\Users\cody\Dev\vispy\vispy\util\event.py", line 319, in connect
    traceback.print_stack(file=sys.stdout)
```

Basically, blah blah blah, the mouse event propagates through all Vispy's code, causing a bunch of events to fire.  The important thing here is the stuff that happens in axis.py:

```
  File "C:\Users\cody\Dev\vispy\vispy\scene\widgets\axis.py", line 88, in _view_changed
    tr = self.node_transform(self._linked_view.scene)
  File "C:\Users\cody\Dev\vispy\vispy\scene\node.py", line 611, in node_transform
    return ChainTransform(self.node_path_transforms(node))
  File "C:\Users\cody\Dev\vispy\vispy\visuals\transforms\chain.py", line 49, in __init__
    self.transforms = trs
  File "C:\Users\cody\Dev\vispy\vispy\visuals\transforms\chain.py", line 94, in transforms
    t.changed.connect(self._subtr_changed)
  File "C:\Users\cody\Dev\vispy\vispy\util\event.py", line 319, in connect
    traceback.print_stack(file=sys.stdout)
```

(My line numbers will be slightly off because I had some debugging prints and stuff)

The [`AxisWidget`](https://github.com/vispy/vispy/blob/master/vispy/scene/widgets/widget.py) class is getting the transform to the view it is associated with whenever the transform changes. This makes sense and is good, I think.  It's trying to figure out what it needs to change its domain to.

There are a few ways to solve this I think.  My PR is one way.  An alternative that works at least for this example is to add a `__del__` to `ChainTransform` to disconnect itself from the events it is associated with.  I tested this example, and adding this method to `ChainTransform` solves this problem:

```python
def __del__(self):
    self.transforms = []
```

It relies on the fact that the setter disconnects from all currently connected events.  I think this is a better solution; I like that the `EventEmitter` wouldn't have to worry about keeping track of the things connected to it, but would just get told when things are done.  I'm going to update my PR to do that and see if tests pass; it seems plausible that this would have unwanted ramifications...

codypiersall:
For the record, I'm pretty happy with the `__del__` approach.  I think it's a lot cleaner to make sure the thing that initially `connect`s finally `disconnect`s.  What do you think?  I'm also happy it passed CI.

djhoese:
I like the `__del__` idea as an elegant/easy solution to what's going on. Would it possibly be more flexible/efficient to add a `__del__` to the BaseTransform? Maybe override it in ChainTransform to not do anything since it would be redundant when the children get garbage collected.

I think there are future optimizations that could be done, but I'm not sure I want to have anyone get in to it at this point. The fact that a new ChainTransform is created every time the view is changed and this ChainTransform connects multiple things to various callbacks sounds like a terrible thing. I'm not sure there is an easy way around it, but even having a special chain transform that can be updated in place sounds better to me. I'm just rambling at this point.

codypiersall:
Adding `__del__` to `BaseTransform` makes sense in general.  I was surprised that it doesn't actually help this situation.  I think the issue is that *some* of the children of the `ChainTransform` still have references laying around on other `EventEmitter`s, so they aren't garbage collected.  My current version of the PR implements `__del__` on `BaseTransform` and on `ChainTransform`.  In `BaseTransform`, _all_ callbacks are removed form the transform--makes sense, since the trasnform isn't around anymore--and in `ChainTransform`, only the `ChainTransform`'s callback is removed from its children.

> The fact that a new ChainTransform is created every time the view is changed and this ChainTransform connects multiple things to various callbacks sounds like a terrible thing.

That was my first impression too!  I think the ChainTransform can be update in place, but the issue is that the relationship from the `AxisWidget` to the `ViewBox` it's linked with can, I think, change in arbitrary ways, so the transform needs to be recalculated every time it's needed... at least that's my understanding.

codypiersall:
Closing and re-opening to retrigger CI: here is the [failure traceback](https://travis-ci.org/vispy/vispy/jobs/504097949#L3406):

```
Example basics/plotting/colorbar.py failed (0):
  File "<string>", line 24, in <module>
  File "/home/travis/build/vispy/vispy/vispy/app/canvas.py", line 509, in __enter__
    self._backend._vispy_warmup()
  File "/home/travis/build/vispy/vispy/vispy/app/backends/_qt.py", line 322, in _vispy_warmup
    self._vispy_canvas.app.process_events()
  File "/home/travis/build/vispy/vispy/vispy/app/application.py", line 78, in process_events
    return self._backend._vispy_process_events()
  File "/home/travis/build/vispy/vispy/vispy/app/backends/_qt.py", line 230, in _vispy_process_events
    app.processEvents()
  File "/home/travis/build/vispy/vispy/vispy/app/backends/_qt.py", line 448, in event
  File "/home/travis/build/vispy/vispy/vispy/app/backends/_qt.py", line 448, in event
  File "/home/travis/build/vispy/vispy/vispy/app/backends/_qt.py", line 720, in paintGL
  File "/home/travis/build/vispy/vispy/vispy/util/event.py", line 455, in __call__
  File "/home/travis/build/vispy/vispy/vispy/util/event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  << caught exception here: >>
  File "/home/travis/build/vispy/vispy/vispy/util/event.py", line 471, in _invoke_callback
    cb(event)
  File "/home/travis/build/vispy/vispy/vispy/scene/canvas.py", line 217, in on_draw
    self._draw_scene()
  File "/home/travis/build/vispy/vispy/vispy/scene/canvas.py", line 266, in _draw_scene
    self.draw_visual(self.scene)
  File "/home/travis/build/vispy/vispy/vispy/scene/canvas.py", line 304, in draw_visual
    node.draw()
  File "/home/travis/build/vispy/vispy/vispy/scene/visuals.py", line 99, in draw
  File "/home/travis/build/vispy/vispy/vispy/visuals/visual.py", line 583, in draw
  File "/home/travis/build/vispy/vispy/vispy/scene/widgets/grid.py", line 197, in _prepare_draw
    self._update_child_widget_dim()
  File "/home/travis/build/vispy/vispy/vispy/scene/widgets/grid.py", line 459, in _update_child_widget_dim
    self._recreate_solver()
  File "/home/travis/build/vispy/vispy/vispy/scene/widgets/grid.py", line 446, in _recreate_solver
    self._grid_widgets)
  File "/home/travis/build/vispy/vispy/vispy/scene/widgets/grid.py", line 366, in _add_widget_dim_constraints
    strength=REQUIRED)
  File "/home/travis/anaconda/envs/testenv/lib/python3.6/site-packages/cassowary/simplex_solver.py", line 63, in add_constraint
    self.add_with_artificial_variable(expr)
  File "/home/travis/anaconda/envs/testenv/lib/python3.6/site-packages/cassowary/simplex_solver.py", line 384, in add_with_artificial_variable
    if e != None:
  File "/home/travis/anaconda/envs/testenv/lib/python3.6/site-packages/cassowary/expression.py", line 302, in __eq__
    raise TypeError('Cannot compare expression with object of type %s' % type(other))
TypeError: Cannot compare expression with object of type <class 'NoneType'>
ERROR: Invoking <bound method SceneCanvas.on_draw of <Fig (PyQt4) at 0x2ace12392e80>> for DrawEvent
```

djhoese:
Does the `ChainTransform` need to delete/remove its children to have them get garbage collected (in addition to disconnecting its own events)?

codypiersall:
> Does the ChainTransform need to delete/remove its children to have them get garbage collected

No, since the garbage collector tracks cycles, the `ChainTransform` won't accidentally keep its children alive.  Plus, if `__del__` is getting called, that means the `ChainTransform` is already getting gc'd, so the `_transforms` list is going to be gc'd too.  If the only reference to the children transform is from the `ChainTransform` and there are no more references to the `ChainTransform`, they'll all get gc'd.

Small aside:  I have an optimization for [`AxisWidget`](https://github.com/vispy/vispy/blob/master/vispy/scene/widgets/axis.py#L12) to not build a new set of shaders every time a mouse event happens.  It *definitely* doesn't need to create a new `ChainTransform` every time; I have a patch that does this optimization, but I couldn't come up with a test where the difference was notable.  Here is the patch anyway:

```
diff --git a/vispy/scene/widgets/axis.py b/vispy/scene/widgets/axis.py
index ffcadb9c..4d9a48c0 100644
--- a/vispy/scene/widgets/axis.py
+++ b/vispy/scene/widgets/axis.py
@@ -28,6 +28,8 @@ class AxisWidget(Widget):
         self.axis = AxisVisual(**kwargs)
         self.orientation = orientation
         self._linked_view = None
+        # keep track of the transformation from us to the linked view
+        self._tr = None
         Widget.__init__(self)
         self.add_subvisual(self.axis)
 
@@ -76,13 +78,19 @@ class AxisWidget(Widget):
                 self._view_changed)
         self._linked_view = view
         view.scene.transform.changed.connect(self._view_changed)
+        self._tr = self.node_transform(self._linked_view.scene)
         self._view_changed()
         
     def _view_changed(self, event=None):
         """Linked view transform has changed; update ticks.
         """
-        tr = self.node_transform(self._linked_view.scene)
-        p1, p2 = tr.map(self._axis_ends())
+        if self._tr is None:
+            raise RuntimeError('_view_changed() was called before link_view()')
+        # if the list of transforms
+        _transforms = self.node_path_transforms(self._linked_view.scene)
+        if _transforms != self._tr.transforms:
+            self._tr.transforms = _transforms
+        p1, p2 = self._tr.map(self._axis_ends())
         if self.orientation in ('left', 'right'):
             self.axis.domain = (p1[1], p2[1])
         else:
```

It's pretty straightforward, and should be correct too, I think. Any changes in the paths between the two nodes will be picked up and applied.  If I (or someone else) can come up with a test case where this actually helps I can submit a PR for it too.

codypiersall:
Just wanted to ping about potentially getting this merged.  What do you all think?

djhoese:
Do you want to make a PR for that other optimization even if you can't get a good test for it? Just so it doesn't get lost in this PR comment? As long as you describe the situation in the PR I think it'll be good.

djhoese:
@larsoner @kmuehlbauer if you can sanity check this for me. I think it is fine, but I may have been looking at this too long.

@codypiersall I didn't see any numbers when I skimmed over this again, but did you have any results for what kind of difference this can make? Sorry if I missed it.

codypiersall:
> I didn't see any numbers when I skimmed over this again, but did you have any results for what kind of difference this can make? Sorry if I missed it.

@djhoese This made a *huge* difference in our application, which is live streaming of the data from a phased array radar system we are developing.  Just yesterday, a coworker who had not yet applied this patch noticed his app crawling with the plotting after being up for maybe 15 minutes.  I had him print the length of the transforms associated with the view: `print(len(view._transform._inverse.changed._callbacks))` and the length of callbacks was about 100k.  The callbacks get iterated over very frequently, so this made a massive difference.

Ultimately, the leak here and from #1592 caused the GUI to first become unusable (plotting at < a couple Hz), and finally to crash due to consuming all the RAM on the system.  These two PRs should increase the stability quite a bit for us.  The main case that this PR fixed was panning and zooming on plots that had linked x/y axes, which was what our plots did, and also any application using the higher-level plotting interface would be affected.  #1592 affected plots that called `set_data` frequently, which our plots did.

Thanks a bunch for the merge!  I'll send a PR for the transforms optimization as well so it doesn't get lost.

PR #1596, opened 2019-03-07
[MRG] Allow XYZAxisVisual override of position, color, connect, and method

### LINKED ISSUE #671: XYZAxisVisuals Override Defaults
It looks like XYZAxisVisual is not overridable in the **init** function for the verts and color arguments?  Passing in `pos=my_custom_verts` results in `TypeError: __init__() got multiple values for keyword argument 'pos'`.

The `**kwds` argument looks like it is being passed through to the Line class, via LineVisual.  Does a method exist to specify the verts, color, and / or connect kwargs?  I am hesitant to submit a PR modifying **kwds since I am not 100% sure how the passing is working.

GuillaumeFavelier:
**What does this implement/fix?**
This PR tries to fix  #671 by allowing the override of defaults `pos` and `color` arguments.

**Additional information**
Maybe overriding `connect` and `method` is too much?

### DIFF
--- a/vispy/visuals/xyz_axis.py
+++ b/vispy/visuals/xyz_axis.py
@@ -10,23 +10,24 @@ class XYZAxisVisual(LineVisual):
     x=red, y=green, z=blue.
     """
     def __init__(self, **kwargs):
-        if 'pos' not in kwargs:
-            kwargs['pos'] = np.array([[0, 0, 0],
-                                      [1, 0, 0],
-                                      [0, 0, 0],
-                                      [0, 1, 0],
-                                      [0, 0, 0],
-                                      [0, 0, 1]])
-        if 'color' not in kwargs:
-            kwargs['color'] = np.array([[1, 0, 0, 1],
-                                        [1, 0, 0, 1],
-                                        [0, 1, 0, 1],
-                                        [0, 1, 0, 1],
-                                        [0, 0, 1, 1],
-                                        [0, 0, 1, 1]])
-        if 'connect' not in kwargs:
-            kwargs['connect'] = 'segments'
-        if 'method' not in kwargs:
-            kwargs['method'] = 'gl'
+        pos = np.array([[0, 0, 0],
+                        [1, 0, 0],
+                        [0, 0, 0],
+                        [0, 1, 0],
+                        [0, 0, 0],
+                        [0, 0, 1]])
+        color = np.array([[1, 0, 0, 1],
+                          [1, 0, 0, 1],
+                          [0, 1, 0, 1],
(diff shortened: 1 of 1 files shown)


vispy/visuals/xyz_axis.py
```diff
@@ -10,17 +10,23 @@ class XYZAxisVisual(LineVisual):
...
-                          [0, 0, 1]])
-        color = np.array([[1, 0, 0, 1],
-                          [1, 0, 0, 1],
-                          [0, 1, 0, 1],
-                          [0, 1, 0, 1],
-                          [0, 0, 1, 1],
-                          [0, 0, 1, 1]])
-        LineVisual.__init__(self, pos=verts, color=color, connect='segments',
-                            method='gl', **kwargs)
+        if 'pos' not in kwargs:
```
kmuehlbauer: @GuillaumeFavelier Not sure which version is more pythonic, but as `kwargs` is a dict, we could use something like:

```kwargs['pos'] = kwargs.pop('pos', default)```

Comment (GuillaumeFavelier) [reply vispy/visuals/xyz_axis.py]: I don't know either which is more pythonic but I can definitely use this syntax. And in this case, all the defaults need to be defined beforehand I suppose?

Comment (kmuehlbauer) [reply vispy/visuals/xyz_axis.py]: @GuillaumeFavelier After thinking a bit, there would be this solution:

```python
pos = np.array([[0, 0, 0],
                [1, 0, 0],
                [0, 0, 0],
                [0, 1, 0],
                [0, 0, 0],
                [0, 0, 1]])
kwargs.update(pos=kwargs.pop('pos', pos)
```
And same with the other kwargs. @larsoner Is there some vispy-style how to do this kind of kwarg-handling?

larsoner: I vaguely recall seeing `.pop`'s used, @kmuehlbauer your code looks good

djhoese: What's wrong with `kwargs.setdefault('pos', pos)`?

kmuehlbauer: @djhoese Believe me or not, I did not know, that something like this exist. This is by far the most minimal solution.

djhoese:
This looks good enough to me. I think in the future we may want to consider making the connect and gl kwargs as something else that can be configured, but I think that will be a larger refactor/rewrite of how all visuals treat those arguments (or at least LineVisual subclasses).

djhoese:
Oops, nevermind. That's already possible. Misread the code. Thanks @GuillaumeFavelier, great job.

PR #1598, opened 2019-03-13
Fix qt5 based backends when moving window from one monitor to another

djhoese:
This is similar to https://github.com/matplotlib/matplotlib/issues/8061 for matplotlib. I work on a macbook with a HiDPI retina display. My external monitor at work is not HiDPI. If I open a PyQt5 canvas on one monitor and drag it to another it gets in to a bad state. This is caused by the two displays having different device pixel ratios. This is the ratio of logical pixels to physical screen pixels. On HiDPI displays this is usually 2 and for regular displays it is 1.

PyQt5 allows you to know this ratio (`self.devicePixelRatio()`) and vispy already uses it to adjust physical versus logic pixel calculations. However, vispy only does this when first creating the window and on resize events. This means that if you get in to the bad state mentioned above you can resize the window and everything goes back to normal...until you move the window to another monitor. This PR moves the check of `self.devicePixelRatio()` to the getter of `physical_size`. This likely has some performance penalty but it should be relatively small if PyQt5 is optimizing things correctly.

This PR fixes this issue 99.9% of the way. There is still an issue if you move the window and don't interact with the canvas (cause a redraw). As far as I know there is no signal in PyQt5 to say "we've moved to a different display" that I have access to. Although, PyQt5 seems to be displaying things just fine since text and window sizes are kept proportional.

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -315,14 +315,20 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
         self._vispy_set_size(*p.size)
         if p.fullscreen is not False:
             if p.fullscreen is not True:
-                logger.warning('Cannot specify monitor number for Qt fullscreen, using default')
+                logger.warning('Cannot specify monitor number for Qt '
+                               'fullscreen, using default')
             self._fullscreen = True
         else:
             self._fullscreen = False
 
         # must set physical size before setting visible or fullscreen
         # operations may make the size invalid
-        self._physical_size = p.size
+        if hasattr(self, 'devicePixelRatio'):
+            # handle high DPI displays in PyQt5
+            ratio = self.devicePixelRatio()
+        else:
+            ratio = 1
+        self._physical_size = (p.size[0] * ratio, p.size[1] * ratio)
 
         if not p.resizable:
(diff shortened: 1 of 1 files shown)


djhoese:
For reference, with my fix I get:

External monitor where the canvas was first opened:

image

Same window immediately moved to other display (HiDPI):

image

And after clicking the canvas in the center on the HiDPI display (a "pan" operation in the PanZoomCamera):

image

larsoner:
> You can connect to the screenChanged(QScreen)* signal of the QWindow to which your widget belongs.

From https://forum.qt.io/topic/52296/detect-changing-devicepixelratio/6

Want to do that instead/in addition?

djhoese:
I'll try it out and let you know.

djhoese:
@larsoner ready for re-review.

larsoner:
LGTM assuming it actually fixes your problem (I don't have an appropriate setup to test)

djhoese:
Yes, it does. It is better than my first solution because it forces the redraw when you switch displays. So besides a small redraw stutter when moving between displays, the user shouldn't see any change in the canvas.

I'll probably squash this merge after I catch up on some emails.

PR #1600, opened 2019-03-16
Avoid recompiling the ChainTransform whenever we don't need to.

codypiersall:
The AxisWidget causes a recompilation of its `ChainTransform` every time
a pan/zoom event occurs. This is not necessary as long as the node path
from the axes to its linked view has not changed.

### DIFF
--- a/vispy/scene/widgets/axis.py
+++ b/vispy/scene/widgets/axis.py
@@ -28,6 +28,8 @@ class AxisWidget(Widget):
         self.axis = AxisVisual(**kwargs)
         self.orientation = orientation
         self._linked_view = None
+        # keep track of the transformation from us to the linked view
+        self._tr = None
         Widget.__init__(self)
         self.add_subvisual(self.axis)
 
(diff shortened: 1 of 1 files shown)


codypiersall:
As I mentioned in #1593, I didn't actually find a test case where this optimization clearly helped.  I thought that the simple case here would benefit:

```python
import vispy.plot as vp
from vispy import app

plotwidget = fig[0, 0]

fig.title = "beep boop"
data = [(x, x**2) for x in range(0, 100)]
line = plotwidget.plot(data, title="y = x^2")

if __name__ == '__main__':
    fig.show(run=True)
```

But it didn't.  It seems like there *should* be a case where this helps but I couldn't find one :smile:

djhoese:
I'm guessing with your other PRs this is even less noticeable? Can you see a difference when panning a plot around or something?

codypiersall:
> I'm guessing with your other PRs this is even less noticeable? Can you see a difference when panning a plot around or something?

That's exactly right: with the last two PRs, this optimization is not noticeable at all.  Can't see a difference at all... at the very least, if this PR is closed, it can be referenced in case someone else notices that it seems weird that the `AxisWidget` does a bunch of unnecessary recompilation.

vispy/scene/widgets/axis.py
```diff
@@ -76,13 +78,20 @@ def link_view(self, view):
                 self._view_changed)
         self._linked_view = view
         view.scene.transform.changed.connect(self._view_changed)
+        self._tr = self.node_transform(self._linked_view.scene)
```
djhoese: Is `_view_changed` triggered/called when the transforms changed? Does this `self._tr` need to get updated because of that?

I'm forgetting everything associated with this, but iirc the performance issue is that `self._tr` gets rebuilt every time the view is changed. This includes transform attributes being updated where `self._tr` wouldn't change (the overall shader code is the same, right?) but `_view_changed` would still be called because it is a change to the transform that requires a redraw. However, if the actual transform object of something in the visual node graph changed (requiring shader recompilation/building) then `_view_changed` is called and `self._tr` should be updated. Does any of this sound reasonable?

rougier:
@codypiersall Did you measure the actual FPS to check for difference? Aren't we stuck at 30fps by default?

PR #1601, opened 2019-03-17
Improve and fix jupyter notebook support

djhoese:
This is the start of my debugging on getting notebook support fixed. It seems ipywidgets is no longer installing the extension (which should be done automatically). I should also take this opportunity to update the notebook examples in `examples/`.

I'm also waiting for freetype-py to get version 2.1.0 on conda-forge: https://github.com/conda-forge/freetype-py-feedstock/pulls
After that is available then we can remove the recently added restriction in .travis.yml.

I also updated the JS dependencies by doing various combinations of `npm install` and `npm update` and adding a default (not sure if it can be overridden in any easy way) "mode" to the webpack configuration.

If things are really confusing I should probably start by looking through differences with the cookie cutter template: https://github.com/jupyter-widgets/widget-cookiecutter


larsoner:
> It seems ipywidgets is no longer installing the extension (which should be done automatically).

Is it possible it's installed but just not enabled? You might need a `jupyter nbextension enable` command like:

https://ipyvolume.readthedocs.io/en/latest/install.html#developer-installation

djhoese:
@larsoner I'll try that, but we are using the following *inside* vispy's backend code:

```python
    install_nbextension(jsdir, overwrite=force, destination='vispy',
                        symlink=(os.name != 'nt'), **kwargs)
```

which used to work without explicitly installing/enabling the extension from the command line.

larsoner:
A quick Google suggests that there is at least an `enable_nbextension` command, too, so it's worth a shot.

Also maybe we should have people install/enable it themselves at some point. It's perhaps not so nice messing with people's `IPython` configs. At some point we should see if there is a modern accepted way to deal with this stuff. I'd probably just do what `ipyvolume` does since it's pretty actively developed (or at least was actively developed more recently than the VisPy extension).

djhoese:
I'll talk to the ipywidgets folks on gitter. We'll have to make sure the conda-forge package sets everything up properly (@kmuehlbauer).

Issue #1456, opened 2018-03-27
Status of EGL backend


landersson:
Hi, I'd like to use the EGL backend for accelerated offscreen rendering into a pbuffer using the NVIDIA OpenGL drivers under Linux. A C++ test program using EGL/GL works fine, but when trying to use the vispy 'egl' backend, I'm getting the message:

RuntimeError: Could not import backend "EGL":
Not ready for testing

What is the status of the EGL backend? If it's currently not functional, what remains to be done to make it usable?

davidh-ssec:
As far as I know this backend should work fine. @larsoner would know more. What code are you running exactly that is producing this error?

landersson:
It's clearly disabled by default due to "Not ready for testing". I've made a few changes here and there and it now seems to work for, at least for headless OpenGL and OpenGL ES2 rendering into a pbuffer. Would you consider merging a PR?

larsoner:
Sure!

PR #1602, opened 2019-03-18
Offscreen EGL rendering code (replacement of #1464)

yogi81:
Additions to and replacement of a Pull Request #1464 
Now the code should work and render properly
It is a most full fix for #1456

### DIFF
--- a/examples/offscreen/simple_egl.py
+++ b/examples/offscreen/simple_egl.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+# vispy: testskip
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+"""
+This is a simple egl example that produces 2d plot.
+
+Was tested on Linux with Nvidia drivers.
+
+Requires libEGLso.1 to be found in the system.
+"""
+import vispy
+import numpy as np
+import vispy.plot as vp
+import vispy.io as io
+vispy.use(app='egl')
+
+# Check the application correctly picked up egl 
+assert vispy.app.use_app().backend_name == 'egl', 'Not using EGL'
+
+if __name__ == '__main__':
+
+    data = np.load(io.load_data_file('electrophys/iv_curve.npz'))['arr_0']
+    time = np.arange(0, data.shape[1], 1e-4)
+
+    fig = vp.Fig(size=(800, 800), show=False)
+
+    x = np.linspace(0, 10, 20)
+    y = np.cos(x)
+    line = fig[0, 0].plot((x, y), symbol='o', width=3, title='I/V Curve',
+                          xlabel='Current (pA)', ylabel='Membrane Potential (mV)')
+    grid = vp.visuals.GridLines(color=(0, 0, 0, 0.5))
+    grid.set_gl_state('translucent')
+    fig[0, 0].view.add(grid)
+
+    fig.show()
+
+    img = fig.render()
+    io.write_png("egl.png", img)
--- a/vispy/app/backends/_egl.py
+++ b/vispy/app/backends/_egl.py
@@ -128,12 +128,6 @@ class ApplicationBackend(BaseApplicationBackend):
 
 # ------------------------------------------------------------------ canvas ---
 
-COMMON_ATTRIB_LIST = [egl.EGL_RED_SIZE, 8,
-                      egl.EGL_BLUE_SIZE, 8,
-                      egl.EGL_GREEN_SIZE, 8,
-                      egl.EGL_ALPHA_SIZE, 8,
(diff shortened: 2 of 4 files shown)


examples/offscreen/simple_egl.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+# vispy: testskip
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+"""
+This is a simple osmesa example that produce an image of a cube
```
kmuehlbauer: Since this is a copy of the osmesa example, please delete any relating lines.

examples/offscreen/simple_egl.py
```diff
@@ -0,0 +1,48 @@
...
+If you have both osmesa and normal (X) OpenGL installed, execute with
+something like the following to pickup the OSMesa libraries::
+
+    VISPY_GL_LIB=/opt/osmesa_llvmpipe/lib/libGLESv2.so \
+    LD_LIBRARY_PATH=/opt/osmesa_llvmpipe/lib/ \
+    OSMESA_LIBRARY=/opt/osmesa_llvmpipe/lib/libOSMesa.so \
+    python examples/offscreen/simple_osmesa.py
+"""
+import vispy
+#vispy.use(app='osmesa')  # noqa
```
kmuehlbauer: remove

examples/offscreen/simple_egl.py
```diff
@@ -0,0 +1,48 @@
...
+import vispy.app
+
+vispy.app.use_app('egl')
+
+import numpy as np
```
kmuehlbauer: This imports brakes flake test, most probably due to not being at the top. Have a look at the osmesa example how to get along without flake8 complaining.

examples/offscreen/simple_egl.py
kmuehlbauer: Or change to something more appropriate.

examples/offscreen/simple_egl.py
kmuehlbauer: See https://travis-ci.org/vispy/vispy/jobs/508119100#L2566-L2571

yogi81: should be done now

djhoese:
This looks good, but I have a couple questions:

1. Do we know what version of EGL that this all changed? Or do we just assume this never really worked?
2. Would it be possible to test this in travis?

codypiersall:
@djhoese I'm in the camp that it never could have worked for most people's setups.  Probably whoever developed it was on a 32-bit build of Python on Windows.  It would work in that case.

yogi81:
> This looks good, but I have a couple questions:
> 
>     1. Do we know what version of EGL that this all changed? Or do we just assume this never really worked?

egl.h defines EGLContext  as void * (https://github.com/KhronosGroup/EGL-Registry/commit/88f2ff1f0d80212cea71a5d4a2404ef7bb3ef64e#diff-df83a81fd8b7f73af8028be41011ae55R60)
and it is like that since it was committed into the repository (around 2017)

The original egl vispy patches were written in 2018. So this means, that it was violating egl specification from the beginning on by having return value of int instead of void * (https://github.com/KhronosGroup/EGL-Registry/commit/88f2ff1f0d80212cea71a5d4a2404ef7bb3ef64e#diff-df83a81fd8b7f73af8028be41011ae55R123)

djhoese:
So `egl.h` was added to the repository in 2017, but it has existed in the project for a while. Maybe it is autogenerated. Not something you need to worry about.

I'm not sure what you mean by "egl vispy patches", but the backend itself was first added in 2014.

codypiersall:
I think `egl.h` is exactly one of the things that @yogi81 needs to worry about; it looks like the header is a modified version of the one released by the khronos group: https://www.khronos.org/registry/EGL/api/EGL/egl.h.  Vispy uses ctypes to use the egl library.  As far as I can tell, the header file isn't actually used by Vispy at all, but it's nice to have on hand to make sure the ctypes definitions are right—which they weren't.

yogi81:
> I'm not sure what you mean by "egl vispy patches", but the backend itself was first added in 2014.

A sorry, what I mean is the previous pull request. It was created in 2018

djhoese:
iirc you can't load/use headers directly in ctypes...right? Or maybe that is cython?

yogi81:
yes, I can not load the headers, but the ctypes function usage must be equivalent to the definition in header.

If not, then you get errors, like we were having

djhoese:
I'm ok merging this, but I'd like to wait until #1634 is finalized and merged back in to this PR. I want to make sure any dependency changes in #1634 in the travis environments don't break this PR.

djhoese:
Closing this in favor of #1637.

PR #1610, opened 2019-03-23
Fix jupyter lab compatibility

djhoese:
Not sure how users' can install this without the source code right now, but this is a start. Do:

```bash
jupyter labextension install js
```

from the root of the repository. This uses the `js` directory contents to build the npm package. When using vispy from `jupyter lab` I get this in the browser console:

image

Sounds like I need to declare the model/view versions properly. Issue with that is we don't have an explicit model: https://github.com/jupyter-widgets/ipyleaflet/issues/95


djhoese:
So this latest commit was a big one. I'm still getting the same error as my initial post but I've done some major changes:

1. Updated vispy.js code to work with jquery 3 (deprecation warnings, etc), removed bundled jquery/ui/mousewheel packages, and removed the need for the jquery-mousewheel package. I'm still not sure if doing this in `events.js` is correct:

```javascript
var VispyCanvas = require('./vispycanvas.js');
var screenfull = require("screenfull");
var $ = require('jquery');
require('jquery-ui');
```

But it works in a notebook.

2. Moved all vispy.js code unbundled to the vispy python repository under `js/lib/` at least for testing right now. Too much weirdness was going on with the `gulp` configuration being used by vispy.js and this was easier than bundling things twice.
3. Rename npm package name for the extension to `vispy_jupyter` just to sanity check that there wasn't a collision of some sort between `vispy.js` and `vispy` the extension.
4. Added a missing `labplugin.js` which I thought was going to be the missing link, but nope.
5. Removed the `install_nbextension` in `vispy/app/backends/_ipynb_webgl.py` because I thought it may be tampering with the lab stuff...just in case.

@jasongrout on gitter has been trying to point me in the right direction. Here is some information that Jason may be interested in for an overview of the process:

1. Python widget class is in `vispy/app/backends/ipython/_widget.py`
2. The `webgl-backend.js` code should be the main javascript code being loaded/exported for the extension.
3. The above webgl-backend.js loads `vispy.js` which connects/loads all of the other low-level parts.
4. An example notebook is in `examples/ipynb/webgl_example_1.ipynb`, just run the cells down to the Canvas instantiation where the widget should be shown.

larsoner:
Sounds like good progress, and I would do whatever @jasongrout suggests :)

djhoese:
Ok I made progress. Now I'm getting actual javascript errors which is more logical. The main fix was the change in the last commit where `package.json` actually declares where jupyterlab should get the plugin from.

Now to fix the js errors.

djhoese:
@kmuehlbauer @codypiersall I think this is ready to play with. The tests on appveyor seem to fail and I think travis might fail from size checks but jupyter lab and notebook work for me. See `examples/ipynb` for notebooks.

Once cloned and if you have jupyter widgets extension installed you can do:

```bash
jupyter labextension install js
```

Which installs all of the javascript under `js/`. There are still a couple best practice things that I want to look at and have asked on jupyter widgets gitter. Like what if we don't want to deal with the javascript stuff when the python package is installed (this is only one of many backends). Or what is the recommended way for users to have to install the nb extension (we have some at runtime stuff currently commented out).

I also would like to put the JS on npm so it can be installed with the above command pointing to `vispy` instead of `js`.

djhoese:
And don't miss the updated README for vispy.js: https://github.com/vispy/vispy.js/blob/master/README.md

kmuehlbauer:
@djhoese I tried to get things running with a fresh vispy-clone using your branch here, but did not manage to get the rotating cube running in a notebook. I suspect, that I miss something in my installation procedure to set everything up correctly.

Here is what I did so far:

- cloned visp
- added your fork and checked out your branch
- installed vispy
- `jupyter nbextension install js --sys-prefix`

After that didn't work (Canvas turns black and then immediately white) I tried to re-create the extension using:

```
jupyter nbextension install vispy --sys-prefix
jupyter nbextension enable --py widgetsnbextension --sys-prefix
```
But that also didn't help.

djhoese:
@kmuehlbauer Try this:

```bash
jupyter labextension install @jupyter-widgets/jupyterlab-manager
```

then install the vispy extension.

djhoese:
@larsoner @kmuehlbauer This is ready for review.

kmuehlbauer:
> And to summarize the conversation from gitter, @kmuehlbauer's main issue was previous configuration in ~/.jupyter. Things seem to work.

Correct. I've had issues with some stale stuff in `.jupyter`.

djhoese:
The GLIR file change was a mistake I made in the geometry shaders PR (where the SHADERS command was removed). The backend fixes have two important parts:

1. Remove the automatic installation of the widget. This should be handled by setup.py from now on by putting the "static" files in the "share" directory of the environment.
2. Add an explicit Model definition to the widget. We were using the default for notebooks by not specifying it, but I think jlab didn't like that. This sets us up for custom model stuff in the future if we'd like.

larsoner:
I couldn't figure out how to get it to work as a plain `nbextension`, but I assume this is me not understanding the install commands properly.

After `pip install jupyterlab`'ing, I could run the `labextension` commands above to install, then ran `jupyter-lab`, ran your examples, then ran [one of mine after copy-pasting commands](https://gist.github.com/larsoner/25fa656c6d6e0b02b56c40a571bfb77c):

Screenshot from 2019-04-04 10-15-43

It's horribly slow, but it plotted! So I'd say we have some work to do on the install instructions (e.g., how to do `nbextension` install instead of `labextension` install), but I'm impressed it worked as well as it did -- I did not expect the `scene` code to even produce a plot!

So +1 for merge from me and we can fix other things later. (IMO we should at least figure out and document well the `nbextension` install instructions before releasing.)

djhoese:
Nice visual @larsoner. I'm a little confused by what you tried. Notebook extensions are a separate process from jupyter lab extensions and one can't be used in the other. Did you follow the installation instructions on the vispy website or just what I posted here?

I doubt it makes a difference but the command to start jupyter lab is technically `jupyter lab`.

larsoner:
Just what you posted here

djhoese:
Try this: http://vispy.org/installation.html#jupyter-extension

I should probably rename that to make it clear this is for notebooks.

larsoner:
When I follow those I get:
```Console
$ jupyter nbextension enable --py vispy --user
Enabling notebook extension vispy/extension...
      - Validating: problems found:
        - require?  X vispy/extension
```

djhoese:
Summary of gitter chat: when installing from source you need to manually install the extension (optionally use the `--symlink` flag) to make it enable-able. A normal pip install doesn't seem to need this because the `vispy/static` files are copied properly. I've cleared up the instructions on my local documentation and will commit them later.

For now, @larsoner has done a great job of pointing out things that don't work:

1. Mouse events (I have an idea how to fix these, I think it is my fault they broke in vispy.js)
2. Shader errors happen in certain cases. From what I can tell it is shader objects trying to re-send themselves when uniforms change. Not sure where to start with this.

djhoese:
@larsoner @kmuehlbauer this PR is ready for re-review. I think I've fixed all of the things that *should* have been working but never worked in the past. @larsoner, it'd be great if you could test your example from above.

larsoner:
Interactivity works great in my example! No more warnings in the Chrome developer console, either. Awesome stuff @djhoese

kmuehlbauer:
@larsoner Your example downloaded 1.5 GB data to my network share and the following automated extraction went straight ahead over my disk quota leaving me with a stalled system :joy: Still recovering...

larsoner:
Oops sorry, yes I probably should have warned you that the download is quite large, then it has to decompress the file...

djhoese:
The travis failure is only due to the size of this PR. Merging now...

Issue #1619, opened 2019-04-16
ValueError: Missing file: vispy/vispy/static/extension.js


os-gabe:
With a fresh clone of the vispy repository, following the directions [here](http://vispy.org/installation.html), I am seeing the error below when trying to install. 

```
Platform: Linux-4.18.0-17-generic-x86_64-with-Ubuntu-18.04-bionic
Python:   3.5.7 (default, Mar 30 2019, 01:05:18)  [GCC 7.3.0]
NumPy:    1.16.2
Backend:  PyQt5
pyqt4:    None
pyqt5:    ('PyQt5', '5.12.1', '5.12.2')


$ pip install -e .
Obtaining file:///home/gabe/code/vispy
    Complete output from command python setup.py egg_info:
    setup.py entered
    $PATH=/home/gabe/Envs/falkor-pointclouds/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
    Compiling vispy/visuals/text/_sdf_cpu.pyx because it changed.
    [1/1] Cythonizing vispy/visuals/text/_sdf_cpu.pyx
    running egg_info
    running jsdeps
    3.5.2
    3.5.2
    Installing build dependencies with npm.  This may take a while...
    
    > nodemon@1.18.11 postinstall /home/gabe/code/vispy/js/node_modules/nodemon
    > node bin/postinstall || exit 0
    
    vispy@0.2.0 /home/gabe/code/vispy/js
    ├─┬ @jupyter-widgets/base@1.2.4
    │ ├─┬ @jupyterlab/services@3.2.1
    │ │ ├─┬ @jupyterlab/coreutils@2.2.1
    │ │ │ ├─┬ ajv@5.1.6
    │ │ │ │ ├── co@4.6.0
    │ │ │ │ ├── json-schema-traverse@0.3.1
    │ │ │ │ └── json-stable-stringify@1.0.1
    │ │ │ ├─┬ comment-json@1.1.3
    │ │ │ │ └─┬ json-parser@1.1.5
    │ │ │ │   └── esprima@2.7.3
    │ │ │ ├── minimist@1.2.0
    │ │ │ ├── moment@2.21.0
    │ │ │ ├── path-posix@1.0.0
    │ │ │ └─┬ url-parse@1.4.6
    │ │ │   ├── querystringify@2.1.1
    │ │ │   └── requires-port@1.0.0
[...]

Issue #1606, opened 2019-03-21
MAINT: Wheels


larsoner:
Now that VisPy uses some compiled code, we should probably produce wheels for releases. (Forcing end users of Windows and macOS to properly set up a dev environment is cruel.)

@djhoese any experience with this? If not, I can look into it. I've been curious to do so at some point anyway.

djhoese:
The aggdraw project that I maintain builds wheels (https://github.com/pytroll/aggdraw) using the cibuildwheels project (https://github.com/joerick/cibuildwheel).

kmuehlbauer:
This aims on users using plain system python installation, no conda etc., right? .

larsoner:
Yes indeed

larsoner:
(really anything other than `conda` I guess, which will take care of building a different way)

kmuehlbauer:
resolved by #1663

PR #1620, opened 2019-04-18
Fix setup.py to install vispy even if javascript compilation fails

djhoese:
In #1610 I fixed setup.py to match the [ipywidgets cookiecutter](https://github.com/jupyter-widgets/widget-cookiecutter) for a notebook/jlab extension. As part of this PR I had removed the `vispy/static/*.js` files which are the bundled output of the javascript build process. Given the rest of the traditional ipywidgets installation process this meant that `setup.py` was now **requiring** that the javascript files be built before installing the rest of vispy. Obviously for a package like vispy this can't be a requirement (a notebook is only one possible backend/use case for our users). This issue is discussed further in #1619 where some users may not have `npm` installed or may have an older version that is unable to build the extensions for whatever reason.

I saw two possible solutions (and asked for more on https://github.com/jupyter-widgets/widget-cookiecutter/issues/49):

1. Add the static javascript files to the git repository so users always have at least one version of the static files. This is not great for file sizes in the git history where each update to the JS likely means a new ~1.5MB (see below). It could also make it harder to tell if `sdist` packages are including the "old" JS builds or are building them new when preparing a package for PyPI.

```bash
$ du -hs vispy/static/
1.5M	vispy/static/
```

2. Modifying the `setup.py` code to allow for JS building to fail. This is what I ended up going with. This puts more work on the maintainers (@kmuehlbauer @larsoner ) to make sure that released PyPI sdists have the javascript files. I updated the [releasing instructions](https://github.com/vispy/vispy/wiki/Releasing-new-version-of-VisPy-python-package#generate-pypi-release-bundles
) already. This choice felt the cleanest to me in the long run and seems less likely that I'll screw it up when building an sdist.

The main difficulty here is that in `setuptools/distutils` there is no difference between the `sdist` that `pip` and `python setup.py install` do and the `sdist` we use when building a release tarball (`python setup.py sdist`). I could make a custom command that we use for building the tarball that makes the javascript build required, but that becomes yet another hack and customization in the vispy build workflow on top of all of the existing complex `doc` and other build processes.

@deciph and @os-gabe could you test this on your own systems by doing:

```bash
pip install git+https://github.com/djhoese/vispy.git@bugfix-setuppy-js
```

Does it install on your systems?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -98,12 +98,17 @@ developers to use the development version on Github (master branch). You need
 to clone the repository and install VisPy with
 ``python setup.py install``.
 
-As a one-liner, assuming `git` is installed ::
+As a one-liner, assuming `git` is installed::
 
-    git clone https://github.com/vispy/vispy.git && cd vispy && python setup.py install --user
+    git clone --recurse-submodules https://github.com/vispy/vispy.git && cd vispy && python setup.py install --user
 
 This will automatically install the latest version of vispy.
 
+If you already have vispy cloned, you may need to update the git submodules
+to make sure you have the newest code::
+
+    git pull
+    git submodule update --init --recursive
 
 Structure of VisPy
 ------------------
--- a/setup.py
+++ b/setup.py
@@ -159,7 +159,14 @@ def js_prerelease(command, strict=False):
                     log.warn('rebuilding js and css failed')
                     if missing:
                         log.error('missing files: %s' % missing)
-                    raise e
+                    # HACK: Allow users who can't build the JS to still install vispy
+                    if not is_repo:
+                        raise e
+                    log.warn('WARNING: continuing installation WITHOUT nbextension javascript')
+                    # remove JS files from data_files so setuptools doesn't try to copy
+                    # non-existent files
+                    self.distribution.data_files = [x for x in self.distribution.data_files
+                                                    if 'jupyter' not in x[0]]
                 else:
                     log.warn('rebuilding js and css failed (not a problem)')
                     log.warn(str(e))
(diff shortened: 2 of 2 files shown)


deciph:
I tried the install using the command you provided but still no success.  The command window output is 1255 lines long which seems obnoxious to post in here (idk, maybe its ok) so I put it out on pastebin.

I did notice a message saying c++  14 is required which I do have installed.

Here's a link to the output:
https://pastebin.com/7stQqDRN

djhoese:
@deciph Thanks for testing. Based on your output it looks like the javascript problems are "solved". As for the C++ issues, vispy has C extensions that *must* be built. Normally if you were installing from PyPI a project would provide binary wheels to install which skip any compilation on the user end. We do not currently provide wheels, but hope to and it is something @larsoner is pushing for (see #1606).

Regardless of our wheels, to install from vispy source you must have a C/C++ compiler available that is compatible with the version of python you are running. However, it seems you are using Anaconda and I would have expected them to bundle a compiler with it. @kmuehlbauer @larsoner am I mis-remembering something about this?

Lastly, @larsoner @kmuehlbauer is it possible (do we want it to be) to make the SDF compilation optional?

kmuehlbauer:
@djhoese I'm running low in time over Easter, so I might not respond fast the next week.
Regarding sdf IIRC @rougier has some solution over at glumpy.

I did not have a problem using anaconda to compile the sdf extension. But good chance that I somehow pulled a compiler in. But we could have a look at Travis what we actually pull in there.

kmuehlbauer:
@deciph Just saw that you are  on Windows. For that I also struggled with installing a compiler. Will have a look the next day.

djhoese:
@os-gabe Have you had a chance to try this?

os-gabe:
Thanks for the reminder. I just tested and and was able to successfully install via `pip install -e .`

deciph:
@kmuehlbauer I was able to get vc++ 14 installed correctly by following the instructions in this link: 
https://stackoverflow.com/questions/49610212/pip-error-even-microsoft-visual-c-14-0-is-installed/50671800#50671800

@djhoese I installed vispy dev using the master branch and it worked perfectly.  
"Successfully installed vispy-0.6.0.dev0"

Thank you for your help on this!

kmuehlbauer:
@deciph Great! Glad you found a working solution! I did not find the time yet to touch my windows box.

PR #1621, opened 2019-04-18
Fix PySide2 qWait monkey patching

mirco-ackermann:
The indention of the line
	QtTest.QTest.qWait = qWait
should be so that it is only executed when the qWait function
needs to be monkey-patched.


hmaarrfk:
I don't have automated testing in my qt code ;) bad bad bad. but I provide it as "samples" of what is possible to end users.

djhoese:
This is a very simple fix and seems to make sense to me.

PR #1622, opened 2019-04-18
Add submodules to one-liner in README

mirco-ackermann:
This updates the one-liner to install from git to also download the submodules.


djhoese:
This is fixed in #1620 already. Just made it this morning. Thanks though.

PR #1623, opened 2019-04-18
MAINT: Use gl-ci-helpers

larsoner:
I recently worked with the `vtki` folks to port over the OpenGL code. This allows us to share a method for getting Windows to have usable OpenGL. No more need to set the VISPY_GL_LIB because the updated solution overwrites `C:\Windows\System32\opengl32.dll` directly.


djhoese:
@larsoner Are you ok with this being merged or would you like to wait for @astrofrog's review?

larsoner:
This is pretty straightforward so can be merged

PR #1624, opened 2019-04-19
Add 'translate_speed' to TurntableCamera and ArcballCamera

### LINKED ISSUE #914: Add support for "TurntableCamera" translate speed.

os-gabe:
Add translate_speed keyword argument to turntable and arcball cameras

### DIFF
--- a/vispy/scene/cameras/arcball.py
+++ b/vispy/scene/cameras/arcball.py
@@ -27,6 +27,8 @@ class ArcballCamera(Base3DRotationCamera):
         The distance of the camera from the rotation point (only makes sense
         if fov > 0). If None (default) the distance is determined from the
         scale_factor and fov.
+    translate_speed : float
+        Scale factor on translation speed when moving the camera center point.
     **kwargs : dict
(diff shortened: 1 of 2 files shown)


djhoese:
Closes #914 (I wanted a comment with the issue mentioned)

os-gabe:
Gotcha. Thanks for updating the title.

Issue #1627, opened 2019-04-23
Potential problem with triangulation


rougier:
I've just fixed some code in glumpy because of a change in the triangle library and it may impact vispy as well (more specifically https://github.com/vispy/vispy/blob/master/vispy/geometry/triangulation.py).

The original bug report: https://github.com/glumpy/glumpy/issues/214
The fix: https://github.com/drufat/triangle/issues/34

djhoese:
This should be fixed with #1628.

PR #1628, opened 2019-04-23
Make sure segments have the right format

rougier:
See #1627


djhoese:
@rougier Thanks for this. I assume this is backwards compatible?

rougier:
Good question. I'm not sure actually.

PR #1634, opened 2019-04-29
Remove forced sdl2 version in travis config

djhoese:
See https://github.com/conda-forge/sdl2-feedstock/issues/13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,7 +119,7 @@ install:
     # EGL and mesa causes segmentation faults. See issue #1401.
     # can remove the github numpydoc install once >0.8.0 is released and get it from conda
     - if [ "${DEPS}" == "full" ]; then
-        conda install --yes pyopengl networkx pysdl2 sdl2=2.0.8;
+        conda install --yes pyopengl networkx pysdl2;
         pip install git+https://github.com/numpy/numpydoc;
         if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
           conda install --yes matplotlib jupyter pyqt=5 pillow decorator six scikit-image glfw;


kmuehlbauer:
@djhoese Maybe this is a problem with SDL2 vs glfw (with regards to the dependencies in conda-forge ).

djhoese:
@kmuehlbauer Are you talking about the additional test dependencies that @matham mentioned on the conda-forge issue?

kmuehlbauer:
@djhoese Yes, something along these lines:

Here is the [glfw recipe](https://github.com/conda-forge/glfw-feedstock/blob/1993837c78eebb64c1848ad0477bb3abe615a4bd/recipe/meta.yaml#L29) and here the [glew recipe](https://github.com/conda-forge/glew-feedstock/blob/97f1cdd47d67e22e16de45e9cc60b7f137f803ce/recipe/meta.yaml#L33)

The latter has the cdt-stuff while the glfw has direct dependencies to xorg-packages. Not sure if this might cause the problems.

djhoese:
What does `cdt` mean/do?

kmuehlbauer:
Good question,, this is all there is in the docs: https://conda-forge.org/docs/maintainer/knowledge_base.html#libgl

djhoese:
So those packages listed are for yum/RHEL and travis is running Ubuntu. We already install quite a few GL libraries through apt: https://github.com/vispy/vispy/blob/master/.travis.yml#L7

Are we thinking these conflict? Maybe these aren't compatible with the versions built on Azure/RHEL and used to build SDL2?

matham:
There could be at least two main causes of the issue from my understanding. 

1. This is a problem specific to the CI where there's no window, hence why `Xvfb` has to be used. And for some reason that broke in the most recent conda install. But on a real system this sdl2 issue won't arise. Has someone confirmed this also fails on a local machine with a similar install?
2. That it fails everywhere due to some linking issue with libgl. We seem to be assuming it's `#2`. If it is, it could be some incompatibility, either because
    1. The conda build is compiled on azure, and the test is on travis. To me this is less likely because I'd trust the conda people to have figured out how to properly link during the compilation stage so that it works on any system at runtime, even with apt installed dependencies.
    2. That you're missing something from your conda install. I.e. you need to install mesa or something or other from conda - like what I mentioned in the other post. This to me seems to be the most likely explanation, given the assumption that the conda people know what they are doing.

But I'm not overly familiar with linux systems and linking etc. so please take it with a grain of salt.

kmuehlbauer:
@matham Thanks for the thoughts. I just updated `sdl2` in my conda env from 2.0.8 to 2.0.9 and get the same error message as in the travis logs (linux box). 

Note, that we access `sdl2` via `pysdl2`.

djhoese:
I've restarted the failed jobs now that there is a new build of sdl2 on conda-forge. Let's see if this passes.

djhoese:
It works! @kmuehlbauer do you have time to double check this on your local machine too?

kmuehlbauer:
@djhoese I'll test tomorrow morning first and can also do the merge afterwards, if everything is OK.

PR #1636, opened 2019-05-07
Fix ValueError exception from ambiguous truth value in PanZoomCamera

mirco-ackermann:
When zooming with the PanZoomCamera, I get a ValueError: The truth value of an array with more than one element is ambiguous. Here is a fix that works when center is none using a line that was changed in commit a8ee4f45a56ff9509546cade6439cdda841b7a04.

I stumbled upon this by running vispy/examples/demo/scene/picking.py using python3.7.2 and numpy 1.16.2.

### DIFF
--- a/vispy/scene/cameras/panzoom.py
+++ b/vispy/scene/cameras/panzoom.py
@@ -86,7 +86,7 @@ class PanZoomCamera(BaseCamera):
             current center.
         """
         # Init some variables
-        center = center or self.center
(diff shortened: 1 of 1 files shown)


djhoese:
Ugh I was suspicious when I merged the recent change for this. Thanks for the fix. I've restarted the failing test. When it passes I'll merge.

mirco-ackermann:
Nice!

PR #1637, opened 2019-05-10
[MRG] Offscreen EGL rendering code (replacement of #1464 and #1602)

djhoese:
This is a replacement mainly for #1602 because I needed to merge current vispy master to that PR but because it is @yogi81's master branch I do not have permissions to do so (if it had been a separate branch I could have). If this passes tests then I would consider this good to be merged.


djhoese:
@yogi81 look right to you? ^

yogi81:
yes. There is still one problem, which is not yet solved. I have just found out, that to make it work completely without X one would need to use additional egl commands, which is not done in the code.

djhoese:
@yogi81 what commands?

yogi81:
static const int MAX_DEVICES = 4;
  EGLDeviceEXT eglDevs[MAX_DEVICES];
  EGLint numDevices;

  PFNEGLQUERYDEVICESEXTPROC eglQueryDevicesEXT =
    (PFNEGLQUERYDEVICESEXTPROC)
    eglGetProcAddress("eglQueryDevicesEXT");

  eglQueryDevicesEXT(MAX_DEVICES, eglDevs, &numDevices);

  printf(“Detected %d devices\n”, numDevices);

  PFNEGLGETPLATFORMDISPLAYEXTPROC eglGetPlatformDisplayEXT =
    (PFNEGLGETPLATFORMDISPLAYEXTPROC)
    eglGetProcAddress("eglGetPlatformDisplayEXT");

  eglDpy = eglGetPlatformDisplayEXT(EGL_PLATFORM_DEVICE_EXT, 
                                    eglDevs[0], 0);

djhoese:
@yogi81 What do you think needs to happen to add those commands to vispy? Is this worth merging without those extra commands? Do they only need to be added when X isn't present? What if X is present and these are still executed?

djhoese:
I'm going to merge this for now. It seems the remaining issue can be addressed in the future. @yogi81 Would you mind making another PR or at least an issue describing the issue and what needs to be done?

PR #1654, opened 2019-06-29
Add pyproject.toml file for better build support

### LINKED ISSUE #1651: setup.py attempts to import numpy before installing it
On master `setup.py` imports numpy on line 46 even though numpy is a dependency of vispy so it is not guaranteed it will be installed yet.

This causes `pip install git+https://github.com/vispy/vispy.git` to fail unless you already have numpy installed.

djhoese:
Closes #1651 

This adds a `pyproject.toml` to match PEP 518 specification so that build dependencies can be handled outside of `setup.py`. I'm hoping that @os-gabe @larsoner @renefritze and @kmuehlbauer could review and maybe test this for me as it is. Mainly, I'm wondering if the numpy/py27 setting is "good enough" for now. We won't be supporting python 2.7 after this next release, but it should at least work (is PEP 518 supported for py27?).

I'd like to add to this later with some changes to how `numpy` is imported in `setup.py` since I don't *think* it is needed the way we have it.

Also, does anyone know if there is a way to optionally require `npm` so that the jupyter widget can be built/compiled?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,10 @@
+[build-system]
+requires = [
+    "wheel",
+    "setuptools",
+    "Cython>=0.29.2",
+    "numpy; python_version=='2.7'",
+    "numpy==1.13.3; python_version=='3.5'",
+    "numpy==1.13.3; python_version=='3.6'",
+    "numpy==1.14.5; python_version>='3.7'",
+]
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,6 @@ from setuptools.command.build_py import build_py
 from setuptools.command.build_ext import build_ext as _build_ext
 from setuptools.command.egg_info import egg_info
 from subprocess import check_call
-import numpy as np
 
 
 try:
@@ -305,7 +304,6 @@ setup(
     },
     packages=find_packages(),
     ext_modules=cythonize(extensions),
-    include_dirs=[np.get_include()],
     package_dir={'vispy': 'vispy'},
     data_files=[
         ('share/jupyter/nbextensions/vispy', [


larsoner:
No idea about npm

kmuehlbauer:
> No idea about npm

Me neither :sad:

renefritze:
> Also, does anyone know if there is a way to optionally require `npm` so that the jupyter widget can be built/compiled?

Not directly about npm, but rather optional requirements in general: https://github.com/pypa/pip/issues/6144
To take use this, I think npm would need some kind of pseudo python-wrapper package that fails to install if npm itself is not installed on the system already.


> is PEP 518 supported for py27?

With `pip` it is, yes. 

Install from source with this branch works for me too in a 3.7 virtualenv on debian.

os-gabe:
Installation worked for me in linux with clean 2.7 and 3.6 virtual environments via:
`pip install git+https://github.com/djhoese/vispy.git@add-pyproject-toml`

Like @kmuehlbauer I also see that it installed numpy 1.16.4 rather than the `numpy==1.13.3` specified in `pyproject.toml`

renefritze:
> Like @kmuehlbauer I also see that it installed numpy 1.16.4 rather than the `numpy==1.13.3` specified in `pyproject.toml`

Yeah, this is somewhat confusing. The older numpy will be installed during the setup time for vispy, while building its wheel in a isolated env. This wheel is then installed in the actually intendend env, where pip will just pick the latest numpy if it's not further restricted from for example `install_requires`.

kmuehlbauer:
> Yeah, this is somewhat confusing. The older numpy will be installed during the setup time for vispy, while building its wheel in a isolated env. This wheel is then installed in the actually intendend env, where pip will just pick the latest numpy if it's not further restricted from for example `install_requires`.

@renefritze This makes perfect sense! Thanks for pointing this out!

djhoese:
My latest commit removes the `import numpy as np` in the global `setup.py` context and removes the redundant `include_dirs` keyword argument. Running this on my local system with a conda environment that doesn't have numpy:

```
(vispy_no_numpy) [davidh@thanatos vispy]$ pip install .
Processing /Users/davidh/repos/git/vispy
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Collecting numpy (from vispy==0.6.0.dev0)
  Downloading https://files.pythonhosted.org/packages/6b/be/608b7f72b851472388eafc010a5d46dae5d41610d0ac5df4c98c2ed1b865/numpy-1.16.4-cp37-cp37m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl (13.9MB)
     |████████████████████████████████| 13.9MB 2.5MB/s 
Building wheels for collected packages: vispy
  Building wheel for vispy (PEP 517) ... done
  Stored in directory: /private/var/folders/k6/nsv7qg197bj58rkw2b7mb2ch0000gn/T/pip-ephem-wheel-cache-vbbyrv7x/wheels/45/63/b3/ca06d2a515f562230111fb6ca8e652fd69aa0b33a1e2bcdbdb
Successfully built vispy
Installing collected packages: numpy, vispy
Successfully installed numpy-1.16.4 vispy-0.6.0.dev0

(vispy_no_numpy) [davidh@thanatos vispy]$ conda list numpy
# packages in environment at /Users/davidh/anaconda/envs/vispy_no_numpy:
#
# Name                    Version                   Build  Channel
numpy                     1.16.4                   pypi_0    pypi
```

So all looks good. Will merge after tests pass.

PR #1660, opened 2019-07-09
Fix reading depth buffers with read_pixels

### LINKED ISSUE #1250: Impossible to read the depth buffer
According to the document http://vispy.org/gloo.html, `vispy.gloo.FrameBuffer.read` supports read the depth buffer.  However in practice it always returns an numpy array whose shape is either `(w, h, 3)` or `(w, h, 4)`, which is not the depth buffer.

### LINKED ISSUE #1297: Unable to draw ellipse in v.0.5.0, while it works in v.0.4.0
Hi,

I am new to VisPy, and in the process of finding out whether it is possible to move over from Matplotlib to VisPy for my application (a viewer for weather radar data), without loss of essential features. I have experienced a huge increase in speed compared to Matplotlib for plotting the radar data only, which makes me enthusiastic, but there are several important features of my application for which I have to check whether I can generate them in VisPy.
While doing so I tried to draw an ellipse using an example I found, which works in v0.4.0, but doesn't in v.0.5.0. A difference between both versions is that the position/center of the ellipse is in the former referred to by 'pos', and in the latter by 'center'. Changing this doesn't solve the issue however, allthough no errors are generated anymore. I also did not found anything in the source code of vispy.scene.visuals.Ellipse that did not seem to be evaluated properly. Below is the code I used, which is the version that works in v.0.4.0. In v.0.5.0 should at least 'center' be substituted for 'pos'. Anyone having ideas about how to solve this?

```
import vispy
from vispy import gloo, app

class Canvas(vispy.scene.SceneCanvas):
    def __init__(self):
        vispy.scene.SceneCanvas.__init__(self, keys='interactive')
        self.size = (800, 800)
        self.show()
        
        try: self.unfreeze();
        except AttributeError: pass
        self.ellipse = visuals.Ellipse(pos=(400,400,0), radius=[30, 30],
                               color=(1, 0, 0, 1),
                               border_color=(1, 1, 1, 1),
                               start_angle=180., span_angle=150.)
        
        self._timer = app.Timer('auto', connect=self.on_timer, start=True)

    def on_timer(self, event):
        self.ellipse.radius[0] += 1
        self.ellipse.radius[1] += 1.5
        self.ellipse.span_angle += 0.6
        self.update()

        self.ellipse.radius = [320, 240]
        self.ellipse.span_angle = 150.
        self.update()

    def on_draw(self, ev):
        gloo.clear(color='black')
        self.draw_visual(self.ellipse)
        self.update();

if __name__ == '__main__':
    win = Canvas()
    import sys
        `vispy.app.run()
```

djhoese:
Closes #1250 

This is a simple hack for fixing reading depth buffers. The problem as described in the related issue was that when you asked to read the depth buffer the underlying `read_pixels` function was still asking for an RGB/RGBA array, meaning the color was being read. This PR fixes that by passing a `mode` parameter which can be color/depth/stencil and will use the appropriate GL constant when calling `glReadPixels`.

Questions for the maintainers:

1. Is `mode` the best way to solve this?
2. Should the return value always be 3D even if it is (rows, columns, 1) for depth and stencil?
3. I also realized that `stencil` mode doesn't work with the `crop` logic in frame buffer's `read` because there is no `self.stencil_buffer` attribute. Not sure the best way to fix this.

TODO:

- [ ] Tests
- [ ] Fix docstrings

### DIFF
--- a/vispy/gloo/framebuffer.py
+++ b/vispy/gloo/framebuffer.py
@@ -250,8 +250,11 @@ class FrameBuffer(GLObject):
         
         """
         _check_valid('mode', mode, ['color', 'depth', 'stencil'])
+        buffer = getattr(self, mode + '_buffer')
+        if buffer is None:
+            raise ValueError("Can't read pixels for buffer {}, "
+                             "buffer does not exist.".format(mode))
         if crop is None:
-            buffer = getattr(self, mode+'_buffer')
             h, w = buffer.shape[:2]
             crop = (0, 0, w, h)
         
--- a/vispy/gloo/framebuffer.py
+++ b/vispy/gloo/framebuffer.py
@@ -257,5 +257,4 @@ class FrameBuffer(GLObject):
         
         # todo: this is ostensibly required, but not available in gloo.gl
         #gl.glReadBuffer(buffer._target)
-        
-        return read_pixels(crop, alpha=alpha)
+        return read_pixels(crop, alpha=alpha, mode=mode)
--- a/vispy/gloo/tests/test_wrappers.py
+++ b/vispy/gloo/tests/test_wrappers.py
@@ -211,13 +211,14 @@ def test_wrappers():
 def test_read_pixels():
     """Test read_pixels to ensure that the image is not flipped"""
     # Create vertices
-    vPosition = np.array([[-1, 1], [0, 1],  # For drawing a square to top left
-                          [-1, 0], [0, 0]], np.float32)
+    vPosition = np.array(
+        [[-1, 1, 0.0], [0, 1, 0.5],  # For drawing a square to top left
+         [-1, 0, 0.0], [0, 0, 0.5]], np.float32)
 
     VERT_SHADER = """ // simple vertex shader
-    attribute vec2 a_position;
+    attribute vec3 a_position;
     void main (void) {
-        gl_Position = vec4(a_position, 0., 1.0);
+        gl_Position = vec4(a_position, 1.0);
     }
     """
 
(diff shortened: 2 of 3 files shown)


kmuehlbauer:
@djhoese The example is working and also the changes LGTM.

> * Is `mode` the best way to solve this?
> 
Not sure, `mode` is already used in `FrameBuffer.read` so it is just a continuation. 

> * Should the return value always be 3D even if it is (rows, columns, 1) for depth and stencil?
> 
I would say yes. Users can squeeze if they want to.

> * I also realized that `stencil` mode doesn't work with the `crop` logic in frame buffer's `read` because there is no `self.stencil_buffer` attribute. Not sure the best way to fix this.
>
I think I do not fully understand. There is `stencil_buffer` property available in `FrameBuffer`. Do you mean it is `None` in that case (which means, there is no stencil buffer available, right)? To my understanding, if the `stencil_buffer` attribute is `None` `read_pixels` should not be called if `mode='stencil'` is requested. We could check and raise some Error or Warning.

vispy/gloo/framebuffer.py
```diff
@@ -257,5 +257,4 @@ def read(self, mode='color', alpha=True, crop=None):
```
kmuehlbauer: We could move the `buffer = getattr(self, mode+'_buffer')` out of the `if crop` and check, if the requested buffer is available or not. If not, we would have to raise an Error I think.

djhoese: Yeah that could make sense and yes this is what I was talking about with stencil_buffer being None. I'll see what I can do.

PR #1670, opened 2019-07-23
Allow changing spectrogram parameters after it has been drawn

cimbi:
The PR adds ability to the spectrogram visual to modify the parameters after the first draw.

### DIFF
--- a/vispy/visuals/spectrogram.py
+++ b/vispy/visuals/spectrogram.py
@@ -54,3 +54,102 @@ class SpectrogramVisual(ImageVisual):
     def freqs(self):
         """The spectrogram frequencies"""
         return fft_freqs(self._n_fft, self._fs)
+
+    @property
+    def data(self):
+        """The data property of parent ImageVisual"""
+        return self._data
+
+    @property
+    def x(self):
+        """The original signal"""
+        return self._x
+
+    @x.setter
+    def x(self, x):
+        nan_mean = np.nanmean(x)
+        idx = np.isnan(x)
+        x[idx] = nan_mean
+        self._x = x
+        self._update_image()
+
+    @property
+    def n_fft(self):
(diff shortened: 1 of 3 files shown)


cimbi:
I think I have addressed all the points.

Regarding 3) - for my application in need the data property to calculate color limits and reset them back to the Image visual to bring up events in the spectrogram. Since I only supply 1D signal to the spectrogram visual I don't have the 2D data that go into the Image visual. I can use _data property of the Image visual but I guess it would make sense to get the data back by calling Image.data.

cimbi:
Do we need tests for the spectrogram visual?

vispy/visuals/spectrogram.py
```diff
@@ -38,19 +40,130 @@ class SpectrogramVisual(ImageVisual):
...
+            data = 20 * np.log10(data) if self._color_scale == 'log' else data
+            if self._normalize:
+                for i in range(data.shape[0]):
+                    data[i, :] -= np.mean(data[i, :])
+                    data[i, :] /= np.std(data[i, :])
+            return data
+        else:
+            return None
+
+    def _update_image(self):
```
djhoese: This method's contents look like they are indented 1 level too much.

djhoese:
Regarding the `data` property, would it be possible for you to subclass this Visual and do the color limit stuff internally? That way the `_data` usage seems less "wrong".

For tests, it would be great if you could find a way to add some tests, preferably tests that don't compare against saved images. My main concern and why I would like tests is that this subclasses from ImageVisual. I have future plans to update ImageVisual to be more flexible (don't hold on to the data array, in-GPU color limits, etc) so I'd like some tests to make sure that I know I didn't break the spectrogram.

djhoese:
Thanks again!

PR #1677, opened 2019-07-26
Add automatic PyPI deployment through Azure

djhoese:
This PR updates the azure configuration to:

1. Add sdist generation
2. Upload the sdist and all wheels to PyPI

This works by setting a special variable in the Azure web UI with a PyPI API token under my account. This token only has access to upload to the vispy project on PyPI, will do it as me, and the secret variable can only be seen by people who have permission on the azure vispy project. By setting the environment variable `TWINE_PASSWORD to this value `twine` should pick it up and use it to upload the packages.

The main thing I think I have wrong is that all the wheels and the sdist file are uploaded to the "wheelhouse" artifacts storage, but I don't think I can access that as a simple directory in a separate job (see the `twine upload` call). If there is no way to do that then I can do the upload in each job separately.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -38,3 +38,4 @@ exclude stdeb.cfg
 exclude .spyderproject
 recursive-exclude vispy *.c
 recursive-exclude js *
+recursive-exclude make *
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,7 +6,7 @@ variables:
   CIBW_BUILD_VERBOSITY: "2"
   CIBW_BEFORE_BUILD: "pip install -U numpy Cython jupyter ipywidgets"
   CIBW_BEFORE_BUILD_MACOS: "npm install npm@latest -g; pip install -U pip setuptools"
-  CIBW_BEFORE_BUILD_LINUX: "curl -sL https://rpm.nodesource.com/setup_8.x | bash -; yum install -y fontconfig xvfb nodejs; pip install -U pip setuptools; if [ `getconf LONG_BIT` == '64' ]; then pip install https://files.pythonhosted.org/packages/f2/00/6f332e63b33d24dc3761916e6d51402a7a82dd43c6ca8a96e24dda32c6b5/freetype_py-2.1.0.post1-py2.py3-none-manylinux1_x86_64.whl; else pip install https://files.pythonhosted.org/packages/b8/ea/adfbeb3762a03c1d1c1ff751101efb1fddb7aadffa991d39bdb486d0557e/freetype_py-2.1.0.post1-py2.py3-none-manylinux1_i686.whl; fi"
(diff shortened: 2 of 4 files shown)


larsoner:
> The main thing I think I have wrong is that all the wheels and the sdist file are uploaded to the "wheelhouse" artifacts storage, but I don't think I can access that as a simple directory in a separate job (see the twine upload call). If there is no way to do that then I can do the upload in each job separately.

I thought Azure had something about conditional jobs that only run if other ones succeed, but maybe that was build steps rather than jobs. If you can't figure it out, let me know and I'll look. Or if you find documentation that it's not possible let me know and I won't bother :) It would be nicer to upload in a single job because it ensures that all or none get uploaded, which is probably nicer when things go wrong.

djhoese:
@larsoner Yes, that's what I'm doing. Look at the last job in the config. That's why I said "if there is no way to do that then", meaning if I can't find a way to easily share artifacts between jobs then I'd have to do it per job where the wheels are already available.

larsoner:
Awesome!

djhoese:
Looks like my latest change made the config unrecognizable by azure.

djhoese:
@larsoner @kmuehlbauer After this passes I think I'm ready to merge and try to release 0.6.1. Any objections?

kmuehlbauer:
No objections, patch numbers are cheap :grinning:

larsoner:
Fine by me

djhoese:
@larsoner @kmuehlbauer So I think we have 3 choices for the linux wheel given the trouble I've had. The issue is that the `manylinux` wheel specification requires compatibility with CentOS 5. We need `npm` to build the jupyter extension javascript files. However, no modern version of node.js (which installs npm) is compatible with CentOS 5. Even semi-modern versions don't want to build from what I've tried.

So our 3 choices:

1. Don't provide linux wheels since they won't have the JS files.
2. Provide the linux wheels without JS files. Users who want the notebook extension will have to install from the source dist (no binary).
3. Build the JS and add the `static` directory to the MANIFEST.in and all wheels. Downside is they would have to be regenerated manually every couple releases to make sure they are up to date.

Option 4: I try harder at finding a CentOS 5 npm install.

larsoner:
For option (3) would we need to actually add these things to the `vispy` repository?

djhoese:
Yes because they are not built on the CIs otherwise. It is three (I think) `.js` files.

larsoner:
Ahh well that is not so painful. Can we somehow use `git submodule` to pull them in / update them when necessary? Even if not, manually copy-pasting a few times a year for security updates is not so bad.

djhoese:
As mentioned on gitter, for the record, I was able to get this to work by building the `sdist` on the "host" Linux machine that `cibuildwheel` is called from *before* calling `cibuildwheel`. Since `cibuildwheel` uses the same cloned repository from its caller, it has access to the `static/*` directory that the `sdist` processing creates. I was already creating the `sdist` in the linux environment, I just needed to move it to before the `cibuildwheel` call.

larsoner:
@djhoese just wanted to let you know that I adapted our `azure-pipelines.yml` to another repo pretty easily:

https://github.com/spatialaudio/python-rtmixer/pull/16

Thanks for figuring out all the upload bits! With this and the lack of complaints about `pip install vispy` failing after 0.6 release, I'm slowly becoming less wary of distributing compiled code :)

PR #1682, opened 2019-07-29
Fixes #1681 by keeping a MatrixTransformation that gets modified on e…

### LINKED ISSUE #1681: PanZoom camera interaction becomes unusable if the up parameter is changed from "+z"
Here is a minimal working example. Try changing the `+z` to `-z` in the line `view.camera.up = "+z"` to see the difference in performance:

```
import numpy as np
import vispy.scene

canvas = vispy.scene.SceneCanvas(keys='interactive', show=True)

epsilon = 0.01

# Make Data
n_markers = 3
p1 = np.zeros((n_markers, 3))
p2 = np.zeros((n_markers, 3))
p1[:, 0] = np.linspace(0.0, 1.0, n_markers) + 0.0
p2[:, 0] = np.linspace(0.0, 1.0, n_markers) + 0.1

# Make Visuals
m1 = visuals.Markers(parent=view.scene)
m2 = visuals.Markers(parent=view.scene)
axis = visuals.XYZAxis(parent=view.scene)

p1[:, 2] += 0 * epsilon
p2[:, 2] += 1 * epsilon

m1.order = 1
m2.order = 2

# Set Data
m1.set_data(
    p1, face_color=[1.0, 0.0, 0.0, 0.5], edge_color=[1, 0, 0, 1], size=80, edge_width=15
)
m2.set_data(
    p2, face_color=[0.0, 1.0, 0.0, 0.5], edge_color=[0, 1, 0, 1], size=80, edge_width=15
)

view.camera = 'panzoom'
view.camera.up = "+z"  # <- This fixes occlusion/ordering/transparency but breaks refresh rate
# view.camera = 'arcball'

if __name__ == '__main__':
    import sys
        vispy.app.run()
```

os-gabe:
Fixes #1681 by keeping a MatrixTransformation that gets modified on each iteration rather than overwritten

### DIFF
--- a/vispy/scene/cameras/panzoom.py
+++ b/vispy/scene/cameras/panzoom.py
@@ -49,6 +49,7 @@ class PanZoomCamera(BaseCamera):
         super(PanZoomCamera, self).__init__(**kwargs)
 
         self.transform = STTransform()
+        self.tf_mat = MatrixTransform()
 
         # Set camera attributes
         self.aspect = aspect
@@ -151,7 +152,8 @@ class PanZoomCamera(BaseCamera):
     @property
     def center(self):
         rect = self._rect
-        return 0.5 * (rect.left+rect.right), 0.5 * (rect.top+rect.bottom), 0
+        return 0.5 * (rect.left + rect.right), 0.5 * (
+            rect.top + rect.bottom), 0
 
     @center.setter
     def center(self, center):
@@ -206,7 +208,7 @@ class PanZoomCamera(BaseCamera):
 
         if event.type == 'mouse_wheel':
             center = self._scene_transform.imap(event.pos)
-            self.zoom((1 + self.zoom_factor) ** (-event.delta[1] * 30), center)
+            self.zoom((1 + self.zoom_factor)**(-event.delta[1] * 30), center)
             event.handled = True
 
         elif event.type == 'mouse_move':
@@ -223,14 +225,14 @@ class PanZoomCamera(BaseCamera):
                 p2 = np.array(event.pos)[:2]
                 p1s = self._transform.imap(p1)
                 p2s = self._transform.imap(p2)
-                self.pan(p1s-p2s)
+                self.pan(p1s - p2s)
                 event.handled = True
             elif 2 in event.buttons and not modifiers:
                 # Zoom
                 p1c = np.array(event.last_event.pos)[:2]
                 p2c = np.array(event.pos)[:2]
-                scale = ((1 + self.zoom_factor) **
-                         ((p1c-p2c) * np.array([1, -1])))
+                scale = ((1 + self.zoom_factor)**
+                         ((p1c - p2c) * np.array([1, -1])))
                 center = self._transform.imap(event.press_event.pos[:2])
                 self.zoom(scale, center)
                 event.handled = True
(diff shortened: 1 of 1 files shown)


os-gabe:
I went back and forth on leaving the if statement and decided the computation was cheap enough that it was a good trade-off for simplifying the logic by removing the if statement.

I like where you are going with only computing the up transform whenever it changes but I'm not seeing a straightforward way to do that at the moment.

vispy/scene/cameras/panzoom.py
```diff
@@ -273,36 +275,32 @@ def _update_transform(self):
...
+                   (rr.right, rr.bottom, 0), (rr.left, rr.bottom, d)]
+        elif self.up[1] == 'y':
+            pp2 = [(rr.left, 0, rr.bottom), (rr.left, 0, rr.top),
+                   (rr.right, 0, rr.bottom), (rr.left, d, rr.bottom)]
+        elif self.up[1] == 'x':
+            pp2 = [(0, rr.left, rr.bottom), (0, rr.left, rr.top),
+                   (0, rr.right, rr.bottom), (d, rr.left, rr.bottom)]
+
+        # Apply
+        self.tf_mat.reset()
```
djhoese: Is this needed? From what I can tell in the `MatrixTransform` sets `self.matrix` in `set_mapping` which then updates the shaders. This (the reset followed by the `set_mapping`) is actually causing to calls to `update` which is doing double the number of draws (it sends a "changed" event out to the vispy system).

os-gabe: Good call

djhoese:
I'm having a hard time being ok with the `if` statement being removed. Could you add it back. I can see how only creating the MatrixTransform once is a huge performance gain (it stops shaders from being recompiled every time), but the other calculations in there aren't nothing. Especially when you consider they aren't needed at all for `+z`.

I spent 30 minutes looking in to a different approach to this and I think it *could* be done, but has so many possible points of failure I'm a little worried. The `up` property isn't actually used anywhere in the base camera and neither is `self.transform`. That should mean all of the changes can be limited to the `panzoom.py` module. I originally though `self._real_rect` was the camera's rectangle oriented for the `self.up` property, but it isn't and is only used in the `_update_transform` method in question. So besides this update method the other use of the transform for the camera is `viewbox_mouse_event` and it is only used to get the current distance the mouse moved or the current pointing position of the mouse. I'm a little confused at why the mouse wheel event uses the Scene's transform but a mouse move uses the camera's transform. We'd have to make sure that the transform isn't linked via event handling to some other piece of code. Otherwise I don't see why it can't be handling this orientation stuff. I'm sure I'm missing something. Just brainstorming, don't feel like you have to do anything about this.

os-gabe:
What do you think about getting this merged @djhoese? I don't think my change is causing the failed CI test is it? I don't really understand the error there.

djhoese:
Darn, I forgot about this. I really wanted to come up with a better overall solution that didn't need the extra logic for changing `up`. I think I'm ok merging this, but am leaving for the weekend and have been kind of swamped with other projects' CIs failing. I restarted the failing job which was due to an HTTP timeout so that should pass. I'll plan on merging next week.

PR #1689, opened 2019-08-04
Mesh wireframe filter

asnt:
Tentative PR for a filter showing the mesh wireframe.
Using the fragment shader technique from [[1], [2]].
Works on top of #1463.
~~Some refinements might still be needed to blend correctly the wireframe with the mesh color or with unfilled faces.~~ Addressed in #2025.

[1]: http://codeflow.org/entries/2012/aug/02/easy-wireframe-display-with-barycentric-coordinates/
[2]: https://stackoverflow.com/a/33004265

Pros:
- control of the line width
- anti-aliased

Cons:
- additional buffer of vertex attributes with barycentric coordinates
- no independent color per edge (always shared with neighbors to some extent)
- does not work with indexed arrays

To do:
- [x] implement base filter
- [x] add example 
- [x] (addressed in #2025) check correct blending with filled mesh faces
- [x] (addressed in #2025) check correct display of unfilled faces

Some variations on this technique for future reference:
- [[3], [4]] ~~works with index buffers~~ (requires non-indexed arrays as well)
- [5] via a geometry shader

[3]: http://www2.imm.dtu.dk/pubdb/edoc/imm4884.pdf
[4]: http://www.imm.dtu.dk/~janba/Wireframe/
[5]: https://developer.download.nvidia.com/whitepapers/2007/SDK10/SolidWireframe.pdf

Demos:
1. wireframe off
image
2. width = 1.5
image
3. width = 3.5
image


vispy/visuals/filters/mesh.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def _detach(self, visual):
+        visual._get_hook('vert', 'post').remove(self.vertex_program)
+        visual._get_hook('frag', 'post').remove(self.fragment_program)
+
+
+wireframe_vertex_template = """
+varying vec3 v_bc;
+
+void preprare_wireframe() {
```
djhoese: Typo here `prepare`, is this ever used?

asnt: If `prepare_wireframe()` is named `main()` instead, it produces a linking error:
```
[...]
RuntimeError: Program linking error:
error: fragment shader varying v_base_color not written by vertex shader
.error: fragment shader varying v_eye_vec not written by vertex shader
.error: fragment shader varying v_light_vec not written by vertex shader
.error: fragment shader varying v_normal_vec not written by vertex shader
.error: fragment shader varying v_pos_scene not written by vertex shader
```
I assume the main function of the original shader gets overridden.

Is there any documentation of the shader parsing and hook system mechanism?

My observations are:
- The `prepare_wireframe()` is being called, otherwise the varying would not be initialised in the fragment shader; the shader would not produce the wireframe.
- Any other unique name seems OK, e.g. "pass_bc_coords()`.
- The first function defined in the hook is used as the entry point. For example, this works:
```
varying vec3 v_bc;

void preprare_wireframe() {
    v_bc = $bc;
}

void other_func () {
}
```
This does not:
```
varying vec3 v_bc;

void other_func () {
}

void preprare_wireframe() {
    v_bc = $bc;
}
```

djhoese: All I'm saying is that the function says `preprare` when it is trying to say `prepare` (note the extra 'r').

asnt: Right. Thanks!

djhoese:
@asnt I'm going through this code and I'm realizing that for both the shading filter and this new wireframe filter the `enabled` property doesn't actually do anything. Are you OK with me removing it?

djhoese:
Actually scratch that, I just added the functionality so it is used. I also removed the wireframe example and instead merged it into the shading example. I've added docstrings that include parameters and point to the example scripts these mesh filters are used.

djhoese:
@asnt @kmuehlbauer If you could give this a review that would be great. I did the following:

* Merged with vispy/main
* Added a basic test of the wireframe filter
* Updated the mesh_shading.py example to use the wireframe filter and removed the now unnecessary mesh_wireframe.py example
* Added docstrings to the various mesh filters
* Removed the 'enabled' kwarg/property for the shading filter as it wasn't being used and is unnecessary (shading=None)
* Added the functionality to make 'enabled' work on the wireframe filter

vispy/visuals/filters/mesh.py
```diff
@@ -347,29 +349,38 @@ def _detach(self, visual):
     vec3 d = fwidth(v_bc);
     vec3 a3 = smoothstep(vec3(0.0), $width * d, v_bc);
     float factor = min(min(a3.x, a3.y), a3.z);
-    gl_FragColor = mix($color, gl_FragColor, factor);
-//    gl_FragColor = vec4(0.0, 0.0, 0.0, 0.95 * (1.0 - factor));
+    if ($enabled == 1) {
```
asnt, line 358: We could exit early, at the top of the function, to avoid unnecessary computations:
```glsl
if ($enabled != 1) {
    return;
}
```

asnt (commented):
Added one suggestion. Looks good to me otherwise. Thank you @djhoese

vispy/visuals/filters/mesh.py
djhoese: :man_facepalming: Of course. Done.

asnt (commented):
:+1:

hausmann1:
Hello everyone,
I have a simple question regarding the wireframe filter, although I am not familiar with shading and only use Vispy for 3D visualization as part of my scientific research. 

Would it be possible to add a boolean array to the filter, containing a flag for each edge, to draw only specific edges? What I have in mind is highlighting feature edges when visualising meshes derived from stp file geometry. I've tried drawing 3D lines over standard mesh objects, but they don't blend well with the original mesh. Is there possibly already another way to achieve this?

Regardless, thanks for the great work and recent updates.

PR #1696, opened 2019-08-19
Fix read_pixels when there are multiple contexts

### LINKED ISSUE #1695: PyQt5 screenshot uses entire window
One of my projects is a PyQt5 GUI that uses VisPy to show data on a canvas along with other widgets to control what's shown. I realized today that my functionality for taking a screenshot of the current canvas using vispy's `_screenshot` is now saving the entire application window. My guess is that this has to do with how PyQt5 draws all of its components with OpenGL now (compare to PyQt4) and our call to `glGetParameter(GL_VIEWPORT)` is now returning the whole window (x, y, width, height).

I came up with a small little workaround that let's my function work by passing `viewport` to `_screenshot`:

```
        geom = self.main_canvas.native.geometry()
        ratio = self.main_canvas.native.devicePixelRatio()
        geom = (geom.x(), geom.y(), geom.width(), geom.height())
        new_pos = self.main_canvas.native.mapTo(self.parent(), QPoint(0, 0))
        new_y = self.parent().height() - (new_pos.y() + geom[3])
        viewport = (new_pos.x() * ratio, new_y * ratio, geom[2] * ratio, geom[3] * ratio)
```

Where `self.parent()` is the main widget and `self.main_canvas` is the Vispy canvas.

I'm wondering if there is a way to turn this off with PyQt5. This is interesting considering that we (vispy) default to not sharing contexts in Qt5 applications.

djhoese:
Closes #1695.

I discovered that the behavior of the semi-hidden `_screenshot` utility function has unexpected results when used as part of a PyQt5 application. Best I can understand is that vispy and pyqt5 are both using OpenGL to draw things to the screen. At any given time the Qt5 GL context could be active. If you ask for any properties of the "current context" when then you may be accidentally getting that information from the Qt context instead of the vispy context. I think that's what was happening with `_screenshot` and `read_pixels`. This PR adds a "set_current" operation to `read_pixels` so the vispy canvas is current when we then ask for the current viewport (`gl.glGetParameter(gl.GL_VIEWPORT)`).

I could see another option being to get the viewport from the actual canvas (which keeps that information cached) and only use the context for checking if it is remote.

@rougier @larsoner Thoughts?

### DIFF
--- a/vispy/gloo/util.py
+++ b/vispy/gloo/util.py
@@ -5,7 +5,7 @@
 # -----------------------------------------------------------------------------
 
 from ..ext.six import string_types
-from .wrappers import read_pixels
+from .wrappers import read_pixels, get_current_canvas
 
 
 def _screenshot(viewport=None, alpha=True):
@@ -27,6 +27,8 @@ def _screenshot(viewport=None, alpha=True):
         3D array of pixels in np.uint8 format
     """
     # gl.glReadBuffer(gl.GL_BACK)  Not avaliable in ES 2.0
+    # Need to set current canvas so context is current
+    get_current_canvas().set_current()
     return read_pixels(viewport, alpha)
 
 
--- a/vispy/gloo/wrappers.py
+++ b/vispy/gloo/wrappers.py
@@ -638,10 +638,7 @@ def read_pixels(viewport=None, alpha=True, mode='color', out_type='unsigned_byte
     _check_valid('mode', mode, ['color', 'depth', 'stencil'])
 
     # Check whether the GL context is direct or remote
-    canvas = get_current_canvas()
-    context = canvas.context
-    # Need to set current canvas so context is current
-    canvas.set_current()
+    context = get_current_canvas().context
     if context.shared.parser.is_remote():
         raise RuntimeError('Cannot use read_pixels() with remote GLIR parser')
 
--- a/vispy/gloo/wrappers.py
+++ b/vispy/gloo/wrappers.py
@@ -638,7 +638,10 @@ def read_pixels(viewport=None, alpha=True, mode='color', out_type='unsigned_byte
     _check_valid('mode', mode, ['color', 'depth', 'stencil'])
 
     # Check whether the GL context is direct or remote
-    context = get_current_canvas().context
+    canvas = get_current_canvas()
+    context = canvas.context
+    # Need to set current canvas so context is current
+    canvas.set_current()
     if context.shared.parser.is_remote():
         raise RuntimeError('Cannot use read_pixels() with remote GLIR parser')
 


rougier:
Your fix seems ok to me.

djhoese:
Do *NOT* merge this! Early results seem to show that it stops PanZoom from working. How? I don't know.

djhoese:
So I think I fixed this, we'll see what the tests think. The issue was that rendered picking (what is used by mouse events) uses `read_pixels` to figure out where we are clicking on the canvas. This functionality already called `set_current` and then `read_pixels`. For some reason calling `set_current` twice called the picking to fail. I didn't spend enough time figuring out what exactly was failing, but the buffer read was always returning the background color (black).

@rougier Any guesses as to why setting the current context twice would cause this kind of behavior? The series of events is:

1. Mouse event
2. Set current context
3. Create FrameBuffer and activate it
4. Draw the scene
5. Read the buffer (via `read_pixels`)
5a. Set current context again
5b. `finish()` to flush GLIR
5c. Get viewport size, if needed, to determine what part of the buffer to read
5d. Call `glReadPixels`

rougier:
No real idea. Do you known where's the code for `set_current`?

djhoese:
The `set_current` function for the Canvas is here:

https://github.com/vispy/vispy/blob/09bd8bee79d30e35a69a6f4959bf5b3e762e06b5/vispy/app/canvas.py#L398-L407

It calls this method on the Qt backend:

https://github.com/vispy/vispy/blob/09bd8bee79d30e35a69a6f4959bf5b3e762e06b5/vispy/app/backends/_qt.py#L773-L777

It seems like the tests related to having multiple canvases failed. I still need to investigate if that is a logical error in the test or if I actually broke something.

djhoese:
My newest guess is that running `setCurrent` on the Qt Context is auto-swapping the buffer bringing an empty buffer "to the front". Not sure if that makes sense.

djhoese:
@rougier I'm starting to think my usage in my own application is just wrong or at least is undefined as far as behavior. I have a vispy canvas inside a PyQt5 application and I click a button for a screenshot of the canvas. If you (the user) don't ask for multiple frames then I just call `_screenshot` immediately. This assumes that the last GL state/context was for the vispy canvas which for Qt5 is not guaranteed since it seems to use OpenGL for its own drawing.

So...I now have an alternative fix which is to force a redraw of the vispy canvas and then take a screenshot. This seems like a reasonable and logical solution. What do you think? Other ideas?

rougier:
There is actually a front and back buffer in OpenGL and there shoudl be a swapbuffer command somewhere. Have  a look at https://stackoverflow.com/questions/6543924/is-double-buffering-needed-any-more

djhoese:
Yes, I'm a little familiar with the idea. Either some part of `set_current` is telling GL to automatically swap or clear the buffer or when `setCurrent` is run on the QOpenGLWidget it is doing one of these. If I put the `set_current` anywhere after the last draw then the buffer is cleared. PyQt5 seems to assume that you are setting the current context because you want to start drawing a new scene/buffer.

I think, based on what I've seen this could be closed, but it does mean users have to be more careful about when they call `_screenshot` on a PyQt5 application because of the "implicit" Qt5 GL context.

djhoese:
Closing this as it doesn't seem like the right solution. If someone has a better way of guaranteeing the correct behavior, let me know. Otherwise, users seeing this "bad" behavior should make sure that their vispy canvas is the last thing to be drawn before getting a screenshot. This may require redrawing the vispy scene, but is likely the same solution we would implement inside vispy if we were going to fix this inside vispy.

PR #1702, opened 2019-09-04
Fix MarkerVisual scaling when rotating in 3D space

### LINKED ISSUE #1166: Scaling not working correctly for markers
When I set `scaling=True` when setting the data in `MarkersVisual#set_data()`, only rotating around the scene (no zooming) makes the points' sizes to vary. I hope this is not the expected behaviour.

Here is a snippet which demonstrates this. Run it and rotate the camera around.

``` python
import sys
import numpy as np
from vispy import scene

# create data cloud
pos = np.random.normal(size=(1000, 3), scale=0.2)


#
# Make a canvas and add simple view
#
canvas = scene.SceneCanvas(keys='interactive', show=True, bgcolor='w')

view.camera = 'turntable'
view.camera.fov = 60

scatter = visuals.Markers(parent=view.scene)
scatter.antialias = 0
scatter.set_data(pos=pos, edge_color=(0.0, 0.0, 0.0, 1.0), face_color=(0.6, 0.5, 0.4, 1.0), size=.03, scaling=True)

scatter.set_gl_state(depth_test=True, blend=True, blend_func=('src_alpha', 'one_minus_src_alpha'))

# Add axes
axis = visuals.XYZAxis(parent=view.scene)

    canvas.app.run()
```

sofroniewn:
This PR fixes #1166 where the scale was not working properly for markers when they were being rotated.

Please see that issue for a detailed discussion of the problem and proposed solution, including example code to reproduce the bug and gifs of the behavior before and after the fix.

### DIFF
--- a/vispy/visuals/markers.py
+++ b/vispy/visuals/markers.py
@@ -619,9 +619,9 @@ class MarkersVisual(Visual):
             return False
         view.view_program['u_px_scale'] = view.transforms.pixel_scale
         if self.scaling:
-            tr = view.transforms.get_transform('visual', 'document')
+            tr = view.transforms.get_transform('visual', 'document').simplified
             mat = tr.map(np.eye(3)) - tr.map(np.zeros((3, 3)))
-            scale = np.linalg.norm(mat)
+            scale = np.linalg.norm(mat[:, :3])
             view.view_program['u_scale'] = scale
         else:
             view.view_program['u_scale'] = 1
--- a/vispy/visuals/markers.py
+++ b/vispy/visuals/markers.py
@@ -619,8 +619,9 @@ class MarkersVisual(Visual):
             return False
         view.view_program['u_px_scale'] = view.transforms.pixel_scale
         if self.scaling:
-            tr = view.transforms.get_transform('visual', 'document').simplified
-            scale = np.linalg.norm((tr.map([1, 0]) - tr.map([0, 0]))[:2])
+            tr = view.transforms.get_transform('visual', 'document')
+            mat = tr.map(np.eye(3)) - tr.map(np.zeros((3, 3)))
+            scale = np.linalg.norm(mat)
             view.view_program['u_scale'] = scale
         else:
             view.view_program['u_scale'] = 1


djhoese:
Based on our discussion I think it would be ok to do `:3` for the points; drop the `w` dimension and keep the `xyz`. You can also go back to using `simplified`. I reviewed the code and everything should be cached so this shouldn't be a performance issue and may benefit things depending on how the transforms are simplified.

Thanks for putting this together and figuring this out.

sofroniewn:
Sounds good, I have made the two requested changes and confirmed performance still works as expected. This is good to go from my end

djhoese:
Thank you! Sorry this got lost in my inbox. Looks good.

sofroniewn:
Thanks for merging - unfortunately, the absolute scaling now seems off. Has something else merged in between `0.6.1` and `0.6.2` that might cause this?

If I have my fix ontop of `0.6.1` it looks like this:

Screen Shot 2019-11-11 at 11 59 36 AM

If I have it ontop of `0.6.2` it looks like this:
Screen Shot 2019-11-11 at 11 56 50 AM
Screen Shot 2019-11-11 at 11 59 04 AM

Note in both cases the rotational invariance is fixed. The behaviour with `0.6.1` is "correct" in the sense that the marker sizes don't change when switching from 2D to 3D mode.

I can look into this now, but wanted to flag before your next release

djhoese:
Hm, what cameras are you using in your tests? The release notes are [here](https://github.com/vispy/vispy/blob/master/CHANGELOG.md#v062-2019-11-04) and should include every PR that was merged between 0.6.1 and 0.6.2. There are some changes to some of the cameras but they were relatively small from what I remember. I could see if this PR was dependent on a transform changing and now that transform isn't (as an optimization) then I could see this being broken.

sofroniewn:
To do my tests I have a port of this PR over in some napari code (see below), and i can then test against the `pip install vispy==0.6.1` and `pip install vispy==0.6.2`. Thanks for pointing to the release notes, I can quickly checked against each PR I think and let you know which one it is

```python
from vispy.scene.visuals import Markers as BaseMarkers
import numpy as np


class Markers(BaseMarkers):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    def _prepare_draw(self, view):
        if self._symbol is None:
            return False
        view.view_program['u_px_scale'] = view.transforms.pixel_scale
        if self.scaling:
            tr = view.transforms.get_transform('visual', 'document').simplified
            mat = tr.map(np.eye(3)) - tr.map(np.zeros((3, 3)))
            scale = np.linalg.norm(mat[:, :3])
            view.view_program['u_scale'] = scale
        else:
            view.view_program['u_scale'] = 1
```

sofroniewn:
It looks like it happens in https://github.com/vispy/vispy/pull/1724 - could be the FOV change?

djhoese:
Just to be clear, it works *before* that PR and fails after?

sofroniewn:
It fails after that PR is merged - setting `fov=0` in the constructor of `ArcballCamera` returns the behavior to be what I expected. The default value changed to `fov=45` in that PR

djhoese:
To me this makes it seem like the fix in this PR doesn't actually fix the issue. The `fov` default changing was only because `fov=0` wasn't behaving properly. The blocks of code that were removed in that PR were only related to `fov=0`. If `fov=45` is breaking things then I would expect your tests to fail *before* #1724 if `fov=45` too.

If this is true, then perhaps `set_perspective` (the transform method being called now at all times) is not triggering events as it should?

sofroniewn:
Yes, that is correct - switching to `fov=45` causes the markers to look too big even on `0.6.1`, I just hadn't been catching it as I was only testing with the old default. It also seems like on `0.6.1` with `fov=45` the marker sizes are also invariant to rotation(!) making this PR maybe unnecessary - though this PR doesn't change anything in that case either, the behavior seems the same.

I'm just looking over #1723 which makes a good case for using `fov=45` as a default.

I guess the only question remaining now is why the markers are so big with `fov=45` (which seems to have always been True), and that could be because of a problem with `set_perspective`, but I am not sure



, the markers just

djhoese:
So looking at the `set_perspective` code and it looks like it sets `self.matrix` on the transform like all of the other methods for changing the transform. I don't see why this would make a difference. So the next question to narrow this down is: do the markers stay the proper relative size to the other markers as you rotate around? What if you resize the window?

Your code here depends on `pixel_scale` from the TransformSystem. Maybe the calculation done in `pixel_scale` isn't being normalized properly or rather the matrix being set in `set_perspective` isn't normalized.

```
    @property
    def pixel_scale(self):
        tr = self._canvas_transform
        return (tr.map((1, 0)) - tr.map((0, 0)))[0]
```

PR #1703, opened 2019-09-05
Fix MarkersVisual.set_data crash when pos is None

proto3:
Replace 'assert' by an 'if' when pos is None. This prevent constructors depending on markers to crash with None data -> For example LinePlotVisual was crashing with data=None.


proto3:
I kept the assert for when pos is not None, otherwise it will silently skip data assignment.

PR #1705, opened 2019-09-08
Add lines_adjacency and line_strip_adjacency OpenGL primitives.

proto3:
I added lines_adjacency and line_strip_adjacency OpenGL primitives, so I can emulate quads with geometry shader.
[https://github.com/proto3/vispy/blob/quad-gif/quad.gif](https://github.com/proto3/vispy/blob/quad-gif/quad.gif)

It worked immediately with GL+ backend, but it won't with GL2 so it raise an explicit error advising to use GL+ with these primitives.
In the mean time, I updated the geometry shader support error message, so it won't blame 'GL2' all the time but gl.current_backend.__name__

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -182,8 +182,9 @@ DRAW
 Applies to: Program
 
 This command is used to draw the program. It has a ``mode`` argument
-which can be 'points', 'lines', 'line_strip', 'line_loop', 'triangles',
-'triangle_strip', or 'triangle_fan' (case insensitive).
+which can be 'points', 'lines', 'line_strip', 'line_loop', 'lines_adjacency',
+'line_strip_adjacency', 'triangles', 'triangle_strip', or 'triangle_fan'
+(case insensitive).
 
 If the ``selection`` argument has two elements, it contains two integers
 ``(start, count)``. If it has three elements, it contains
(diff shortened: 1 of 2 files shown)


vispy/gloo/glir.py
```diff
@@ -1320,7 +1321,16 @@ def draw(self, mode, selection):
...
-        mode = as_enum(mode)
+        try:
+            mode = as_enum(mode)
+        except ValueError as err:
+            if mode == 'lines_adjacency' or mode == 'line_strip_adjacency':
+                raise RuntimeError(gl.current_backend.__name__ + " backend does"
+                " not support lines_adjacency and line_strip_adjacency"
+                " primitives. Try gloo.gl.use_gl('gl+').")
+            else:
+                raise err
```
djhoese: Could you remove the `as err` part and change this to `raise`, that way the original traceback is preserved.

djhoese: Actually you can remove the `else:` block and just do `raise`:

```python
except ValueError:
    if mode ...:
        raise RuntimeError(...)
    raise
```

djhoese:
I'm finally getting back to this. Some of the tests are failing because of some styling issues:

```
vispy/gloo/glir.py:1004:13: E128 continuation line under-indented for visual indent
vispy/gloo/glir.py:1329:17: E128 continuation line under-indented for visual indent
vispy/gloo/glir.py:1330:17: E128 continuation line under-indented for visual indent
```

Do you think you could take a look and fix these? You maybe have to merge or rebase with master to make the tests pass (given some conda-forge changes).

proto3:
I cannot figure out why example/basics/plotting/colorbar.py test fails :confused:, I see no relation between this and my modifications. Someone ?

djhoese:
Thanks for the updates. The test failure is a random failure caused by the old version of the cassowary library we depend on (and bundle with vispy). We have a pull request half finished to replace cassowary but it isn't done yet. Restarting the test for now usually fixes it. I've restarted it now so we'll see.

djhoese:
And now it failed because of an HTTP download hang up. Restarting again...

djhoese:
Oops, forgot about this. Thanks again!

PR #1706, opened 2019-09-09
Switch to setuptools_scm for automatic version numbering

djhoese:
This updates vispy's version number handling to be automatic based on git tags instead of having to manually edit the value in `vispy/__init__.py`. The one major change to this is that `setup.py` no longer updates the `__doc__` attribute of `setup.py`, but I don't see the point of this anyway. This is even less useful now that `pip install` (and variants) is the preferred method of installing python packages instead of `python setup.py ...`.

More info on setuptools_scm: https://pypi.org/project/setuptools-scm/
Requires https://github.com/vispy/vispy-website/pull/26

Note: In previous non-github conversations I had mentioned "caching" the version number in something like a `vispy/version.py` file. I was going to try this with my Satpy library but one of the other maintainers didn't like it. I'm doing the same thing here that I did with Satpy which is asking setuptools_scm to get the version from git every time it is imported. This has a slight performance penalty but nothing noticeable in my opinion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ vispy.egg-info
 *.swp
 .ipynb_checkpoints
 .cache/
+.eggs/
 examples/basics/scene/animation.gif
 .pytest_cache/
 js/node_modules
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,9 @@
 [build-system]
 requires = [
     "wheel",
-    "setuptools",
+    "setuptools>=30.3.0",
+    "setuptools_scm",
+    "setuptools_scm_git_archive",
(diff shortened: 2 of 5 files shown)


larsoner:
Looks reasonable to me. Takes import time from ~100ms to ~200ms on my system, which is not awesome but probably okay for the benefits. CIs are not happy, though.

djhoese:
So the failures are because some of the import tests make sure that nothing else is accidentally imported, but now `pkg_resources` is imported so that needs to be accounted for. However, the tests are only failing on python 2.7, so I need to look at how/why the importing is different in python 2.7.

larsoner:
Neat, thanks

PR #1722, opened 2019-10-13
[WIP] BayerImageVisual

### LINKED ISSUE #1721: Raw Image Shader
Would there be any interest in a raw image shader????

I managed to port https://casual-effects.com/research/McGuire2009Bayer/index.html#shaders

to the `ImageVisual`.

THat said, I really implemented the bare functionality, and didn't take good care of getting feature parity.

Mostly, I don't understand opengl too much, and would need help implementing interpolation and understanding what `method` means.

hmaarrfk:
Still a WIP, running on CIs to test code.
Fixes #1721

I think this is good enough for a first pass.

I would rather better understand:

- [ ] impostor vs subdivide
- [ ] interpolation (maybe just remove it in this case????)
- [ ] build texture

Before cleaning up the docs.


hmaarrfk:
For some reason, the tests fail with segfaults on my local computer when trying to run them automatically on my computer.

djhoese:
Any interest/time in bringing this PR back to life @hmaarrfk? Maybe merge with master to test/fix the tests?

hmaarrfk:
i can try, but i may ask for your help trying to find a bug. When I run this for long enough, my whole python system crashes. :/ weird variables start to disappear, and strange things continue to happen. Not too sure if it is due to this, or just general Qt/OpenGL

hmaarrfk:
Here is my attempt at making a minimal reconstruction of my bug, but it doesn't seem to be causing the intended effect. So this PR is likely ready to go if it passes tests.

The bug i'm seeing is likely due to something else.
[qt_bug_try.zip](https://github.com/vispy/vispy/files/3936525/qt_bug_try.zip)

hmaarrfk:
Do you have some standard images i can use to load to test this automatically? I'm trying to use imageio, but it seems it isn't a dependency.

hmaarrfk:
Do you also know why these lines delete my window decorations

```
        self.vispy_scene = SceneCanvas(
            size=size, keys='interactive')
        self.vispy_scene.bgcolor = '#e2e2e2'

        self.vispy_view = self.vispy_scene.central_widget.add_view()
        native = self.vispy_scene.native
        self.canvas.setLayout(QVBoxLayout())
        self.canvas.layout().addWidget(native)

hmaarrfk:
Without (has window decorations):
image

with (missing window decorations):
image

djhoese:
It depends what `self.canvas` is. Setting a whole new layout on a widget may be removing something that was on the old layout in a weird way. I also think I always keep an external handle on a layout object even though I'm adding/setting it on a widget.

hmaarrfk:
> I also think I always keep an external handle on a layout object even though I'm adding/setting it on a widget.

Good try, but doesn't help.

The code is mostly the same as the one in the zip. Not sure if you'll have time or patience to look at it.

hmaarrfk:
Ok. I'm convinced my earlier error was not due to my Bayer code. I can probably finish pushing this through with confidence when I get more time

hmaarrfk:
I guess I should debase. I was having a hard time getting the rest correct. Subsequently, we've found it was hard to get predictable behavior on all platforms.

We've seen differences in results based on different OpenGL versions

djhoese:
Very interesting. Yeah a rebase or merge would be smart if you want to continue this work.

Issue #1723, opened 2019-10-16
Zero FOV leads to poor depth buffer precision


h3fang:
For some scene cameras, the default `fov=0.0` causes `PerspectiveCamera` to use orthogonal projection with `zNear=0.0`, which leads to poor depth buffer precision.

The problematic `PerspectiveCamera`

https://github.com/vispy/vispy/blob/8b61cd439076aa3f50ac5f6dacb4c0af8c1d0684/vispy/scene/cameras/perspective.py#L169-L175

Cameras with default `fov=0.0`

https://github.com/vispy/vispy/blob/8b61cd439076aa3f50ac5f6dacb4c0af8c1d0684/vispy/scene/cameras/turntable.py#L56-L58

https://github.com/vispy/vispy/blob/8b61cd439076aa3f50ac5f6dacb4c0af8c1d0684/vispy/scene/cameras/arcball.py#L48-L49

Related issue: #1716

Detailed explanation:
https://www.khronos.org/opengl/wiki/Depth_Buffer_Precision

Suggested fixes:
- Set the camera default `fov` to a non zero value, e.g. `fov=45.0`.
- Remove the `fov == 0` `if` case in `PerspectiveCamera`, since https://github.com/vispy/vispy/blob/8b61cd439076aa3f50ac5f6dacb4c0af8c1d0684/vispy/scene/cameras/perspective.py#L172
already takes care of it.

djhoese:
Thanks for this issue. Well written too. I think everything you've said makes sense. Would you be able to make a pull request for this yourself? Otherwise I can add the "hacktoberfest" label and see if someone else wants to pick it up.

djhoese:
As far as I can tell this was fixed with #1724 and maybe #1702 a little bit. Someone let me know if this is still an issue and we'll reopen.

PR #1724, opened 2019-10-17
Fix depth buffer precision

h3fang:
For #1723


djhoese:
Thanks for this. The linux builds on azure failed for an odd reason (an attribute error internal to python typing) so I'm restarting in hopes that it fixes itself.

djhoese:
According to this the `typing` package may be getting installed when we should only be using python's built in typing module. Will have to look at this later.

PR #1725, opened 2019-10-20
Fix Azure Linux builds

djhoese:
Noticed in #1724 our tests on the linux container on Azure is failing. It fails with an AttributeError related to the `typing` module. Looking at the paths involved it seems `site-packages/typing.py` is being used instead of the built in typing in CPython.

All of the environments are generally the same except for Linux include installing extra dependencies so we can build the `sdist` tarball of vispy (iirc it uses the extension .so's built from the wheel generation). The easiest logical leap is that these extra dependencies are for some reason installing the third-party typing package. Otherwise, it is an issue with the cibuildwheel linux which could also be because none of these dependencies have been recently updated.

This PR is me trying to figure this out...yay.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,9 +11,9 @@ variables:
   CIBW_TEST_REQUIRES: "pytest pytest-sugar nose"
   CIBW_TEST_COMMAND: "python -c \"import vispy; vispy.test()\""
   CIBW_BUILD_VERBOSITY: "2"
-  CIBW_BEFORE_BUILD: "pip install -U numpy Cython jupyterlab ipywidgets; pip uninstall -y typing"
+  CIBW_BEFORE_BUILD: "pip install -U numpy Cython jupyterlab ipywidgets"
   CIBW_BEFORE_BUILD_MACOS: "npm install npm@latest -g; pip install -U pip setuptools"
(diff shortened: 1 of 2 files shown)


djhoese:
@larsoner You mentioned you adopted these azure recipes for other projects. Are those projects failing in a similar fashion?

Other things: numpy has been updated recently and I've seen issues with numpy 1.17 being installed in python 2.7 environments in some cases. However, we seem to be failing in python 3.6.

larsoner:
@djhoese no I haven't had a problem thus far since those repos, presumably because they do not depend on `jupyterlab` or fontconfig

djhoese:
@larsoner Do they use pytest?

larsoner:
Yes. But they are also not super active so it's possible there has not been a PR that would have triggered the failure in the last week or two

djhoese:
So I think something is going on in the import order of things. Importing `from typing import Optional` works fine outside of vispy's tests.

djhoese:
As discussed in the linked to issues, the main issue here seems to revolve around the `typing` *package* having issues when you use it in a process started with `subprocess`. This is how we start all of our backend-based tests so we hit this case. The suggestion by the `typing` package developers in an old issue was to uninstall `typiing` from environments that don't need it; just like our Python 3.7 environment.

I made an issue and a PR for the manylinux repository [here](https://github.com/pypa/manylinux/pull/378) and it was merged today. The docker images have been built and I've restarted the failing PR mentioned above. So...this should be fixed upstream.

larsoner:
Yikes

djhoese:
So looks like I'll still have to wait. cibuildwheel doesn't use the new version of the manylinux2010 images (which is what has updated already). It still uses manylinux1 (which is semi-deprecated). I've asked the manylinux people if they could update the image. Otherwise we wait for the cibuildwheel newest release. We could also point cibuildwheel at the new wheel manually with an env variable.

PR #1728, opened 2019-10-22
Fix VolumeVisual modifying user provided data in-place

### LINKED ISSUE #1727: VolumeVisual.set_data method and copy=False may lead to unexpected behavior
An issue recently arose in the napari repo (see napari/napari#575) in which calling the `VolumeVisual.set_data()` method was altering the input data and causing unexpected results (but only with float32 input data).  We eventually realized it was caused by line 477 below in the `set_data()` method, in which the data is converted to float32 with `copy=False`.

https://github.com/vispy/vispy/blob/8b61cd439076aa3f50ac5f6dacb4c0af8c1d0684/vispy/visuals/volume.py#L477-L483

I realize we can simply copy the data on our end before calling the `VolumeVisual.set_data()` method, but I'm wondering whether the default behavior shouldn't be to copy the volume before applying the clim normalizations, so as to avoid unexpectedly overwriting the input data)?  (it was particularly hard to find due to the fact that the memory address of the input volume is unchanged _only_ for float32 data)

tlambert03:
This PR copies data by default in `VolumeVisual.set_data`, and adds a copy parameter to the method for those who want to maintain the existing behavior (wherein float32 data is not copied).
see #1727 for full discussion.

### DIFF
--- a/vispy/visuals/tests/test_volume.py
+++ b/vispy/visuals/tests/test_volume.py
@@ -57,4 +57,25 @@ def test_volume_draw():
         assert_image_approved(c.render(), 'visuals/volume.png')
 
 
+@requires_pyopengl()
+def test_set_data_does_not_change_input():
+    # Create volume
+    V = scene.visuals.Volume(np.zeros((20, 20, 20)))
+
+    # calling Volume.set_data() should NOT alter the values of the input array
+    # regardless of data type
+    vol = np.random.randint(0, 200, (20, 20, 20))
+    for dtype in ['uint8', 'int16', 'uint16', 'float32', 'float64']:
+        vol_copy = np.array(vol, dtype=dtype, copy=True)
+        # setting clim so that normalization would otherwise change the data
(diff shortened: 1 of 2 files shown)


djhoese:
Thanks for this. Looks like the CIs don't like your docstring.

tlambert03:
whoops.  forgot to add the new parameter to the docstring... sorry.  let's see if that fixes it.

djhoese:
Are there any existing tests where you can add a check for this?

tlambert03:
I'll try to look through your tests later today to see where it might fit best.
Thanks!

tlambert03:
OK added a test to confirm data integrity for a variety of dtypes, and also one that makes sure calling `set_data(..., copy=False)` actually _does_ change the data if it's `float32` (indicating that no copy operation was performed).  Hope that will help!

tlambert03:
don't understand why the colorbar example test is failing... don't see any connections to the changes in this PR.  closing and reopening to trigger a rebuild...

Issue #1742, opened 2019-11-04
axis labels problem


pcpatfire:
I was using the plot.py example and I got the following problems:
1. the xaxis labels are overlapped to the x axis
2. the xaxis labels values do not correspond to the actual x axis domain; they are adjusted to the correct values after resizing the window.
3. some other problems with yaxis labels and title (see attached picture)

![plot_py_screenshot](https://user-images.githubusercontent.com/50824984/68126482-34191a80-ff14-11e9-9690-9885c0ddc735.JPG)

djhoese:
Some of this seems like the default font size is too large. We've seen the ticks not updating issue before, but I thought we fixed it. What backend are you using?

djhoese:
Here is what it looks like with the current master branch on my Ubuntu system with PyQt5:

![image](https://user-images.githubusercontent.com/1828519/68128912-3ceef980-fede-11e9-9442-3f34bbb2685c.png)

pcpatfire:
I am using PyQt5 on Windows 10. Yes there is also a font size problem. It would be better if the axis_font_size and the tick_font_size could be set from the user application. 
Your plot is better than mine, but, as you can see, the x axis labels are wrong (they should be from 0 to 10) and they touch the tick marks.

djhoese:
You're right about the ticks being wrong. I guess this is still an issue.

The tick labels touching the tick marks may be something that has always existed and was never completely made to look "perfect". @larsoner may remember more.

pcpatfire:
As I noticed that if you move the plot the xaxis is updated with the correct tick labels, I was trying to force an xaxis update setting _need_update=True and calling the update() method, but it didn't work. 

```
fig[0,0].xaxis.axis._need_update = True
fig[0,0].xaxis.axis.update()
```

djhoese:
If I remember correctly, the last time this came up it was caused by a race condition where the ticks would be updated to the correct values and then immediately set back to the incorrect (initial) values. I'm sorry I don't have time to look in to it more right now.

pcpatfire:
Thank you for the link to the bug fixes related to AxisVisual. They seem to concern some other problems.
In my Qt application I found that the x axis domain is correct until the show() method is called. After, it gets wrong. If I update the plot after the show() method is executed, the x axis domain is correct.
Something happens while executing the show() method. 
A strange thing is that only the x axis has this kind of problems.

djhoese:
[...]

Issue #1198, opened 2016-04-01
Issue with get_transform when camera fov!=0


astrofrog:
I'm trying to convert the coordinates of markers from the visual coordinates to canvas coordinates. If I do this with a camera with fov=0, everything works fine:

```python
import numpy as np
from vispy import scene, app
from vispy.scene import visuals

canvas = scene.SceneCanvas(keys='interactive', show=True, size=(400,400))
view = canvas.central_widget.add_view()


pos = np.random.uniform(-0.2, 0.2, size=(100000, 3))

scatter = visuals.Markers()
scatter.set_data(pos, edge_color=None, face_color=(1, 1, 1, .5), size=5)
view.add(scatter)

view.camera = scene.cameras.TurntableCamera(parent=view.scene,
                                                 distance=2.0, fov=0)

tr = scatter.get_transform(map_from='visual', map_to='canvas')

trpos = tr.map(pos)

print(trpos.min(axis=0)[:2])
print(trpos.max(axis=0)[:2])

app.run()
```

gives

```
[ 63.3475892   45.53460351]
[ 337.69183274  351.44768515]
```

i.e. all the points are inside the visual, which matches what can be seen:

<img width="514" alt="screen shot 2016-04-01 at 3 45 37 pm" src="https://cloud.githubusercontent.com/assets/314716/14218215/cefec2c4-f820-11e5-88ce-2ca0023aefcf.png">

However, if I now run the same example as above but with ``fov=60``, what I see is:

<img width="514" alt="screen shot 2016-04-01 at 3 45 18 pm" src="https://cloud.githubusercontent.com/assets/314716/14218204/c0611cbc-f820-11e5-8d66-4e7309e61bd7.png">

but the values I get back should not be inside the canvas (which has size 400x400):

```
[ 274.89976609  320.00162177]
[ 523.80557382  479.99838575]
```

Is this a bug in ``get_transform``, or am I mis-using this function? If the latter, what is the correct way to get the final canvas coordinates of the markers?

campagnola:
I think these coordinates are actually correct, they're just represented in homogeneous coordinates. You can normalize them by dividing by the 4th vector element:

    trpos = tr.map(pos)
    trpos /= trpos[:,3:]

It's not intuitive, but this is the coordinate system OpenGL lives in. I'm not sure there is any way to make this easier..

PR #1743, opened 2019-11-05
Bug fixes related to AxisVisual

astrofrog:
This fixes a few bugs I ran into while using ``AxisVisual``:

* Changing the ``.rotation`` on a ``TextVisual`` didn't previously update the actual shader attribute for rotation. This became apparent when rotating the axes in 3-d, as the labels didn't follow the rotation.
* The calculation of the angle for the axis label didn't quite work when projection is present
* Warnings were emitted when the min/max for axes were the same

It's not clear to me how to properly test the first two but I think I could add a test for the last one if needed.

### DIFF
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -171,7 +171,9 @@ class AxisVisual(CompoundVisual):
             # TODO: make sure we only call get_transform if the transform for
             # the line is updated
             tr = self._line.get_transform(map_from='visual', map_to='canvas')
-            x1, y1, x2, y2 = tr.map(self.pos)[:, :2].ravel()
+            trpos = tr.map(self.pos)
+            trpos /= trpos[:, 3:]
+            x1, y1, x2, y2 = trpos[:, :2].ravel()
             if x1 > x2:
                 x1, y1, x2, y2 = x2, y2, x1, y1
             self._axis_label.rotation = math.degrees(math.atan2(y2-y1, x2-x1))
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -505,6 +505,10 @@ def _get_ticks_talbot(dmin, dmax, n_inches, density=1.):
     # the density function converts this back to a density in data units
     # (not inches)
     n_inches = max(n_inches, 2.0)  # Set minimum otherwise code can crash :(
+
+    if dmin == dmax:
+        return np.array([dmin, dmax])
+
     m = density * n_inches + 1.0
     only_inside = False  # we cull values outside ourselves
     Q = [1, 5, 2, 2.5, 4, 3]
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -168,17 +168,25 @@ class AxisVisual(CompoundVisual):
         if self._pos is None:
             return False
         if self.axis_label is not None:
-            # TODO: make sure we only call get_transform if the transform for
-            # the line is updated
-            tr = self._line.get_transform(map_from='visual', map_to='canvas')
-            trpos = tr.map(self.pos)
-            trpos /= trpos[:, 3:]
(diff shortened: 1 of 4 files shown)


djhoese:
Thanks @astrofrog. Any idea if these might help with the tick labels being incorrect (see #1742)? CC @pcpatfire

astrofrog:
@djhoese - I don't **think** so. I've just added a regression test for the third bug.

djhoese:
At least one test environment is failing from linting issues. Maybe we need to start using black and/or a bot like stickler to do this on the github side instead of waiting for test environments.

vispy/visuals/axis.py
```diff
@@ -171,7 +171,9 @@ def _prepare_draw(self, view):
             # TODO: make sure we only call get_transform if the transform for
             # the line is updated
             tr = self._line.get_transform(map_from='visual', map_to='canvas')
-            x1, y1, x2, y2 = tr.map(self.pos)[:, :2].ravel()
+            trpos = tr.map(self.pos)
+            trpos /= trpos[:, 3:]
```
djhoese: What is this line actually doing?

vispy/visuals/tests/test_axis.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
```
djhoese: Could you copy the header from the other modules and add it here?

vispy/visuals/axis.py
astrofrog: This is the fix that is required for the resulting display coordinates to be correct in the presence of non-zero FOV in the turntable camera. I wrote this a few years ago so I don't remember how exactly it works 🤔

vispy/visuals/tests/test_axis.py
astrofrog: Done

vispy/visuals/axis.py
djhoese: I'm trying to figure out what the `3:` is for. If `trpos` is (N, 4)` then I suppose this is dividing `x, y, z` coordinates by a `w` coordinate? I feel like I'm missing something obvious.

astrofrog: Yes, we're basically renormalizing the coordinates based on this last fourth coordinate - I could have written it ``trpos[:, 3]`` instead of ``3:``. I can't remember the details of this but when there are projection effects the last ``w`` coordinate ends up being != 1.

astrofrog: @djhoese - ah I found this old issue that discusses it! https://github.com/vispy/vispy/issues/1198#issuecomment-212790199

djhoese: Ah, so the answer is math. Got it. (ok, fine, homogeneous coordinates)

djhoese:
I think I'm willing to pass on tests for the first two bullet points. We don't currently have tests (that I know of) that do complex things like changing the rotation of a TextVisual and comparing the results. We could have an expected image that we compare against, but I'm not a huge fan of static image tests. You could mock the update call (using the mock `wraps` keyword argument to keep the original functionality of the call) and check that it is called at least once when the rotation is changed. Thoughts?

Is the rotation calculation something that could/should move to the transform object itself? Then a test for it would be a simple unit test.

astrofrog:
@djhoese - I added a couple of tests:

* For the rotation calculation, I don't think it makes sense for this to live on the transform, but I split it out into a dedicated property on ``Axis`` (``_rotation_angle``) which makes the code cleaner and easier to test (I added a unit test to make sure the homogeneous coordinate normalization works properly)

* For the updating of the rotation I realized that I could just check the OpenGL attribute ``a_rotation`` so I've done that and it seems to work nicely.

So I think everything here now has a test in some form or other now.

djhoese:
The failing test should be "fixed" on master by dropping py3.8 wheel builds. Merging this since it isn't the thing triggering the failures. Thanks again @astrofrog!

PR #1744, opened 2019-11-05
Improve AxisVisual visual by providing property getters and setters

astrofrog:
This adds getters and setters for a number of properties related to the appearance of the ``AxisVisual``. Also enables anti-aliasing for axis lines and ticks.

### DIFF
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -100,21 +100,22 @@ class AxisVisual(CompoundVisual):
         self.tick_direction = np.array(tick_direction, float)
         self.tick_direction = self.tick_direction
         self.scale_type = scale_type
-        self.axis_color = axis_color
-        self.tick_color = tick_color
 
         self.minor_tick_length = minor_tick_length  # px
         self.major_tick_length = major_tick_length  # px
         self.tick_label_margin = tick_label_margin  # px
         self.axis_label_margin = axis_label_margin  # px
 
-        self.axis_label = axis_label
+        self._axis_label_text = axis_label
 
         self._need_update = True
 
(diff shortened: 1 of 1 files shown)


vispy/visuals/axis.py
```diff
@@ -100,21 +100,22 @@ def __init__(self, pos=None, domain=(0., 1.), tick_direction=(-1., 0.),
...
         self.axis_label_margin = axis_label_margin  # px
 
-        self.axis_label = axis_label
+        self._axis_label_text = axis_label
 
         self._need_update = True
 
-        self._line = LineVisual(method='gl', width=axis_width)
+        self._line = LineVisual(method='gl', width=axis_width, antialias=True,
+                                color=axis_color)
```
djhoese: Should we take this opportunity to switch to 'agg' lines which will always support `width`? The GL line width is not supported on modern GPUs (it is deprecated from the specification).

astrofrog: The issue is that the ``LineVisual`` below uses ``connect='segments'`` which doensn't work with agg. We could still change the main axis line but not the ticks to agg, though that might look inconsistent.

djhoese: Wow, I didn't even realize that. I really thought I was using Agg for everything in my own work, but I guess now.

djhoese:
Just to be clear, this doesn't change any existing functionality, right? It only adds new properties which were previously "private".

astrofrog:
Yes that’s correct!

djhoese:
I know this is technically not a "bug fix" but I'd like to get this in the next 0.6.x release so I'm merging it. It's such a simple feature. Of course I'll regret it if this breaks something, but let's just pretend that won't happen.

PR #1746, opened 2019-11-07
Fail gracefully if OpenGL version can't be determined

astrofrog:
I ran into issues on Azure Pipelines where ``self.capabilities['gl_version']`` is an empty string which causes the version comparison below to crash. This PR adds a workaround to show a warning and not crash.

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -839,6 +839,12 @@ class GlirParser(BaseGlirParser):
             self.capabilities['gl_version'] = gl.glGetParameter(gl.GL_VERSION)
             self.capabilities['max_texture_size'] = \
                 gl.glGetParameter(gl.GL_MAX_TEXTURE_SIZE)
+
+            if self.capabilities['gl_version'] == '':
+                logger.warning('Could not determine OpenGL version. Some '
+                               'functionality may fail.')
+                return
+
             this_version = self.capabilities['gl_version'].split(' ')
             if this_version[0] == "OpenGL":
                 # For OpenGL ES, the version string has the format:


djhoese:
What was the error message you were getting (so people googling will find this conversation)?

In what cases is OpenGL properly installed and available but doesn't have the GL version string available? Is a warning "good enough" or does this need to be a more logical exception and error message?

astrofrog:
The error message was:

```
self = <[AttributeError("'LooseVersion' object has no attribute 'vstring'") raised in repr()] LooseVersion object at 0x1f1dd998bc8>

    def __str__ (self):
>       return self.vstring
E       AttributeError: 'LooseVersion' object has no attribute 'vstring'

c:\hostedtoolcache\windows\python\3.7.4\x64\Lib\distutils\version.py:324: AttributeError
```

which occurs if ``LooseVersion`` is given an empty string and compared to another version.

This is occurring on Azure Pipelines with Windows builds - presumably we'd see the same issues if we ran the test suite here on Azure. It's not easy to fully debug this but looking at ``glGetParameter`` (which is used to get the version string) it's possible to get an empty string back under some circumstances.

astrofrog:
Just for info, I double checked, and the following:

```
nativefunc = _get_gl_func("glGetString", ctypes.c_char_p, (ctypes.c_uint,))
res = nativefunc(gl.GL_VERSION)
```

returns ``None`` on Azure Pipelines with Windows.

djhoese:
> under some circumstances.

When I've seen this type of thing in the past it was because of an incorrectly installed GPU/OpenGL drivers or with someone trying to run OpenGL over an SSH/VNC connection.

You have to have a valid GL context to access gl parameters like the version, right? Even with this change, won't this still cause issues with other parts of the code running GL functions? If you can't get a valid GL version then how can we confidently move forward with the rest of the processing/tests/visualizations?

@larsoner @kmuehlbauer Thoughts?

astrofrog:
Ok so I was missing the Mesa OpenGL library which I was able to install with the script from http://github.com/vtkiorg/gl-ci-helpers (the one for AppVeyor) and this avoids the error above.

However, I still think it might be better to have a better error message than the one related to the version parsing - shall I make the warning I added here into an error instead?

djhoese:
> shall I make the warning I added here into an error instead?

Yes, I like that idea.I think this is one of those cases where the error message can be nice and long. It should probably mention that the version returned by the OpenGL drivers was empty and that this could mean missing or old OpenGL drivers.

Thoughts?

astrofrog:
So after investigating some more, I ran into further issues - when I install the Mesa OpenGL dll, things sometimes work and the version string is not empty, but sometimes fail with this kind of error:

```
Thread 0x00000730 (most recent call first):
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\vispy\app\backends\_qt.py", line 785 in _vispy_swap_buffers
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\vispy\app\canvas.py", line 417 in swap_buffers
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\vispy\util\event.py", line 471 in _invoke_callback
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\vispy\util\event.py", line 455 in __call__
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\vispy\app\backends\_qt.py", line 818 in paintGL
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\vispy\app\backends\_qt.py", line 501 in event
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\vispy\app\backends\_qt.py", line 501 in event
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\glue\app\qt\application.py", line 1193 in start
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\glue\app\qt\application.py", line 1123 in restore_session
  File "d:\a\1\s\.tox\py37-test\lib\site-packages\Windows fatal exception: gaccess violationl
```

I have no idea what's causing this, but it seems to happen when swapping the buffers. So that makes it unusable to me.

I then tried to set the DLL path explicitly to the one that ships with PyQt5 and that 'works' in the sense that I don't get access violations, but I'm back again to the problem of the OpenGL version string being empty. So I need to be able to gracefully ignore the fact the version string is empty when running the CI for my package.

My main motivation in running the CI for my package is to make sure that all the calls to VisPy work correctly but I don't check the rendering, so for me it's acceptable for the version check to not work and for some OpenGL calls to then fail afterwards, provided that they don't actually result in Python-level exceptions. So one option would be to make the behavior of the error/warning here dependent on an environment variable?

djhoese:
Could/should vispy perhaps provide a mocked gl module so that OpenGL doesn't have to be installed to run certain tests (in vispy or in downstream libraries/applications)? There are some vispy tests that need OpenGL and do actually test the buffer contents. Others depend on the GL context being sane and the order of operations being correct. Maybe there aren't that many vispy tests that can work without OpenGL. That doesn't necessarily mean that it wouldn't work for your application.

I would be ok with an environment variable check before error'ing out, but it still feels "wrong" to be testing an OpenGL application without knowing that the OpenGL calls are correct/working. Have you tried with this warning and your tests pass? Even tests that result in drawing operations?

astrofrog:
@djhoese - no there are exceptions in the drawing due to some shader compilation errors, so I'm not able to test that everything runs successfully on Windows, but I'm able to test everything else except the drawing. Having some kind of mocked gl module could be really handy for testing on headless machines with no OpenGL capabilities.

djhoese:
So from what I gather of our conversation, if the GL version is an empty string that usually means that there are no OpenGL drivers installed. This means that future drawing or other GL calls will fail. This means that tests doing anything with drawing (from a user's application that uses vispy) will fail and can't be tested.

So I feel like updating this PR to error out is the best course of action, if there is a kwarg or environment variable to control error versus warning, fine.

A new thought though, what if you used vispy's `vispy.testing._testing.requires_application` and/or `requires_pyopengl` decorators? These would skip tests if a vispy application backend could not be imported or if pyopengl is not available. I'm not sure if these gracefully detect no-GL drivers, but that is something that could be added. We could also remove the leading `_` from the testing utilities module.

astrofrog:
@djhoese - I agree the decorator idea might be useful. I'm not sure if I'll have time to work on this in the near future, so in the interest of avoiding stale PRs, I'm going to close this (since we aren't going with the current approach).

PR #1748, opened 2019-11-10
Fix xaxis labels being wrong initially in PlotWidget

### LINKED ISSUE #1742: axis labels problem

tlambert03:
This PR Fixes #1742 by adding the xaxis `AxisWidget` to the `PlotWidget.grid` before the `ViewBox` gets added.

Explanation (hope this is clear):
The order that widgets get added to the `PlotWidget.grid` in `PlotWidget._configure_2d` is currently problematic, because the xaxis `AxisWidget` (but not the yaxis `AxisWidget`) is getting added to the grid after the `ViewBox` is added.  As a result, when the `Grid._prepare_draw` method triggers `Grid._update_child_widget_dim` and it [loops through the grid widgets](https://github.com/vispy/vispy/blob/6f4e52aeecc40c29bf2fde3d217c0a3e6ea6b438/vispy/scene/widgets/grid.py#L481-L502) to update their sizes, the `ViewBox` `rect.setter` gets called before the xaxis `AxisWidget` has had a chance to update its size.  It is the `ViewBox` `rect.setter` that ultimately triggers the `resize()` event that is hooked to the `AxisWidget._view_changed` method that changes the axis domain & labels.  And though the ViewBox `rect.setter` _does_ get called a second time after the xaxis widget has changed its size, this time it fails the `self._pos_or_size_changed` test and does not trigger the `resize` event necessary for the xaxis redraw.  (this also explains why the problem was only affecting the x axis, because the y axis was added to the grid before the viewbox).

### DIFF
--- a/vispy/plot/plotwidget.py
+++ b/vispy/plot/plotwidget.py
@@ -142,7 +142,7 @@ class PlotWidget(scene.Widget):
         self.cbar_bottom = self.grid.add_widget(None, row=5, col=4)
         self.cbar_bottom.height_max = 1
 
-        # This needs to be added to the grid last (to fix #1743)
+        # This needs to be added to the grid last (to fix #1742)
         self.view = self.grid.add_view(row=2, col=4,
                                        border_color='grey', bgcolor="#efefef")
         self.view.camera = 'panzoom'
--- a/vispy/plot/plotwidget.py
+++ b/vispy/plot/plotwidget.py
@@ -122,6 +122,14 @@ class PlotWidget(scene.Widget):
         yaxis_widget = self.grid.add_widget(self.yaxis, row=2, col=3)
         yaxis_widget.width_max = 40
 
+        # row 3
+        # xaxis - column 4
+        # This needs to be added to the grid before the viewbox (to fix #1743)
+        self.xaxis = scene.AxisWidget(orientation='bottom', text_color=fg,
+                                      axis_color=fg, tick_color=fg)
+        xaxis_widget = self.grid.add_widget(self.xaxis, row=3, col=4)
+        xaxis_widget.height_max = 40
+
         self.view = self.grid.add_view(row=2, col=4,
                                        border_color='grey', bgcolor="#efefef")
         self.view.camera = 'panzoom'
@@ -129,13 +137,6 @@ class PlotWidget(scene.Widget):
 
         self.cbar_right = self.grid.add_widget(None, row=2, col=5)
         self.cbar_right.width_max = 1
-
-        # row 3
-        # xaxis - column 4
-        self.xaxis = scene.AxisWidget(orientation='bottom', text_color=fg,
-                                      axis_color=fg, tick_color=fg)
-        xaxis_widget = self.grid.add_widget(self.xaxis, row=3, col=4)
-        xaxis_widget.height_max = 40
 
         # row 4
         # xlabel - column 4
--- a/vispy/plot/plotwidget.py
+++ b/vispy/plot/plotwidget.py
@@ -124,16 +124,10 @@ class PlotWidget(scene.Widget):
 
         # row 3
         # xaxis - column 4
-        # This needs to be added to the grid before the viewbox (to fix #1743)
         self.xaxis = scene.AxisWidget(orientation='bottom', text_color=fg,
                                       axis_color=fg, tick_color=fg)
         xaxis_widget = self.grid.add_widget(self.xaxis, row=3, col=4)
         xaxis_widget.height_max = 40
-
-        self.view = self.grid.add_view(row=2, col=4,
-                                       border_color='grey', bgcolor="#efefef")
-        self.view.camera = 'panzoom'
-        self.camera = self.view.camera
 
         self.cbar_right = self.grid.add_widget(None, row=2, col=5)
         self.cbar_right.width_max = 1
@@ -147,6 +141,12 @@ class PlotWidget(scene.Widget):
         # row 5
         self.cbar_bottom = self.grid.add_widget(None, row=5, col=4)
         self.cbar_bottom.height_max = 1
+
+        # This needs to be added to the grid last (to fix #1743)
+        self.view = self.grid.add_view(row=2, col=4,
+                                       border_color='grey', bgcolor="#efefef")
+        self.view.camera = 'panzoom'
+        self.camera = self.view.camera
 
         self._configured = True
         self.xaxis.link_view(self.view)
--- a/vispy/plot/tests/test_plot.py
+++ b/vispy/plot/tests/test_plot.py
@@ -40,11 +40,11 @@ def test_plot_widget_axes():
         # note: fig.show() must be called for this test to work... otherwise
         # Grid._update_child_widget_dim is not triggered and the axes aren't updated
         fig.show(run=False)
+        # currently, the AxisWidget adds a buffer of 5% of the 
+        # full range to either end of the axis domain
         buffer = (point[1] - point[0]) * 0.05
         expectation = [point[0] - buffer, point[1] + buffer]
         for call in domain_setter.call_args_list:
-            # currently, the AxisWidget adds a buffer of 5% of the 
-            # full range to either end of the axis domain
             assert [round(x, 2) for x in call[0][1]] == expectation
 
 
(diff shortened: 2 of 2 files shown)


tlambert03:
Actually... yeah, I suppose it wouldn't be a bad idea to put all that stuff before the viewbox.  I haven't looked at the redraw events of all the other widgets involved in the `PlotWidget`, but it seems possible that other widgets might also be listening to that viewbox `rect.setter` event.  I'll take a look and update the PR accordingly.

djhoese:
Or should the view go before all of the children? Not sure it makes a difference overall.

The Windows Azure builds are failing but looks unrelated to your changes. I've restarted the builds but Azure doesn't seem to want to start them quite yet.

tlambert03:
Ok, looks like it won't work if the view is added before all the children.  And I do think it's safer to move it all the way to the end, so updated the PR to that effect.

vispy/plot/plotwidget.py
```diff
@@ -147,6 +142,12 @@ def _configure_2d(self, fg_color=None):
         self.cbar_bottom = self.grid.add_widget(None, row=5, col=4)
         self.cbar_bottom.height_max = 1
 
+        # This needs to be added to the grid last (to fix #1743)
```
jni: Are you fixing 1742 or 1743? 🤔

jni:
Should there be a new test to check that this is working...?

djhoese:
> Should there be a new test to check that this is working...?

I think technically there is in the form of an example script (the one mention in the related issue), but we may be skipping it do to issues between platforms. Or we may be running it but aren't actually checking the final visual result (by comparing to an output image). I'll look at this tomorrow. Going offline for the night now.

vispy/plot/plotwidget.py
tlambert03: whoops!  thanks :)

tlambert03:
I tried for a while to get a test to work, but I think it requires better mocking skills than I possess.  What I was hoping to do was to either mock `Viewbox.rect.setter`, or connect to the `PlotWidget.view.events.update()` event to assert that `PlotWidget.xaxis.axis.domain` is what would be expected at the moment of that `update` event.  But it's tricky because all of those instance are created inside that big `configure_2d` method... and by the time that method is done, the `axis.domain` will have eventually been set correctly.  And as I mentioned, I failed with the mocking strategy.

pcpatfire:
Thanks for the fix. With the view after all the children the plot.py and my application work well!

pcpatfire:
There is another problem with the x axis tick labels which is not solved by moving the view after all the children. The tick labels are over the axis. I modified `fig[0,0].xaxis.axis.tick_label_margin` setting it to 30 and this made the plot look better, but it's not the fix for the problem. Any idea about the reason why this happens? I think also `fig[0,0].xaxis.axis.axis_label_margin` has the same problem. As in the previous case, for the y axis everything works fine.

tlambert03:
> this made the plot look better, but it's not the fix for the problem

can you clarify what you mean here?  in my experience, tuning the tick_label_margin fixes the issue of overlapping tick labels.  For example

```python
import numpy as np
from vispy import plot as vp

fig = vp.Fig(size=(800, 800), show=False)
x = np.linspace(0, 10, 20)
y = np.cos(x)
ax = fig[0, 0]
line = ax.plot((x, y), symbol='o')
ax.xaxis.axis.tick_label_margin = 15
fig.show(run=True)
```

it might be a bit back-end dependent... making it potentially difficult to find a one-sized fits all tick_label_margin.  Does something like the above example not work for you?

pcpatfire:
Yes. It's what I am doing to fix the problem. As I need to change the tick label font size, I'm using a bit more complicated type of tuning, dependent on the thick_font_size:

`ax.xaxis.axis.tick_label_margin = int((self.tick_font_size)*28./6.)`

I empirically found the factor 28/6. It works quite well on Windows. I have to test it on Linux too. As you say, this problem can be back-end dependent (I'm using PyQt5), but probably is also operating system dependent. There should be a general fix, because other packages, such as matplotlib, do not show this problem, even using the same back-end and operating system. I still wonder way for the y axis there is not the same problem. The difference is that in this case a `width ` is used instead of a `height`, to calculate the distance of the tick labels from the axis.

djhoese:
Anything dealing with fonts and their location/size is likely going to end up being a very deep rabbit hole of hacky-ness.

@tlambert03 Any more thoughts on a test for this? Let me know if I can answer any questions on mocking if you had a specific way you wanted to test this. If you use do:

```
try:
    from unittest import mock
except ImportError:
    # python 2.7
    import mock
```

Then use mocks `patch` with the `wraps=<the actual function being wrapped>`, then you can run your tests and look through the mock objects calls:

```
rect_mock.assert_has_calls([mock.call(expected_x, expected_y)])
```

Or something like that.

tlambert03:
sorry!  I forgot this was waiting on a test.  let me work with your `mock` suggestion and get back to you.

tlambert03:
ok, finally figured something out.  I ended up directly patching the `AxisVisual` domain setter to make sure that it is set with the appropriate value given a particular range of data.  The test fails on the current master branch, but passes with this PR.

djhoese:
@tlambert03 Did you uncheck the box on the right hand side of the pull request which would allow me to push to your fork/branch? I tried to make a small update related to numpy warnings (and to trigger another test related to #1765) but it says I don't have permission.

tlambert03:
> @tlambert03 Did you uncheck the box on the right hand side of the pull request which would allow me to push to your fork/branch? I tried to make a small update related to numpy warnings (and to trigger another test related to #1765) but it says I don't have permission.

no, here's what my screen looks like:
Screen Shot 2019-11-24 at 9 22 22 PM

I think we actually struggled with this recent at napari...  @sofroniewn is that true?
I could try to close and reopen?

tlambert03:
i unchecked and rechecked... maybe try again?

djhoese:
I tried again with pycharm:

```
ERROR: Permission to tlambert-forks/vispy.git denied to djhoese.
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
```

I made #1766 for the specific changes I was going to make here. We'll see if @astrofrog responds to #1765 with any ideas, otherwise I'll just restart the tests some time this week and move forward with this.

djhoese:
Regarding push access, github seems **really** slow for me tonight. Even commenting is taking forever. I'll try again tomorrow.

tlambert03:
> Wow, you're right that this mocking got messy fast. Nice job figuring it out.

yeah, that took me a long time!  but it was a good mock-learning experience

djhoese:
It seems that restarting the OSX job with the failure isn't fixing it this time. I'll merge this for now and will have to work with @astrofrog in another PR to fix it.

tlambert03:
sorry about that :/
for what it's worth... the box is actually now _unchecked_ for allowing edits.  perhaps the "rechecking" I did last night trying to fix it didn't take (though it definitely did visually recheck last night).  i think something definitely was going on with github last night.  anyway, sorry for the trouble, and let me know if I can help in any way.  Thanks for merging!

pcpatfire:
It seems there is a general problem when using Grid objects. 
For instance the **size** of the `view ` widget created by PlotWidget is not updated, if the `ax.grid._update_child_widget()` is not called. I found that an alternative to that method is `ax.grid._recreate_solver()`, which updates child widgets as well (enough for what I need).

Another problem is related to Axis objects. I want to change the `tick_font_size` and adiust the `tick_label_margin` accordingly, as follows:
```
  self.ax.xaxis.axis._text.font_size = self.tick_font_size
  self.ax.xaxis.axis.tick_label_margin = int((self.tick_font_size)*28./6.)
```
The update works only if I call `ax.xaxis.axis._update_subvisuals()`.
I don't know if this last problem is related to the grid one.

djhoese:
@pcpatfire What do you think the best way forward is. It sounds like you and @tlambert03 understand this issue fairly well. @pcpatfire Maybe you could make a pull request with the changes you think are best and work for you and @tlambert03 could review them with their own use cases?

PR #1749, opened 2019-11-11
mesh bounds index error

sofroniewn:
This PR side-steps an `IndexError: list index out of range` error that happens when swapping cameras on 2D mesh. If desired I can come up with a minimal reproducible example purely using vispy code, but we've been encountering it quite a bit in napari - see https://github.com/napari/napari/issues/672 for the latest time. We might be using the cameras in a bad way to be triggering it, but what is happening is that we're asking for scene bounds for `axis=[0, 1, 2]`, but our mesh is 2D and so one is missing. We get the following error:

```bash
  File "/Users/nicholassofroniew/Github/napari/napari/_qt/qt_viewer.py", line 177, in _update_camera
    aspect=1, name="PanZoomCamera"
  File "/anaconda3/lib/python3.7/site-packages/vispy/util/frozen.py", line 17, in __setattr__
    object.__setattr__(self, key, value)
  File "/anaconda3/lib/python3.7/site-packages/vispy/scene/widgets/viewbox.py", line 111, in camera
    cam.view_changed()
  File "/anaconda3/lib/python3.7/site-packages/vispy/scene/cameras/base_camera.py", line 375, in view_changed
    self.set_range(*args)
  File "/anaconda3/lib/python3.7/site-packages/vispy/scene/cameras/base_camera.py", line 273, in set_range
    bounds = self._viewbox.get_scene_bounds()
  File "/anaconda3/lib/python3.7/site-packages/vispy/scene/widgets/viewbox.py", line 151, in get_scene_bounds
    b = ob.bounds(axis)
  File "/anaconda3/lib/python3.7/site-packages/vispy/visuals/visual.py", line 251, in bounds
    self._vshare.bounds[axis] = self._compute_bounds(axis, view)
  File "/anaconda3/lib/python3.7/site-packages/vispy/visuals/mesh.py", line 537, in _compute_bounds
    return self._bounds[axis]
IndexError: list index out of range
Abort trap: 6
```

This PR fixes that by returning `(np.inf, -np.inf)` instead of trying to index into the list at a coordinate that doesn't exist, which allows the camera scene bound code to do the right thing. It is backwards compatible as previous code would have just given an error.

Please let me know if you think this is an OK fix, or if you'd rather we try and find a fix somewhere else in the code. Please also let me know if you'd like me to add a test, and where that test should go.

### DIFF
--- a/vispy/visuals/mesh.py
+++ b/vispy/visuals/mesh.py
@@ -535,6 +535,6 @@ class MeshVisual(Visual):
         if self._bounds is None:
             return None
         if axis >= len(self._bounds):
-            return (np.inf, -np.inf)
+            return (0, 0)
         else:
             return self._bounds[axis]
--- a/vispy/visuals/mesh.py
+++ b/vispy/visuals/mesh.py
@@ -534,4 +534,7 @@ class MeshVisual(Visual):
     def _compute_bounds(self, axis, view):
         if self._bounds is None:
             return None
-        return self._bounds[axis]
+        if axis >= len(self._bounds):
+            return (np.inf, -np.inf)
+        else:
+            return self._bounds[axis]
--- a/vispy/visuals/tests/test_mesh.py
+++ b/vispy/visuals/tests/test_mesh.py
@@ -26,4 +26,28 @@ def test_mesh_color():
     np.testing.assert_allclose(vertices['position'], new_vertices)
 
 
+
+@requires_pyopengl()
+def test_mesh_bounds():
+
+    # Create 3D visual
+    vertices, filled_indices, outline_indices = create_cube()
+    axis = scene.visuals.Mesh(vertices['position'], outline_indices,
+                              color='black', mode='lines')
+
+    # Test bounds for all 3 axes
+    for i in range(3):
+        np.testing.assert_allclose(axis.bounds(i), (-1.0, 1.0))
+
+    # Create 2D visual using projection of cube
+    axis = scene.visuals.Mesh(vertices['position'][:, :2], outline_indices,
+                              color='black', mode='lines')
+
+    # Test bounds for first 2 axes
+    for i in range(2):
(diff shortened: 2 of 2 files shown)


djhoese:
Do you get this error *only* when switching from any camera to any other camera?

After skimming the relevant code, what I see is that the Mesh bounds are based on the shape of the data (the vertices). If that is true and you are viewing a 2D Mesh in a 3D environment (a camera that understands 3D space) then a fix like yours makes sense. Is `inf, -inf` the right answer though? First, is infinity the right limit versus `0, 0` or something like that? Also, why isn't the order `-inf, inf`?

sofroniewn:
I also get the error when just trying to view 2D mesh data with a 3D camera, which I now try and avoid as much as possible (I pad if I can), though in this case during the data switch and camera switch - from 3D to 2D I can't seem to prevent the call from getting made.

`(0, 0)` is probably a better choice than using `np.inf` though they were being overwritten (and order corrected) here https://github.com/vispy/vispy/blob/2c3ab163b7e90eddb14c2cd98e91dec615acae57/vispy/scene/widgets/viewbox.py#L152
as `(0, 0)` is what you'd get if you had zeros padded the data before hand.

I'll update this now

djhoese:
@sofroniewn Could you add a test for this? You may want to also rebase with master to avoid the failing azure windows build.

djhoese:
Oops spoke too soon. Flake8 doesn't like one of your changes:

```
vispy/visuals/tests/test_mesh.py:30:1: E303 too many blank lines (3)
```

sofroniewn:
Extra line dropped! Let me know if there's anything else. The new test includes both 2D and 3D meshes and would have failed without this change, and makes sure previous behavior is preserved

djhoese:
Thanks. Everything looks good now. Merging...

PR #1751, opened 2019-11-12
use Rect properties in PanZoomCamera

misterjoa:
Hello everyone,

Some small changes (first PR in this project...)

Added default value for rect
Use Rect class to get local rect properties

### DIFF
--- a/vispy/scene/cameras/panzoom.py
+++ b/vispy/scene/cameras/panzoom.py
@@ -43,9 +43,11 @@ class PanZoomCamera(BaseCamera):
 
     """
 
+    DEFAULT_RECT_TUPLE = (0, 0, 1, 1)
+
     _state_props = BaseCamera._state_props + ('rect', )
 
-    def __init__(self, rect=(0, 0, 1, 1), aspect=None, **kwargs):
+    def __init__(self, rect=DEFAULT_RECT_TUPLE, aspect=None, **kwargs):
         super(PanZoomCamera, self).__init__(**kwargs)
 
         self.transform = STTransform()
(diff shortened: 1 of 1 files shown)


misterjoa:
It is just to avoid calculations of rect that are already defined in the `Rect` class. It should return exactly the same results as before. Maybe I should write a test for this ?

vispy/scene/cameras/panzoom.py
```diff
@@ -152,22 +154,23 @@ def rect(self, args):
     @property
     def center(self):
         rect = self._rect
-        return 0.5 * (rect.left + rect.right), 0.5 * (rect.top +
-                                                      rect.bottom), 0
+        return (*rect.center, 0)
```
djhoese: Is this syntax python 2 compatible? We are dropping python 2 support in the next minor release but it'd be nice to not deliberately lose it for this small change.

djhoese:
Oh also the tests on travis seem to be failing because they have some issue with the styling of your changes (flake8). I'm not sure I understand the error; could be a bug in the flake8 checks.

kmuehlbauer:
> Oh also the tests on travis seem to be failing because they have some issue with the styling of your changes (flake8). I'm not sure I understand the error; could be a bug in the flake8 checks.

Seems like some glitch with flake8, I can't reproduce locally.

djhoese:
@kmuehlbauer Is it maybe a rule that you have globally ignored? Or maybe a different version of flake8? Maybe we should tell it to ignore it?

vispy/scene/cameras/panzoom.py
misterjoa: Indeed, it's not python2 compatible, it's been a long time since I've dropped python2...

djhoese: Yeah, it would be nice if you could explicitly do the first and second element for this small change. I don't like it, but this type of change wasn't really the thing I envisioned as breaking our python 2 support.

djhoese:
@misterjoa Do you think you'll have time to make this python 2 compatible any time soon?

misterjoa:
Hello @djhoese I'll wait for the new minor release to have less work :-)

kmuehlbauer:
@misterjoa Next minor release will be `0.7.0`. 

JFTR:
- 0.5.0: 24 Oct 2017
- 0.6.0: 28 Jul 2019
- 0.7.0: sometime next year (after this countdown is zero -> https://pythonclock.org/)

djhoese:
@misterjoa I'm asking for a one line change on a pull request that doesn't add a new feature or fix any bugs so that I can include this in a 0.6.x release. As mentioned the python 3 tests are failing because of a flake8 issue (although @kmuehlbauer says he can't reproduce it locally) so you need to make updates to this to get it to pass anyway.

misterjoa:
yes ok, will do !

djhoese:
Any updates on this? We have dropped python 2 support so no need to make that change. However, the tests are failing so if you could merge with master it would be appreciated.

djhoese:
Replaced by #2079. Thanks.

PR #1758, opened 2019-11-18
Fix texture alignment to use data itemsize

djhoese:
See https://github.com/napari/napari/issues/576 for the main discussion of this.

## OpenGL PACK/UNPACK Alignment

When writing or reading data to/from the GPU with OpenGL we need to specify the row alignment of our data. For writing data to the GPU we set the "GL_UNPACK_ALIGNMENT" parameter to 1, 2, 4, or 8. This tells OpenGL if the next row of data will be in the next byte, next 2-bytes, 4-bytes, or 8-bytes.

**Edit for clarification:** I meant the next aligned 2-bytes or 4-bytes, or 8-bytes. So if the data goes up to byte 30 and our alignment is 8 then OpenGL will get our next from starting at byte offset 32.

## How this applies to Vispy

In general Vispy requires or copies the data it is given by the user to be memory contiguous. So you might argue that the alignment can always be 1 (the next row starts in the next immediate byte). It isn't clear, but it seems there are some performance improvements if you choose a higher alignment. I have not tested this, but based on the Vispy code that was written a long time ago this seems to be the idea.

## The issue being fixed

Vispy currently assumes that all data are bytes (unsigned 8-bit integers), but this is often not true (32-bit floats, etc). Alignment should be based on width of a row, number of "bands" per pixel (3 for RGB, 1 for Luminance, etc), number of bytes per element (4 for 32-bit float, 1 for uint8, etc). This is the first bug fixed by this PR (use `data.itemsize` when determining alignment).

So for a 2D texture we should be getting the highest possible alignment for:

```
bytes_per_row = width * num_bands * itemsize
```

For 3D textures we should still only care about the width of a "row" (the last data dimension not including number of bands). This is the second bug being fixed by this PR where the Texture3D was using the "height" and "width".

Let's see what tests fail and then I'll come up with some tests to specifically check this based on the discussion in https://github.com/napari/napari/issues/576.

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -1503,7 +1503,7 @@ class GlirTexture1D(GlirTexture):
         if gtype is None:
             raise ValueError("Type %r not allowed for texture" % data.dtype)
         # Set alignment (width is nbytes_per_pixel * npixels_per_line)
-        alignment = self._get_alignment(data.shape[-1])
+        alignment = self._get_alignment(data.shape[-1] * data.itemsize)
         if alignment != 4:
             gl.glPixelStorei(gl.GL_UNPACK_ALIGNMENT, alignment)
         # Upload
@@ -1536,7 +1536,7 @@ class GlirTexture2D(GlirTexture):
         if gtype is None:
             raise ValueError("Type %r not allowed for texture" % data.dtype)
         # Set alignment (width is nbytes_per_pixel * npixels_per_line)
-        alignment = self._get_alignment(data.shape[-2]*data.shape[-1])
+        alignment = self._get_alignment(data.shape[-2] * data.shape[-1] * data.itemsize)
         if alignment != 4:
             gl.glPixelStorei(gl.GL_UNPACK_ALIGNMENT, alignment)
         # Upload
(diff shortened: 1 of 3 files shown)


tlambert03:
just tested this and it fixes the problems we've seen in napari.  Thank you!

djhoese:
FYI to any of the napari folks watching this PR, I've rebased this to get some changes to the azure build config that were sitting in master. None of the code for this PR has changed, but if you had a copy of this you'll need to probably re-pull.

djhoese:
@kmuehlbauer @larsoner Any good ideas of how/where to add a test for this? I could mock the `gl` module and use the glir Texture objects directly but that won't give anything to the GPU so we won't get a GPU. Eh I guess just checking the alignment is the proper number is good enough.

larsoner:
Yeah I would just check the number

djhoese:
There haven't been any complaints (not that I asked for feedback), but tests have been added and everything is passing so I'm going to merge this. Feel free to leave feedback and we can revert changes if needed in a separate PR.

sofroniewn:
I can confirm all the napari issues in https://github.com/napari/napari/issues/576 are now fixed with the current version of vispy master. Thanks so much for getting this in so quickly!

PR #1691, opened 2019-08-06
Fix numpy futurewarning in dtype creation

djhoese:
In future versions of numpy defining a dtype as `('name', np.float32, 1)` will be interpreted differently. You currently get this warning:

```
/home/travis/build/vispy/vispy/vispy/visuals/markers.py:564: FutureWarning: Passing (type, 1) or '1type' as a synonym of type is deprecated; in a future version of numpy, it will be understood as (type, (1,)) / '(1,)type'.
  ('a_edgewidth', np.float32, 1)])
```

I fixed a couple of these after it was pointed out by some napari developers, but these should be all the rest now that travis can complain about them.

If I remember correctly vispy is currently using `, 1)` as "scalar" and in the future this will be treated as a length 1 array. But now I don't remember what `, 0)` means.

Issue #1765, opened 2019-11-25
Recent AxisVisual cause random test failures


djhoese:
@astrofrog recently made some changes to the AxisVisual in #1743 related to rotation handling. The test that was added seems to be failing as seen in the CI runs for #1748 (see https://travis-ci.org/vispy/vispy/jobs/616039829).

```
>       assert_allclose(axis2._rotation_angle, -17.056795)

E       AssertionError: 

E       Not equal to tolerance rtol=1e-07, atol=0

E       

E       Mismatch: 100%

E       Max absolute difference: 0.36324159

E       Max relative difference: 0.021296

E        x: array(-16.693553)

E        y: array(-17.056795)
```

I've seen this fixed by just restarting the test. @astrofrog could you look at what's going on?

djhoese:
And now I can't get them to pass by an update. This may be a OSX versus linux thing. @astrofrog what operating system do you use for development?

djhoese:
As mentioned in #1767 this seems to be passing now in travis with new pull requests that people have made. I'll close this for now and reopen if the failures pop up again. I thought it looked like a pytest/pluggy dependency issue, but obviously that isn't the problem given the recent successes.

djhoese:
Ok, this is still a problem. I merged a pull request that was passing tests and then it started failing in master. It doesn't seem to be tied to dependency versions which means it is some kind of race condition or random-ness in the drawing. I've seen issues like this before in other tests. Last resort is to skip (xfail) the test on travis OSX environments.

astrofrog:
We could just increase the tolerance of the test to ~5%?

djhoese:
Yeah that is the other second-to-last option. the master branch was failing and I restarted it 3 times and it started passing.

PR #1766, opened 2019-11-25
Fix additional numpy warnings when dealing with dtypes of size 1

djhoese:
These were missed by #1691 but I started noticing them in some recent PR CI tests.


djhoese:
Failing test is caused by #1765. This is the problem so merging now.

PR #1769, opened 2019-11-26
Fix automatic limits of ImageVisual with non-finites values

EtienneCmb:
Hello @djhoese ,

Here's the PR that only contains the patch when there's non finite values inside the data.

Small code for reproducibility :

```python
import numpy as np

from vispy import scene
from vispy import app

# Camera and scene canvas :
cam = PanZoomCamera(rect=(0, 0, 4, 5))
sc = scene.SceneCanvas(bgcolor='black')

# 2d array of data
data = np.arange(20).reshape(5, 4).astype(float)
data[2, 0] = -np.inf
data[2, -1] = np.inf
data[-1, 2:3] = np.nan

# Red image :
im_red = scene.visuals.Image(data, parent=wc.scene, cmap='viridis')

sc.show()
app.run()
```

Basically, if the data contains NaN or inf, the minimum and maximum of clim are not corrects which forces the render to be a single color :
Screenshot_20191126_191530

After modifications :
Screenshot_20191126_191744

**Note** : on matplotlib those pixels are discarded which is not the case here.
Screenshot_20191126_191913

### DIFF
--- a/vispy/visuals/image.py
+++ b/vispy/visuals/image.py
@@ -401,7 +401,8 @@ class ImageVisual(Visual):
             # can eventually do this by simulating 32-bit float... maybe
             clim = self._clim
             if isinstance(clim, string_types) and clim == 'auto':
-                clim = np.min(data), np.max(data)
+                finite_data = data[np.isfinite(data)]
+                clim = np.min(finite_data), np.max(finite_data)
             clim = np.asarray(clim, dtype=np.float32)
             data = data - clim[0]  # not inplace so we don't modify orig data
             if clim[1] - clim[0] > 0:


djhoese:
Nice. Thanks for separating this out.

I'm a little concerned with the possible performance hit of always doing `np.isfinite` on the entire image array. Using `nanmin` and `nanmax` seems reasonable if we assume that numpy is doing this as fast as possible. Creating the new array from `isfinite` could use a lot of memory and time. This could be a big performance hit for people who are using the ImageVisual with large data arrays.

EtienneCmb:
Agree, but `nanmin` and `nanmax` doesn't detect infinite values

Screenshot_20191126_193839

djhoese:
:thinking: 

Is it too much to ask of the user to set the color limits themselves in these cases? How common is it that an image has infinite values?

EtienneCmb:
Hard to say, I guess it depends on the field. If you prefer, we can just use `np.nanmin` and `np.nanmax`. Up to you.

I made a quick test, on relatively large images of shape (10000, 10000) the finite version is three times slower than the nan version. On smaller arrays, computing time are equivalents.

kmuehlbauer:
After thinking about this for a while, my suggestion would be to re-use `clim` keyword argument.

Currently if we use `clim='auto'` the limits are gathered from the given data automatically. 
https://github.com/vispy/vispy/blob/16fb53df702af7f6018877b761d5b4e0a4978f5e/vispy/visuals/image.py#L402-L410
My interpretation of the code: The `else` is here to set the data to either all 0 or all 1 in case anything is wrong with the limits.

I'm strongly in favour of keeping this behaviour, because this is the only way for the user to see **_something_** went wrong. Just considering the user doesn't know about `NaN` and/or `INF` in his data, silently setting the limits to finite values can and will lead to misinterpretation of the data.

Instead we could add more meanings to the `clim`-keyword, like `fixnan`, `fixinf` or `fixall` (just example naming), which handle the other cases. The user would have to request this and it would be clear that `NaN` and `-INF` will be set to lower color bound and `+INF` to the upper bound (which should be reflected in the docstring).

Just my 2c.

EtienneCmb:
Hello @kmuehlbauer ,

I do agree with you that silent modifications are always bad. Now, setting 1 or 0 for the entire array is a bit extreme in my opinion and it's a case where data are modified and the user don't know it.

About the `clim='auto'`, the way I understand it is to "find the best way for plotting my data". An alternative could be :
- `clim=None` : use `np.min` and `np.max` and if `clim[1] < clim[0]` raise a value error
- `clim='auto'` : find optimal boundaries using the finite array (like this is proposed now)

kmuehlbauer:
Best solution would be to have this handled in the shader, but as already discussed this might be tricky.

To move this forward i think this proposal (@EtienneCmb :+1:):

> About the `clim='auto'`, the way I understand it is to "find the best way for plotting my data". An alternative could be :
> 
>     * `clim=None` : use `np.min` and `np.max` and if `clim[1] < clim[0]` raise a value error
> 
>     * `clim='auto'` : find optimal boundaries using the finite array (like this is proposed now)

might be an elegant solution too. We should check for `NaN`/`INF` in `clim` instead before the current if-clause. After re-reading I think that  was for cases where all data values are the same.

In any case we would need to make `clim=None` the default and clarify the rest in the docstring and error message. At least users will be notified then.

djhoese:
My only concern with changing the default to `None` and having `auto` mean something else is that I think `auto` is used in other Visuals in vispy so consistency would be nice.

kmuehlbauer:
`ImageVisual`, `MeshVisual` and `SpectrogramVisual` are using `clim` and `auto`. Where `SpectrogramVisual` is inherited from `ImageVisual`. In `MeshVisual` `clim`  is not derived from data but from created `MeshData` which should not suffer from `NaN` or `INF`.

If we change the current behaviour of `auto` we would need to use `None` as default and raise an error. At least if we want to make the user aware of the situation.

jni:
Just weighing in here because we coincidentally just this morning were discussing nans in napari in the context of [this tweet](https://twitter.com/TobiasAdeJong/status/1200928293291274240). It would be super awesome if nans, if present, were mapped to alpha=0.

I'll also point out that NumPy's nan{min,max,mean,etc} are significantly slower than the corresponding non-safe functions. So I would suggest not using them by default in vispy.

djhoese:
@jni Is napari focused on supporting any older or mobile platforms at the moment? A lot of this logic could be "fixed" by storing the float data directly in the GPU, but it may not be available on all platforms...yet.

Side note: I also saw that tweet as you can see by all my comments. Something I've wanted in dask-based notebooks for a while. Not quite as pretty as I'd like, but it's getting there.

jni:
@djhoese 

> Is napari focused on supporting any older or mobile platforms at the moment?

Define "older"? No, we are not focused on mobile for the foreseeable future, nor very old platforms, but I don't know exactly what you mean by "old". Is my 2012 Macbook Pro 15" Retina, which is still working great and works a treat with vispy/napari, "old"? Or do you mean older OSs like Mac OS X Lion/Windows XP? We already don't support those.

djhoese:
Good question. At this point OpenGL 3.x features should be available in any GPU from the last 7 or more years so we shouldn't have to worry about it. Worst case a user tries to use it on their old system and it fails.

So I don't have time to look at implementing this right now, but I did do a hacky version of it in my [own project](https://github.com/ssec/sift/blob/master/uwsift/view/visuals.py#L344). We'll have to provide a keyword argument to let the user control the texture format. *Or* we have a kwarg that is a toggle to say "do color limit calculations on the CPU" versus "on the GPU".

djhoese:
So coming back to this, the end goal is to have:

1. NaNs/Infs be displayed as transparent pixels
2. Not include NaNs/Infs in auto color-limiting

Right? In the end we'd like to match matplotlib as much as possible to avoid surprises for users.

So this PR handles 2 at the cost of performance. Point 1 can only really be handled with floating point data in shader-land and trying to detect nan/inf there (GLSL: `discard` for not drawing the pixel). The only other way to handle 2 is with designating a fill value and I don't want to go down that road.

My point about performance was easy to make since a user could always specify the min/max themselves outside the ImageVisual. I'm not realizing that this is true for the opposite case though too. if we adopt this PR then users who want the performance/memory usage from before can just calculate the clims themselves...right?

So all of that said I think I'm ok merging this PR as-is, but want to know if anyone else has ideas or wants to point out something I'm missing.

vispy/visuals/image.py
```diff
@@ -401,7 +401,8 @@ def _build_texture(self):
             # can eventually do this by simulating 32-bit float... maybe
             clim = self._clim
             if isinstance(clim, string_types) and clim == 'auto':
-                clim = np.min(data), np.max(data)
+                finite_data = data[np.isfinite(data)]
```
kmuehlbauer, line 404: @djhoese As I wrote above, it would be nice to issue a Warning if the source data contains NaN/INF, so the user has at least an information that the image might not be what he expects. Other wise I'm fully in favour of adding this.

djhoese: Is there a way to do that without an even bigger performance hit? You said that not warning about it and doing it this way will lead to misinterpretation of the data. I get that, but I'm on the fence as to whether this is that likely. Of course now that I've said that it is going to happen to me.

EtienneCmb: Could looks like this :

```python
isfinite = np.isfinite(data)
if isfinite.any():
    from warnings import warn
    warn("Image contains non-finite values that have been ignored to compute colorbar limits")
    # maybe logger.warning() would be better?
finite_data = data[isfinite]
clim = np.min(finite_data), np.max(finite_data)
```

EtienneCmb:
At this point I think it's a question of what is / should be Vispy :
- If Vispy is a "low-level" package with the main focus on efficiency, in that case I would not merge this PR. And you let the entire responsability on the user to check for NaN 
- If Vispy is also targeting non-expert users, in that case I think this PR is worth it

Now, I also think that trying to mimic the behavior of Matplotlib is not a necessity, it's only a question of what is the purpose of Vispy.

vispy/visuals/image.py
larsoner: One way to stay performant is just to penalize the people who need penalizing, e.g.:
```
min_ = np.nanmin(data)
max_ = np.nanmax(data)
if not np.isfinite(min) or not np.isfinite(max_):
    warn(...)
```
then people with only nan+finite pay no performance penalty. In this case I wouldn't even bother with the `warn` and just say in the docstring that if you have infinite values in your data, the number of passes for computing min/max will double

EtienneCmb: That's a nice solution, however here you need to perform `nanmin` and `nanmax`, which by definition ignore NaN and are just looking for inf. Maybe simply `np.min` and `np.max` could work? I've heard that nan NumPy operators are slower

EtienneCmb: Ok, I get it, `np.min` will return `nan`

jni: Just to add some data to the discussion: @larsoner's approach is indeed faster:

```python
In [5]: %%timeit  
   ...: finite = np.isfinite(image) 
   ...: min_ = np.min(image[finite]) 
   ...: max_ = np.max(image[finite]) 
   ...:                                                                                             
25.6 ms ± 178 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: %%timeit 
   ...: min_ = np.nanmin(image) 
   ...: max_ = np.nanmax(image) 
   ...:                                              
7.25 ms ± 43.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I guess this is because boolean indexing performs a copy, regardless of the contents of the array (in this case, all the values are finite, non-nan).

nanmin is indeed slower than min:

```python
In [3]: %timeit np.min(image)                                                                                                 
2.7 ms ± 26.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit np.nanmin(image)                                                                                              
3.69 ms ± 46.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

So, perhaps it should be a cascade:
- try min and max
- if they are not finite, try nanmin, nanmax
- if they are not finite, use isfinite

This would make the performance penalty be ~0 for finite arrays (which constitutes all inputs so far), and ~2.3x for finite + nan arrays, and ~∞ for arrays containing inf, which is somewhat appropriate. =P For me, this is preferable to adding a >30% penalty to all users by using nanmin/nanmax by default. As @djhoese points out, full performance can be restored for users with nans by passing the the clims up front.

Regarding @EtienneCmb's comment about what VisPy is, I would put it somewhere in the middle: for me, VisPy is a convenience layer for *fast* OpenGL-based visualisation. Usability is important, but speed even more so.

almarklein:
Superseded by #2085

Issue #1771, opened 2019-11-30
Test suite fails on Ubuntu Due to Font Noto Emoji


hmaarrfk:
Font Noto Color Emoji is not scalable

```
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― test_font_glyph ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

    def test_font_glyph():
        """Test loading glyphs"""
        # try both a vispy and system font
        sys_fonts = set(list_fonts()) - set(_vispy_fonts)
        # assert_true(len(sys_fonts) > 0) not always the case!
        for face in ['OpenSans'] + sorted(list(sys_fonts)):
            print(face)  # useful for debugging
            font_dict = dict(face=face, size=12, bold=False, italic=False)
            glyphs_dict = dict()
            chars = 'foobar^C&#'
            for char in chars:
                # Warning that Arial might not exist
                with warnings.catch_warnings(record=True):
                    warnings.simplefilter('always')
>                   _load_glyph(font_dict, char, glyphs_dict)

vispy/util/fonts/tests/test_font.py:37: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
vispy/util/fonts/_freetype.py:50: in _load_glyph
    face = _load_font(f['face'], f['bold'], f['italic'])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

face = 'Noto Color Emoji', bold = False, italic = False

[...]

PR #1772, opened 2019-11-30
Skip bad font in test_font

hmaarrfk:
Fixes: https://github.com/vispy/vispy/issues/1771

Also uses fixtures so as not to change the global state of the the warnings filter.

### DIFF
--- a/vispy/util/fonts/tests/test_font.py
+++ b/vispy/util/fonts/tests/test_font.py
@@ -8,9 +8,20 @@ import warnings
 
 from vispy.testing import assert_in, run_tests_if_main
 from vispy.util.fonts import list_fonts, _load_glyph, _vispy_fonts
+import pytest
 
-warnings.simplefilter('always')
+known_bad_fonts = set([
+    'Noto Color Emoji',  # https://github.com/vispy/vispy/issues/1771
+])
 
+@pytest.fixture(scope='module')
+def setup():
+    with warnings.catch_warnings():
+        warnings.simplefilter('always')
+        yield
+
+# try both a vispy and system font   <--- what does this mean???
+sys_fonts = set(list_fonts()) - set(_vispy_fonts)
 
 def test_font_list():
     """Test font listing"""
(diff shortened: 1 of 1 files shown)


djhoese:
Thanks @hmaarrfk.

@kmuehlbauer @larsoner I was talking with a coworker who found out that vispy's tests technically don't need pytest. I recently added some tests and this PR adds tests that import `pytest` from the top of the module. This seems like the easiest time to ask if you are OK with me removing the fallback code in vispy's tests so that it will now require `pytest` to run.

kmuehlbauer:
> This seems like the easiest time to ask if you are OK with me removing the fallback code in vispy's tests so that it will now require `pytest` to run.

@djhoese Removing code sounds good. I'm fine with whatever makes live easier .

hmaarrfk:
yeah, you guys definitely require `nose` which is quite old, and will likely to stop being supported
https://nose.readthedocs.io/en/latest/#note-to-users

many projects including numpy and scipy have moved to completely supporting and requiring pytest
https://github.com/numpy/numpy/blob/master/test_requirements.txt#L2

djhoese:
@hmaarrfk I'm 90% sure I've looked at this and we don't depend on `nose` but do depend on parts of numpy.testing that require nose.

hmaarrfk:
Yeah, seems like you are right. I do see lots of logic in there that is trying to work around the myriad of different testing backends.

larsoner:
Yes a requirement for pytest is fine

vispy/util/fonts/tests/test_font.py
```diff
@@ -3,39 +3,42 @@
...
 
-def test_font_glyph():
+@pytest.mark.parametrize('face', ['OpenSans'] + sorted(sys_fonts))
+def test_font_glyph(face):
     """Test loading glyphs"""
-    # try both a vispy and system font
-    sys_fonts = set(list_fonts()) - set(_vispy_fonts)
-    # assert_true(len(sys_fonts) > 0) not always the case!
-    for face in ['OpenSans'] + sorted(list(sys_fonts)):
-        print(face)  # useful for debugging
```
hmaarrfk: The print statement is no longer useful since the font name will appear sure to parameterization

djhoese: Awesome!

vispy/util/fonts/tests/test_font.py
```diff
@@ -3,39 +3,42 @@
...
-        glyphs_dict = dict()
-        chars = 'foobar^C&#'
-        for char in chars:
-            # Warning that Arial might not exist
-            with warnings.catch_warnings(record=True):
-                warnings.simplefilter('always')
-                _load_glyph(font_dict, char, glyphs_dict)
-        assert_equal(len(glyphs_dict), np.unique([c for c in chars]).size)
+    if face in known_bad_fonts:
+        pytest.skip()
```
hmaarrfk:
```suggestion
        pytest.xfail()
```

PR #1774, opened 2019-11-30
Fixup decorator to enable full use of pytest.

hmaarrfk:
I get that moving backends is a "decision" that must be thought of. 

This is just a little fix that enables decorators to be stacked and used with pytest.

Otherwise, enabling the gc destroys other pytest decorators.


djhoese:
This is passing so I'm going to merge it. Something is up with the Azure build but I'm not going to debug it in this PR.

hmaarrfk:
Hey thanks for this!

PR #1775, opened 2019-11-30
Update tests to use pytest better

hmaarrfk:
Needs: https://github.com/vispy/vispy/pull/1774

Which has been folded into this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,7 @@ install:
     - conda create -n testenv --yes --quiet pip python=$PYTHON;
     - source activate testenv;
     # add needed packages common to all backends/tests
-    - conda install --yes --quiet numpy scipy nose pytest cython coveralls pytest-cov pytest-sugar pytest-faulthandler flake8;
+    - conda install --yes --quiet numpy scipy pytest cython coveralls pytest-cov pytest-sugar pytest-faulthandler flake8;
 
     # On Python3, install system-wide copies of bundled libraries instead
     # Also install PyQt5, imaging (PIL or pillow), scipy, mpl, egl
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,10 +117,8 @@ install:
     # WX requires OSMesa (mesa on conda) which has typically been an
     # additional test environment. With llvm=3.3 the combination of
     # EGL and mesa causes segmentation faults. See issue #1401.
-    # can remove the github numpydoc install once >0.8.0 is released and get it from conda
     - if [ "${DEPS}" == "full" ]; then
-        conda install --yes pyopengl networkx pysdl2;
-        pip install numpydoc;
+        conda install --yes pyopengl networkx pysdl2 numpydoc;
         if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
           conda install --yes matplotlib jupyter pyqt=5 pillow decorator six scikit-image glfw;
           if [ "${PYTHON}" == "3.7" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-language: python
+language: generic
+os: linux
 dist: xenial
 addons:
   apt:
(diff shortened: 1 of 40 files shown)


djhoese:
Thanks for looking in to this. It looks like `numpydoc` is still needed by some tests to look at docstrings. Could we use `flake8-docstring` for this?

djhoese:
What do you think about removing the `run_tests_in_object` function which is used when pytest is not imported (lower in the _testing.py module I think)?

hmaarrfk:
> What do you think about removing the run_tests_in_object function which is used when pytest is not imported (lower in the _testing.py module I think)?

depends on how much you want to review

Why is numpydoc simply not a test dependency?

appveyor.yml
```diff
@@ -10,10 +10,10 @@ environment:
 
   matrix:
       - PYTHON_VERSION: "2.7"
-        CONDA_DEPENDENCIES: "numpy scipy setuptools nose pytest coverage pytest-cov pytest-sugar pytest-faulthandler cython coveralls mock"
+        CONDA_DEPENDENCIES: "numpy scipy setuptools pytest coverage pytest-cov pytest-sugar pytest-faulthandler cython coveralls mock"
         PIP_DEPENDENCIES: "numpydoc https://files.pythonhosted.org/packages/71/5f/07aad120ca6e4339a5c127631341787bc6bc74add39a1f41223bda949721/freetype_py-2.1.0.post1-py2.py3-none-win_amd64.whl"
       - PYTHON_VERSION: "3.7"
-        CONDA_DEPENDENCIES: "numpy scipy setuptools nose pytest coverage pytest-cov pytest-sugar pytest-faulthandler cython coveralls"
+        CONDA_DEPENDENCIES: "numpy scipy setuptools pytest coverage pytest-cov pytest-sugar pytest-faulthandler cython coveralls"
         PIP_DEPENDENCIES: "numpydoc https://files.pythonhosted.org/packages/71/5f/07aad120ca6e4339a5c127631341787bc6bc74add39a1f41223bda949721/freetype_py-2.1.0.post1-py2.py3-none-win_amd64.whl"
```
hmaarrfk: Why do you need this numpydoc dependency?

.travis.yml
```diff
@@ -105,7 +105,7 @@ install:
     - conda create -n testenv --yes --quiet pip python=$PYTHON;
     - source activate testenv;
     # add needed packages common to all backends/tests
-    - conda install --yes --quiet numpy scipy nose pytest cython coveralls pytest-cov pytest-sugar pytest-faulthandler flake8;
+    - conda install --yes --quiet numpy scipy pytest cython coveralls pytest-cov pytest-sugar pytest-faulthandler flake8;
```
hmaarrfk:
```suggestion
    - conda install --yes --quiet numpy scipy pytest cython coveralls pytest-cov pytest-sugar pytest-faulthandler flake8 numpydoc;
```

appveyor.yml
hmaarrfk: i see, you needed bleeding edge at one point. Seems like both freetype-py and numpydoc got released on conda-forge as you required.

hmaarrfk:
now that you have embraced the wild wild west of conda-forge, you should be able to fully test everything on conda soon: https://github.com/conda-forge/staged-recipes/pull/10342

djhoese:
> you should be able to fully test everything on conda soon

You mean, no pip installs? At first I thought you meant run vispy tests on the conda-forge builds.

> Why is numpydoc simply not a test dependency?

Just missed dependencies.

> depends on how much you want to review

I assumed it would be a small change along with the other import changes you did. Since pytest is required I thought it made sense and was going to do it myself, but since you're already changing this I thought I'd ask.

Thanks again.

djhoese:
The other thing about numpydoc is that for rendering the sphinx docs we could probably just use napolean that now comes with sphinx.

djhoese:
Apparently flake8-docstring is much more strict that numpydoc. If you install flake8-docstrings and run the normal flake checks:

```
python -m make test flake | wc -l
6315
```

djhoese:
A lot of this was handled similarly in another PR by @larsoner, but there were some things in here I didn't want to lose. The stuff discussed here as alternative solutions (flake8-docstrings, napolean, etc), we can handle in another PR.

PR #1757, opened 2019-11-18
make it possible to update text font face/bold/italic

thomasdeneux:

PR #1777, opened 2019-12-01
Make it possible to get and set the face, bold, and italic properties of Text via properties

astrofrog:
For a use case I have, I needed to be able to find out whether a ``Text`` object had a bold font, and needed to be able to set it, so I decided to add properties for the face, bold, and italic properties of Text (since these can be set via the ``__init__``, I thought having properties that match these arguments would make sense).

### DIFF
--- a/vispy/visuals/text/text.py
+++ b/vispy/visuals/text/text.py
@@ -414,7 +414,10 @@ class TextVisual(Visual):
         # Init font handling stuff
         # _font_manager is a temporary solution to use global mananger
         self._font_manager = font_manager or FontManager(method=method)
-        self._font = self._font_manager.get_font(face, bold, italic)
+        self._face = face
+        self._bold = bold
+        self._italic = italic
+        self._update_font()
         self._vertices = None
         self._color_vbo = None
         self._anchors = (anchor_x, anchor_y)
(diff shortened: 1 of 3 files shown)


vispy/visuals/text/text.py
```diff
@@ -606,6 +609,36 @@ def _prepare_transforms(self, view):
...
+    def italic(self):
+        return self._italic
+
+    @italic.setter
+    def italic(self, value):
+        self._italic = value
+        self._update_font()
+
+    def _update_font(self):
+        self._font = self._font_manager.get_font(self._face, self._bold, self._italic)
```
astrofrog: What is the proper way of triggering a redraw here?

djhoese:
Could you please compare this with https://github.com/vispy/vispy/pull/1757 which from the title seems to be the same feature set. @thomasdeneux FYI.

astrofrog:
Oops, sorry for the duplication of effort! #1757 looks very similar - the main difference is that here I use separate private properties to store the current state instead of a tuple, and this PR also has a test. To move forward, the test here could be cherry picked over to #1757, or alternatively I can include the commit from #1757 here as the first commit. Let me know how you'd like to proceed!

thomasdeneux:
hello @astrofrog, interesting that we had the same need at almost the same time! I did not have the time to write tests yet, and probably is it better to store the current state in a more explicit way as you do. So i don't see anything that #1757 has which #1777 does not have and it is fine with me to drop #1757, which probably is the simpler.
(and thanks @djhoese for seing this and avoiding that we go further in writing redundant code)

djhoese:
Thanks @astrofrog and @thomasdeneux. I agree that storing the state in separate attributes is better. I'll close #1757 as a duplicate. Thanks again both of you for reviewing this coincidence.

astrofrog:
@djhoese - just to check, are you planning on merging this? Thanks!

djhoese:
Yep :wink: Thanks for reminding me.

PR #1780, opened 2019-12-03
Switch to setuptools_scm version.py usage to avoid import overhead

### LINKED ISSUE #1779: Slow import
On my Windows workstation with >1200 of Python packages installed, `import vispy` takes about 3.4 seconds. 

98% of that time is spend at     
https://github.com/vispy/vispy/blob/ded293841c6438ab54af10c561f372155bb77edc/vispy/__init__.py#L26

On this system, more than half of `napari`'s import time is spent on determining the vispy version. 

Please consider not using `pkg_resources` during runtime.

Loosely related to https://github.com/napari/napari/pull/745

djhoese:
Closes #1779 

This switches vispy to use a feature of setuptools_scm which writes the generated version number to a `vispy/version.py` file. This is different from what was previously done which was to use setuptools to read the installed package's metadata and extract the version number. As discussed in #1779, this can cause major overhead on systems with a lot of packages installed. This PR is the simplest workaround for that.

The downside to this is that developers installing vispy in a "development mode" may be confused by a `vispy/version.py` that is ignored by `.gitignore` and was generated just by them installed the package.

## Additionally

This starts to put in place some functionality that setuptools_scm recently documented and added to enable setuptools_scm from the pyproject.toml file. However, this requires setuptools 42 or above which is not widely available so I'm not sure it is smart to require that for building quite yet. Maybe version 0.7 or 0.8 (assuming this gets included in a 0.6.x release).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ js/node_modules
 js/dist
 node_modules/
 vispy/static/
+vispy/version.py
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -16,6 +16,7 @@ recursive-include vispy *.vert
 include vispy/app/tests/qt-designer.ui
 include vispy/io/_data/spatial-filters.npy
 include vispy/util/fonts/data/*.ttf
+include vispy/version.py
 
 recursive-include examples *.py
 recursive-include examples *.ipynb
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,9 @@ requires = [
     "wheel",
     "setuptools>=30.3.0",
     "setuptools_scm",
+    # Use these when setuptools>=42 is more widely available
+    # "setuptools>=42",
+    # "setuptools_scm[toml]>=3.4",
     "setuptools_scm_git_archive",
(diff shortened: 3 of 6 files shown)


djhoese:
@cgohlke If you have time to try this branch out and install it on your system it would be greatly appreciated.

cgohlke:
Perfect: `import vispy` is down to ~0.2s from > 3s.

Btw, you may want to exclude `make` from binary distributions:
```diff
diff --git a/setup.py b/setup.py
index 53e8ee55..9ae0c74d 100644
--- a/setup.py
+++ b/setup.py
@@ -223,7 +223,7 @@ setup(
         'wx': ['wxPython'],
         'doc': ['sphinx_bootstrap_theme', 'numpydoc'],
     },
-    packages=find_packages(),
+    packages=find_packages(exclude=['make']),
     ext_modules=cythonize(extensions),
     package_dir={'vispy': 'vispy'},
     data_files=[

```

djhoese:
Done. Thank you.

larsoner:
Works great, thanks @djhoese

Issue #1787, opened 2019-12-12
QSocketNotifier: Multiple socket notifiers for same socket 1864 and type Read


hmaarrfk:
When using spyder (3.x and 4), with qt5 as the plotting backend, and vispy with qt5, one gets this mind wrecking warning

```
Python 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 22:01:29) [MSC v.1900 64 bit (AMD64)]
Type "copyright", "credits" or "license" for more information.

IPython 7.10.1 -- An enhanced Interactive Python.

%gui qt

import vispy

vispy.use('pyqt5')
WARNING: QSocketNotifier: Multiple socket notifiers for same socket 1864 and type Read
WARNING: QSocketNotifier: Multiple socket notifiers for same socket 1864 and type Read

print('hello')
hello
WARNING: QSocketNotifier: Multiple socket notifiers for same socket 1864 and type Read

```

that keeps popping up, they consistently pop up, and never stop, really annoying and hinders workflow.

https://vimeo.com/379109772

djhoese:
What about from vanilla ipython in the command line? What about doing the `print` before importing vispy in spyder or vanilla ipython?

hmaarrfk:
That print statement just keeps popping up, at a semi periodic interval.

It is definitely related to these lines of code:

```
try:
    QtCore.qInstallMsgHandler(message_handler)
except AttributeError:
    QtCore.qInstallMessageHandler(message_handler)  # PyQt5
```

hmaarrfk:
what is the harm, for myself, as a quick hot fix, to remove these lines from my distribution of vispy while we find a long term fix?

This is a pretty serious impedement in my team's workflow.

hmaarrfk:
I think it is because spyder installs its own message handler:
https://github.com/spyder-ide/spyder/blob/master/spyder/app/mainwindow.py#L286

@ccordoba12 any ideas here?

ccordoba12:
You can blacklist the QSocketNotifier message in our handler, I think.

djhoese:
FYI: https://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler

hmaarrfk:
I think that documentation is bogus. That documentation states that it should be OK to overwrite the message handler, but my guess is that it is broken.

```%gui qt
# %%
from vispy.util import logger
from PyQt5 import QtCore


# Properly log Qt messages
# Also, ignore spam about tablet input
[...]

PR #1789, opened 2019-12-12
Filter away annoying qnotifier warning.

hmaarrfk:
Workaround for https://github.com/vispy/vispy/issues/1787

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -158,7 +158,6 @@ BUTTONMAP = {0: 0, 1: 1, 2: 2, 4: 3, 8: 4, 16: 5}
 
 
 # Properly log Qt messages
-# Also, ignore spam about tablet input
 def message_handler(*args):
 
     if qt_lib in ("pyqt4", "pyside"):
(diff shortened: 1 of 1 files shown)


djhoese:
Thoughts on a blacklist like spyder:

https://github.com/spyder-ide/spyder/blob/25a788288e33d7e210e3cf1647c57db10bf863f6/spyder/app/mainwindow.py#L278-L282

hmaarrfk:
seems a little over the top, but whatever gets this in faster ;)

hmaarrfk:
The logic for it is so complicated due to the backward compatibility with PyQt4, ;)

hmaarrfk:
Well blacklist logic added.

djhoese:
Thank you.

djhoese:
If this turns out to be something in QtConsole that gets fixed in the future (https://github.com/jupyter/qtconsole/issues/332) then we can remove this ignore since I assume it is trying to tell us something useful in the usual case.

hmaarrfk:
wow that was fast. Is it too much to ask for 0.6.4?

djhoese:
No I was actually wanting to do it for the `version.py` stuff I changed recently. Probably today or this weekend depending on how some of the rest of my work goes.

hmaarrfk:
All good, the workaround I provided in the post kinda works, although there is a slight chance of race condition, but pretty minor in the grand scheme of things!

PR #1792, opened 2019-12-15
Make all AxisVisual parameters easily updatable after instantiation

tlambert03:
I am starting to work with the `PlotWidget` a bit more for use in `napari`.  This is a small PR, partially for that and also to fix part of https://github.com/vispy/vispy/pull/1748#issuecomment-559789319.  It modifies `AxisVisual` to ensure that all of the parameters in the `__init__` method can be easily changed as attributes/properties after instantiation, and that changing them immediately updates the necessary subvisuals.

Rather than make a whole bunch of relatively empty `@property` and `@setter` methods just to set `_need_update` and call `self.update()` each time, I chose to modify the `__setatr__` method instead, calling `update()` for all parameters that wouldn't have otherwise induced a visual update.  (Note, parameters that are _not_ listed in there, such as `tick_width`, already update the visual without needing `update()`).  Let me know if you prefer the more explicit but verbose properties & setters.

There is one slight API change here: the `label_color` property is renamed to `text_color`, since that's what it was called in the function signature and class docstring.  Also added a couple more properties to access axis & tick widths/direction after instantiation.

### DIFF
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -75,7 +75,7 @@ class AxisVisual(CompoundVisual):
                  scale_type="linear", axis_color=(1, 1, 1),
                  tick_color=(0.7, 0.7, 0.7), text_color='w',
                  minor_tick_length=5, major_tick_length=10,
-                 tick_width=2, tick_label_margin=5, tick_font_size=8,
+                 tick_width=2, tick_label_margin=12, tick_font_size=8,
                  axis_width=3,  axis_label=None,
                  axis_label_margin=35, axis_font_size=10,
                  font_size=None, anchors=None):
@@ -88,6 +88,7 @@ class AxisVisual(CompoundVisual):
             tick_font_size = font_size
             axis_font_size = font_size
 
+        self._initialized = False
         self._pos = None
         self._domain = None
 
@@ -97,8 +98,7 @@ class AxisVisual(CompoundVisual):
         self._stop_at_major = (False, False)
 
         self.ticker = Ticker(self, anchors=anchors)
-        self.tick_direction = np.array(tick_direction, float)
-        self.tick_direction = self.tick_direction
+        self.tick_direction = tick_direction
         self.scale_type = scale_type
 
         self.minor_tick_length = minor_tick_length  # px
@@ -124,13 +124,14 @@ class AxisVisual(CompoundVisual):
         if pos is not None:
             self.pos = pos
         self.domain = domain
+        self._initialized = True
 
     @property
-    def label_color(self):
+    def text_color(self):
         return self._text.color
 
-    @label_color.setter
-    def label_color(self, value):
+    @text_color.setter
+    def text_color(self, value):
         self._text.color = value
         self._axis_label.color = value
 
@@ -143,6 +144,14 @@ class AxisVisual(CompoundVisual):
         self._line.set_data(color=value)
 
     @property
+    def axis_width(self):
+        return self._line.width
+
+    @axis_width.setter
+    def axis_width(self, value):
+        self._line.set_data(width=value)
+
+    @property
     def tick_color(self):
         return self._ticks.color
 
@@ -151,12 +160,44 @@ class AxisVisual(CompoundVisual):
         self._ticks.set_data(color=value)
 
     @property
+    def tick_width(self):
+        return self._ticks.width
+
+    @tick_width.setter
+    def tick_width(self, value):
+        self._ticks.set_data(width=value)
+
+    @property
     def tick_font_size(self):
         return self._text.font_size
 
     @tick_font_size.setter
     def tick_font_size(self, value):
         self._text.font_size = value
+
+    @property
+    def tick_direction(self):
+        return self._tick_direction
+
+    @tick_direction.setter
+    def tick_direction(self, value):
+        self._tick_direction = np.array(value, float)
+
+    def __setattr__(self, name, val):
+        super().__setattr__(name, val)
+        if self._initialized:
+            if name in (
+                "minor_tick_length",
+                "major_tick_length",
+                "tick_label_margin",
+                "axis_label_margin",
+                "tick_direction",
+                "axis_label",
+                "pos",
+                "scale_type", 
+            ):
+                self._need_update = True
+                self.update()
 
     @property
     def axis_font_size(self):
@@ -173,8 +214,6 @@ class AxisVisual(CompoundVisual):
     @axis_label.setter
     def axis_label(self, axis_label):
         self._axis_label_text = axis_label
-        self._need_update = True
-        self.update()
 
     @property
     def pos(self):
@@ -183,8 +222,6 @@ class AxisVisual(CompoundVisual):
     @pos.setter
     def pos(self, pos):
         self._pos = np.array(pos, float)
-        self._need_update = True
-        self.update()
 
     @property
     def domain(self):
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -8,7 +8,7 @@ import math
 
 import numpy as np
 
-from .visual import CompoundVisual
+from .visual import CompoundVisual, updating_property
 from .line import LineVisual
 from .text import TextVisual
 
@@ -88,7 +88,6 @@ class AxisVisual(CompoundVisual):
             tick_font_size = font_size
             axis_font_size = font_size
 
-        self._initialized = False
         self._pos = None
         self._domain = None
 
@@ -98,15 +97,15 @@ class AxisVisual(CompoundVisual):
         self._stop_at_major = (False, False)
 
         self.ticker = Ticker(self, anchors=anchors)
-        self.tick_direction = tick_direction
+        self.tick_direction = np.array(tick_direction, float)
         self.scale_type = scale_type
 
-        self.minor_tick_length = minor_tick_length  # px
-        self.major_tick_length = major_tick_length  # px
-        self.tick_label_margin = tick_label_margin  # px
-        self.axis_label_margin = axis_label_margin  # px
+        self._minor_tick_length = minor_tick_length  # px
+        self._major_tick_length = major_tick_length  # px
+        self._tick_label_margin = tick_label_margin  # px
+        self._axis_label_margin = axis_label_margin  # px
 
-        self._axis_label_text = axis_label
+        self._axis_label = axis_label
 
         self._need_update = True
 
@@ -117,14 +116,13 @@ class AxisVisual(CompoundVisual):
                                  color=tick_color)
 
         self._text = TextVisual(font_size=tick_font_size, color=text_color)
-        self._axis_label = TextVisual(font_size=axis_font_size,
-                                      color=text_color)
+        self._axis_label_vis = TextVisual(font_size=axis_font_size,
+                                          color=text_color)
         CompoundVisual.__init__(self, [self._line, self._text, self._ticks,
-                                       self._axis_label])
+                                       self._axis_label_vis])
         if pos is not None:
             self.pos = pos
         self.domain = domain
-        self._initialized = True
 
     @property
     def text_color(self):
@@ -133,7 +131,7 @@ class AxisVisual(CompoundVisual):
     @text_color.setter
     def text_color(self, value):
         self._text.color = value
-        self._axis_label.color = value
+        self._axis_label_vis.color = value
 
     @property
     def axis_color(self):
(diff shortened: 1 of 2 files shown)


vispy/visuals/axis.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, pos=None, domain=(0., 1.), tick_direction=(-1., 0.),
                  scale_type="linear", axis_color=(1, 1, 1),
                  tick_color=(0.7, 0.7, 0.7), text_color='w',
                  minor_tick_length=5, major_tick_length=10,
-                 tick_width=2, tick_label_margin=5, tick_font_size=8,
+                 tick_width=2, tick_label_margin=12, tick_font_size=8,
```
djhoese: Was this intended for this PR?

tlambert03:
Yeah, I'm a little conflicted about the approach myself.  I agree it's harder to follow... probably not worth it.  I like the decorator idea!  let me think about that for a while.

I couldn't think of any reason not to update the plot when changing tick properties... but if you think that should be omitted from the update cycle, that's fine with me.  (but: all tick props?  just length? direction?)

vispy/visuals/axis.py
tlambert03: i do think 12 should be the default (since the `major_tick_length` defaults to 10, I think the labels will almost always [be on top of the ticks](https://github.com/vispy/vispy/issues/1742#issuecomment-549382586) with a margin of 5).  but the change is a bit out of place here.  what do you think?  leave it in, or consider a different PR?

djhoese: I am OK with this staying here. There is a very good chance people are already changing this themselves to make it look better.

djhoese:
About the tick properties, I was just noticing that things like `tick_font_size` is not part of the list to cause an update.

djhoese:
Wait, what is `self._text`? The tick labels?

As for decorator versus other methods, I'm not sure a decorator will actually work but maybe I haven't thought about it enough. The other option would be to put them in the base Visual class like I mentioned. This may be our way to simplify a lot of the ugliness with properties in Vispy. A base class with methods for defining what properties to make. Like a CompoundVisual that knows what properties to make available for its sub-visuals. Ok...I need to stop brainstorming on this. One thing at a time.

tlambert03:
this really underscores just how uninterpretable my implementation is! 🤪
setting `tick_font_size` actually _does_ update the graph already, without being specified in `__setattr__` because, as you noticed next, the `AxisVisual.tick_font_size` setter is actually passing on the value to the `self._text` (`TextVisual` subvisual) `@font_size.setter` method... and that _does_ call `update()`...  so I left it out from the `__setattr__` list so as not to duplicate the update.

The problem here is a little bit systemic... there's no obvious consensus on which properties should auto-update their own object.  That seems quite hard really!  I doubt we _always_ want properties to update their objects...
I certainly don't claim to know how best to approach that... so this PR is sort of a "cop-out", focusing only on the `AxisVisual` for what is almost certainly an issue that would pop up elsewhere.

In any case, I think I'm convinced that the `__setattr__` approach is too uninterpretable.  The actual behavior is working correctly, but the approach needs to be rethought.  The quickest approach would be to just fall back to `@property` for the existing attributes that don't have `_private` values (such as `tick_label_margin`, and `axis_label_margin`), and leave a better and more pervasive fix for another PR.  I'm not in a hurry though.

codypiersall:
Vispy could implement a custom descriptor for things that should update when being set:

```python3
class UpdatingProperty(property):
    """Updates the visual after setting the attribute"""
    def __set__(self, instance, value):
        print('beep')
        super().__set__(instance, value)
        instance._needs_update = True
        instance.update()
```

Since it inherits from `property` the usage would be the same.

tlambert03:
i like that!
Seems like a great intermediate solution (short of modifying the base `compoundVisual` class).  @djhoese?

djhoese:
That looks like a good solution. Thanks @codypiersall. @tlambert03 Maybe you could add a version of that property class to `vispy/visuals/visual.py` and add a section to the module docstring about using it in custom visual classes? You could mention it being an experimental "feature" or something.

I like this in the sense of moving towards an easier to implement custom Visual class. The only thing this doesn't implement is a corresponding `_`-prefixed instance attribute, but maybe that's ok? There are some classes that I know of right now that use the `_` attribute to set something without updating, but we can probably work around that in the future if we want this to become the new go-to way of property-ifying a custom visual class. Over the past couple years I would say besides "vispy is slow because I'm using it wrong" issues, the next most popular questions are "how can I change this property of a visual?" where my answer has to be "sorry, that's not implemented". This could be a step in simplifying that.

Oh! This should let us do `isinstance` stuff too if needed in the future ("is this a property that updates the visual?").

CC @almarklein @larsoner @kmuehlbauer

tlambert03:
yeah, I think it's got a lot of potential.   Before actually updating the PR, here are a couple strategies that you can consider.

1. The first is just like @codypiersall  suggested:
    ```python
    class UpdatingProperty(property):
        """Updates the visual after setting the attribute"""
        def __set__(self, instance, value):
            super().__set__(instance, value)
            instance._needs_update = True
            instance.update()
    ```
    in which case usage would be a slightly verbose, like this:
    ```python
    class SomeVisual:
        def __init__(self, someprop=None):
            self._someprop = someprop

        @UpdatingProperty
        def someprop(self):
            return self._someprop
        
        @someprop.setter
        def someprop(self, value):
            self._someprop = value
    ```
2. The second option puts more of the logic in the descriptor itself, automatically handling the corresponding private attribute and `setter` methods.
    ```python
    class UpdatingProperty:
        """A property descriptor that autoupdates the Visual during setting"""

        def __init__(self, fget, doc=None):
            self.attr_name = f'_{fget.__name__}'
            self.__doc__ = doc or fget.__doc__

        def __get__(self, obj, objtype=None):
            if obj is None:
                return self
            return getattr(obj, self.attr_name)

        def __set__(self, obj, value):
            setattr(obj, self.attr_name, value)
            obj._needs_update = True
            obj.update()  # could prevent update during obj.__init__

        def __delete__(self, obj):
            raise AttributeError("can't delete attribute")
    ```
    in which case usage looks like this
    ```python
    class SomeVisual:
        def __init__(self, someprop=None):
            self.someprop = someprop

        @UpdatingProperty
        def someprop(self):
            """docstring all that is needed"""
    ```

not necessarily endorsing either approach, but curious if you have thoughts.
(note, in option 2, `UpdatingProperty` could still implement a `setter` method that could override the default behavior if we wanted it to behave more like a regular `@property` with the option for `@someprop.setter`).

djhoese:
Oh I was hoping something like 2 was possible. So you're saying that if the setter had to have complicated logic (like converting from the provided data to something more complex) it could be implemented like normal (`@my_prop.setter`) and then call the `super` method?

My biggest issue with option 2 is that it takes away from the normal python property creation. But I like the auto-`_` attribute. I'll sleep on it.

tlambert03:
> Oh I was hoping something like 2 was possible. So you're saying that if the setter had to have complicated logic (like converting from the provided data to something more complex) it could be implemented like normal (`@my_prop.setter`) and then call the `super` method?

yep.  I actually learned a lot here, [this SO post](https://stackoverflow.com/questions/17330160/how-does-the-property-decorator-work/17330273#17330273) was particularly interesting in understanding the construction of the `property` decorator and descriptors.

> My biggest issue with option 2 is that it takes away from the normal python property creation. But I like the auto-`_` attribute. I'll sleep on it.

I agree.  It's a bit "magical" and not immediately obvious what's going on.

vispy/visuals/visual.py
```diff
@@ -678,3 +679,76 @@ def __init__(self, visual):
...
+
+
+class updating_property:
+    """A property descriptor that autoupdates the Visual during attribute setting.
+
+    Use this as a decorator in place of the @property when you want the attribute to trigger
+    an immediate update to the visual upon change.
+    For example, the following code examples are equivalent:
+
+    class SomeVisual1:
```
djhoese: Could you update these examples to base themselves off of the BaseVisual class (or whatever the actual name is)? Just in case in the future this class depends more on actual BaseVisual functionality we know it is documented here. Thoughts?

vispy/visuals/axis.py
```diff
@@ -175,64 +173,53 @@ def tick_font_size(self):
...
     @pos.setter
     def pos(self, pos):
         self._pos = np.array(pos, float)
 
-    @property
-    def domain(self):
-        return self._domain
+    @updating_property
+    def minor_tick_length(self):
+        """The length of minor ticks, in pixels"""
```
djhoese: This (an empty method) is probably the worst side effect of this, but I'm still OK with it. Looks odd at first glance though.

vispy/visuals/visual.py
tlambert03: yeah, can do.  I'm actually reviewing the inheritance chain for Visuals and how they draw themselves right now... cause I'm still not 100% sure just how generally this decorator actually is.  (That is, I don't know how widely used that `_need_update` and `.update()` pattern is in the Visuals... and how many other visuals would actually use this.).  Once I do, I'll try to update with a few more modular commits that implement this pattern in a couple more Visuals.

vispy/visuals/axis.py
tlambert03: yeah I agree.  It's definitely the worst part.  looks odd and requires additional research to understand it.

vispy/visuals/visual.py
tlambert03: actually... can you help me on that?  It's still not clear to me why `Visual.update()` even does anything.  From what I have found, it will ultimately call `BaseVisual.update()` which simply emits the update event (`BaseVisual.events.update()`)... but I am unable to find anywhere in vispy where that event is actually connected (except in `CompoundVisual`, where it simply is connected to calling update on the subvisuals).  How does `BaseVisual.update()` eventually trigger the program to draw?

djhoese: Give me a bit and I'll try to find the parts you're looking for, but I would keep this PR simple with only the necessary changes (don't change other visuals).

djhoese: So without looking too much harder (I always forget the exact calling sequence), the individual visual `update` calls should trigger the call on `Canvas.update` which ends up calling `_vispy_update` on the backend. For some backends like Qt this is just a redraw (which I think is also called `update` in Qt).

I can't find the connection between the update event and the canvas update either...

djhoese: Ok, I feel a little less dumb now. When working with a SceneCanvas the important piece of code is the `Node` base class which is used for all `vispy.scene.visuals` versions of the Visual objects:

https://github.com/vispy/vispy/blob/9dbe751dcdbaaa1d17d0a5885f15872b9b0741a8/vispy/scene/node.py#L325-L333

So that's what gets called when visuals do `self.update` on a Visual in a SceneCanvas. In a normal `Canvas` situation though there is nothing that automatically connects `visual.events.update` to any other methods. I found examples that did this themselves (connect the update event to Canvas.update).

So in summary, `Node` overrides the default `update` method for the Visual objects.

tlambert03: aha!  thanks!  

OK, I won't make any changes to other visuals in this PR, but I do want to just look around a bit to make sure that this isn't introducing a whole new API for creating "updatable" visuals that really doesn't apply very far beyond the `AxisVisual`.

djhoese:
@tlambert03 What's the current status of this?

tlambert03:
So sorry for dropping this.  I think it’s basically ready... just still wasn’t sure if I felt like it was the “right” way to do it.  I’ll look again later today

djhoese:
No problem. I was asking because I completely forgot and dropped this too :wink:

tlambert03:
yeah @djhoese, this is "ready" in the sense that it achieves what the PR set out to do: all of the parameters in the `AxisVisual.__init__` method can be changed as attributes/properties after instantiation, and the change will be reflected on the canvas.   I think the remaining decisions are more "executive" decisions for vispy core devs to make (or request changes) and come down to this:

- While the `@updating_property` descriptor introduced here is nice, and will be usable for other visuals, as it stands after this PR, it's a little bit of an "orphaned" pattern.  It will take a little bit of doing (or someone who is very familiar with it) to dig through the rest of the codebase to find places where this pattern could be used.  (i.e. are you ok with a slightly orphaned pattern).
- one of the consequences of this pattern is that `@updating_property` getters _do_ need to be declared (in order for them to autoupdate), but the body of the methods needn't be populated.  leaving slightly strange looking "empty" methods.  I added an additional note to the docstring of `updating_property` to make that even clearer.

So this is ready to go, but let me know if you want anything else addressed...

djhoese:
Thanks for the update. I'm ok with the way this is and having this dangling pattern. I may regret this for being "too clever for my own good" and I'm not sure I suggested this already, but would it be possible to have the base Visual class loop over all of its attributes to see if a property is a `updating_property` and create a `self._property_name = None` if it doesn't already exist (ex. `setattr(self, '_property_name', None)`)?

Lastly, do you have an example of the `tick_label_margin` change and the difference that makes?

tlambert03:
> would it be possible to have the base `Visual` class loop over all of its attributes to see if a property is a updating_property and create a self._property_name = None if it doesn't already exist?

it _is_ possible, but it might be even uglier than you anticipated.  As a [descriptor](https://docs.python.org/3/howto/descriptor.html), the actual updating property instance is a _class_ attribute, whereas the private `_property_name` attribute is on the object.  It would look something like this:
```python
class BaseVisual(Frozen):
    def __init__(self, vshare=None):
        ...
        for attr in dir(self):
            # the updating_property instance actually lives on the object class
            class_attr = getattr(type(self), attr, None)
            if class_attr and isinstance(class_attr, updating_property):
                # the private attribute lives on the on the object instance
                if not hasattr(self, f"_{attr}"):
                    setattr(self, f"_{attr}", None)
```

I'm not sure I love it though ... It's a bit confusing, and in most cases, the attribute is going to be set in the init method of the subclass anyway.  For instance, adding this there wouldn't change anything for the `AxisVisual`, since there are no `self._property_name = None` declarations in `AxisVisual.__init__`.

> Lastly, do you have an example of the tick_label_margin change and the difference that makes?

As you've said, this might be platform specific, but for me `examples/demo/plot/plot.py` before:

Screen Shot 2020-03-23 at 11 36 00 AM
and after:
Screen Shot 2020-03-23 at 11 36 18 AM

djhoese:
Awesome! Thanks.

Maybe it isn't worth it to have that for loop, but if it is wrapped up in a separate hidden method then it isn't too bad. It is a little to implicit and not explicit so I can see why not putting it in would be smarter. My thought was that using the decorator is already a little magical so what's one more thing. Plus, that way you could have the property by itself and not have to worry about putting stuff in other parts of the code. Anyway, looks good then.

tlambert03:
your call.  for me, it's a bit too magical and implicit (and I think I have a relatively high tolerance for magic 😂)... and I'm inclined to leave it out until there is an immediate use case.  At least the code is there in the comment of this PR, in case we want to revisit it later.

djhoese:
@kmuehlbauer @larsoner What are your thoughts on this PR? Part of me doesn't think it is that big of a deal to include this interface that very few people will see, but another part would rather see this separated out to another PR where it gets used by multiple visuals to drive the point home that it is used and how it works. That said, I'm also open to the idea that this interface was me and @tlambert03 being too magical and it shouldn't be implemented/used.

tlambert03:
> I'm also open to the idea that this interface was me and @tlambert03 being too magical and it shouldn't be implemented/used.

I'm open to that as well, and if preferred, I'm happy to add the originally desired functionality just to the `AxisVisual` with more verbose/less-magical syntax.

kmuehlbauer:
@djhoese @tlambert03 This looks very nice. 

One concern is that this might affect performance if many of these setter functions are called in short succession resulting in an update for each. If we want to go this road we would need to think about some switch at visual level (like for .visible) which decides if this setting should do the update or just wait. I think this just makes sense in user interactive mode, so we could name this `interactive`. Does this make sense?

djhoese:
> One concern is that this might affect performance if many of these setter functions are called in short succession resulting in an update for each.

I would agree except this is how most (all?) properties currently work. A lot of properties are setup as:

```
@property
def some_prop(self):
    return self._some_prop

@some_prop.setter
def some_prop(self, val):
    self._some_prop = val
    self.update()
```

As part of this PR it moves this pattern in to a base class (or attempts to). In theory the event emitter stuff is supposed to allow you to "block" events so that redraws don't happen. But I can never remember how to do it and it doesn't always work (iirc). Something like:

```
with canvas.events.block('draw'):
    my_visual.some_prop = 5
```

That's not what it actually is, but I don't have time to look up the real way. That's the idea of how the existing framework is supposed to work. I don't remember if this is SceneCanvas-specific either.

kmuehlbauer:
@djhoese OK, then we should go for it and merge. We can think about possible performance issues (if any) later.

vispy/visuals/axis.py
larsoner: the decorator makes it pretty clear that it must be (is the only option as) the culprit so this seems fine

tlambert03:
thanks @larsoner!

djhoese:
Oh boy. Seems like we merge it and see how it involves in the future. Thanks again @tlambert03 for dealing with the back and forth on this.

Issue #1795, opened 2019-12-19
Unpin numpy in pyproject.toml


anntzer:
Currently, pyproject.toml pins the supported numpy for Py3:
```
    "numpy==1.13.3; python_version=='3.5'",
￼    "numpy==1.13.3; python_version=='3.6'",
￼    "numpy==1.14.5; python_version>='3.7'",
```

Locally (Arch Linux Py38, system python), this causes a build failure in a clean venv:
```
Processing /tmp/foo/vispy
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /tmp/testenv/bin/python /tmp/testenv/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmpe2zfosrb
       cwd: /tmp/pip-req-build-yhub7l48
  Complete output (42 lines):
  Traceback (most recent call last):
    File "/tmp/pip-build-env-r9nw4ay2/overlay/lib/python3.8/site-packages/numpy/core/__init__.py", line 16, in <module>
      from . import multiarray
  ImportError: libopenblas.so.0: cannot open shared object file: No such file or directory
  
  During handling of the above exception, another exception occurred:
  
  Traceback (most recent call last):
    File "/tmp/testenv/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py", line 257, in <module>
      main()
    File "/tmp/testenv/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py", line 240, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/tmp/testenv/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py", line 91, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/tmp/pip-build-env-r9nw4ay2/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 146, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=['wheel'])
    File "/tmp/pip-build-env-r9nw4ay2/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 127, in _get_build_requires
      self.run_setup()
    File "/tmp/pip-build-env-r9nw4ay2/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 236, in run_setup
      super(_BuildMetaLegacyBackend,
[...]

PR #1796, opened 2019-12-19
Unpin numpy for Py>=3.7.

anntzer:
See discussion at #1795.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ requires = [
     "numpy; python_version=='2.7'",
     "numpy==1.13.3; python_version=='3.5'",
     "numpy==1.13.3; python_version=='3.6'",
-    "numpy==1.14.5; python_version>='3.7'",
+    "numpy>=1.14.5; python_version>='3.7'",
 ]
 
 [tools.setuptools_scm]


djhoese:
Just to double check, this fixes the library issues on your system?

anntzer:
yes

djhoese:
Thanks for the quick fix for this!

anntzer:
thanks for the quick review :)

hmaarrfk:
Do you compile C code with numpy?
If so, you should not leave the numpy pinning loose.

If you don't, you should likely loosen the pinning on all the other versions of numpy as well.

The correct fix is to follow something like scipy's pinning
https://github.com/scipy/scipy/blob/master/pyproject.toml

IMO, pyproject.toml is just asking for trouble. I always use the `--no-build-isolation` flag anyway.

anntzer:
I guess that would work too?
I personally don't like the pyproject.toml approach either but I've given up on fighting that, I just want things to compile fine locally :)

hmaarrfk:
> I guess that would work too?

it isn't so much that it would work, it is that without the pinning, the package might not install.

If you for example have numpy 1.15, and compile C code with numpy 1.17+, as pyproject.toml would create for you, it may not work due to how the C-API changes.

https://github.com/scikit-image/scikit-image/issues/4261#issuecomment-557415094

You probably want the line:

```
 pip config set no-build-isolation False
```
yes, False is the actual line to set.

djhoese:
@hmaarrfk Thanks for the insight on this. Based on what you've said it sounds like a loose numpy dependency will always be a problem since `pip` will always build/install C dependencies against the newest numpy but may not update the version of numpy in your final environment *or* of the C libraries built from other dependencies.

As far as pyproject.toml goes, as far as I understand it is the recommended way of doing build dependencies for the pip community. If there are better suggestions, let me know, but this was the easiest solution at the time.

hmaarrfk:
It may be recommended way, but I personally don't like it.

I've experienced the value in having separate build control and source control, and pyproject encourages the combination, which isn't always advantageous.

djhoese:
> having separate build control and source control

What are your preferences for a python package with C extensions?

hmaarrfk:
So pyproject.toml avoids users having the command:

```
pip install -e . -vv
```

fail when because numpy installed and `setup.py` requires it before being parsed. But it doesn't ensure consistency with the isolated environment and the users' currently virtual envinronment.

I would rather users error on that command, with a non-descript message, and be forced to read the development documentation.

I don't think pyproject.toml doesn't actually help any users install from source on pypi. That often requires other system dependencies, compilers, correctly configured path, as well as obviously working with unsupported versions of the python interpretter and endianess. 

Anyway, scipy and numpy started using it, so my opinion is much less valid than theirs, but I personally have this:
```
$ cat ~/.config/pip/pip.conf 
[global]
no-build-isolation = False

# [install]
# no-dependencies = yes
```

essentially disabling `pyproject.toml` for all the projects I hack on.

PR #1799, opened 2019-12-22
Updated pinnings for python 3.8

hmaarrfk:


hmaarrfk:
It seems that this is important for
https://github.com/vispy/vispy/blob/master/vispy/visuals/text/_sdf_cpu.pyx

djhoese:
So what I'm understanding from the original PR and issue (#1799 and #1796) is that the main issue regarding numpy versions is if you have a loose definition you can have some packages built with one version and other packages built with another, but that won't determine what version is actually installed in the environment.

This looks good to me. I'll merge this next time I'm online and give others time to comment if they'd like.

hmaarrfk:
You want to ensure that all packages were built with older versions of numpy than the one that is ultimately installed in the users enviroent.

djhoese:
Thanks for this. Sorry I lost track of it.

hmaarrfk:
It was meant to be easy to pull in when you found time.

Thanks for finding it!

No issue on the delay

PR #1802, opened 2020-01-01
Remove isosurface green color in VolumeVisual

sofroniewn:
This PR improves the isosurface rendering in the volume visual by getting rid of the default greenish tinge that was always being applied and allowing the isosurface to also respond to the changing of the colormaps. You can see it in action here https://github.com/napari/napari/pull/840 and below.

Original behavior:
orig_iso_surface

Improved behavior:
better_iso_surface

I realize this change will change how peoples renderings currently look, but I hope the majority of people will find the new behavior and improvement.

### DIFF
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -7,19 +7,19 @@ About this technique
 --------------------
 
 In Python, we define the six faces of a cuboid to draw, as well as
-texture cooridnates corresponding with the vertices of the cuboid.
+texture cooridnates corresponding with the vertices of the cuboid. 
 The back faces of the cuboid are drawn (and front faces are culled)
-because only the back faces are visible when the camera is inside the
+because only the back faces are visible when the camera is inside the 
 volume.
 
-In the vertex shader, we intersect the view ray with the near and far
+In the vertex shader, we intersect the view ray with the near and far 
 clipping planes. In the fragment shader, we use these two points to
 compute the ray direction and then compute the position of the front
 cuboid surface (or near clipping plane) along the view ray.
 
 Next we calculate the number of steps to walk from the front surface
 to the back surface and iterate over these positions in a for-loop.
-At each iteration, the fragment color or other voxel information is
+At each iteration, the fragment color or other voxel information is 
 updated depending on the selected rendering method.
 
 It is important for the texture interpolation is 'linear', since with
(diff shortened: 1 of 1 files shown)


sofroniewn:
> Does the Volume visual require that textures be RGB right now? What if I have a single band texture (L)?

I think the Volume visual right now actually requires (L) currently and this PR won't change that. It might be nice to extend support for the Volume visual to RGB, but that would be followup PRs.

My guess is that `$sample(u_volumetex, iloc)` actually returns a `vec4` where `.g`, `.r`, `.b` are all the same for the (L) data and so

> could you explain why the shaders here use the .g color instead of .r

I think this choice is arbitrary right now because of the note above. I have tried `.r` and `.b` and it gave identical visual performance for the data I used.

> Do you see any side effects popping up from this?

The use of `.g` is also present in the original code (now just on a different line) and so I don't think we'll be causing any unexpected side effects here.

> The other big change being changing the "u_ambient" to be the same between R, G, and B. Any idea why this may have been done this way? Aren't humans more sensitive to green?

I don't really know about this one. I havn't noticed any other changes in rendering by changing this, and I think it makes more sense to have a gray default, but we could make it a property too. I think the mesh visual has an [ambient_light_color](http://vispy.org/visuals.html#vispy.visuals.MeshVisual.ambient_light_color) property

djhoese:
>> Do you see any side effects popping up from this?
> 
> The use of .g is also present in the original code (now just on a different line) and so I don't think we'll be causing any unexpected side effects here.

I was more thinking about the order of getting the colormap first and then getting the "best color". I'm not sure what that best color function is doing so I wanted to make sure it was intentional by you to switch the order that these are being called.

sofroniewn:
> I'm not sure what that best color function is doing so I wanted to make sure it was intentional by you to switch the order that these are being called.

I agree it's a little odd right now because the `calculateColor` code is really just being used to transform a luminance before it gets passed to the colormap, but I think it works well in practice. There might be a way to keep that order and get similar (even better behaviour), which feels more natural, but in practice I was just getting saturation and everything was looking white instead of colored. The `calculateColor` code is also needed to give the surface the nice shinny 3D look - otherwise it looks flat.

djhoese:
Looks like @almarklein added the original Volume visual. @almarklein Any idea why the hardcoded ambient light had a green component of 0.4 while red and blue had 0.2?

almarklein:
> Any idea why the hardcoded ambient light had a green component of 0.4 while red and blue had 0.2?

I have no idea. I think/fear that this is one of the loose ends that we left behind  😢

djhoese:
Coming back to this after a decent amount of time and I fear I don't fully remember every concern I had. Based on the last couple comments and what I do remember, I think all the changes are good but maybe the code could be cleaned up. @sofroniewn I'm guessing you've been using this with napari for a while now? How has it been going? Any issues?

sofroniewn:
We've been using it fine @djhoese - it's ready as far as I'm concerned. Let me know if you want any changes.

PR #1803, opened 2020-01-01
Add nearest interpolation to volume visual

sofroniewn:
This PR adds an `interpolation` property to the Volume visual and exposes a `nearest` and `linear`. It defaults to the `linear` behavior, which is what is currently supported, but provides the option for a `nearest` - which is particularly useful when viewing discrete (say integer labeled) regions in 3D.

Before with `linear`:
orig_linear_interpolation

Now with `nearest`:
better_3D_labels

You can see these in use here https://github.com/napari/napari/pull/841 too

This PR adds documentation for the new property and a test for it.

### DIFF
--- a/vispy/visuals/tests/test_volume.py
+++ b/vispy/visuals/tests/test_volume.py
@@ -35,8 +35,8 @@ def test_volume():
     assert V.method == 'iso'
 
     # Interpolation
-    V.method = 'nearest'
-    assert V.method == 'nearest'
+    V.interpolation = 'nearest'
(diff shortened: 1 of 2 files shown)


sofroniewn:
Sounds good! I just updated variable name internal to a function call from `i` -> `interp` for more readability

djhoese:
The tests were failing but this shouldn't be your fault and should be fixed on master. I'll merge it and if they end up failing on master...then I'll deal with it.

sofroniewn:
Thanks!!

PR #1805, opened 2020-01-07
MAINT: GLFW workaround

larsoner:
Hopefully makes GLFW 3.3.1 tests work. I think they might have some bug with `decorate` as not setting the value makes things pass on my system (at least on 3.4.0 dev that I installed). Probably worth filing an upstream bug if this actually works.

EDIT: Also replaced Travis custom setup with cleaner `ci-helpers` interface
EDIT: Also silenced some errors about bad division in `axis` tests
EDIT: Also configured Travis's runs to use `--tb=short` to shorten tracebacks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
 
     # also tests file sizes, style, line endings
     - env: PYTHON_VERSION=3.7 DEPS=minimal TEST=standard
-           CONDA_DEPENDENCIES=numpy scipy pytest cython coveralls pytest-cov pytest-sugar pytest-faulthandler flake8
+           CONDA_DEPENDENCIES="numpy scipy pytest cython coveralls pytest-cov pytest-sugar pytest-faulthandler flake8"
            CONDA_CHANNEL_PRIORITY=strict
            PIP_DEPENDENCIES="numpydoc"
       addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,20 +13,20 @@ addons:
 matrix:
   include:
     - env: PYTHON_VERSION=3.7 DEPS=minimal TEST=standard
-           CONDA_DEPENDENCIES="numpy scipy pytest cython coveralls pytest-cov pytest-sugar pytest-faulthandler flake8"
+           CONDA_DEPENDENCIES="numpy scipy pytest cython coveralls pytest-cov pytest-sugar"
            CONDA_CHANNEL_PRIORITY=strict
       os: osx
       osx_image: xcode10.1
 
     - env: PYTHON_VERSION=3.7 DEPS=full TEST=standard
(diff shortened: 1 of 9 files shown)


larsoner:
Also removed the use of `np.testing.dec` in favor of `pytest.mark.skipif` since the `numpy` one requires `nose`.

larsoner:
Okay I have to stop now. I'm not sure what else to do since GLFW tests now pass locally. I'll try a bit more tomorrow, but in the meantime this has some changes that should make the test suite stuff a bit cleaner if you want to merge @djhoese (assuming the latest commit at least gets us back to the success rate of `master`).

Also if you have ideas feel free to push

kmuehlbauer:
Pinging @hmaarrfk as per gitter chat. @hmaarrfk [this build](https://travis-ci.org/vispy/vispy/jobs/633507859?utm_medium=notification&utm_source=github_status) fails on master (using new glfw 3.3.1). [This build before](https://travis-ci.org/vispy/vispy/jobs/631110057?utm_medium=notification&utm_source=github_status) works and uses glfw 3.3.0.

@larsoner tried to get things running on Travis, but without luck so far.

@hmaarrfk Any help much appreciated.

hmaarrfk:
i odn't know much about glfw, maybe look at the patches that were accepted in 3.3.1???

larsoner:
Filed an upstream bug after a `git bisect`:

https://github.com/glfw/glfw/issues/1620

Set this PR just to do `glfw!=3.3.1` for now. Really I'm not too optimistic that even if they fix my reported bug that things will work since with the current version of this PR things do pass on my system but not Travis, so there might be more than one bug at work. But we'll see!

@djhoese ready for review/merge from my end.

djhoese:
Thanks @larsoner for trying to figure this out. I've subscribed to your issue. We'll see where that discussion goes and I'll wait to merge this in a day or two. I'm not in a hurry to figure this out *right* now and am (unjustifiably) optimistic that a glfw patch will fix things for us.

larsoner:
> I'm not in a hurry to figure this out right now and am (unjustifiably) optimistic that a glfw patch will fix things for us.

Even if they do it will require them pushing a release and then conda updating. The time between 3.3 and 3.3.1 appears to have been 9 months. So I'm not holding my breath...

djhoese:
> appears to have been 9 months

Darn it.

kmuehlbauer:
If it is an easy fix, we could patch the conda-forge version (although this isn't recommended).

hoechenberger:
> If it is an easy fix, we could patch the conda-forge version (although this isn't recommended).

That would definitely be an option. I'm co-maintaining `glfw` on `conda-forge`, so just ping me if you need a quick action on this once you know how to fix the issue :)

djhoese:
Merging this now until glfw 3.3.2 can come up. Too many new PRs are coming out and failing. Thanks again @larsoner.

PR #1806, opened 2020-01-09
DOC: Drop installation instructions from readme

hoechenberger:
Refer to documentation on the website instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ js/dist
 node_modules/
 vispy/static/
 vispy/version.py
+.vscode
--- a/README.rst
+++ b/README.rst
@@ -72,46 +72,9 @@ and experimental OpenGL backend for matplotlib.
 Installation
 ------------
 
-VisPy runs on Python 2.7+ and Python 3.3+ and depends on NumPy. You also
-need a backend (PyQt4/PySide, PyQt5/PySide2, glfw, pyglet, SDL, or wx).
-
-PyQt5/PySide2 should be considered more experimental than PyQt4/PySide.
-
-VisPy can be installed either via `pip`::
-
-    pip install vispy
-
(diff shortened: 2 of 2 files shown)


djhoese:
Thanks! Any chance you could take a screenshot of the difference? Switching between github markdown and restructuredtext and slack I never remember what looks best.

hoechenberger:
Before:
Screen Shot 2020-01-09 at 21 33 17

After:
Screen Shot 2020-01-09 at 21 33 51

djhoese:
Interesting, it doesn't look that way on the generated docs: http://vispy.org/installation.html

hoechenberger:
@djhoese These docs appear to have been generated from https://github.com/vispy/vispy-website/blob/09586e86066f51547a29bd080771d7084e566c9e/installation.rst, and not from the readme file this PR is concerned with :)

hoechenberger:
Which makes me wonder why there's such a discrepancy _in content_ anyway …

djhoese:
:man_facepalming: 

Ok, so we're mixing markdown and rst in the README.

As for the "in content" comment, do you mean you think we should just point to the official docs in the README?

hoechenberger:
> As for the "in content" comment, do you mean you think we should just point to the official docs in the README?

That would probably make things easier to maintain, yes…

djhoese:
Agreed. @larsoner @kmuehlbauer any objections to having the README point to the main docs site for installation instructions?

kmuehlbauer:
@djhoese Yeah, sure, DRY is a good idea.

hoechenberger:
@djhoese WDYT?

djhoese:
Don't worry about the test failures. Those are being worked on in #1805.

djhoese:
Thanks again!

PR #1807, opened 2020-01-12
Fix LinePlotVisual not remembering styles

tlambert03:
Currently, if someone generates a line plot with no markers, for instance:
```python
line = scene.LinePlot((x,y), marker_size=0)
```
and then later updates the data with 
```python
line.set_data((x,y/2))
```
the markers come back.

This PR modifies the behavior of `LinePlotVisual` visual slightly to store the styles set during `__init__` or calls to `set_data`, such that styles do not need to be repeatedly given to `set_data` if one only wants to update the data itself.

### DIFF
--- a/vispy/visuals/line_plot.py
+++ b/vispy/visuals/line_plot.py
@@ -81,8 +81,14 @@ class LinePlotVisual(CompoundVisual):
             Keywoard arguments to pass to MarkerVisual and LineVisal.
         """
         # remember these kwargs for future updates
+        bad_keys = []
+        for key in kwargs:
+            if (key not in self._line_kwargs) and (key not in self._marker_kwargs):
+                bad_keys.append(key)
+        if bad_keys:
+            raise TypeError("Invalid keyword arguments: {}".format(",".join(bad_keys)))
+
         self._kwargs.update(kwargs)
-        kwargs = self._kwargs.copy()
         if data is None:
             pos = None
         else:
@@ -101,7 +107,7 @@ class LinePlotVisual(CompoundVisual):
 
                 # if both args and keywords are zero, then there is no
                 # point in calling this function.
-                if len(kwargs) == 0:
+                if len(self._kwargs) == 0:
                     raise TypeError("neither line points nor line properties"
                                     "are provided")
             elif pos.shape[1] == 1:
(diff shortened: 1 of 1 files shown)


tlambert03:
yeah, I didn't particularly love the look of that either.  but I guess I reasoned that, with the current logic in master, any kwargs that made it all the way to the end without being "`pop`ed" somewhere along the way were junk, and all the others were valid.  So updating `_kwargs` and then deleted unused ones seems logically sound, if not particularly attractive.

I could do the same logic upfront, by only updating `self._kwargs` with keys from `kwargs` that are also in `_line_kwargs` or `_marker_kwargs`.

To get rid of the `copy`, I think i'd need to chop up the existing method a bit more... because it currently relies on eating up `kwargs` (popping keys out along the way).  So we basically need to create an object that has the previous defaults, but which can also be destroyed, right?

djhoese:
Off the top of my head we/you could change the kwarg for loops from:

```
        marker_kwargs = {}
        for k in self._marker_kwargs:
            if k in kwargs:
                k_ = self._kw_trans[k] if k in self._kw_trans else k
                marker_kwargs[k_] = kwargs.pop(k)
```

to:

```
marker_kwargs = {}
for k in self._marker_kwargs:
    k_ = self._kw_trans.get(k, k)
    val = kwargs.get(k_, self._kwargs.get(k_))
    if val is not None:
        marker_kwargs[k_] = val
```

But now that I write that I see that this assumes that `None` is not the desired value that we want to set and also does multiple `get` calls regardless of whether or not the key is in the dictionary (compared to possibly short-circuited if statements).

tlambert03:
Hi again, let's see if we can get this one finished up :)
reminder: the only remaining issue here was that, in `LinePlotVisual.set_data`, I first updated the instance `self._kwargs` attribute, but only at the end of the method do we check to make sure that all of the `kwargs` provided to `set_data` were valid, which required me to then go back and remove the invalid ones from `self._kwargs`.
I just updated it to throw the invalid keyword arguments error at the beginning of the method instead, which avoids that.  (the method would have errored anyway, so might as well get it over with).  Let me know what you think

djhoese:
Thanks. The only negative to this approach is having to iterate over the keyword arguments multiple times. At this point you could probably do set operations `bad_keys = set(kwargs) - set(self._line_kwargs) - set(self._markers_kwargs)`. Just popped in to my head. If you like it better, go for it. Otherwise what you have is fine.

Travis OSX is having problems lately. Not your fault.

tlambert03:
i like sets much better, thanks.  I added a new `_valid_kwargs` class attribute, (since it never changes) so checking for bad keys should be very inexpensive with: `set(kwargs) - self._valid_kwargs`

djhoese:
Thanks again for this. The tests were failing for something that wasn't your fault. Merging now.

tlambert03:
thanks @djhoese!

PR #1821, opened 2020-02-18
Fix GitHub raw file download base URL

djhoese:
The VisPy cron job for Travis CI is failing on OSX. Not sure why specifically OSX, but while investigating I noticed that the GitHub raw file URL gets redirected. This PR changes VisPy's default download URL to be this new base URL. Let's see if the tests pass with it...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,17 +62,6 @@ matrix:
             - *mesa_apt
             - *full_apt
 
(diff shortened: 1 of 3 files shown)


djhoese:
Example of failing job: https://travis-ci.org/vispy/vispy/jobs/651736530

djhoese:
Well now Python 2.7 tests are failing. Since we aren't supporting python 2.7 I guess it is time to remove them.

PR #1825, opened 2020-02-22
Fix DataBuffer.set_subdata docstring with wrong offset units

asnt:
The `offset` is in "item count" unit not in "byte count" unit as is the case in the dtype-agnostic `Buffer` class.
The doc was probably copied over from the `Buffer` class without adaptation.

### DIFF
--- a/vispy/gloo/buffer.py
+++ b/vispy/gloo/buffer.py
@@ -162,7 +162,8 @@ class DataBuffer(Buffer):
         data : ndarray
             Data to be uploaded
         offset: int
-            Offset in buffer where to start copying data
+            Offset in buffer where to start copying data (i.e. index of
+            starting element).
         copy: bool
             Since the operation is deferred, data may change before
             data is actually uploaded to GPU memory.
--- a/vispy/gloo/buffer.py
+++ b/vispy/gloo/buffer.py
@@ -299,9 +299,9 @@ class DataBuffer(Buffer):
         elif data.size > stop - start:
             raise ValueError('Data too big to fit GPU data '
                              '(%d > %d-%d).' % (data.size, stop, start))
-        
+
         # Set data
-        offset = start  # * self.itemsize
+        offset = start
         self.set_subdata(data=data, offset=offset, copy=True)
 
     def __repr__(self):
--- a/vispy/gloo/buffer.py
+++ b/vispy/gloo/buffer.py
@@ -162,7 +162,7 @@ class DataBuffer(Buffer):
         data : ndarray
             Data to be uploaded
         offset: int
-            Offset in buffer where to start copying data (in bytes)
+            Offset in buffer where to start copying data
         copy: bool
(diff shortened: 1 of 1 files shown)


djhoese:
Don't we want this to be in item-space rather than byte-space?

asnt:
> Don't we want this to be in item-space rather than byte-space?

This is what I mean. The [computation of the offset](https://github.com/vispy/vispy/blob/bcec7fd11b11d870d29dee9aacbd31d64592296b/vispy/gloo/buffer.py#L174) is fine but the doc did not reflect that.

vispy/gloo/buffer.py
```diff
@@ -299,9 +299,9 @@ def __setitem__(self, key, data):
         elif data.size > stop - start:
             raise ValueError('Data too big to fit GPU data '
                              '(%d > %d-%d).' % (data.size, stop, start))
-        
+
         # Set data
-        offset = start  # * self.itemsize
+        offset = start
```
djhoese: So `start` is in item-space, right? It is then passed to `set_subdata` as `offset`. You then updated the `set_subdata` method to say that `offset` is in "bytes", right? Or is `start` in number of bytes (if so where does that happen)?

asnt: Here, yes,  `start` is in item space. It is passed below to `DataBuffer.set_subdata()`, which takes `offset` in item space, converts it to byte space and forwards it `Buffer.set_subdata()`.
Note the difference between space of `offset` in `DataBuffer.set_subdata()` and `Buffer.set_subdata()`.
I did not touch the code, just the doc.
The comment at the line above maybe comes from this confusion.

djhoese: OMG :man_facepalming: 

I read your change to the docstring as you *adding* the `in bytes`. I now see you removed it. Thank you for your patience.

vispy/gloo/buffer.py
```diff
@@ -162,7 +162,7 @@ def set_subdata(self, data, offset=0, copy=False, **kwargs):
         data : ndarray
             Data to be uploaded
         offset: int
-            Offset in buffer where to start copying data (in bytes)
+            Offset in buffer where to start copying data
```
djhoese: Could you maybe say that this is an index or is reference to the items or some other clear way of mentioning that this isn't bytes?

vispy/gloo/buffer.py
asnt: No worry. You are welcome!

vispy/gloo/buffer.py
asnt: Yes, indeed, it would be clearer. Not sure about how to best express it, though.
"(index into the underlying array)"?

asnt: See tentative in the commit below.

djhoese: That works for me. I'm not sure how else to phrase this. At least we tried.

PR #1829, opened 2020-02-28
solve ssl problem

rancheng:
RuntimeError: Error while fetching file http://github.com/vispy/demo-data/raw/master/galaxy/star-particle.png.
Dataset fetching aborted (<urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:841)>)

### DIFF
--- a/vispy/util/fetching.py
+++ b/vispy/util/fetching.py
@@ -236,6 +236,11 @@ def _fetch_file(url, file_name, print_destination=True):
     n_try = 3
     for ii in range(n_try):
         try:
+            # https://github.com/scikit-learn/scikit-learn/issues/10201#issuecomment-365734582
+            import os, ssl
+            if (not os.environ.get('PYTHONHTTPSVERIFY', '') and
+                    getattr(ssl, '_create_unverified_context', None)):
+                ssl._create_default_https_context = ssl._create_unverified_context
             data = urllib.request.urlopen(url, timeout=15.)
         except Exception as e:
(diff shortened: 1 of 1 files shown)


djhoese:
Interesting. What operating system are you using? Reading the scikit issue in your comment it looks like this was a solvable issue for Macs.

From what I've found this is a problem with how Python is built with (or without) the SSL library. Based on [PEP 493](https://www.python.org/dev/peps/pep-0493/#feature-environment-based-configuration) the `PYTHONHTTPSVERIFY` environment variable is only part of python 2.7.x; VisPy no longer supports Python 2. Even if it does, the `not os.environ` part of your code would miss the case where someone specifies the variable with the value `"0"` (which resolves to `True`).

I'm not really excited about including this if it is a problem with how Python is built. I'm wondering if this solution would work for you: https://community.netapp.com/t5/Software-Development-Kit-SDK-and-API-Discussions/Python-How-to-disable-SSL-certificate-verification/m-p/113698/highlight/true#M1691

Lastly, the tests are failing due to flake8 issues (you have multiple imports in one line).

djhoese:
Hi @rancheng any updates or other opinions on this? As I mentioned in my previous comment, I'm not sure this solution is the best solution for VisPy now that we've dropped Python 2.7 support. I'm closing this for now since you didn't respond. If you'd like to continue working on this please comment and we can reopen it.

PR #1835, opened 2020-03-09
Make demos easier to switch to PySide2 from PyQt5

fedepell:
While most of examples/demos have hardcoded PyQt5 usage, it is generally trivial to pass to PySide2 by just replacing PyQt5 with PySide2 in the imports. But there are a few exceptions, like the two demos touched by this PR, that use signals where the PyQt5 and PySide2 API differs. 
This PR makes also those demos easily switchable.

I'm not sure a general revision of demo code, or just a plain explanation, would be advisable given that PySide2 is getting more important with the backing of Qt. 
If so I can try to contribute in this sense.

Federico

### DIFF
--- a/examples/basics/scene/isocurve_for_trisurface_qt.py
+++ b/examples/basics/scene/isocurve_for_trisurface_qt.py
@@ -41,6 +41,7 @@ except Exception:
 # Provide automatic signal function selection for PyQtX/PySide2
 pyqtsignal = QtCore.pyqtSignal if hasattr(QtCore, 'pyqtSignal') else QtCore.Signal
 
+
 class ObjectWidget(QWidget):
     """
     Widget for editing OBJECT parameters
--- a/examples/basics/scene/isocurve_for_trisurface_qt.py
+++ b/examples/basics/scene/isocurve_for_trisurface_qt.py
@@ -25,22 +25,27 @@ except ImportError:
     pass
 
 try:
-    from PyQt4.QtCore import pyqtSignal, Qt
+    from PyQt4 import QtCore
+    from PyQt4.QtCore import Qt
     from PyQt4.QtGui import (QApplication, QMainWindow, QWidget, QLabel,
                              QSpinBox, QComboBox, QGridLayout, QVBoxLayout,
                              QSplitter)
 except Exception:
-    from PyQt5.QtCore import pyqtSignal, Qt
+    # To switch between PyQt5 and PySide2 bindings just change the from import
+    from PyQt5 import QtCore
+    from PyQt5.QtCore import Qt
     from PyQt5.QtWidgets import (QApplication, QMainWindow, QWidget, QLabel,
                                  QSpinBox, QComboBox, QGridLayout, QVBoxLayout,
                                  QSplitter)
 
+# Provide automatic signal function selection for PyQtX/PySide2
+pyqtsignal = QtCore.pyqtSignal if hasattr(QtCore, 'pyqtSignal') else QtCore.Signal
 
 class ObjectWidget(QWidget):
     """
     Widget for editing OBJECT parameters
     """
-    signal_object_changed = pyqtSignal(name='objectChanged')
+    signal_object_changed = pyqtsignal(name='objectChanged')
 
     def __init__(self, parent=None):
         super(ObjectWidget, self).__init__(parent)
--- a/examples/demo/gloo/primitive_mesh_viewer_qt.py
+++ b/examples/demo/gloo/primitive_mesh_viewer_qt.py
@@ -22,8 +22,6 @@ except ImportError:
 
 # To switch between PyQt5 and PySide2 bindings just change the from import
 from PyQt5 import QtCore, QtWidgets
-# Provide automatic signal function selection for PyQt5/PySide2
-pyqtsignal = QtCore.pyqtSignal if hasattr(QtCore, 'pyqtSignal') else QtCore.Signal
 
 import sys
 
@@ -32,6 +30,10 @@ from vispy import app, gloo
 from vispy.util.transforms import perspective, translate, rotate
 from vispy.geometry import meshdata as md
 from vispy.geometry import generation as gen
+
+# Provide automatic signal function selection for PyQt5/PySide2
+pyqtsignal = QtCore.pyqtSignal if hasattr(QtCore, 'pyqtSignal') else QtCore.Signal
+
 
 OBJECT = {'sphere': [('rows', 3, 1000, 'int', 3),
                      ('cols', 3, 1000, 'int', 3),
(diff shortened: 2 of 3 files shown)


djhoese:
Thanks for this. One other suggestion that has come up is to switch to using `qtpy` (https://pypi.org/project/QtPy/) in the VisPy backend. I'm not sure using this in the examples and adding another dependency is wise unless the backend is also switched. Just wanted to mention it.

I'm wondering if a try/except import might be simpler for users skimming the code:

```
try:
    from PySide2 import QtCore
    Signal = QtCore.Signal
except ImportError:
    from PyQt5 import QtCore
    Signal = QtCore.pyqtSignal
```

Thoughts?

fedepell:
I was also unsure first about using a _try/except_ instead, but I chose the _dir_ way not to put imports in the middle of the code, so to switch between the two bindings one would hopefully have everything more visible on the top.
Also the _try/except_ has the (IMO) negative fact that it can be deceiving on systems with both installed (which happens often, especially in distributions such as Anaconda for example) as it could give the appearance of working but choose what is first picked.

But indeed the nice thing would be to rethink/refactor the demos overall to easily switch between the various bindings now supported (ie. I also saw a few cases when the even older PyQt4 is imported directly).
qtpy sounds like a good alternative but indeed brings one more dependency, so I'm not sure how the project feels about it. Otherwise a general rewrite with a common shared init part could be an idea.

djhoese:
The import could happen at the top. Additionally your `dir` check could be replaced with `hasattr(QtCore, 'Signal')`.

As for PyQt4, we only semi-recently (the last year?) switched all defaults to PyQt5. Yeah I know, we were late to the party by quite a few years.

I see now that in a system with both pyqt5 and pyside2 installed, my try/except would be a problem because it prefers pyside2 first but VisPy currently prefers pyqt5 when choosing a backend automatically. My preference is that examples be switched to try/excepts with pyqt5 and fallback to pyside2. We can leave qtpy for the future. @kmuehlbauer @larsoner chime in if you have an opinion on the matter.

djhoese:
Not sure how this has passed in the past, but the Azure failures are due to no `mock` module being available. That's because it should now be `from unittest import mock`. I would not worry about this in this PR and I'll fix it in a separate one later.

fedepell:
> The import could happen at the top. Additionally your `dir` check could be replaced with `hasattr(QtCore, 'Signal')`.

Ok! Let me fix my PR with this, just time to do it and I commit!

> I see now that in a system with both pyqt5 and pyside2 installed, my try/except would be a problem because it prefers pyside2 first but VisPy currently prefers pyqt5 when choosing a backend automatically. My preference is that examples be switched to try/excepts with pyqt5 and fallback to pyside2. We can leave qtpy for the future. @kmuehlbauer @larsoner chime in if you have an opinion on the matter.

Should we do this in a separate PR that takes care of all (or at least ie. by directory, but more than just random two ;-) ) the examples in one go?

fedepell:
Force pushed now (with squash) using _hasattr_ instead of _dir_

djhoese:
> Should we do this in a separate PR that takes care of all (or at least ie. by directory, but more than just random two ;-) ) the examples in one go?

I'd say either this PR becomes the PR that fixes all the examples to allow for both or we don't do it at all. However, I think you said originally that the ones you changed are the *only* ones that can't be handled by a simple import change (changing PyQt5 to PySide2)?

fedepell:
> > Should we do this in a separate PR that takes care of all (or at least ie. by directory, but more than just random two ;-) ) the examples in one go?
> 
> I'd say either this PR becomes the PR that fixes all the examples to allow for both or we don't do it at all. However, I think you said originally that the ones you changed are the _only_ ones that can't be handled by a simple import change (changing PyQt5 to PySide2)?

Yes that two are the only two demos I came across that required a bit more than the trivial import change, due to the different API for signals.


Now I grepped also a a level lower and found one more outside of demos: _basics/scene/isocurve_for_trisurface_qt.py_. I could fix in case also this one in a similar fashion if useful!

djhoese:
Adding it to the other example would be nice. I'd prefer the try/except though. *Or* you could do `pyqtsignal = QtCore.pyqtsignal if hasattr(QtCore, 'pyqtsignal') else QtCore.Signal` at the top of the script. Maybe that is the cleanest way and it doesn't require the try/except for pyqt5 and pyside2?

I appreciate you bringing this overall problem to my attention and figuring out a solution, but I'm worried that it might add complexity to the "meat" of the example to have those `if/else` in the class definition.

fedepell:
@djhoese  thanks for the input! I just committed a new version with as little "meat" as possible but still make it trivial to pass between the two bindings I hope. 
I tried to make something more elegant (ie. a for loop from the list of bindings) but indeed it was getting too much distracting code. 

I also corrected the third example that used signals. In that case the example already had also archaic PyQt4 support. I didn't remove it (not to remove "functionality" already there) but just adapted the solution including also that. If you want I can of course remove it.

I hope it looks fine now!

djhoese:
> a for loop from the list of bindings

Just because it makes the code shorter doesn't mean it is easier to read :wink: 

Thanks for this. I think your PR is a good balance between "I want this to work" and "I don't want to rewrite every example that use PyQt with a try/except". I'll double check the failing tests later today (hopefully) and merge this. Thanks again.

fedepell:
@djhoese thanks to you for the feedback (and all the work on vispy!) and very glad if this small PR can help someone when evaluating vispy!

djhoese:
Looks like the Travis tests are complaining about some of the imports:

```
Running flake8... 
examples/basics/scene/isocurve_for_trisurface_qt.py:44:1: E302 expected 2 blank lines, found 1
examples/demo/gloo/primitive_mesh_viewer_qt.py:28:1: E402 module level import not at top of file
examples/demo/gloo/primitive_mesh_viewer_qt.py:30:1: E402 module level import not at top of file
examples/demo/gloo/primitive_mesh_viewer_qt.py:31:1: E402 module level import not at top of file
examples/demo/gloo/primitive_mesh_viewer_qt.py:32:1: E402 module level import not at top of file
examples/demo/gloo/primitive_mesh_viewer_qt.py:33:1: E402 module level import not at top of file
examples/demo/gloo/primitive_mesh_viewer_qt.py:34:1: E402 module level import not at top of file
Failed: flake8 failed
```

Could you try to fix those? If you install `flake8` locally and run `flake8 examples/` then it should print these same errors. If you fix them and run the command again then it won't print them.

fedepell:
Fixed! I didn't squash but I guess you can just immediately at merge (or just let me know and I do it if needed!)

djhoese:
I think I've fixed the problem on master now. Thanks for this. Merging now.

PR #1839, opened 2020-03-16
Fix PRs not being run on Azure

djhoese:
Taking the advice from @larsoner for fixing this. See https://github.com/mne-tools/mne-python/pull/7457


djhoese:
@larsoner @kmuehlbauer Any opinion on dropping Python 3.5 support? If against it, what about just not building wheels for it (for time)?

larsoner:
Fine with me

kmuehlbauer:
It's EOL in September, not sure if it's worth the effort.

PR #1842, opened 2020-03-18
Implement volume contrast limits in shader, add gamma

tlambert03:
This began as an internal PR at https://github.com/napari/napari/pull/1056, but I'd like to know if it's something you would consider.  In typical [napari](https://github.com/napari/napari) usage, it would be pretty normal to repeatedly change the min/max contrast limits while viewing a 3D volume.  Unfortunately, each change incurs a renormalization on the CPU in [`Volume.set_data`](https://github.com/vispy/vispy/blob/3ac28ff0daf8836b44c6eba8d3c98541e37c5702/vispy/visuals/volume.py#L487-L493), leading to very laggy performance.  This PR moves most of the _re_-calculation of clims to the shader as follows:

- `Volume.clim`→ `Volume.clim_range`: `Volume.clim` was previously just serving a one-time role of data normalization (0-1) prior to moving the volume to Texture memory.  Seems to me that normalization needn't be repeated often, if the user is able to provide the maximum _range_ of min/max values that they might want to use.  So `Volume.clim` becomes `Volume.clim_range`, and `set_data(vol, clim)` becomes `set_data(vol, clim_range)`
- `Volume.clim` now becomes a property that is passed to the shaders where the final scaling is done.
- this PR also adds `Volume.gamma`, and performs gamma calculations in the shaders at the moment of colormap lookup.

For typical volume sizes in napari (~256-1024 pixels<sup>3</sup>), this speeds up rendering after changing the contrast almost 5 fold (on my computer)

### breaking change
this is a **breaking** change, in that it changes the semantics of `Volume.clim` and adds a new property `Volume.clim_range` (that has the same meaning and usage as the previous `Volume.clim`.  The `clim` argument in `Volume.set_data` has been renamed to `clim_range` to reflect what it is actually doing (though a deprecation warning catches previous usage and warns).
It _would_ be possible to implement this without breaking the API, by just re-interpreting the meaning of the existing `Volume.clim` and creating a new property (maybe `display_limits`?) that controls the final lookup on the shader.  But I felt like that was much less clear.  Anyway, let me know what you think of the basic idea and we can discuss how you would prefer to handle breakages.

### DIFF
--- a/vispy/visuals/tests/test_volume.py
+++ b/vispy/visuals/tests/test_volume.py
@@ -62,6 +62,7 @@ def test_volume_draw():
         assert_image_approved(c.render(), 'visuals/volume.png')
 
 
+@requires_pyopengl()
 @requires_application()
 def test_volume_clims_and_gamma():
     """Test volume visual with clims and gamma on shader.
--- a/vispy/visuals/tests/test_volume.py
+++ b/vispy/visuals/tests/test_volume.py
@@ -26,7 +26,7 @@ def test_volume():
     
     # Clim
     V.set_data(vol, (0.5, 0.8))
-    assert V.clim == (0.5, 0.8)
+    assert V.clim_range == (0.5, 0.8)
     with raises(ValueError):
         V.set_data((0.5, 0.8, 1.0))
     
@@ -73,14 +73,14 @@ def test_set_data_does_not_change_input():
     for dtype in ['uint8', 'int16', 'uint16', 'float32', 'float64']:
         vol_copy = np.array(vol, dtype=dtype, copy=True)
         # setting clim so that normalization would otherwise change the data
-        V.set_data(vol_copy, clim=(0, 200))
+        V.set_data(vol_copy, clim_range=(0, 200))
         assert np.allclose(vol, vol_copy)
 
     # for those using float32 who want to avoid the copy operation,
     # using set_data() with `copy=False` should be expected to alter the data.
     vol2 = np.array(vol, dtype='float32', copy=True)
     assert np.allclose(vol, vol2)
-    V.set_data(vol2, clim=(0, 200), copy=False)
+    V.set_data(vol2, clim_range=(0, 200), copy=False)
     assert not np.allclose(vol, vol2)
 
 
--- a/vispy/visuals/tests/test_volume.py
+++ b/vispy/visuals/tests/test_volume.py
@@ -6,7 +6,7 @@ from vispy import scene
 from vispy.testing import (TestingCanvas, requires_application,
                            run_tests_if_main, requires_pyopengl,
                            raises)
-from vispy.testing.image_tester import assert_image_approved
+from vispy.testing.image_tester import assert_image_approved, downsample
 
 
 @requires_pyopengl()
@@ -60,6 +60,53 @@ def test_volume_draw():
         scene.visuals.Volume(vol, parent=v.scene)  # noqa
         v.camera.set_range()
         assert_image_approved(c.render(), 'visuals/volume.png')
+
+
+@requires_application()
+def test_volume_clims_and_gamma():
+    """Test volume visual with clims and gamma on shader.
+    
+    currently just using np.ones since the angle of view made more complicated samples
+    challenging, but this confirms gamma and clims works in the shader.
+    """
+
+    with TestingCanvas(size=(40, 40), bgcolor="k") as c:
+        v = c.central_widget.add_view(border_width=0, size=(40, 40))
+        data = np.ones((40, 40, 40)) / 2.5
+        volume = scene.visuals.Volume(
+            data,
+            clim=(0, 1),
+            interpolation='nearest',
+            parent=v.scene,
+            method='mip',
+        )
+        v.camera = 'arcball'
+        v.camera.fov = 0
+        v.camera.scale_factor = 40
+        v.camera.center = (19.5, 19.5, 19.5)
+
+        rendered = c.render()[..., 0]
+        _dtype = rendered.dtype
+        shape_ratio = rendered.shape[0] // data.shape[0]
+        rendered1 = downsample(rendered, shape_ratio, axis=(0, 1)).astype(_dtype)
+        predicted = np.round(data.max(0) * 255).astype(np.uint8)
+        assert np.allclose(predicted, rendered1, atol=1)
+
+        # adjust contrast limits
+        new_clim = (0.3, 0.8)
+        volume.clim = new_clim
+        rendered2 = downsample(c.render()[..., 0], shape_ratio, axis=(0, 1)).astype(_dtype)
+        scaled_data = np.clip((data - new_clim[0]) / np.diff(new_clim)[0], 0, 1)
+        predicted = np.round(scaled_data.max(0) * 255).astype(np.uint8)
+        assert np.allclose(predicted, rendered2)
+        assert not np.allclose(rendered1, rendered2, atol=10)
+
+        # adjust gamma
+        volume.gamma = 1.5
+        rendered3 = downsample(c.render()[..., 0], shape_ratio, axis=(0, 1)).astype(_dtype)
+        predicted = np.round((scaled_data ** volume.gamma).max(0) * 255).astype(np.uint8)
+        assert np.allclose(predicted, rendered3)
+        assert not np.allclose(rendered2, rendered3, atol=10)
 
 
 @requires_pyopengl()
--- a/vispy/visuals/tests/test_volume.py
+++ b/vispy/visuals/tests/test_volume.py
@@ -26,7 +26,7 @@ def test_volume():
     
     # Clim
     V.set_data(vol, (0.5, 0.8))
-    assert V.clim_range == (0.5, 0.8)
+    assert V.clim == (0.5, 0.8)
     with raises(ValueError):
         V.set_data((0.5, 0.8, 1.0))
     
@@ -73,14 +73,14 @@ def test_set_data_does_not_change_input():
     for dtype in ['uint8', 'int16', 'uint16', 'float32', 'float64']:
         vol_copy = np.array(vol, dtype=dtype, copy=True)
         # setting clim so that normalization would otherwise change the data
-        V.set_data(vol_copy, clim_range=(0, 200))
+        V.set_data(vol_copy, clim=(0, 200))
         assert np.allclose(vol, vol_copy)
 
     # for those using float32 who want to avoid the copy operation,
     # using set_data() with `copy=False` should be expected to alter the data.
     vol2 = np.array(vol, dtype='float32', copy=True)
     assert np.allclose(vol, vol2)
-    V.set_data(vol2, clim_range=(0, 200), copy=False)
+    V.set_data(vol2, clim=(0, 200), copy=False)
     assert not np.allclose(vol, vol2)
 
 
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -34,6 +34,7 @@ The ray is expressed in coordinates local to the volume (i.e. texture
 coordinates).
 
 """
+import warnings
 
 from ..gloo import Texture3D, TextureEmulated3D, VertexBuffer, IndexBuffer
 from . import Visual
@@ -82,6 +83,8 @@ FRAG_SHADER = """
 // uniforms
 uniform $sampler_type u_volumetex;
 uniform vec3 u_shape;
+uniform vec2 clim;
+uniform float gamma;
 uniform float u_threshold;
 uniform float u_relative_step_size;
 
@@ -280,7 +283,9 @@ MIP_SNIPPETS = dict(
             maxval = max(maxval, $sample(u_volumetex, loc).g);
             loc += step * 0.1;
         }
-        gl_FragColor = $cmap(maxval);
+        maxval = max(maxval - clim.x, 0);
+        maxval = maxval / (clim.y - clim.x);
+        gl_FragColor = $cmap(pow(maxval, gamma));
         """,
 )
 MIP_FRAG_SHADER = FRAG_SHADER.format(**MIP_SNIPPETS)
@@ -291,7 +296,9 @@ TRANSLUCENT_SNIPPETS = dict(
         vec4 integrated_color = vec4(0., 0., 0., 0.);
         """,
     in_loop="""
-            color = $cmap(val);
+            val = max(val - clim.x, 0);
+            val = val / (clim.y - clim.x);
+            color = $cmap(pow(val, gamma));
             float a1 = integrated_color.a;
             float a2 = color.a * (1 - a1);
             float alpha = max(a1 + a2, 0.001);
@@ -323,7 +330,9 @@ ADDITIVE_SNIPPETS = dict(
         vec4 integrated_color = vec4(0., 0., 0., 0.);
         """,
     in_loop="""
-        color = $cmap(val);
+        val = max(val - clim.x, 0);
+        val = val / (clim.y - clim.x);
+        color = $cmap(pow(val, gamma));
         
         integrated_color = 1.0 - (1.0 - integrated_color) * (1.0 - color);
         """,
@@ -348,7 +357,9 @@ ISO_SNIPPETS = dict(
                 color = $sample(u_volumetex, iloc);
                 if (color.g > u_threshold) {
                     color = calculateColor(color, iloc, dstep);
-                    gl_FragColor = $cmap(color.g);
+                    val = max(color.g - clim.x, 0);
+                    val = val / (clim.y - clim.x);
+                    gl_FragColor = $cmap(pow(val, gamma));
                     iter = nsteps;
                     break;
                 }
@@ -378,9 +389,14 @@ class VolumeVisual(Visual):
     vol : ndarray
         The volume to display. Must be ndim==3.
     clim : tuple of two floats | None
-        The contrast limits. The values in the volume are mapped to
+        The current contrast limits. The values in the volume are mapped to
         black and white corresponding to these values. Default maps
         between min and max.
+    clim_range : tuple of two floats | None
+        The maximum range contrast limits. This value is used to normalize
+        ``vol`` to 0-1 prior to transfer to the GPU.  It determines the
+        minimum and maximum possible values that can be used for clim.  
+        Default maps between min and max.
     method : {'mip', 'translucent', 'additive', 'iso'}
         The render method to use. See corresponding docs for details.
         Default 'mip'.
@@ -402,8 +418,8 @@ class VolumeVisual(Visual):
 
     _interpolation_names = ['linear', 'nearest']
 
-    def __init__(self, vol, clim=None, method='mip', threshold=None, 
-                 relative_step_size=0.8, cmap='grays',
+    def __init__(self, vol, clim=None, clim_range=None, method='mip', threshold=None, 
+                 relative_step_size=0.8, cmap='grays', gamma=1,
                  emulate_texture=False, interpolation='linear'):
         
         tex_cls = TextureEmulated3D if emulate_texture else Texture3D
@@ -411,6 +427,8 @@ class VolumeVisual(Visual):
         # Storage of information of volume
         self._vol_shape = ()
         self._clim = None
+        self._clim_range = None
+        self._gamma = gamma
         self._need_vertex_update = True
 
         # Set the colormap
@@ -439,6 +457,7 @@ class VolumeVisual(Visual):
         self.shared_program['u_volumetex'] = self._tex
         self.shared_program['a_position'] = self._vertices
         self.shared_program['a_texcoord'] = self._texcoord
+        self.shared_program['gamma'] = self._gamma
         self._draw_mode = 'triangle_strip'
         self._index_buffer = IndexBuffer()
 
@@ -448,7 +467,8 @@ class VolumeVisual(Visual):
         self.set_gl_state('translucent', cull_face=False)
         
         # Set data
-        self.set_data(vol, clim)
+        self.set_data(vol, clim_range)
+        self.clim = clim or self.clim_range
         
         # Set params
         self.method = method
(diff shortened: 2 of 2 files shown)


vispy/visuals/tests/test_volume.py
```diff
@@ -73,14 +73,14 @@ def test_set_data_does_not_change_input():
     for dtype in ['uint8', 'int16', 'uint16', 'float32', 'float64']:
         vol_copy = np.array(vol, dtype=dtype, copy=True)
         # setting clim so that normalization would otherwise change the data
-        V.set_data(vol_copy, clim=(0, 200))
+        V.set_data(vol_copy, clim_range=(0, 200))
```
tlambert03: here's an example of the API breakage... though technically, with the `if "clim" in kwargs` conditional that I added to `Volume.set_data` this test would have still worked

vispy/visuals/volume.py
```diff
@@ -348,7 +357,9 @@
                 color = $sample(u_volumetex, iloc);
                 if (color.g > u_threshold) {
                     color = calculateColor(color, iloc, dstep);
-                    gl_FragColor = $cmap(color.g);
+                    val = max(color.g - clim.x, 0);
```
tlambert03: wasn't entirely sure if this is the best place to implement clims for the ISO rendering... thoughts welcome

tlambert03:
> This seems like a good compromise between switching to float textures and the current behavior.

yeah, with the obvious caveat that if someone chooses a clim_range much large than the current clim setting, they will start seeing some posterization.  I think in most use cases, that will be hard to see, but it's easy to see it if you force it (by setting a `clim_range` much large than required for the dataset).  Am I correct in understanding that the texture bit depth is hardcoded [right here](https://github.com/vispy/vispy/blob/cf244a266fe71a5b306ef518f0f8a05040502939/vispy/gloo/glir.py#L1623-L1624)?  I assume that's to conserve VRAM and imagine there's been a lot of discussion on that ... is it something you've talked about exposing as a variable?


> What if the original clim was used as `clim_range` and then everything else was the same as you have it _except_ if someone specified a `clim` outside of the current `clim_range` it would re-normalize the data. Thoughts?

interesting!  so it wouldn't be a breaking change, right?  Basically, we'd just be adding a `@clim.setter` that updates the shared_program, and calls `set_data` if out of range?  I think I like that!   The only (very minor) downside i guess would be that if the `clim` starts ratcheting upwards, posterization _may_ appear in some cases, so there has to be a way to reduce `clim_range` as well.   The user could always call `set_data` again, to bring it back down, and we'd just have to document that the normalization range "automatically" increases when you change `Volume.clim` but can only be manually decreased using `set_data`.  Do I have that right?

tlambert03:
ok, have a look again at your convenience.  I think I like that a lot.  Doesn't break anything, and it works fine for our needs in `napari`.  On unfortunate consequence of this approach however (and tell me if you can see a workaround) is that we have to store a copy of the data last provided to `set_data`, because the `clim.setter` will need to call `set_data(last_data, new_clims)` if it needs to renormalize the data.

djhoese:
> Am I correct in understanding that the texture bit depth is hardcoded right here? I assume that's to conserve VRAM and imagine there's been a lot of discussion on that ... is it something you've talked about exposing as a variable?

Keep in mind I'm not one of the original vispy developers so I don't know the historical reasons for things. I also am not an OpenGL expert so I'm not familiar with every specification and its limitations. I think you are right that it is limited here. However, it is not limited in 2D textures which I've customized myself in the past in my own applications to upload float data. It is possible the byte type here is hardcoded because that's all that was supported for 3D textures at the time (especially since 3D textures themselves are not part of the OpenGL 2.0 standard iirc).

> interesting! so it wouldn't be a breaking change, right? Basically, we'd just be adding a @clim.setter that updates the shared_program, and calls set_data if out of range?

Right. We could have a `clim_range` setter still that lets the user force a new normalization or I guess that wouldn't be needed if they recalled `set_data`.

> One unfortunate consequence of this approach however (and tell me if you can see a workaround) is that we have to store a copy of the data last provided to set_data, because the clim.setter will need to call set_data(last_data, new_clims) if it needs to renormalize the data.

Darn, you're right. I didn't think about that. I'm not sure how ugly it would be to hold on to that.

tlambert03:
I may submit a different PR to open up access to some of those texture bit depths.  Even just bumping it to GL_SHORT would prevent posterization in most cases.

> Darn, you're right. I didn't think about that. I'm not sure how ugly it would be to hold on to that [data].

I guess then that's the biggest question for this PR.  For all of our stuff in napari (and I suspect a lot of other users), that extra copy is no big deal, since we're already holding it in memory elsewhere (and it's the "uncopied" version that I store here)... but it seems like something we should get some addition opinions on.  Feel free to ping anyone you think should review it.  I'll wait to hear more before continuing to work on it.  (For those just tuning in, in [this line](https://github.com/vispy/vispy/blob/0bcfb3d2d5be78201b2698680f7cae77e5b6173b/vispy/visuals/volume.py#L502) of this PR, I store a reference to the data provided to `set_data` for later re-normalization)

djhoese:
Coming back to this, comparing it to the existing Volume visual and your new ImageVisual that you worked on in #1844, and I'm still leaning towards my suggestion where:

1. A reference to the original passed data is kept in VolumeVisual. This matches how the current ImageVisual works. Unless we can find a past reasoning for why the VolumeVisual is different or can think of any other reason to *not* hold on to a reference of the original data then I think this makes the most sense and is the most flexible.
2. Data is scaled according to the provided clims or default to min/max of the provided data.
3. Scaled data is sent to the GPU's texture and the clims for that texture data are stored in the Visual. The scaled data is not held on to (or at least it is up to the lower-level gloo interfaces if it needs to be).
4. Changes in clims are immediately reflected in the shader parameters. If a clim is outside of the current textures clims then the original data (which we have a reference to) is rescaled.

I'm still open to arguments and other options but so far this seems like the best way to keep the existing interface(s) while improving performance on the amount of rescalings that have to be done, the quality of the image in the GPU, and the speed for future clim changes. I also think it is important that the VolumeVisual and ImageVisual have similar interfaces. I think a PR for the ImageVisual in the future could perform a similar "caching" of the color limits. And then a future-future PR could add the option to specify the dtype for the VolumeVisual's and the ImageVisual's textures (default uint8 to match current behavior). I'm not sure if using numpy dtypes or GL constants is best/easiest for this. That said, there are two parameters that I think need to be set on the textures to fully control what's stored in them so a single numpy dtype might not be enough. For example:

```
format="LUMINANCE", internalformat="R32F",
```

This would break OpenGL 2.0 compatibility but not by default which is OK with me.

tlambert03:
great!  I agree, and I think that's the general state that this PR is currently in.  I'll go through it again and look for places where I can make it as similar to ImageVisual as possible and will ping again when ready for review

tlambert03:
OK, I had another look and I think it's ready to go.  Can't totally mimic what we did on the ImageVisual, since colormaps are handled differently there, but I think the shader code looks cleaner here anyay.

I added a basic test to make sure clims and gamma are responding correctly on in the shaders.
this is ready for review.  Thanks again!

sofroniewn:
@djhoese i was wondering if you could take another look at this.

We're planning a big napari release soon, it would be great to have this merged an a `0.6.5` vispy release before then if possible. Thanks as always for being open to these contributions!!

djhoese:
When is soon? This is a new feature so this would require a 0.7.0 release. I'd like to include those Mesh-based PRs too which require a lot more time to review. This one is probably fine. I meant to review things this last weekend but chose sanity over volunteer work :wink: (finished a 2000 piece puzzle though).

sofroniewn:
> I meant to review things this last weekend but chose sanity over volunteer work 😉 (finished a 2000 piece puzzle though).

Definitely the right life choices!!!

> When is soon? This is a new feature so this would require a 0.7.0 release. 

We're looking to make the `0.3.0` release this week - but for us this won't be a public API change, so we can easily add it in to `0.3.1` or a later release. It should lead to a big performance boost for us though which will be exciting!

djhoese:
@tlambert03 The tests are failing because of a bad docstring:

```
―――――――――――――――――――――――――― test_docstring_parameters ―――――――――――――――――――――――――――
vispy/util/tests/test_docstring_parameters.py:112: in test_docstring_parameters
    raise AssertionError(msg)
E   AssertionError: 
E   vispy.visuals.volume.__init__ arg mismatch: ['gamma']
E   
E   1 docstring violations found
```

vispy/visuals/volume.py
```diff
@@ -114,6 +116,11 @@
     return color1.g; // todo: why did I have this abstraction in visvis?
 }}
 
+vec4 applyColormap(float data) {{
+    float val = max(data - clim.x, 0) / (clim.y - clim.x);
```
djhoese, line 120: Are users allowed to have flipped color limits? It is common in the data I work with to flip the color limits so that `clim = (50, 25)` and the data is then inverted. Does this `max` interfere with that? Is it needed? Is it this way in the ImageVisual PR that you made?

vispy/visuals/volume.py
```diff
@@ -506,13 +522,75 @@ def set_data(self, vol, clim=None, copy=True):
...
+        ``value`` should be a 2-tuple of floats (min_clim, max_clim), where each value is
+        within the range set by self.clim. If the new value is outside of the (min, max)
+        range of the clims previously used to normalize the texture data, then data will
+        be renormalized using set_data.
+        """
+        clim = np.array(value, float)
+        if not (clim.ndim == 1 and clim.size == 2):
+            raise ValueError('clim must be a 2-element array-like')
+        self._clim = tuple(clim)
+        if (clim[0] < self._texture_limits[0]) or (clim[1] > self._texture_limits[1]):
```
djhoese, line 559: I'm realizing something and I know this is probably too late to change this given you've made the ImageVisual changes already, but one problem with this approach is that if I make the color limits smaller I probably want a better color/bit depth of my image. This current approach if initialized with large-ish values and then changed to smaller limits would not update the texture and could result in only a few colors being shown on the screen.

I'm wondering if there should be an overall flag in `__init__` that controls this behavior. Like maybe it defaults to the optimized method that you've done here, but then can be turned off for "optimized" colors.

vispy/visuals/volume.py
tlambert03: hmm, good question.  Does the existing vispy normalization code work for your application?  Yes, I believe there is a `max()` call in the ImageVisual PR as well.  But I can submit a followup PR if we identify the desired behavior.

For what it's worth, if I run the `examples/basics/scene/volume.py` in the master branch, with a flipped `clim` argument (.e.g `(1000, 0)`), it doesn't work: the volume is all white.  Same goes for the image example, even if I revert back before the ImageVisual PR, like a22df65b.  Is this something you often do with vispy?  or a new feature that would be nice to support?

vispy/visuals/volume.py
tlambert03: Yeah that's what I was describing in https://github.com/vispy/vispy/pull/1842#issuecomment-600901420.
One potential solution is to have a threshold that also causes an automatic recalculation, perhaps if the clim range gets to lower than 25% of the original range? (that would correspond to ~64 colors which qualitatively feels to be about where obvious posterization starts to appear).

something like this in `clim.setter`
```python

if (clim[0] < self._texture_limits[0]) or (clim[1] > self._texture_limits[1]):
    self.rescale_data()
    return

range_ratio = np.subtract(*clim) / np.subtract(*self._texture_limits)
if np.abs(range_ratio) < self._min_clim_range_ratio:
    self.rescale_data()
else:
   ...
```

vispy/visuals/volume.py
djhoese: It's possible I have it custom coded in my own application since I'm using floats there:

https://github.com/ssec/sift/blob/4874d971811b16502b23b3837069ed25cd454942/uwsift/view/visuals.py#L147-L153

To support it you'd have to make sure the python code checks that the new clims are within the entire previous clim range instead of just `new_clim_max > old_clim_max`.

As for this PR, is the `max` doing anything?

tlambert03: yeah, on this PR the max is making sure that we don't get wraparound... Without it, if you bring the clim min above the minimum values in the image, those pixels don't go black, they get bright again

vispy/visuals/volume.py
djhoese: I think this would be overkill. At this point I'm worried we're (you're) doing all this work just for a small optimization, but if the napari folks think this will improve performance then I suppose it is worth it.

vispy/visuals/volume.py
djhoese: Thoughts on switching to what I'm doing at the above linked code where I use `clamp` to clip the data to the provided color limits?

vispy/visuals/volume.py
tlambert03: I can definitely vouch for the performance improvement when changing the contrast limits on volumes.  for volumes larger than about 512x512x256 (pretty common in microscopy) I see ~5fold improvement in FPS on my system (some numbers [here](https://github.com/napari/napari/pull/1056)).  I'm actually very excited to eventually take advantage of these two PRs.

I was already implementing something like this in the napari code to prevent posterization, so unless you're worried about it being overkill from a perspective of code cleanliness or performance in vispy, I definitely don't mind doing it.  (but if you think it's confusing or weird, that's another story! 😄)

On the other hand, if you'd rather keep it opt-in-or-nothing, can you clarify what you meant with "it defaults to the optimized method that you've done here, but then can be turned off for "optimized" colors."  Do you mean someone would be able to opt out of _any_ rescaling (whether it by upscaling or downscaling), or that they would opt in/out of rescaling if the clim range got dangerously low?  If the latter, `_min_clim_range_ratio` could be that opt-in/out value: if it was 0, no low-end rescaling would ever occur.

Anyway, I'm happy to follow your lead here!

djhoese: Sounds good. As for "optimized" colors, I meant that the person could have it always rescale or do your current "smart" rescaling. You are write that the clim range ratio could be this flag (`1` means always rescale?).

tlambert03: sure!  yeah with the `range_ratio` logic above, `0` would mean never rescale, `1` would mean always rescale, and anything in between would only rescale if the current clim `range_ratio` was smaller  than `_min_clim_range_ratio`.  (and of course, it will always rescale if the new clims are outside of the previous texture limits)

tlambert03: and fwiw, a "future goal" is definitely to have another look at full float support for textures, in which case we can just kinda avoid all of this right?

djhoese: For napari, yes. For people targeting older GPUs, no, they'll still want to use 8-bit textures.

tlambert03: yeah def.  and I'm sure we have plenty of CPU/old-GPU using users as well, so I recognize it doesn't remove the need for this approach too.

vispy/visuals/volume.py
tlambert03: looks good to me! will work on fixing this PR to match it, probably tomorrow.

tlambert03:
@djhoese, can you tell what I'm doing wrong with the last [failing test](https://travis-ci.org/github/vispy/vispy/jobs/680816155#L6574-L6611)?  It seems to be only my test, but it's unclear why 'OpenGL' is not found for me...

djhoese:
My best guess is you need the requires_pyopengl decorator too.

vispy/visuals/volume.py
tlambert03: hey @djhoese, sorry for the long delay, though I gather you've been busy too :)

I updated the code to match your clamp pattern instead of using max.  If it's ok with you, I'd propose that the clims inverting behavior be addressed in a separate PR, since that's a new feature altogether.  Let me know if you disagree, or if you see anything else you'd like addressed in this PR.

kmuehlbauer: > hey @djhoese, sorry for the long delay, though I gather you've been busy too :)
> 

@tlambert03 FYI, David is AFK for two weeks or so. I can have a look into this PR by mid of next week. We might have to wait until David is back.

djhoese: @tlambert03 Thanks for the updates. Correct me if I'm wrong, but don't your recent changes already handle the clim inversion?

tlambert03: I don't think so?  at least, not in the sense that providing clims (1, 0) will give you an inverted lookup-table.  I thought that was also going to require some changes to `Volume.set_data`, which I mostly hadn't touched in this PR:

https://github.com/vispy/vispy/blob/77f016a0c6121e1b0c024bdf4e03d5bf84aa9879/vispy/visuals/volume.py#L486-L493

it wouldn't be too hard to support... but that's why I was suggesting that we maybe address that in a follow-up PR.  but let me know if you want it all done here and now

djhoese: Ah I think you're right. I was thinking of my own application where I have floating point data in the texture and there is not CPU handling of the clims. I'd like it in this PR, but if it is too complex I am fine moving it to another PR. I'd really like it to be included in the next release (I don't want the existing functionality released without the inverse clims possible).

tlambert03: works for me.  let's leave this PR open then as motivation :) I'll work on it soon

tlambert03: ok just pushed an update with [inversion](https://github.com/vispy/vispy/pull/1842/commits/1c71091c0be60fd03c34a3dee7cba18657302ce3).  I have to say, my brain got a little turned around... and while I think this code achieves the correct behavior (in my tests), I may have made it more complicated than it needs to be, and would appreciate a critical review.  

Also, adding support for inversion  _and_ shader-based clim changes definitely seems to increase the possibility of bugs if someone switches from a standard clim to an inverted one after instantiating the volume...  not 100% I've caught all of those.

djhoese:
@tlambert03 Care to catch me up on this PR? Sorry for losing track again. I see @sofroniewn has vendored this PR in napari. Any issues discovered?

sofroniewn:
> Any issues discovered?

https://github.com/napari/napari/issues/1585 might be related, not sure yet, could be purely on the napari end - but probably worth looking if it can be reproduced in this PR too

tlambert03:
> @tlambert03 Care to catch me up on this PR? Sorry for losing track again. I see @sofroniewn has vendored this PR in napari. Any issues discovered?

no problem.  I'm relatively confident about the core functionality at this point, I think the main thing that we were considering was the implementation of the clims-inverting behavior [here](https://github.com/vispy/vispy/pull/1842/files#diff-82dd371139f2855ea1e97309d1fa4186R525-R528).  I think you were [concerned it wouldn't work](https://github.com/vispy/vispy/pull/1842#commitcomment-40356815), but if I remove the `elif` block as suggested, it breaks (but it works for me as is).  I'm inclined to say this is ready if you're ok with it.

I did just check and napari/napari#1585 _does_ get introduced in napari/napari#1456... but there's a lot of statefulness going on there that could be on the napari side too... will try to determine whether it's something you need to care about here or not

tlambert03:
couple more observations: it's not related to this PR... it only affects 2D images and `ImageVisuals`.  But I actually don't think it has anything to do with #1844 either, I think it's on the napari side....  The reason I say that is that if I toggle image visibility with `ImageVisual.visible = False/True` then it doesn't happen.  But if I toggle visibility with `napari.layers.Image.visible = False/True` then it _does_ happen unless I force `_need_colortransform_update` to be true on the `ImageVisual`.  So I think it's on our end.

so @djhoese, I think that means the only thing you need to consider is whether your ok with my clim inversion logic

djhoese:
Thanks @tlambert03 and thanks for your patience. I did a simple test by modifying the ````volume.py` example and commenting out that small block seems to work fine for me. Am I not testing the right thing? Here are my changes which allow me to hit the `r` key to swap the clims...or maybe this is bad because I never change the clim range and never recompute the texture?

```
Index: vispy/visuals/volume.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- vispy/visuals/volume.py	(revision 1c71091c0be60fd03c34a3dee7cba18657302ce3)
+++ vispy/visuals/volume.py	(date 1599443685542)
@@ -522,10 +522,10 @@
         if self._clim[1] == self._clim[0]:
             if self._clim[0] != 0.:
                 vol *= 1.0 / self._clim[0]
-        elif self._clim[0] > self._clim[1]:
-            vol *= -1
-            vol += self._clim[1]
-            vol /= self._clim[1] - self._clim[0]
+        # elif self._clim[0] > self._clim[1]:
+        #     vol *= -1
+        #     vol += self._clim[1]
+        #     vol /= self._clim[1] - self._clim[0]
         else:
             vol -= self._clim[0]
             vol /= self._clim[1] - self._clim[0]
Index: examples/basics/scene/volume.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- examples/basics/scene/volume.py	(revision 1c71091c0be60fd03c34a3dee7cba18657302ce3)
+++ examples/basics/scene/volume.py	(date 1599444158913)
@@ -51,7 +51,7 @@
 
 # Create the volume visuals, only one is visible
 volume1 = scene.visuals.Volume(vol1, parent=view.scene, threshold=0.225,
-                               emulate_texture=emulate_texture)
+                               emulate_texture=emulate_texture, clim=[900, 1100])
 volume1.transform = scene.STTransform(translate=(64, 64, 0))
 volume2 = scene.visuals.Volume(vol2, parent=view.scene, threshold=0.2,
                                emulate_texture=emulate_texture)
@@ -114,7 +114,9 @@
 def on_key_press(event):
     global opaque_cmap, translucent_cmap
-    if event.text == '1':
+    if event.text == 'r':
+        volume1.clim = volume1.clim[::-1]
+    elif event.text == '1':
         cam_toggle = {cam1: cam2, cam2: cam3, cam3: cam1}
         view.camera = cam_toggle.get(view.camera, cam2)
         print(view.camera.name + ' camera')

```

tlambert03:
Thanks for testing!  I will try commenting that out tomorrow morning and let you know what I see

djhoese:
I think I just realized I'm not testing it properly/fully. I have a 2.5 month old daughter. Give me a break.

tlambert03:
> I think I just realized I'm not testing it properly/fully. I have a 2.5 month old daughter. Give me a break. 😴

lol.  well, if it makes you feel any better, it just took me an hour to find a case where commenting that bit out doesn't work. :joy:
But here it is... with that bit commented out, the simple inversion case works, but if I try to change the clim to a small range:
```
volume1.clim = (2000,0)  # inverts ok at first
volume1.clim = (200, 0)  # no good
volume1.clim = (2000, 0)  # can't recover step 1
volume1.clim = (0, 2000)  # can't even recover good image
```

I'm sorry that I'm not doing a better job of clarifying what's going on here. feels so basic, and I've probably over-complicated it here.  but at the moment, it's only working with the code as is.  maybe we go with it and refactor later if need be?

djhoese:
So I've been playing with this all morning and I'm feeling dumber by second. Good news is I hacked things enough that it worked with floating point textures in the GPU (format='r', internalformat='r32f'). That'll have to be another PR.

Anyway I came up with two important points:

1. Your way of inverting the calculation isn't doing what we want it to do. Run the calculations on `np.linspace(0, 2000, 100)`. If you run your special elif calculations on this array you'll see that (0, 2000) becomes (0, 1), but if you run the `else` block of calculations you'll end up with (0, 2000) becomes (1, 0) which *is* what we want. Right? Put another way, your special `elif` 
2. I couldn't even get to `(2000, 0)` to work when passing it as the original clim. That gets around half of your new code anyway and it still didn't work. It just gives me an all white volume. I went back and forth and I *think* I've decided it isn't possible to do invert clims the way I want because the internal luminance format (again, I *think*, not know). I think the clim calculations in the shader are overflowing because they are internally representing unsigned integers instead of floats.

So...I'm not sure why this is. With the `else` block of code it should be scaling and inverting the data that goes into the texture. I've tried flipping the clims that get passed to the shader and still either an all black volume or an all white one.

I suppose your normalized clim is doing the inverse while the volume scaling in your `elif` is doing the scaling?

tlambert03:
> If you run your special elif calculations on this array you'll see that (0, 2000) becomes (0, 1), but if you run the else block of calculations you'll end up with (0, 2000) becomes (1, 0) which is what we want

there's enough variables that i feel like it all depends.  I guess my approach was indeed to invert the texture (silly as that may be?), but given the normalized_clims and the way the actual cmap is applied in the shader... I thought the visual outcome was still correct with my code, no?  were you able to find a case that didn't work with the code as it stands?  or is there a particular performance concern in there?

if you want, we could consider reverting back to the commit before adding the clim inversion feature and opening another PR for inverted clims?  Or do you feel like we're _that_ close ... and you'd like to have the inversion option?


> Good news is I hacked things enough that it worked with floating point textures in the GPU (format='r', internalformat='r32f'). That'll have to be another PR.

now that's exciting :) ... ultimately, I'd rather just use that than my little hack here, so if you have a WIP branch to share, I could potentially take a look at that.

djhoese:
> now that's exciting :) ... ultimately, I'd rather just use that than my little hack here, so if you have a WIP branch to share, I could potentially take a look at that.

I could throw one together. Would you rather this entire PR doesn't get merged into vispy and instead we just use the floating point textures for what you guys want in napari?

tlambert03:
> I could throw one together. Would you rather this entire PR doesn't get merged into vispy and instead we just use the floating point textures for what you guys want in napari?

I'm torn.  I can definitely see the benefit of using byte textures in terms of device support.  So I think that as long as vispy continues to support CPU-scaling to generate byte textures, it'd be nice for napari to be able to offer that mode to those with smaller/older GPUs... in which case this PR would still be useful even if we also enable the other mode, right?

djhoese:
Yes I think this PR offers a lot and is kind of an automatic performance improvement in most cases for most people with little unexpected results (except someone dealing with clim changes within the threshold and not seeing the bit depth they expect). I think this works as it should, but may have some redundant flipping of sign. How about I merge this, then this week (I'm 50% off this week) I'll work on another PR that will allow the floating point textures in the ImageVisual and VolumeVisual and napari folks can tell me what you think. Sound good?

tlambert03:
> How about I merge this, then this week (I'm 50% off this week) I'll work on another PR that will allow the floating point textures in the ImageVisual and VolumeVisual and napari folks can tell me what you think. Sound good?

yeah, that'd be great.  I'm with you that this isn't the most "stunning" bit of code :joy: ... but I do think it's "working" (behaviorally), and I'm more than happy to help refine it in the future.

I know we have some other applications that would really benefit from floating point textures as well.  that was on the list of things I wanted to get back to, so I'm more than happy to help push along whatever you get started.

djhoese:
Thanks @tlambert03

PR #1843, opened 2020-03-18
Add MacOS 10.14 and manylinux2014 wheel builds

Czaki:
Build wheels against more modern standards allow to use newer compilators and api which my produce better/faster shared libs. So I suggest to add builds of manylinux2014 and macos 10.14 wheels.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -17,7 +17,7 @@ variables:
   CIBW_BUILD_VERBOSITY: "2"
   CIBW_BEFORE_BUILD: "pip install -U numpy Cython jupyter ipywidgets"
   CIBW_BEFORE_BUILD_MACOS: "npm install npm@latest -g; pip install -U pip setuptools"
-  CIBW_BEFORE_BUILD_LINUX: "yum install -y fontconfig; pip install -U pip setuptools; if [ `getconf LONG_BIT` == '64' ]; then pip install https://files.pythonhosted.org/packages/f2/00/6f332e63b33d24dc3761916e6d51402a7a82dd43c6ca8a96e24dda32c6b5/freetype_py-2.1.0.post1-py2.py3-none-manylinux1_x86_64.whl; else pip install https://files.pythonhosted.org/packages/b8/ea/adfbeb3762a03c1d1c1ff751101efb1fddb7aadffa991d39bdb486d0557e/freetype_py-2.1.0.post1-py2.py3-none-manylinux1_i686.whl; fi"
+  CIBW_BEFORE_BUILD_LINUX: "yum install -y fontconfig; pip install -U pip setuptools; pip install freetype-py"
 jobs:
 - job: linux
   pool: {vmImage: 'Ubuntu-16.04'}
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,7 +45,6 @@ jobs:
         python -m pip install --upgrade pip
         pip install cibuildwheel
         cibuildwheel --output-dir wheelhouse .
-    - bash: rm wheelhouse/*manylinux1* || true
     - task: PublishPipelineArtifact@1
       inputs:
         path: $(System.DefaultWorkingDirectory)/wheelhouse
@@ -109,7 +108,7 @@ jobs:
       inputs:
         patterns: |
           vispyDeployLinux/*
-          vispyDeployLinux2014/*2014*
+          vispyDeployLinux2014/*2014*.whl
           vispyDeployMacOS/*.whl
           vispyDeployWindows/*.whl
           vispyDeployMacOS_10_14/*.whl
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -17,7 +17,7 @@ variables:
   CIBW_BUILD_VERBOSITY: "2"
   CIBW_BEFORE_BUILD: "pip install -U numpy Cython jupyter ipywidgets"
   CIBW_BEFORE_BUILD_MACOS: "npm install npm@latest -g; pip install -U pip setuptools"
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -17,7 +17,7 @@ variables:
   CIBW_BUILD_VERBOSITY: "2"
   CIBW_BEFORE_BUILD: "pip install -U numpy Cython jupyter ipywidgets"
   CIBW_BEFORE_BUILD_MACOS: "npm install npm@latest -g; pip install -U pip setuptools"
-  CIBW_BEFORE_BUILD_LINUX: "yum install -y fontconfig xvfb; pip install -U pip setuptools; if [ `getconf LONG_BIT` == '64' ]; then pip install https://files.pythonhosted.org/packages/f2/00/6f332e63b33d24dc3761916e6d51402a7a82dd43c6ca8a96e24dda32c6b5/freetype_py-2.1.0.post1-py2.py3-none-manylinux1_x86_64.whl; else pip install https://files.pythonhosted.org/packages/b8/ea/adfbeb3762a03c1d1c1ff751101efb1fddb7aadffa991d39bdb486d0557e/freetype_py-2.1.0.post1-py2.py3-none-manylinux1_i686.whl; fi"
+  CIBW_BEFORE_BUILD_LINUX: "yum install -y fontconfig; pip install -U pip setuptools; if [ `getconf LONG_BIT` == '64' ]; then pip install https://files.pythonhosted.org/packages/f2/00/6f332e63b33d24dc3761916e6d51402a7a82dd43c6ca8a96e24dda32c6b5/freetype_py-2.1.0.post1-py2.py3-none-manylinux1_x86_64.whl; else pip install https://files.pythonhosted.org/packages/b8/ea/adfbeb3762a03c1d1c1ff751101efb1fddb7aadffa991d39bdb486d0557e/freetype_py-2.1.0.post1-py2.py3-none-manylinux1_i686.whl; fi"
```
djhoese: Is `xvfb` not available? We may want to have this in the future if we run actual tests on Azure.

Czaki: it fail on centos7 docker image. I do not see anny difference in number of executed test. So I remove this instead of workaround. I can work on this.

I have other question. Why you install something with pip in this steep? This step is executed onece on begin and does not affect all python interpreters.

Czaki:
In my experience modern version of libc offers faster implementation of functions. But I do not have idea how to properly test how It change performance.

azure-pipelines.yml
djhoese: If I remember correctly it was that some of the vispy cython extensions need freetype and the freetype_py package (which we also use) had it bundled with. According to the docs this is run before each wheel build: https://cibuildwheel.readthedocs.io/en/stable/options/#before-build

djhoese: Oh and we install this specific version of freetype_py because for some reason `pip` wasn't giving us the right build when run as `pip install freetype-py`.

Czaki: Ok. My bad. I wrong remember execution order in cibuildwheel. When https://github.com/joerick/cibuildwheel/issues/54 will be finished then yum call can be executed onece.

I test with `pip install freetype-py` and build pass https://dev.azure.com/bokota/open%20source%20contrib/_build/results?buildId=419&view=results

Second observation. manylinux2014 produce wheel which is compatibile with manylinux1. So maybe not add manylinux2014 build but switch to it?

djhoese: Unless I'm wrong, manylinux2014 is for CentOS7 and above. Switching to only using 2014 would mean the wheels would no longer work/be available for CentOS 6.

djhoese: If you want to replace the freetype-py complicated if statement with the simpler pip install that would be fine with me.

djhoese: Do you know of a popular python package that is building linux wheels on something other than manylinux1 right now?

Czaki: if you look at result of manylinux2014 https://dev.azure.com/vispy/vispy/_build/results?buildId=369&view=artifacts&type=publishedArtifacts then you see that it produce also manylinux1. It works in this way because auditwheel checks if wheel is manylinux1 compatible. And if it is then produce two wheels. Also manylinux2014 is faster.

djhoese: Ok so you are saying if the `manylinux1` specific job is removed and `manylinux2014` is the *only* linux environment, then it should produce (and upload) a `manylinux2014` wheel *and* a `manylinux1` wheel? But that *only* happens if it is compatible right?

Czaki: in current configuration you not have manylinux1 job. You have manylinux2010 job. 

currently manylinux1 wheel is produced only if manylinux2010 wheel is compatibile with manylinux1. 

If you would like to have manylinux1 job you need to use:
```
  variables:
    CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
    CIBW_MANYLINUX_I686_IMAGE: manylinux1
```
in build specification. 

In this PR configuration. Both manylinux2010 and manylinux2014 produces manylinux1 wheel.

djhoese: Ah ok. So the question is then do we turn off the 2010 job and only have the 2014 job, right? Since the 2014 job is producing manylinux1 wheels that must mean they are compatible? Which means they'll work on CentOS 5+. That's ok with me then.

Czaki: current PR configuration produce `manylinux1`, `manylinux2010` and `manylinux2014` maybe `maylinux2010` is not needed. 

To be sure that `manylinux1`is produced something like this may be added.
```bash
ls *cp36*manylinux1* &&  ls *cp37*manylinux1* &&  ls *cp38*manylinux1* 
```

djhoese: Since it seems like `manylinux1` is currently being produced and our cython code rarely gets updated, I'm not too worried about making sure it is always created. This is with the assumption that if it ever stops getting produced (because it isn't compatible anymore) that CentOS5 will be long dead and CentOS6 will also be deprecated. I think you can get rid of the manylinux2010 based on what we've talked about here.

Czaki: ready

djhoese: Yep. Thanks again.

Czaki:
from release 1.0.0 cibuildwheel install python on windows, so I remove obsolete tasks.

> Windows now uses Python as installed by nuget, rather than the versions installed by the various CI providers. (joerick/cibuildwheel#180)

djhoese:
> from release 1.0.0 cibuildwheel install python on windows, so I remove obsolete tasks.

Very cool. In your opinion is this ready for full review and merge?

Czaki:
Yes

Czaki:
@djhoese You ask about packages without manylinux1. PyQt5 has only manylinux2014 wheels. PySide2 mark wheels as manylinux1, but author of PyQt5 told that new version of Qt5 depends on such new version of libc that it is fake marking and need quite new system to run.

djhoese:
Thanks again!

Issue #1850, opened 2020-04-17
Shader compilation error with #version 300 es


diricxbart:
The below code generates a rectangular PNG image with red on the left side and blue on the right side. The shaders for this are  implemented using both the default shader version as well as 300 es.

```
from vispy import gloo
from vispy import app
from vispy import io
import numpy as np


vertex = """
precision mediump float;

attribute vec2 a_position;
varying vec2 normalizedCoord;

void main(void)
{
    gl_Position = vec4(a_position, 0.0, 1.0);
    normalizedCoord = (a_position * 0.5 + 0.5).xy;
}
"""   

fragment = """
precision mediump float;

varying vec2 normalizedCoord;

void main()
{
    if (normalizedCoord.x <= 0.5) {
        gl_FragColor = vec4(1.0, 0.0, 0.0, 1.0);
    } else {
        gl_FragColor = vec4(0.0, 0.0, 1.0, 1.0);
    }
}

"""
    
vertex_300es = """
#version 300 es

precision mediump float;

in vec2 a_position;
out vec2 normalizedCoord;

void main(void)
{
    gl_Position = vec4(a_position, 0.0, 1.0);
    normalizedCoord = (a_position * 0.5 + 0.5).xy;
}
"""   

fragment_300es = """
#version 300 es

precision mediump float;

in vec2 normalizedCoord;
out vec4 fragmentColor;

void main()
{
    if (normalizedCoord.x <= 0.5) {
        fragmentColor = vec4(1.0, 0.0, 0.0, 1.0);
    } else {
        fragmentColor = vec4(0.0, 0.0, 1.0, 1.0);
    }
}

"""

class Canvas(app.Canvas):
    def __init__(self):
        h, w = (400, 800)
        
        app.Canvas.__init__(self, size=(w, h), keys='interactive')
        ps = self.pixel_scale

        self.program = gloo.Program(vert=vertex, frag=fragment)
        #self.program = gloo.Program(vert=vertex_300es, frag=fragment_300es)
        
        # gloo needs 32bit
        y = np.linspace(-1 + 2.0/h, 1, h)
        x = np.linspace(-1 + 2.0/w, 1, w)
        a, b = np.meshgrid(x, y)
        data = np.c_[a.flatten(), b.flatten()]
        self.program['a_position'] = data.astype('float32')

        self.show()

    def on_resize(self, event):
        width, height = event.size
        gloo.set_viewport(0, 0, width, height)

    def on_draw(self, event):
[...]

PR #1854, opened 2020-04-28
Allow for precision definitions in shaders

djhoese:
While looking at #1850 we realized that vispy is still removing `precision` definitions from shaders. As far as I can tell this has been this way since almost the beginning of vispy. The best guess from Almar was that some drivers may not have handled the precision identifiers back then even though it is part of the OpenGL ES 2.0 spec. It's been at least 5 years since that code was written so let's see if the tests pass and allow more complex shaders to work out of the box.

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -618,13 +618,9 @@ def _convert_desktop_shader(shader):
     for line in shader.lstrip().splitlines():
         line_strip = line.lstrip()
         has_version = has_version or line.startswith('#version')
-        if line_strip.startswith('precision '):
-            line = ''
         if line_strip.startswith('#extension'):
             extensions.append(line_strip)
             line = ''
-        for prec in (' highp ', ' mediump ', ' lowp '):
-            line = line.replace(prec, ' ')
         lines.append(line.rstrip())
     # Write
     # Make sure extensions are at the top, but after version
--- a/vispy/gloo/tests/test_glir.py
+++ b/vispy/gloo/tests/test_glir.py
@@ -13,6 +13,7 @@ import numpy as np
 
 
 def test_queue():
+    """Test GlirQueue and shader conversions."""
     q = glir.GlirQueue()
     parser = glir.GlirParser()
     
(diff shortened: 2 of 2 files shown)


almarklein:
It fails on one example, where (according to the comment) a precision qualifier was added on purpose to verify that it is removed. https://github.com/vispy/vispy/blob/master/examples/demo/gloo/fireworks.py#L67-L74

My best guess is that precision qualifiers are not available in GLSL 120. And this may well be the reason for stripping them (so perhaps not driver related like I said earlier). Can anyone else remember why we wrote most our examples with `#VERSION 120`? Maybe for compatibility with WebGL at the time?

djhoese:
I mentioned this in #1850, but it seems like `highp` is the only problem. With `highp` in that `fireworks.py` example it fails to run on my laptop in the same way as on travis. If I change all the `highp` to `mediump` then it works fine.

However, I'm realizing now why this exists and should have before. I thought `precision` was a newer GLSL feature, but it is actually only an ES feature that some versions of regular GLSL support for backwards compatibility. See https://www.khronos.org/opengl/wiki/Type_Qualifier_(GLSL)#Precision_qualifiers which mentions that precision is a no-op in desktop compilers. The bigger "bug" here is that vispy doesn't detect when you are using a `es` specific GLSL version in your shader like @diricxbart was doing in #1850. It relies on the "shader compatibility" of the GL backend (different than the GUI backend) you are using:

https://github.com/vispy/vispy/blob/8c0c58a3762f22d1530c1225ef7b32ff7feac229/vispy/gloo/glir.py#L721-L726

https://github.com/vispy/vispy/blob/a9f04f12027b56a444b827c9b8708ff57b1f1e5b/vispy/gloo/gl/__init__.py#L61-L83

As far as I can tell this `.es` functionality isn't actually used in VisPy. The only GL ES supported parser is the WebGL one (which provides its own `shader_compatibility` property).

I don't think this PR should be merged, but before I close it I'll ask some additional questions on #1850.

larsoner:
@djhoese do you want to rebase?

djhoese:
I'll have to look at this more closely. Re-reading comments I think this should be closed. I think my solution here didn't actually work. I'll have to look at this after the sprint probably. If I don't close this in 5 days it should be closed. (Can I tell github to send me a reminder email?)

djhoese:
After rereading this pull request's comments and the comments in #1850 I don't think this should be merged. I'll comment on #1850 with more details.

PR #1866, opened 2020-05-23
Fix jupyter lab extension to use newest vispy.js

mjlbach:
Bump ipython widget to match package.json

### DIFF
--- a/vispy/app/backends/ipython/_widget.py
+++ b/vispy/app/backends/ipython/_widget.py
@@ -48,10 +48,10 @@ def _stop_timers(canvas):
 class VispyWidget(DOMWidget):
     _view_name = Unicode("VispyView").tag(sync=True)
     _view_module = Unicode('vispy').tag(sync=True)
-    _view_module_version = Unicode('~0.2.0').tag(sync=True)
+    _view_module_version = Unicode('~0.3.0').tag(sync=True)
     _model_name = Unicode('VispyModel').tag(sync=True)
     _model_module = Unicode('vispy').tag(sync=True)
-    _model_module_version = Unicode('~0.2.0').tag(sync=True)
+    _model_module_version = Unicode('~0.3.0').tag(sync=True)
 
     #height/width of the widget is managed by IPython.
     #it's a string and can be anything valid in CSS.


djhoese:
Just so I have a better grasp on this and the failures you were seeing...were you running the master branch of this vispy repository? What was the error you were originally getting? I'm wondering if there is some piece of code that is installing `vispy@latest` that shouldn't be? I know we talked on gitter about needing to install `vispy@latest` and that was probably wrong. I wonder if things would have worked for you if you had forced the install of vispy.js 0.2.0 to match the python code?

mjlbach:
When I install the latest vispy from pypi
```
pip install -U vispy pytest jupyterlab
jupyter labextension install @jupyter-widgets/jupyterlab-manager vispy@0.2.0
````

I get the errror in the browser JS debugger:
```
Error: "Module vispy, semver range ~0.2.0 is not registered as a widget module"
```

When I look in the package.json, the version is 0.2.0, and the module version in the python extension is   _model_module_version = Unicode('~0.2.0').tag(sync=True), so I'm not exactly sure what is going on there. When I switched to master and applied the PR (along with locally installing the extension) it works.

**Comment by djhoese** [comment]
Hm that is odd. It seems like 0.2.0 should have worked. I was really hoping to not require people have vispy.js 0.3.0 since it is only adding some optional functionality. I have no problem merging this, just a little annoyed that it is needed :wink:

**Comment by mjlbach** [comment]
Understandable, I can try to debug what's going on with the 0.2.0 extension, I've been installing 0.3.0 from the git submodule (which is currently versioned at 0.3.0), so maybe it's something with building locally vs jupyter fetching the extension.

**Comment by mjlbach** [comment]
Ok figured out what happened, so the latest version of the 0.2.x extension, 0.2.2, has the following dependency in package.json which is causing it to not work with the latest jupyter

```
    "@jupyter-widgets/base": "^1.0.0 || ^2.0.0",
```

This was updated in 0f5a40e2bb7784b5121a2fc15184c8840100c7c8 to:

```
"@jupyter-widgets/base": "^1 || ^2 || ^3",
```

and building the above commit locally allows use on vispy 0.6.4 and vispy git. It looks like vispy@0.2.2 was made from the commit immediately before this (da24411), which explains why installing via jupyterlab yields a broken extension. If a 0.2.3 extension is made at 0f5a40e2bb7784b5121a2fc15184c8840100c7c8, I think everything will work without this PR (although, there are only two subsequent commits before 0.3.0)

djhoese:
Ok so because you were using jupyter widgets 3.x, you also needed the version of the vispy.js package that said it supported that version. Right?

I'll just need to get the next version of VisPy out soon.

PR #1870, opened 2020-06-02
emit transform_change on view_changed

Czaki:
Emit transform_change on view_changed to get information about view changed.


djhoese:
As mentioned on gitter, this shouldn't be needed. Could you create an issue with a minimal example that isn't producing the event you think it should? In a recently created example script we used:

https://github.com/vispy/vispy/blob/b5f6228c7c4ec734206e9556033f5c3e7941616c/examples/basics/scene/mesh_texture.py#L50-L53

Czaki:
This example clarify mi situation I try to connect to camera.events.transform_change, ans should connect to scene.transform.changed.

PR #1881, opened 2020-06-27
flip images vertically or horizontally

### LINKED ISSUE #1880: vispy.gloo.io.imread() don't has an argument for flipping the image
Textures are sometimes imported upside down, there is no **flip** argument to do so.

tushar5526:
Closes #1880 
- vispy.gloo.io.imread didn't had an option to flip the image.
- Used np.flip to flip it horizontally or vertically for both PIL and imageio

### DIFF
--- a/vispy/io/image.py
+++ b/vispy/io/image.py
@@ -156,7 +156,7 @@ def write_png(filename, data):
         f.write(_make_png(data))  # Save array with make_png
 
 
-def imread(filename, format=None):
+def imread(filename, format=None, flipVertically=False, flipHorizontally=False):
     """Read image data from disk
 
     Requires imageio or PIL.
@@ -167,6 +167,10 @@ def imread(filename, format=None):
         Filename to read.
     format : str | None
         Format of the file. If None, it will be inferred from the filename.
+    flipVertically : bool | False
+        Flip the image vertically, default is false
+    flipHorizontally : bool | False
+        Flip the image horizontally, default is false
 
     Returns
     -------
@@ -179,7 +183,12 @@ def imread(filename, format=None):
     """
     imageio, PIL = _check_img_lib()
     if imageio is not None:
-        return imageio.imread(filename, format)
+        img = imageio.imread(filename, format)
+        if flipHorizontally:
+            img = np.flip(img, 1)
+        if flipVertically:
+            img = np.flip(img, 0)
+        return img
     elif PIL is not None:
         im = PIL.Image.open(filename)
         if im.mode == 'P':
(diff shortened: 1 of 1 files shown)


djhoese:
Hi. Thanks for the contribution. I'm a little confused why this is needed inside this utility function. Why can't this be done by the user after the data is already loaded but before it is provided to another vispy function? I'm not sure I approve of this being added here unless you can give me more context for why this is needed here.

tushar5526:
I am doing a python version of www.learnopegl.com tutorials using vispy, so I had to load a texture following the tutorials. Textures are generally flipped when imported that's why I needed the option for flipping horizontally or vertically. I agree with your view, but for beginners who are new to vispy or not familiar much with numpy had to search for **np.flip**. Also, libraries in other languages have this option to flip the texture while loading. It would be quite handy to have this and better code readibility

```
img = np.flip(io.imread('img.jpg'), 0) or //1
```
```
img = io.imread('img.jpg', flipVertically=True)
```
I agree with your point too. You can keep this PR or close :)

djhoese:
> Also, libraries in other languages have this option to flip the texture while loading.

They do? Do you have examples? I think this `imread` function is modeled after the function by the same name in the scipy library which is why I'm hesitant to provide this. Are there use cases where someone would need to flip in both dimensions? Flipping it vertically (top becomes bottom) makes sense given different image conventions with where the "origin" of the image is, but horizontally sounds less likely.

Additionally if we were to include this the keyword arguments should follow PEP8 (standard) python naming and be something like `flip_vertically`.

tushar5526:
Yeah sorry for the confusion, they don't have the option in their 'imread' but they have some different function to flip the image. 

Yeah we can remove horizontal flip, I added it just in case if someone needs it, didn't put much thought into it.
I added these feature because it felt handy but we can go with np.flip too.

If you want it other way I will change the name according to PEP8 standards.
Thanks

djhoese:
Another thought: If you are seeing flipped images, this could also be caused by "bad" defaults for the camera you are using. Cameras have their own methods for flipping the axes. I think I'd still like to not include this in this function and just make users know that there is an easy way to flip the image either with the camera or with numpy. @kmuehlbauer thoughts?

kmuehlbauer:
>  I think I'd still like to not include this in this function and just make users know that there is an easy way to flip the image either with the camera or with numpy.

I totally agree here. This would create another point of possible misinterpretation of data layout.

djhoese:
Ok then I think it would be best to close this. @tushar5526 if you can't seem to make changes to the camera work for you or you can't find a way to add the numpy flip call in a way that isn't confusing to users, let us know. Maybe we can come up with a clean way to describe it.

PR #1883, opened 2020-06-30
Fix OpenGL to ctypes type mapping

cgohlke:
See https://www.khronos.org/opengl/wiki/OpenGL_Type

### DIFF
--- a/codegen/createglapi.py
+++ b/codegen/createglapi.py
@@ -223,7 +223,7 @@ HARDER_TYPES = {
                 'GLvoid*':('', 'c_void_p'),  # or c_voidp?
                 'GLvoid**':('', 'POINTER(ctypes.c_void_p)'),
                 'GLintptr':('', 'c_ssize_t'), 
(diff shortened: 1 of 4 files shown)


djhoese:
Interesting. Thanks. I don't think I've ever actually run this code myself. And just to clarify, this is functionally the same (c_int versus c_ssize_t) in most cases/systems, right?

jni:
My interpretation is the same as yours @djhoese: resolving to the same hardware type in most systems, meaning it masks a logic error in most systems. =)

djhoese:
@kmuehlbauer you recently played with the codegen, any concerns about this?

kmuehlbauer:
We would need to rerun the codegen to make this permanent. Should this be committed to this PR?

djhoese:
Probably. Give me a sec. I'll do it since it should be merged with the changes on master from a couple days ago.

djhoese:
My browser shows one of two travis tasks (?) is queued but when I click on it it all passes. Thanks @cgohlke. Merging now.

Issue #1878, opened 2020-06-21
Slow TurntableCamera


povik:
Hello!

While trying out vispy (v0.6.4), I set up two view boxes with linked turntable cameras, like so:

```
from vispy import app, gloo, visuals, scene, io
canvas = scene.SceneCanvas(keys='interactive')
canvas.size = 800, 600
canvas.show()
vb1 = scene.widgets.ViewBox(border_color='yellow', parent=canvas.scene)
vb2 = scene.widgets.ViewBox(border_color='blue', parent=canvas.scene)
grid = canvas.central_widget.add_grid()
grid.padding = 6
grid.add_widget(vb1, 0, 0)
grid.add_widget(vb2, 0, 1)
vb1.camera, vb2.camera = scene.TurntableCamera(), scene.TurntableCamera()
vb1.camera.link(vb2.camera)
```

I found the result to be sluggish in that there's noticeable delay between mouse input and redraw when manipulating the camera. Looking at `--vispy-cprofile` output, I noticed there's significant time spent in calculating matrix inverses:

```
         3229935 function calls (2994911 primitive calls) in 10.236 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    9.277    9.277 _default_app.py:58(run)
        1    0.000    0.000    9.277    9.277 application.py:135(run)
        1    0.000    0.000    9.277    9.277 _qt.py:274(_vispy_run)
        1    1.691    1.691    9.277    9.277 {built-in method exec_}
  267/210    0.008    0.000    7.599    0.036 _qt.py:507(event)
  267/210    0.043    0.000    7.590    0.036 {function QtBaseCanvasBackend.event at 0x7f1234a18560}
32621/887    0.274    0.000    7.549    0.009 event.py:405(__call__)

...

     5642    0.036    0.000    2.637    0.000 linear.py:403(inv_matrix)
     5638    0.012    0.000    2.599    0.000 <__array_function__ internals>:2(inv)
    12149    0.225    0.000    2.574    0.000 function.py:184(__setitem__)
     5638    2.414    0.000    2.571    0.000 linalg.py:482(inv)
      704    0.012    0.000    1.919    0.003 turntable.py:137(_rotate_tr)
     1408    0.011    0.000    1.757    0.001 linear.py:446(rotate)
```

[...]

PR #1884, opened 2020-06-30
Speed up arcball and turntable cameras

povik:
In Base3DRotationCamera, instead of gradually constructing the view
matrix inside MatrixTransform, do prepare the final matrix first, then
assign it to MatrixTransform.

This is done to avoid the relatively costly operations each time the
matrix inside MatrixTransform changes.

See #1878.

### DIFF
--- a/vispy/scene/cameras/arcball.py
+++ b/vispy/scene/cameras/arcball.py
@@ -6,6 +6,8 @@ from __future__ import division
 
 import numpy as np
 
+
+from ...util import transforms
 from ...util.quaternion import Quaternion
 from ...visuals.transforms import MatrixTransform
 from .perspective import Base3DRotationCamera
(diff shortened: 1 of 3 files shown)


povik:
Well, as you say, I took care of arcball and turntable cameras, which are the only users of Base3DRotationCamera inside vispy. I see `vispy.scene.cameras.perspective.Base3DRotationCamera` is not in the API reference, and so consider it ok to break. But that may not be vispy's policy.

djhoese:
>  I see vispy.scene.cameras.perspective.Base3DRotationCamera is not in the API reference, and so consider it ok to break. But that may not be vispy's policy.

I agree with that (or at least I will when users start filing bugs about the next version of vispy breaking their code).

djhoese:
Thanks again!

povik:
Happy to contribute!

PR #1886, opened 2020-06-30
Let camera link be limited to specified properties

povik:
I suggest adding an argument to BaseCamera.link letting the user specify the camera properties to be linked.

The use case I have in mind is two turntable cameras with only the elevation and azimuth, not the scale_factor, being linked.

### DIFF
--- a/vispy/scene/cameras/base_camera.py
+++ b/vispy/scene/cameras/base_camera.py
@@ -49,7 +49,7 @@ class BaseCamera(Node):
         self._viewbox = None
 
         # Linked cameras
-        self._linked_cameras = []
+        self._linked_cameras = {}
         self._linked_cameras_no_update = None
 
         # Variables related to transforms
@@ -340,7 +340,7 @@ class BaseCamera(Node):
                 raise KeyError('Not a valid camera state property %r' % key)
             setattr(self, key, val)
 
-    def link(self, camera):
+    def link(self, camera, props=None):
         """ Link this camera with another camera of the same type
 
         Linked camera's keep each-others' state in sync.
@@ -349,16 +349,25 @@ class BaseCamera(Node):
         ----------
         camera : instance of Camera
             The other camera to link.
+        props : list of strings | tuple of strings | None
+            List of camera state properties to keep in sync between
+            the two cameras. If None, all of camera state is kept in sync.
         """
+        if props is None:
+            props = self._state_props
         cam1, cam2 = self, camera
+        # Check both cameras have props
+        available_props = set(cam1._state_props).intersection(set(cam2._state_props))
+        if not set(props).issubset(available_props):
+            raise KeyError('Not a state property of both cameras: %s' % (set(props) - available_props))
         # Remove if already linked
         while cam1 in cam2._linked_cameras:
-            cam2._linked_cameras.remove(cam1)
+            del cam2._linked_cameras[cam1]
         while cam2 in cam1._linked_cameras:
-            cam1._linked_cameras.remove(cam2)
+            del cam1._linked_cameras[cam2]
         # Link both ways
-        cam1._linked_cameras.append(cam2)
-        cam2._linked_cameras.append(cam1)
+        cam1._linked_cameras[cam2] = props
+        cam2._linked_cameras[cam1] = props
 
     ## Event-related methods
 
@@ -472,6 +481,10 @@ class BaseCamera(Node):
                 continue
             try:
                 cam._linked_cameras_no_update = self
-                cam.set_state(self.get_state())
+                state = self.get_state()
+                # Only apply state props we were asked to link
+                linked_props = self._linked_cameras[cam]
+                state_filt = { prop: state[prop] for prop in linked_props }
+                cam.set_state(state_filt)
             finally:
                 cam._linked_cameras_no_update = None
--- a/vispy/scene/cameras/base_camera.py
+++ b/vispy/scene/cameras/base_camera.py
@@ -333,14 +333,14 @@ class BaseCamera(Node):
         """
         if props is None:
             props = self._state_props
-        D = {}
+        state = {}
         for key in props:
             # We support tuple keys to accomodate camera linking.
             if isinstance(key, tuple):
-                D[key] = nested_getattr(self, key)
+                state[key] = nested_getattr(self, key)
             else:
-                D[key] = getattr(self, key)
-        return D
+                state[key] = getattr(self, key)
+        return state
 
     def set_state(self, state=None, **kwargs):
         """ Set the view state of the camera
@@ -357,8 +357,8 @@ class BaseCamera(Node):
             Unused keyword arguments.
         """
 
-        D = state or {}
-        D.update(kwargs)
+        state = state or {}
+        state.update(kwargs)
   
         # In first pass, process tuple keys which select subproperties. This
         # is an undocumented feature used for selective linking of camera state.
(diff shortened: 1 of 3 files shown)


djhoese:
Interesting idea. This PR will need to be coordinated with #1876 by @Bliss3d otherwise we're going to have a not so fun time merging both of them. Maybe there is a chance these can be combined and they will end in an overall cleaner solution.

povik:
I see. One option that comes to my mind of combining the two is extending the `props` argument to accept property names with subattributes, e.g. `["rect.pos[1]", "rect.top", "rect.bottom"]` to link the Y axis as @Bliss3d would like to. Then we could implement the `axis` argument as proposed in the other PR as a shortcut adding the right attribute paths to `props`.

djhoese:
Oh that could work. I'd be worried about it getting pretty complicated parsing the properties for sub-properties. @bliss3d what are your thoughts on this?

Bliss3d:
At the moment i need to learn for a bunch of exams, so i dont have a much time. @povik do you have a working example, you could share? I'd like to check it out. 
I am working on plotting (with updates) ohlcv trading data, i was going to add it to the repo some time

povik:
I will prepare something. Not necessarily with arbitrary subattributes as I described in my previous comment, but it will be a version of the link method that supports both our needs.

povik:
@djhoese @Bliss3d Take a look at the new commits. Linking now supports the `axis=` argument from the other PR. I made changes to `set_state` and `get_state` and implemented subproperties through tuple keys. I think it is best we keep the support for subproperties in `get_state`, `set_state` and `link` undocumented, as it only works for root properties with such type that its constructor can be used to make copies of the object, as is the case for `Rect`. Let me know what you think.

povik:
@djhoese I am glad you approve. I have now added some tests.

djhoese:
Nice. Thanks. The travis tests are failing because of some flake8 issues:

```
Running flake8... 
vispy/scene/cameras/tests/test_link.py:6:1: F401 'sys' imported but unused
vispy/scene/cameras/tests/test_link.py:8:1: F401 'pytest' imported but unused
vispy/scene/cameras/tests/test_link.py:12:1: F401 'vispy.geometry.Rect' imported but unused
vispy/scene/cameras/tests/test_link.py:55:33: E201 whitespace after '('
Failed: flake8 failed
```

Could you try fixing those?

povik:
Oh, okay. That should be fixed now.

djhoese:
@Bliss3d What do you think about this solution? Do you think you could have a couple minutes to test @povik's branch with your use case? As-is, I think I'm OK with merging this assuming it works. @kmuehlbauer let me know if you have any thoughts.

kmuehlbauer:
@djhoese Looks very elegant! I do not want to be too nit-picking, but I prefer a more speaking name than `D` for the dict, Why not use `state`, @povik?

vispy/scene/cameras/base_camera.py
```diff
@@ -307,16 +318,28 @@ def set_default_state(self):
...
+
+        Parameters
+        ----------
+        props : list of strings | None
+            List of properties to include in the returned dict. If None,
+            all of camera state is returned.
         """
+        if props is None:
+            props = self._state_props
         D = {}
```
kmuehlbauer, line 336: As already mentioned, why not use `state` instead of `D`?

vispy/scene/cameras/base_camera.py
```diff
@@ -333,14 +356,38 @@ def set_state(self, state=None, **kwargs):
         **kwargs : dict
             Unused keyword arguments.
         """
+
         D = state or {}
```
kmuehlbauer, line 360: Same here, I suggest to use `state` instead of `D`.

povik:
OK. I will be AFK until Friday, then I will do those renamings.

Bliss3d:
can confirm working with #1876, could add my script as an example

kmuehlbauer:
@Bliss3d We could wait until this is merged and then you can update your PR #1876, adding the example.

Bliss3d:
The example is not that sophisticated, its just two viewboxes and link(axis="x")

povik:
@kmuehlbauer Done.

kmuehlbauer:
@povik Thanks for your contribution. The error on windows looks unrelated (font-stuff). @djhoese will have a last look later, before merge.

djhoese:
Freetype-py just had a release yesterday: https://pypi.org/project/freetype-py/
Very likely caused by this PR: https://github.com/rougier/freetype-py/pull/118

@larsoner was the last person to work on the font tests I think. @rougier is one of the freetype-py maintainers. Any ideas guys why this isn't working on Azure: https://dev.azure.com/vispy/vispy/_build/results?buildId=475&view=logs&jobId=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=168f295b-0553-5364-35f7-923225ecd8b3

Actually, I just notice we aren't explicitly installing freetype-py for Windows, but it is likely still being installed given freetype-py's release and the start of these failures.

rougier:
For windows, you need to install freetype-py with specific a flag to install the DLL. That might be the cause. Not  sure why it worked before though.

djhoese:
@rougier Looks like the windows wheel either doesn't exist or isn't being found. Here's the log from the last passing build:

```
2020-07-09T15:45:16.8204087Z Processing c:\users\vssadm~1\appdata\local\temp\cibuildwheelm_u_eoov\repaired_wheel\vispy-0.6.5.dev174+gbb2cd96e-cp36-cp36m-win32.whl
2020-07-09T15:45:18.9945781Z Collecting freetype-py
2020-07-09T15:45:19.0751362Z   Downloading freetype_py-2.1.0.post1-py2.py3-none-win32.whl (574 kB)
```

Here's the same bit from the most recent failing:

```
2020-07-10T06:49:12.4882365Z Processing c:\users\vssadm~1\appdata\local\temp\cibuildwheellsdl0shy\repaired_wheel\vispy-0.6.5.dev180+g0d327773-cp36-cp36m-win32.whl
2020-07-10T06:49:15.1888439Z Collecting freetype-py
2020-07-10T06:49:15.2727931Z   Downloading freetype-py-2.2.0.zip (827 kB)
2020-07-10T06:49:15.8486322Z   Installing build dependencies: started
2020-07-10T06:50:12.2919886Z   Installing build dependencies: finished with status 'done'
2020-07-10T06:50:12.2970646Z   Getting requirements to build wheel: started
2020-07-10T06:50:12.6828915Z   Getting requirements to build wheel: finished with status 'done'
2020-07-10T06:50:12.6908656Z     Preparing wheel metadata: started
2020-07-10T06:50:13.3316555Z     Preparing wheel metadata: finished with status 'done'
2020-07-10T06:50:14.1620947Z Collecting numpy
2020-07-10T06:50:14.2129283Z   Using cached numpy-1.19.0-cp36-cp36m-win32.whl (10.9 MB)
2020-07-10T06:50:15.1307165Z Building wheels for collected packages: freetype-py
2020-07-10T06:50:15.1372346Z   Building wheel for freetype-py (PEP 517): started
2020-07-10T06:50:16.2299631Z   Building wheel for freetype-py (PEP 517): finished with status 'done'
2020-07-10T06:50:16.2403682Z   Created wheel for freetype-py: filename=freetype_py-2.2.0-py3-none-any.whl size=74530 sha256=5c37cc337739c85d5830728a4aae1712751a800d80e337fab67535cae6f1ea2e
2020-07-10T06:50:16.2405529Z   Stored in directory: c:\users\vssadministrator\appdata\local\pip\cache\wheels\44\34\68\ca9d7179dccd4c5150e7feb3c79ca24c44d8ca825f79a16d2c
2020-07-10T06:50:16.2417905Z Successfully built freetype-py
2020-07-10T06:50:16.3413257Z Installing collected packages: freetype-py, numpy, vispy
2020-07-10T06:50:29.1396677Z Successfully installed freetype-py-2.2.0 numpy-1.19.0 vispy-0.6.5.dev180+g0d327773
```

I know we had an issue with this in the past where we had to hardcode the exact wheel URL to install because pip wasn't finding the wheel. Eventually we removed the hardcode and it still "just worked". Any ideas? I can specify the environment variable so the sdist (.zip) pulls the .dll in with it, but we should be getting the wheel.

rougier:
In the release note: Drop Python 2 support and 32 bit wheels.

djhoese:
Ah and I see that it was just recently fixed to declare the python 2 part of that: https://github.com/rougier/freetype-py/commit/f09dd56b21ae1779161baef17af582d6b221f227#diff-2eeaed663bd0d25b7e608891384b7298

Is there a way to declare that pip shouldn't install a version for 32-bit environments? And is it that you dropped support for 32-bit or that you aren't providing wheels for 32-bit?

rougier:
Ony the wheels

jni:
@djhoese what's the status here? Could this get merged and the failing CI fixed separately?

djhoese (approved):
This looks good. Any issues I have with this are very small and a matter of opinion. You made a really good PR and I've made you wait long enough. Let's merge this.

PR #1888, opened 2020-07-02
Delay testing import with proxy function

### LINKED ISSUE #1887: Pytest missed in dependecies
in top level `__init__` file there is testing import 

https://github.com/vispy/vispy/blob/b5f6228c7c4ec734206e9556033f5c3e7941616c/vispy/__init__.py#L32

which import `testing._testing`

https://github.com/vispy/vispy/blob/b5f6228c7c4ec734206e9556033f5c3e7941616c/vispy/testing/__init__.py#L43-L51

which import pytest 

https://github.com/vispy/vispy/blob/b5f6228c7c4ec734206e9556033f5c3e7941616c/vispy/testing/_testing.py#L14

It looks intentional (`# noqa` after import) but pytest is not listed in package requirements in `setup.cfg` 

IT should be fixed, or `pytest` should be added to package requirements?

Czaki:
Fixes #1887 with delay import  of testing module.


vispy/__init__.py
```diff
@@ -29,13 +29,20 @@
 
 from .util import config, set_log_level, keys, sys_info  # noqa
 from .util.wrappers import use  # noqa
```
djhoese, line 31: I wonder if this `test` function should be put in `util.wrappers` based on the docstring for that module (meant to avoid circular dependencies and other similar issues). Thoughts?

Czaki: You mean about moving whole test function, or only wraper?

djhoese: Moving the `test` wrapper you created here in `vispy/__init__.py` into `vispy/util/wrappers.py`.

djhoese:
Travis is failing with some flake errors:

```
Running flake8... 
vispy/util/check_environment.py:57:15: W292 no newline at end of file
vispy/util/wrappers.py:159:22: E261 at least two spaces before inline comment
Failed: flake8 failed
```

Czaki:
Done

PR #1891, opened 2020-07-09
Fix typo

BioGeek:


kmuehlbauer:
@BioGeek Thanks for catching this.

djhoese:
Thanks. @kmuehlbauer we should really do better flake8 docstring checks (even go as far as black).

PR #1911, opened 2020-09-06
Fix ImageVisual not updating color transform after texture update

tlambert03:
fixes a bug resulting from #1844  that can cause inconsistent contrast (only in specific cases) in when rebuilding an image texture with `ImageVisual._build_texture()`  (originally observed/described at napari/napari#1585)

### DIFF
--- a/vispy/visuals/image.py
+++ b/vispy/visuals/image.py
@@ -487,6 +487,7 @@ class ImageVisual(Visual):
                 self._clim = (0, 1)
 
         self._texture_limits = np.array(self._clim)
+        self._need_colortransform_update = True
         self._texture.set_data(data)
         self._need_texture_upload = False
 


djhoese:
So to put this another way, the color transform needs to be rebuilt because updating the texture can sometimes update the clims. Right?

tlambert03:
> So to put this another way, the color transform needs to be rebuilt because updating the texture can sometimes update the clims. Right?

yeah.  specifically, because we're always scaling the texture to 0-1 prior to sending it to the GPU, the `normalized_clims` has to relative to the `texture_limits` used for the last texture uploaded to the GPU.  So if `texture_limits` has possibly changed, we have to make sure to call `self.shared_program.frag['color_transform'] = _build_color_transform(...)` during the next draw.

djhoese:
Makes sense. Thanks!

PR #1912, opened 2020-09-08
Add 'texture_format' kwarg to VolumeVisual for floating point textures

djhoese:
See #1842 for a longer discussion on this topic.

Currently the VolumeVisual must scale the provided data to fit in an 8-bit integer. This is needed due to limits in OpenGL 2.0/ES but causes some major performance limitations. Data must be scaled for any color limit changed which means not only recalculations of the data but reupload of it to the GPU. Data must also be copied if the user does not want their original data array modified.

Newer versions of OpenGL allow for non-integer texture formats and this PR tries to make them available to the VolumeVisual. A future PR will do the same thing for the ImageVisual.

Some issues:

1. I don't have any tests that specifically check that this behavior is working and I'm not really sure how to do it without writing tests that test implementation rather than actual functionality.
2. I really don't like the keyword argument being a boolean or the name of it `texture_float`. The VolumeVisual is relatively limited (I have limited experience with it as well so someone correct me) in what formats it accepts as input data, but it seems a luminance (single color value) arrays are all that are accepted. That means that the only logical in-GPU format is really R32F. There are other formats mentioned [here](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexImage2D.xhtml) but I'm not sure if someone would really want to use them (maybe the compressed formats). The ImageVisual is another story as someone may have a lot of different formats they want to use, but maybe that doesn't matter and should require subclassing by the user. I'd like VolumeVisual and ImageVisual to have the same keyword argument for this if possible. Opinions welcome.

CC @sofroniewn @tlambert03


vispy/visuals/volume.py
```diff
@@ -416,20 +416,27 @@ class VolumeVisual(Visual):
...
     interpolation : {'linear', 'nearest'}
-        Selects method of image interpolation. 
+        Selects method of image interpolation.
+    texture_float : bool
+        Whether to store data on the GPU as floating point data or as scaled
+        integers. Floating point data may not be supported on older versions
+        of OpenGL or with older GPUs. Default is ``False`` to scale data on the
+        CPU and store the result in the GPU. If True, data is scaled on the
+        GPU which allows for faster color limit changes. Additionally, when
+        32-bit float data is provided it won't be copied.
```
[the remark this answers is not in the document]
djhoese: Agreed. Thanks.

kmuehlbauer:
>     2\. I really don't like the keyword argument being a boolean or the name of it `texture_float`.

Not sure, if this makes sense, but could we use `internalformat` for this?  Then it would conform with the texture classes.

djhoese:
@kmuehlbauer I thought about that. The downside is we are bringing low-level concepts into the high-level interfaces. With that keyword argument name would we expect users to pass `'r32f'` or something else?

kmuehlbauer:
@djhoese Yes, this keyword would be `None` as it is now for the Texture-classes. Those who need this feature would use `internalformat='r32f'`. For most of the users nothing would change. But you made a point there and I have no other suggestion.

djhoese:
I talked to a coworker about this. What about `texture_format` that takes a numpy dtype and makes the best decision for the user. Additionally, advanced users could provide the internalformat string if they wanted to do something fancy like compressed texture formats. Thoughts?

kmuehlbauer:
@djhoese `texture_format` looks good, way better than `texture_float` and `internalformat`. :+1:

tlambert03 (approved):
This is _awesome_!  I just plugged it into napari, and it works great.  It's way better than my previous solution for adjust clims & gamma rapidly (particularly when _increasing_ clims from a small range to a larger range ... and the `texture_float=False` fallback continues to work nicely for those with smaller GPUs.

thanks @djhoese!  Would you consider a PR that mimics this behavior for the `ImageVisual`?

(I don't have a strong opinion on the keyword arguments, so fine with whatever you decide there)

vispy/visuals/volume.py
```diff
@@ -572,6 +592,26 @@ def texture_is_inverted(self):
...
+    def _check_clim_requires_rescale(self, clim):
+        if self._texture_float:
+            return False
+        if self._check_clims_in_texture_range(clim):
+            return True
+        if self._check_clims_in_threshold_range(clim):
+            return True
+        return False
+
     @clim.setter
```
tlambert03, line 615: looks like these got split in a previous PR... but it might be nice to bring `@clim.setter` back together with the original `@property` getter declaration

djhoese: You mean that visually the two parts of the `clim` property are not together? Yeah I can move that.

djhoese:
@tlambert03 any opinions against `texture_format` as described above? As for `ImageVisual`, read the original description. That was going to be my next plan.

tlambert03:
> @tlambert03 any opinions against `texture_format` as described above? 

nope, I think I probably like it best of the options mentioned

> As for `ImageVisual`, read the original description. That was going to be my next plan.

whoops, missed it.  great!

vispy/visuals/volume.py
tlambert03: yep... some linters will actually yell at you for that as well (for instance, flake8 is telling me this is a redefinition, unless I put them together)

djhoese:
@kmuehlbauer @larsoner @tlambert03 Ok I've made some updates. It now uses `texture_format` and I kind of blindly/naively made it that if you provide the format at all (as a numpy dtype or string) that it lets you do it. I included dtype conversion that may or may not work; I've never actually tested them (I plan to do a little testing). I made it so it assumes that if you specify the format you don't want it scaled on the CPU. I was thinking this might make it possible for you to provide uint8 data and then not perform any scaling of it. Just let the GPU take the clims and do what it needs to to clip them. Maybe this is too much of an assumption and I should just stick with float32 support only?

tlambert03:
> It now uses `texture_format` and I kind of blindly/naively made it that if you provide the format at all (as a numpy dtype or string) that it lets you do it. I included dtype conversion that may or may not work; I've never actually tested them (I plan to do a little testing)

just tested with a handful of data types (via napari) and it works great.  I like this more than `texture_float`.

i know someone can always do this manually, but what would you think of one more special case for `texture_format='auto'` ... where:
```python
if texture_format == 'auto':
    texture_format = vol.dtype.type
```

As an aside, I have a `ComplexImageVisual` class working locally to display complex data as mag/phase/real/imag.  I can see that fitting in nicely with either `texture_format = 'auto'` or `np.complex64`

vispy/visuals/volume.py
```diff
@@ -495,6 +548,7 @@ def set_data(self, vol, clim=None, copy=True):
             Colormap limits to use. None will use the min and max values.
         copy : bool | True
             Whether to copy the input volume prior to applying clim normalization.
+            Has no effect if visual was created with 'texture_float' set to True.
```
tlambert03, line 551:
```suggestion
            Has no effect if visual was created with 'texture_format' not `None`.
```

djhoese:
Interesting idea. Would make it easy to change the default in the future. For complex numbers, what format does that map to for `internalformat`? Or how do you expect that mapping?

tlambert03:
> For complex numbers, what format does that map to for internalformat? Or how do you expect that mapping?

The only way I’ve done it so far is to use two r32f textures (real/imag). Then I add a complex_mode parameter that decides how to combine the two numbers (abs, angle, real, imag) that then goes through clim, gamma, and cmap.   I’ve got ideas/plans for color-based mappings too, but we can discuss that elsewhere.

vispy/visuals/volume.py
```diff
@@ -458,8 +491,16 @@ def __init__(self, vol, clim=None, method='mip', threshold=None,
...
+        tex_kwargs = {}
+        if texture_format and not isinstance(texture_format, str):
+            if texture_format not in self._texture_dtype_format:
+                raise ValueError("Can't determine internal texture format for '{}'".format(texture_format))
+            texture_format = self._texture_dtype_format[texture_format]
+        if isinstance(texture_format, str):
+            tex_kwargs['internalformat'] = texture_format
+            tex_kwargs['format'] = 'luminance'
+        self._tex = tex_cls((10, 10, 10), interpolation=self._interpolation,
+                            wrapping='clamp_to_edge', **tex_kwargs)
```
jni, line 503: Reading this as a naive newcomer, this line is very mysterious. What is (10, 10, 10)? Why clamp to edge?

vispy/visuals/volume.py
```diff
@@ -495,6 +548,7 @@ def set_data(self, vol, clim=None, copy=True):
             Colormap limits to use. None will use the min and max values.
         copy : bool | True
             Whether to copy the input volume prior to applying clim normalization.
+            Has no effect if visual was created with 'texture_format' not equal to None.
```
jni, line 551: This bit is surprising. How do we update a visual if we want to change the data? In napari, for example, when we move the slider, e.g. to get a different timepoint of a 3D volume, we call `.set_data()`. We would still want to do this even if we are using this new zero-CPU-copy method.

... And anyway from my reading of the code below, this method *does* update the texture value even in that case?

jni (approved):
@djhoese this PR makes me so so happy. =) It's actually critical functionality for some parts of napari, like displaying segmentations. (Which are currently super hacky.) I've left a couple of comments where things were unclear to me.

vispy/visuals/volume.py
tlambert03: @jni, this isn't so much about whether the visual gets update by `.set_data()` (it always does).  It's whether the data array needs to be copied before scaling it prior to GPU upload (so as not to mutate the input data).  You may recall in https://github.com/napari/napari/issues/575#issuecomment-544753893, we learned that the `VolumeVisual` was mutating data.  So this copy step was added in https://github.com/vispy/vispy/commit/876a0d3 ... but if the raw data is uploaded to the GPU texture... it's no longer needed! 🎉

@djhoese  ... this also reminds me: technically, we no longer need to retain the `self._last_data` pointer if `self._scale_texture_gpu is True`, since it will never be used in that case.

djhoese: @jni to clarify, this "Has no effect" statement refers only to the `copy` keyword argument, not `set_data` as a whole. I'll reword some of it.

vispy/visuals/volume.py
djhoese: I'll add some comments. I think the 10 is just a placeholder size so that the texture is created. It will get resized when actual data is uploaded. The clamp_to_edge...I don't remember.

vispy/visuals/volume.py
djhoese: Updated.

vispy/visuals/volume.py
djhoese: Separate the texture creation out and added some comments so hopefully this is a little more clear.

vispy/visuals/volume.py
jni: Right, sorry for the noise, the newline messed with my brain's numpydoc parser. 😂 🤦 😂

vispy/visuals/volume.py
jni: Awesome, thank you!

djhoese:
@tlambert03 is this last commit for 'auto' what you had in mind?

tlambert03:
Yep, exactly.  Whole thing looks great to me!

djhoese:
Great! I think I'll make a separate PR for the ImageVisual and wait for that one to work out before merging this in case something needs to change in the interfaces. I'd like them to be consistent or at least kind of similar.

vispy/visuals/volume.py
```diff
@@ -483,7 +518,40 @@ def __init__(self, vol, clim=None, method='mip', threshold=None,
...
+        # set later this will be resized.
+        # clamp_to_edge means any texture coordinates outside of 0-1 should be
+        # clamped to 0 and 1.
+        return tex_cls((10, 10, 10), interpolation=self._interpolation,
+                       wrapping='clamp_to_edge', **tex_kwargs)
+
+    def _cpu_scale_data(self, vol):
+        if self._clim[1] == self._clim[0]:
+            if self._clim[0] != 0.:
+                vol *= 1.0 / self._clim[0]
```
rougier, line 546: Why note `vol /= self._clim[0]` ?

vispy/visuals/volume.py
```diff
@@ -483,7 +518,40 @@ def __init__(self, vol, clim=None, method='mip', threshold=None,
...
+        elif self._clim[0] > self._clim[1]:
+            vol *= -1
+            vol += self._clim[1]
+            vol /= self._clim[1] - self._clim[0]
```
rougier, line 550: `self._clim[1] - self._clim[0]` is negative and will reverse sign of vol, is that on purpose ?

vispy/visuals/volume.py
```diff
@@ -416,23 +416,55 @@ class VolumeVisual(Visual):
...
+        formatted data (single band). This is equivalent to `GL_RED` format
+        in OpenGL 4.0.
     """
 
     _interpolation_names = ['linear', 'nearest']
-
-    def __init__(self, vol, clim=None, method='mip', threshold=None, 
+    _texture_dtype_format = {
+        np.float32: 'r32f',
+        np.float64: 'r32f',
```
rougier, line 454: This will induce precision lost, does this need a warning or exception  ? Is it allowed ?

vispy/visuals/volume.py
```diff
@@ -483,7 +518,40 @@ def __init__(self, vol, clim=None, method='mip', threshold=None,
         self.relative_step_size = relative_step_size
         self.threshold = threshold if (threshold is not None) else vol.mean()
         self.freeze()
-    
+
+    def _create_texture(self, texture_format, emulate_texture, data):
+        tex_cls = TextureEmulated3D if emulate_texture else Texture3D
+
+        tex_kwargs = {}
+        if isinstance(texture_format, str) and texture_format == 'auto':
```
rougier, line 526: These 3 tests could probably be factorized. It is made of three consecutive conditionals making it hard to see how texture format is transformed

vispy/visuals/volume.py
```diff
@@ -558,6 +626,31 @@ def rescale_data(self):
...
+    def _check_clims_in_threshold_range(self, clim):
+        # if the clim range is too small of a percentage of the last-used texture range,
+        # posterization may be visible, so downscale the texture range.
+        range_ratio = np.subtract(*clim) / abs(np.subtract(*self._texture_limits))
+        return np.abs(range_ratio) < self._clim_range_threshold
+
+    def _check_clim_requires_rescale(self, clim):
+        if self._scale_texture_gpu:
+            return False
+        if self._check_clims_in_texture_range(clim):
```
rougier, line 648: elif

vispy/visuals/volume.py
```diff
@@ -558,6 +626,31 @@ def rescale_data(self):
...
+            return True
+        if self._check_clims_in_threshold_range(clim):
```
rougier, line 650: elif

vispy/visuals/volume.py
```diff
@@ -585,32 +673,25 @@ def clim(self, value):
...
-                return
-        else:
-            if (clim[0] < self._texture_limits[0]) or (clim[1] > self._texture_limits[1]):
-                self.rescale_data()
-                return
-        # if the clim range is too small of a percentage of the last-used texture range,
-        # posterization may be visible, so downscale the texture range.
-        range_ratio = np.subtract(*clim) / abs(np.subtract(*self._texture_limits))
-        if np.abs(range_ratio) < self._clim_range_threshold:
+        if self._check_clim_requires_rescale(clim):
```
rougier, line 676: Comment for above but we should accept a tuple. An array is a bit overkill.
`try: cmin, cmax = value except: ...`

rougier (changes requested):
I've started a review and left some comments.

vispy/visuals/volume.py
djhoese: This was in the original code and was just copied to a new method. I can update this when I get back to this PR. I hope to finish the ImageVisual PR first.

vispy/visuals/volume.py
djhoese: Yes I think so. It has been a while since I've looked at this code, but @tlambert03 and I had a long discussion about this in one of his PRs. There was something funky going on with the operations and what the data ended up as in the shader that we couldn't track down at the time. This was just the way we/he ended up implementing it and it worked so we kept it. The multiply by `-1` at the top and then divide by this negative result produces the expected result. I'll try re-evaluating after I've dealt with the ImageVisual PR and added real tests to this PR.

vispy/visuals/volume.py
djhoese: This seems like a style opinion. I'm not sure I favor one over the other. I think some style checkers don't like `elif` when the previous branch is a `return`.

vispy/visuals/volume.py
djhoese: Yeah this code does seem weird. Convert to array just to check the shape then back to a tuple on the instance attribute. I'll have to walk through this code to make sure I'm not missing something with the array usage.

If I'm understanding you correctly, you'd like things refactored to just use cmin/cmax rather than a tuple or an array. And if we wanted to use a tuple, then don't accept an array at all. Right?

vispy/visuals/volume.py
tlambert03: yeah, I agree it looks weird.  I can't remember the details at this point... but I definitely spent a while trying to make it look like it "should" without success.  no reason not to try again though!

vispy/visuals/volume.py
djhoese: Good point. Warning would be good.

vispy/visuals/volume.py
rougier: Yes. But it's mostly cosmetic, not really that important as long as the code is correct.

vispy/visuals/volume.py
djhoese: This now uses the same check function as the ImageVisual. Thanks for pointing out the issue.

djhoese:
@tlambert03 Seeking your opinion: The VolumeVisual has the logic of only rescaling data if it is within a certain threshold. With this pull request that shouldn't really be necessary. Do you see this logic as important enough to retain and maintain in the VolumeVisual? I ask out of my own laziness for not wanting to port the logic to the classes I'm working on. How lazy do you think I should be?

tlambert03:
> Do you see this logic as important enough to retain and maintain in the VolumeVisual?

nope, that logic was always an ugly workaround to avoid posterization in the absence of 32-bit textures.  would be happy to see that die :)

djhoese:
Thanks @tlambert03. Next question for everyone: Does VolumeVisual still need to support emulated 3D textures? According to docstrings this is for OpenGL ES 2.0 compatibility. @almarklein @rougier @kmuehlbauer @larsoner Opinions?

jni:
From my limited understanding, for us (napari) it is not important to support OpenGL ES 2.0.

almarklein:
No I think we can drop that requirement.

rougier:
It will make our life easier for sure.

djhoese:
Ok, I think this is finally complete. My last commits add a test that should perform a sanity check for @alisterburt's #2046 so he shouldn't need to add one. The test attempts to draw with each available render method and just proves that the output isn't all black or all white for a basic volume.

This PR has really grown since I first started it, but I think it is overall much cleaner. I've combined the "scalable texture" concept that I used in #1920 with the necessary features of the scalable textures needed by the VolumeVisual. The result is a series of "mixin" classes that can be shared for (theoretically) a texture class of any dimensionality. This should help in the future if someone wants to optimize, for example, the CPU scaling the way @tlambert03 originally did in the VolumeVisual. We wouldn't need to find a way to copy the logic to the VolumeVisual and the ImageVisual; just create a new mixin class.

A couple things that should work the same, but were odd to me that I want to note:

1. I kept confusing myself about what it mean to pass the original data to the Texture class. It seems pretty "standard" in the VisPy visuals to set the texture initially with some representative array with a smaller size and then set the actual data later on. This isn't a problem necessarily, but does require a resize on the GPU when you could theoretically not need it. However, you end up with confusing code if you do it any other way where you want to prepare the data to go in the texture, but you need to know what type of texture you're going to be making first. Anyway, I didn't think it was worth it to optimize this resize out of the procedure. It doesn't seem to be a big deal.
2. The color limit setting in the volume visual is weird. You set the color limits and then you might need to update the texture data so it updates the color limits then calls `.set_data` with the original data and the new color limits so the color limits can be set again and then the texture scaled based on those color limits. And the way the code/docstrings are written makes it seem like color limits can be set to `None` to have the min/max of the data used, but this only happens on the initial time. After that, the color limits will just be unchanged if you change the data. It's just odd, but I'm not sure how much this actually affects most users.
3. The VolumeVisual and ImageVisual set their data at different times. We may need to resolve this in the future. The ImageVisual uses instance attributes to keep track of data and color limit changes and then waits for draw time to actually apply the changes to the GPU/gloo elements. The VolumeVisual however applies these changes immediately (except for vertex coordinate changes). I'm not sure if there is a reason for this, but that's how it is and I didn't feel like changing it in this PR.

So please review and provide feedback. Let's get this merged and start working on a 0.7 release.

vispy/visuals/tests/test_volume.py
```diff
@@ -64,49 +67,124 @@ def test_volume_draw():
...
+        rendered3 = downsample(c.render(), shape_ratio, axis=(0, 1)).astype(_dtype)
+        predicted = (scaled_data ** 2).max(axis=1)
+        compare_render(predicted, rendered3, previous_render=rendered2)
+
+
+@requires_pyopengl()
+@requires_application()
+@pytest.mark.parametrize('method_name', volume_render_methods.keys())
+def test_all_render_methods(method_name):
+    """Test that render methods don't produce any errors."""
```
alisterburt, line 142: great stuff, thanks for adding! 👍🏼

alisterburt (approved):
I'm guessing the diff here has become smaller because of #1920? some of the discussion doesn't seem to map onto the changes

lots of useful testing utilities in here - I can't comment on the networkx stuff but everything else looks good to me, definitely no specific concerns!

djhoese:
Yeah some of the networkx stuff was discovered as part of making the "central" testing utilities for these ImageVisual and VolumeVisual changes. I didn't want to try to move it to another PR or forget about it. It was just easier to put here. Thanks for the review.

@alisterburt @sofroniewn @jni @tlambert03 Would it be possible for one of you to test basic napari functionality with this PR? Not sure how hard it will be to add your custom VolumeVisual stuff that's in napari on top of this.

alisterburt:
Makes sense! I will see how napari behaves with this PR as is and report back

alisterburt:
Works like a charm locally with napari on master and this PR for vispy - no modification required

https://user-images.githubusercontent.com/7307488/121263770-0a716380-c8ae-11eb-9213-60ef0fe68a3f.mp4

changing gamma and clims works fine (not shown in the video)

jni:
So pretty, @alisterburt! Just to clarify, do you mean to say that you used the Volume visual from this PR rather than our vendored copy? Because I don't think we use the VisPy Volume visual at all on master right now because of our totally rad custom mods 😂. (Thank you @alisterburt for pushing those upstream to here!)

sofroniewn:
> So pretty, @alisterburt! Just to clarify, do you mean to say that you used the Volume visual from this PR rather than our vendored copy? Because I don't think we use the VisPy Volume visual at all on master right now because of our totally rad custom mods 😂. (Thank you @alisterburt for pushing those upstream to here!)

I just confirmed this works for me too - I checkout out this branch and then copied and pasted the volume visual that we had previously vendored into napari, fixed the imports - added the `u_attenuation` parameter (which was why we needed to vendor, and will be pushed upstream in https://github.com/vispy/vispy/pull/2047) and everything worked great! Though I know get this message (which I think is fine, but just noting it for completeness sake).

```
GPUs can't support floating point data with more than 32-bits, precision will be lost due to downcasting to 32-bit float.
```

Screen Shot 2021-06-08 at 9 27 25 PM

Thanks for pushing on this @djhoese, thanks for checking it too @alisterburt! Very excited for this progress 🚀

vispy/visuals/_scalable_textures.py
```diff
@@ -0,0 +1,441 @@
...
+        clim_min = self.normalize_value(self.clim[0], self._data_dtype)
+        clim_max = self.normalize_value(self.clim[1], self._data_dtype)
+        return clim_min, clim_max
+
+    @property
+    def is_normalized(self):
+        """Whether the in-shader representation of this texture is normalized or not.
+
+        Formats ending in 'f' (float), 'ui' (unsigned integral), or 'i'
+        (integral) are not normalized in the GPU. Formats ending in "_snorm"
```
almarklein, line 100:
```suggestion
        Formats ending in 'f' (float), 'ui' (unsigned integer), or 'i'
        (signed integer) are not normalized in the GPU. Formats ending in "_snorm"
```

vispy/visuals/_scalable_textures.py
```diff
@@ -0,0 +1,441 @@
...
+    # 'r' will be replaced (if needed) with rgb or rgba depending on number of bands
+    _texture_dtype_format = {
+        np.float32: 'r32f',
+        np.float64: 'r32f',
+        np.uint8: 'r8',
+        np.uint16: 'r16',
+        # np.uint32: 'r32ui',  # not supported texture format in vispy
+        np.int8: 'r8',
+        np.int16: 'r16',
+        # np.int32: 'r32i',  # not supported texture format in vispy
```
almarklein, line 331: Why are these not supported? Is this a matter of adding a field somewhere in gloo?

alisterburt:
> do you mean to say that you used the Volume visual from this PR rather than our vendored copy? Because I don't think we use the VisPy Volume visual at all on master right now because of our totally rad custom mods 😂.

Uh oh - you're right @jni! Sleepy brain -> I (mistakenly) assumed our `BaseVolume` was imported directly from vispy. Thankfully @sofroniewn stepped up to *actually* test it 😆

vispy/visuals/_scalable_textures.py
almarklein: Ah: https://github.com/vispy/vispy/blob/a20297a34bf5de3204111842b0d77f09f7a59500/vispy/gloo/texture.py#L68-L76

almarklein (commented):
Looking good! I test-drove the `scene/volume.py` example and with the `GPUScaledTextured3D` it all works as expected for uint8, uint16, and float32. However, I get all black for int16. Will look into it.

vispy/visuals/_scalable_textures.py
djhoese: This wording comes from the khronos wiki: https://www.khronos.org/opengl/wiki/Image_Format

I suppose "integer" will make more sense to Python programmers.

djhoese:
With signed integers, it is possible the color limits need to be flipped or handled specially before going to the shader.

vispy/visuals/_scalable_textures.py
```diff
@@ -0,0 +1,441 @@
...
+    def _scale_data_on_cpu(data, clim, copy=True):
+        if copy:
+            should_cast_to_f32(data.dtype)
+            data = np.array(data, dtype=np.float32, copy=copy)
+        elif not copy and not np.issubdtype(data.dtype, np.floating):
+            raise ValueError("Data must be of floating type for no copying to occur.")
+
+        if clim[0] == clim[1]:
+            if clim[0] != 0:
+                data /= clim[0]
```
almarklein, line 261: What is this for? :)

djhoese: I honestly don't know. It was in the original code and I think is to make it so if the limits are the same you don't end up with a divide by zero. I guess it "normalizes" the data to about the same range as the user expected with their color limits being set to the same thing? I didn't change it because I assumed that someone might be depending on it. I'm definitely open to changing it and it doesn't make a lot of sense to me.

almarklein: haha, ok lets not touch it for now :)

vispy/visuals/_scalable_textures.py
```diff
@@ -0,0 +1,441 @@
...
+    _texture_dtype_format = {
+        np.float32: 'r32f',
+        np.float64: 'r32f',
+        np.uint8: 'r8',
+        np.uint16: 'r16',
+        # np.uint32: 'r32ui',  # not supported texture format in vispy
+        np.int8: 'r8',
+        np.int16: 'r16',
+        # np.int32: 'r32i',  # not supported texture format in vispy
+    }
```
almarklein, line 332:
```suggestion
    _texture_dtype_format = {
        np.float32: 'r32f',
        np.float64: 'r32f',
        np.uint8: 'r8',  # uint8 normalized
        np.uint16: 'r16',  # uint16 normalized
        # np.int8: 'r8',  # not supported, there are no signed-integer norm formats
        # np.int16: 'r16',
        # np.uint32: 'r32ui',  # not supported, no normal formats for 32bit ints
        # np.int32: 'r32i',
    }
```

vispy/visuals/_scalable_textures.py
kmuehlbauer: see https://github.com/vispy/vispy/commit/8d280f2b50c0fce92a1bc35e5229f223293eadba for context ;-)

vispy/visuals/_scalable_textures.py
almarklein: So I looked into this. The r8 and r16 formats are also normalized unsigned integer formats: the texture data is an uint, but when sampling you get a float in the shader. There are no such formats for signed integers or 32bit integers.

You suggested converting the clim for these signed integer cases, but that is tricky: the data is interpreted as uint, so as you go from 2, 1, 0, -1, there is a jump from 0 to 65525. We could maybe handle it with additional shader code, but not simply by changing the clim values.

almarklein: So I think we'd need to disable these formats for the GPUScaled texture for now.

We can maybe enable them later by enabling support for real integer tectures, but that would require some fundamental changes how we sample such textures.

vispy/visuals/_scalable_textures.py
almarklein: Thanks! That makes sense. Let's add a comment while we're here.

almarklein, line 261: Adding comment and turn the numpy-division into a (faster) numpy-multiplication:
```suggestion
        if clim[0] == clim[1]:
            # Prevent zero division when all elements have the same value
            if clim[0] != 0:
                data *= 1 / clim[0]
```

djhoese: I still don't get it. Like I get what it is talking about. If clims are the same then we get a divide by zero at some point. If clims are the same what should the expected output be? This line `vol *= 1.0 / self._clim[0]` from the original code (same as my `data /= clim[0]` above) just doesn't make sense as far as logical output in my opinion. I guess why divide at all?

vispy/visuals/_scalable_textures.py
djhoese: If we know how the data will appear in the GPU then can't we convert the clim to the equivalent number space before it gets used in the shader (`clim_normalized` in this PR)?

vispy/visuals/_scalable_textures.py
almarklein: > I guess why divide at all?

Good point. The division is ... arbitrary.

vispy/visuals/_scalable_textures.py
almarklein: No, because the int8 (or int16) binary data is interpreted by the GPU as unsigned and this transformation is non-linear. Basically, values -128..0 map to 0.5..1 and values 0..127 map to 0..0.5.

almarklein: In other words: signed ints use the most significant bit to expres the sign. It's set to 1 for negative numbers.

djhoese: Yeah...you're right. It could be done with a lot of special casing with no real benefit.

vispy/visuals/_scalable_textures.py
djhoese: We could do something like: if clims are the same, make data all 0s (`np.zeros_like`) and make clims 0 and 1. Or...we could detect the clims being the same in the shader and discard the fragment. That should make it work for CPU or GPU scaled textures. I'd just have to make sure that `.clim_normalized` returns clims that are equal to each other. Thoughts?

almarklein: If the clims are equal, we'd expect the result to be all gray, I guess. We should not have to change the data. So maybe something like:
```py
if clim[0] == clim[1]:
  clim = clim[0] - 0.5, clim[0] + 0.5
```

Another thought: let's do this in a separate PR :P

almarklein: oh wait, this code is in the CPUScaledTexture. So I guess setting the data to all 127 makes sense, and something like what I wrote above for the GPUScaledTexture.

djhoese: > If the clims are equal, we'd expect the result to be all gray, I guess.

Eh. That's like saying `x / 0` is a real number. I'm starting to think the two texture subclasses shouldn't do any scaling if the clims are the same and leave it to the shader code to decide what to do. I think for both the ImageVisual and the VolumeVisual it should do one of the following:

1. discard (nothing there, fully transparent)
2. return 0 so the colormap will get its lowest value
3. return 0.5 so the colormap will get its middle value

Thoughts? @jni @tlambert03 @alisterburt Any thoughts on this? What do you expect as output for the Image or Volume if the color limits are equal to each other? Does that result change if the limits remain equal to each other but change in value (ex. `(0, 0)` versus `(0.5, 0.5)` versus `(1.0, 1.0)`)?

almarklein: > Eh. That's like saying x / 0 is a real number

I think what I meant to say is what you call option nr 3 :)

djhoese: In matplotlib with no vmin/vmax on imshow:

image

With vmin/vmax both set to 0, or both set to 255, or both set to 128, you get this for each:

image

So matplotlib always shows the lowest value in the colormap.

djhoese: So based on my twitter poll (18 responses) most people think this should raise an error, while the second most popular option follows mpl's behavior above. I'm starting to think I agree with mpl. This makes sense in the case the limits are automatically detected and all input pixels are the same. You'd want to see *something* on the screen even if it doesn't mean anything. I think in vispy the easiest way to do this is to not scale the data, detect the equal clims in the shader, and return 0 (to be colormapped).

djhoese:
So latest commits I think fix everything regarding the equal color limits issue including adding tests. Note that for RGB/A ImageVisual cases, they don't use the colormap so the actual result will always be full opaque black.

Note @alisterburt the test I wrote did actually fail for minip, but it was due to the volume I was making in the test. The test volume had `0` along every path so the final output was always all black. I updated the test data to have at least one portion that is all white so the tests passed. I also updated your shader to use the red component (`.r`) instead of (`.g`) to be consistent with the other shaders. I think @sofroniewn had noticed that `.g` was being used a while back, but we couldn't figure out why it was used over `.r`. With floating point textures we need to use `.r` now as I believe `.g` and `.b` will always be 0.

I think once tests pass this is ready for re-review and merge.

sofroniewn:
> About the .r vs .g mystery: IIRC with good old GL_LUMINANCE the rgb are all equal. I may have defaulted to g there, out of a habit of using the green channel as a cheap way to turn rgb images into grayscale (which is a good habit with photographs because CCD chips have as many green sensors as R and G combined). Anyway ... with R8unorm or R32float only the .r gets a value.

Sorry for a naive question, but does this now mean that the behaviour of the isosurface will be "grayscale" like I made it for napari? That would be my preference. I can pull PR and check again, I don't recall it not being "grayscale"

djhoese:
@sofroniewn I think you already contributed that work back to vispy. This was only to update what @alisterburt had contributed for his other render methods. Don't have the time right now to find all the PR numbers.

sofroniewn:
Oh ok great, yes, that makes sense! Apologies for my confusion/ forgetting

djhoese:
@sofroniewn Here it is: https://github.com/vispy/vispy/pull/1802

vispy/visuals/image.py
```diff
@@ -451,8 +105,11 @@ def set_data(self, data, offset=None, copy=False):
     vec4 apply_clim(vec4 color) {
         if ($clim.x < $clim.y) {{
             color.rgb = clamp(color.rgb, $clim.x, $clim.y);
-        }} else {{
+        }} else if ($clim.x > $clim.y) {{
             color.rgb = clamp(color.rgb, $clim.y, $clim.x);
+        }} else {{
+            // clims are the same, show minimum colormap value
+            return vec4(0.0, 0.0, 0.0, 1.0);
```
almarklein, line 112: To avoid branching in the shader, it would be good to do checks like this on the CPU (in clim_normalized) and output a clim that can be used in the shader using a one-liner. (Branches can make shaders harder to run in parallel). But we may do that in another PR.

djhoese: > CPU (in clim_normalized) and output a clim that can be used in the shader using a one-liner.

Not sure what you mean with the one-liner comment. I agree and understand that branches should be avoided, but I'm not sure any clim on the CPU can be produced that makes the whole image 0. That means the data would have to be uploaded to be all 0s and I was trying to avoid having to reupload or change the texture data. Other option?

almarklein: Sorry for being a bit cryptic. I mean that the logic here in the shader can become a onelined without control flow (if's).

For the CPUScaledTexture we already scale the data, so we can e.g. subtract the (common) clim value, and set clim `(value, value+1)`. For the GPUScaledTexture maybe make the second clim value huge?

almarklein: Anyway, let's not make this hold up the current PR.

almarklein: I could have a look at that clim tweaking after this is merged, and also include Eric's solution of #1769, perhaps.

djhoese: That sounds good. I still feel weird about that whole discussion. I've been working with dask so long the idea of operations that require looking at the entire array at once feel like a bad idea. Anyway...we can resolve this comment then right? Anything left on this PR?

almarklein: I think it should be good to go? Github says that Rougier has requested changes, have you covered those?

djhoese: Yes. Quite a bit has actually changed since his review and I haven't heard anything lately. Given my urgency to make a release and include this news in the CZI report I have due at the end of the month, I'd like to merge this today or tomorrow. @rougier if you have time to re-review it that'd be great, otherwise I'll plan on merging it tomorrow.

djhoese: Correction: turns out my report isn't do until September 1st. Even though I have more time I'd still like to release 0.7 next week which means merging this this week.

sofroniewn:
> Even though I have more time I'd still like to release 0.7 next week which means merging this this week.

So exciting! When you have the release ready let us know over at napari and we'll do some testing to make sure it's a smooth transition for us!! Thanks again for all the great work over here

djhoese:
It is done. No going back now. :fearful:

PR #1914, opened 2020-09-10
Fix ImportError on Python 3.9

cgohlke:
The `fractions.gcd()` function has been removed in Python 3.9. It was deprecated since Python 3.5: use math.gcd() instead. See https://docs.python.org/3.9/whatsnew/3.9.html#removed


djhoese:
Thanks!

PR #1918, opened 2020-09-22
Add Tkinter backend

Wosser1sProductions:
For my current project, it would be beneficial to be able to use VisPy within a Tkinter context, i.e. an OpenGL canvas controlled by VisPy embedded inside a Tkinter widget (tk.Frame).

For this purpose, I attempted to create a new backend to target Tkinter. However, it needs more work to to actually draw something.

The implementation tries to use [Jon Wright's pyopengltk](https://github.com/jonwright/pyopengltk) `OpenGLFrame`.

You can refer to [this Gist with example code](https://gist.github.com/Wosser1sProductions/620d2422a8a2f2529d01bf34e229d7ca) that shows how I might use it. For reference, everything works as expected using the Glfw backend, but this creates a new window naturally.
Using the tkinter backend, creates the frame that can be embedded, but it seems to stay black. Running `vispy.app.process_events()` is doing *something* as the GPU gets used, but the canvas stays black.

Any help would be appreciated, and would of course benefit the community as well.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ environment:
   matrix:
       - PYTHON_VERSION: "3.7"
         CONDA_DEPENDENCIES: "numpy scipy setuptools pytest coverage pytest-cov pytest-sugar pytest-faulthandler cython coveralls freetype-py numpydoc meshio"
-        PIP_DEPENDENCIES: "pyopengltk"
+        PIP_DEPENDENCIES: ""
 
 platform:
     -x64
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ environment:
   matrix:
       - PYTHON_VERSION: "3.7"
         CONDA_DEPENDENCIES: "numpy scipy setuptools pytest coverage pytest-cov pytest-sugar pytest-faulthandler cython coveralls freetype-py numpydoc meshio"
-        PIP_DEPENDENCIES: ""
+        PIP_DEPENDENCIES: "pyopengltk"
 
 platform:
     -x64
--- a/vispy/app/backends/_tk.py
+++ b/vispy/app/backends/_tk.py
@@ -23,27 +23,35 @@ try:
     import sys
     import os
 
-    _tk_on_linux = sys.platform.startswith('linux')
-    _tk_on_darwin = sys.platform == 'darwin'
-    _tk_on_windows = sys.platform.startswith('win')
+    _tk_on_linux, _tk_on_darwin, _tk_on_windows = \
+        map(sys.platform.startswith, ("linux", "darwin", "win"))
+    _tk_pyopengltk_imported = False
 
-    try:
-        import tkinter as tk  # Python >= 3
-    except ModuleNotFoundError:
-        import Tkinter as tk  # Python < 3
+    import tkinter as tk
     from OpenGL import GL
     import pyopengltk
-    OpenGLFrame = pyopengltk.OpenGLFrame
 except (ModuleNotFoundError, ImportError):
     available, testable, why_not, which = \
         False, False, "Could not import Tkinter or pyopengltk, module(s) not found.", None
+else:
+    which_pyopengltk = getattr(pyopengltk, "__version__", "???")
+    which = f"Tkinter {tk.TkVersion} (with pyopengltk {which_pyopengltk})"
 
+    if hasattr(pyopengltk, "OpenGLFrame"):
+        _tk_pyopengltk_imported = True
+        available, testable, why_not = True, True, None
+    else:
+        # pyopengltk does not provide an implementation for this platform
+        available, testable, why_not = \
+            False, False, f"pyopengltk {which_pyopengltk} is not supported on this platform ({sys.platform})!"
+
+if _tk_pyopengltk_imported:
+    # Put OpenGLFrame class in global namespace
+    OpenGLFrame = pyopengltk.OpenGLFrame
+else:
+    # Create empty placeholder class
     class OpenGLFrame(object):
         pass
-else:
-    available, testable, why_not = True, True, None
-    which_pyopengltk = getattr(pyopengltk, "__version__", "???")
-    which = "Tkinter " + str(tk.TkVersion) + " (with pyopengltk " + which_pyopengltk + ")"
 
 
 def _fix_tcl_lib():
@@ -51,6 +59,7 @@ def _fix_tcl_lib():
     It is possible that the Tcl library cannot be found when the Python environment
     does not have the proper variables, so we force them here (mostly when running tests).
     From: https://github.com/enthought/Python-2.7.3/blob/master/Lib/lib-tk/FixTk.py
+    (This is rarely encountered in conjunction with PyInstaller.)
     """
     # Delay import _tkinter until we have set TCL_LIBRARY,
     # so that Tcl_FindExecutable has a chance to locate its
--- a/vispy/app/backends/_tk.py
+++ b/vispy/app/backends/_tk.py
@@ -263,7 +263,7 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
 
         OpenGLFrame.__init__(self, parent, **kwargs)
         if not hasattr(self, "_native_context") or self._native_context is None:
-            self.tkCreateContext()
+            self.tkMap(None)
             # Workaround to get OpenGLFrame.__context for reference here
             # if access would ever be needed from self._native_context.
             self._native_context = vars(self).get("_CanvasBackend__context", None)
--- a/vispy/app/backends/_tk.py
+++ b/vispy/app/backends/_tk.py
@@ -8,9 +8,12 @@ vispy backend for Tkinter.
 
 from __future__ import division
 
+from time import sleep
+
 from ..base import (BaseApplicationBackend, BaseCanvasBackend,
                     BaseTimerBackend)
 from ...util import keys
+from ...util.ptime import time
 from ... import config
 
 
@@ -18,13 +21,103 @@ from ... import config
 
 # Import
 try:
-    import tkinter as tk
+    try:
+        import tkinter as tk  # Python >= 3
+    except:
+        import Tkinter as tk  # Python < 3
     from OpenGL import GL
     from pyopengltk import OpenGLFrame
 except:
-    available, testable, why_not = False, False, "Import error"
+    available, testable, why_not, which = False, False, "Import error", None
 else:
     available, testable, why_not = True, True, None
+    which = "Tkinter " + str(tk.TkVersion)
+
+
+def _fix_tcl_lib():
+    """
+    It is possible that the Tcl library cannot be found when the Python environment
+    does not have the proper variables, so we force them here (mostly when running tests).
+    From: https://github.com/enthought/Python-2.7.3/blob/master/Lib/lib-tk/FixTk.py
+    """
+    import sys, os
+
+    # Delay import _tkinter until we have set TCL_LIBRARY,
+    # so that Tcl_FindExecutable has a chance to locate its
+    # encoding directory.
+
+    # Unfortunately, we cannot know the TCL_LIBRARY directory
+    # if we don't know the tcl version, which we cannot find out
+    # without import Tcl. Fortunately, Tcl will itself look in
+    # <TCL_LIBRARY>\..\tcl<TCL_VERSION>, so anything close to
+    # the real Tcl library will do.
+
+    # Expand symbolic links on Vista
+    try:
+        import ctypes
+        ctypes.windll.kernel32.GetFinalPathNameByHandleW
+    except (ImportError, AttributeError):
+        def convert_path(s):
+            return s
+    else:
+        def convert_path(s):
+            assert isinstance(s, str)   # sys.prefix contains only bytes
+            udir = s.decode("mbcs")
+            hdir = ctypes.windll.kernel32.\
+                CreateFileW(udir, 0x80, # FILE_READ_ATTRIBUTES
+                            1,          # FILE_SHARE_READ
+                            None, 3,    # OPEN_EXISTING
+                            0x02000000, # FILE_FLAG_BACKUP_SEMANTICS
+                            None)
+            if hdir == -1:
+                # Cannot open directory, give up
+                return s
+            buf = ctypes.create_unicode_buffer(u"", 32768)
+            res = ctypes.windll.kernel32.\
+                GetFinalPathNameByHandleW(hdir, buf, len(buf),
+                                          0) # VOLUME_NAME_DOS
+            ctypes.windll.kernel32.CloseHandle(hdir)
+            if res == 0:
+                # Conversion failed (e.g. network location)
+                return s
+            s = buf[:res].encode("mbcs")
+            # Ignore leading \\?\
+            if s.startswith("\\\\?\\"):
+                s = s[4:]
+            if s.startswith("UNC"):
+                s = "\\" + s[3:]
+            return s
+
+    prefix = os.path.join(sys.prefix,"tcl")
+    if not os.path.exists(prefix):
+        # devdir/../tcltk/lib
+        prefix = os.path.join(sys.prefix, os.path.pardir, "tcltk", "lib")
+        prefix = os.path.abspath(prefix)
+    # if this does not exist, no further search is needed
+    if os.path.exists(prefix):
+        prefix = convert_path(prefix)
+        if "TCL_LIBRARY" not in os.environ:
+            for name in os.listdir(prefix):
+                if name.startswith("tcl"):
+                    tcldir = os.path.join(prefix,name)
+                    if os.path.isdir(tcldir):
+                        os.environ["TCL_LIBRARY"] = tcldir
+        # Compute TK_LIBRARY, knowing that it has the same version
+        # as Tcl
+        import _tkinter
+        ver = str(_tkinter.TCL_VERSION)
+        if "TK_LIBRARY" not in os.environ:
+            v = os.path.join(prefix, 'tk'+ver)
+            if os.path.exists(os.path.join(v, "tclIndex")):
+                os.environ['TK_LIBRARY'] = v
+        # We don't know the Tix version, so we must search the entire
+        # directory
+        if "TIX_LIBRARY" not in os.environ:
+            for name in os.listdir(prefix):
+                if name.startswith("tix"):
+                    tixdir = os.path.join(prefix,name)
+                    if os.path.isdir(tixdir):
+                        os.environ["TIX_LIBRARY"] = tixdir
 
 
 # Map native keys to vispy keys
@@ -108,7 +201,7 @@ capability = dict(
     resizable=True,       # can toggle resizability (e.g., no user resizing)
     decorate=True,        # can toggle decorations
     fullscreen=True,      # fullscreen window support
-    context=True,         # can share contexts between windows
+    context=False,        # can share contexts between windows
     multi_window=True,    # can use multiple windows at once
     scroll=True,          # scroll-wheel events are supported
     parent=True,          # can pass native widget backend parent
@@ -119,7 +212,7 @@ capability = dict(
 # ------------------------------------------------------- set_configuration ---
 def _set_config(c):
     """Set gl configuration for template"""
-    raise NotImplementedError
+    return []
 
 
 # ------------------------------------------------------------- application ---
@@ -129,12 +222,12 @@ _tk_inst_owned = False  # Whether we created the Tk instance or not
 _tk_toplevels = []      # References to created CanvasBackend Toplevels
 
 
-def _new_toplevel(*args, **kwargs):
+def _new_toplevel(self, *args, **kwargs):
     """Create and return a new withdrawn Toplevel."""
     global _tk_inst, _tk_toplevels
     tl = tk.Toplevel(_tk_inst, *args, **kwargs)
     tl.withdraw()
-    _tk_toplevels.append(tl)
+    _tk_toplevels.append(self)
     return tl
 
 
@@ -166,7 +259,18 @@ class ApplicationBackend(BaseApplicationBackend):
         return tk.__name__
 
     def _vispy_process_events(self):
-        self._vispy_get_native_app().update_idletasks()
+        # Update idle tasks first (probably not required)
+        app = self._vispy_get_native_app()
+        app.update_idletasks()
+
+        # Update every active Canvas window
+        for c in _tk_toplevels:
+            c._delayed_update()
+
+        # Process some events in the main Tkinter event loop
+        # And quit so we can continue elsewhere (call blocks normally)
+        app.after(0, lambda: app.quit())
+        app.mainloop()
 
     def _vispy_run(self):
         self._vispy_get_native_app().mainloop()
@@ -186,6 +290,7 @@ class ApplicationBackend(BaseApplicationBackend):
                 _tk_inst_owned = False
             else:
                 # Create our own top level Tk instance
+                _fix_tcl_lib()
                 _tk_inst = tk.Tk()
                 _tk_inst.withdraw()
                 _tk_inst_owned = True
@@ -213,7 +318,7 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
         # Deal with config
         # ... use context.config
         # Deal with context
-        p.context.shared.add_ref('tkinter', self)
+        p.context.shared.add_ref('tk', self)
         if p.context.shared.ref is self:
             self._native_context = None
         else:
@@ -234,7 +339,7 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
 
         if p.parent is None:
             # Create native window and master
-            self.top = _new_toplevel()
+            self.top = _new_toplevel(self)
 
             if p.title:    self._vispy_set_title(p.title)
             if p.size:     self._vispy_set_size(p.size[0], p.size[1])
@@ -248,7 +353,7 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
                 self.top.overrideredirect(True)
             if p.always_on_top:
                 self.top.wm_attributes("-topmost", "True")
-            self._fullscreen = p.fullscreen
+            self._fullscreen = bool(p.fullscreen)
 
             self.top.protocol("WM_DELETE_WINDOW", self._vispy_close)
             parent = self.top
@@ -261,11 +366,12 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
         self._init = False
         self.is_destroyed = False
 
-        OpenGLFrame.__init__(self, parent, **kwargs)
+        c = OpenGLFrame.__init__(self, parent, **kwargs)
         if not hasattr(self, "_native_context") or self._native_context is None:
             self.tkMap(None)
             # Workaround to get OpenGLFrame.__context for reference here
             # if access would ever be needed from self._native_context.
+            # ERROR: Context sharing this way seems unsupported.
             self._native_context = vars(self).get("_CanvasBackend__context", None)
 
         if self.top:
@@ -334,6 +440,13 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
         self._vispy_canvas.events.initialize()
         self.update_idletasks()
         self._on_configure(Coord(self._vispy_get_size()))
+
+    def _vispy_warmup(self):
+        etime = time() + 0.3
+        while time() < etime:
+            sleep(0.01)
+            self._vispy_canvas.set_current()
+            self._vispy_canvas.app.process_events()
 
     def _parse_state(self, e):
         return [ key for mask, key in KEY_STATE_MAP.items() \
@@ -421,7 +534,8 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
 
     def _vispy_set_current(self):
         # Make this the current context
-        self.tkMakeCurrent()
+        if not self.is_destroyed:
+            self.tkMakeCurrent()
 
     def _vispy_swap_buffers(self):
         # Swap front and back buffer
@@ -457,8 +571,8 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
         # Has no effect for widgets. If you want it to become fullscreen,
         # while embedded in another Toplevel window, you should make that
         # window fullscreen instead.
+        self._fullscreen = bool(fullscreen)
         if self.top:
-            self._fullscreen = bool(fullscreen)
             self._vispy_set_visible(True)
 
     def _vispy_update(self):
@@ -478,7 +592,7 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
         if self.top and not self.is_destroyed:
             self.is_destroyed = True
             self._vispy_canvas.close()
-            _del_toplevel(self.top)
+            _del_toplevel(self)
 
     def destroy(self):
         self._vispy_canvas.close()
--- a/vispy/app/backends/_tk.py
+++ b/vispy/app/backends/_tk.py
@@ -53,94 +53,6 @@ else:
     class OpenGLFrame(object):
         pass
 
-
-def _fix_tcl_lib():
-    """
-    It is possible that the Tcl library cannot be found when the Python environment
-    does not have the proper variables, so we force them here (mostly when running tests).
-    From: https://github.com/enthought/Python-2.7.3/blob/master/Lib/lib-tk/FixTk.py
-    (This is rarely encountered in conjunction with PyInstaller.)
-    """
-    # Delay import _tkinter until we have set TCL_LIBRARY,
-    # so that Tcl_FindExecutable has a chance to locate its
-    # encoding directory.
-
-    # Unfortunately, we cannot know the TCL_LIBRARY directory
-    # if we don't know the tcl version, which we cannot find out
-    # without import Tcl. Fortunately, Tcl will itself look in
-    # <TCL_LIBRARY>\..\tcl<TCL_VERSION>, so anything close to
-    # the real Tcl library will do.
-
-    # Expand symbolic links on Vista
-    try:
-        import ctypes
-        ctypes.windll.kernel32.GetFinalPathNameByHandleW
-    except (ImportError, AttributeError):
-        def convert_path(s):
-            return s
-    else:
-        def convert_path(s):
-            use_dec = hasattr(s, "decode")
-            udir = s.decode("mbcs") if use_dec else s
-            hdir = ctypes.windll.kernel32.\
-                CreateFileW(udir, 0x80,  # FILE_READ_ATTRIBUTES
-                            1,           # FILE_SHARE_READ
-                            None, 3,     # OPEN_EXISTING
-                            0x02000000,  # FILE_FLAG_BACKUP_SEMANTICS
-                            None)
-            if hdir == -1:
-                # Cannot open directory, give up
-                return s
-            buf = ctypes.create_unicode_buffer(u"", 32768)
-            res = ctypes.windll.kernel32.\
-                GetFinalPathNameByHandleW(hdir, buf, len(buf),
-                                          0)  # VOLUME_NAME_DOS
-            ctypes.windll.kernel32.CloseHandle(hdir)
-            if res == 0:
-                # Conversion failed (e.g. network location)
-                return s
-            s = buf[:res].encode("mbcs") if use_dec else buf[:res]
-            # Ignore leading \\?\
-            if s.startswith("\\\\?\\"):
-                s = s[4:]
-            if s.startswith("UNC"):
-                s = "\\" + s[3:]
-            return s
-
-    prefix = os.path.join(sys.prefix, "tcl")
-    if not os.path.exists(prefix):
-        # devdir/../tcltk/lib
-        prefix = os.path.join(sys.prefix, os.path.pardir, "tcltk", "lib")
-        prefix = os.path.abspath(prefix)
-
-    # if this does not exist, no further search is needed
-    if os.path.exists(prefix):
-        prefix = convert_path(prefix)
-
-        if "TCL_LIBRARY" not in os.environ:
-            for name in os.listdir(prefix):
-                if name.startswith("tcl"):
-                    tcldir = os.path.join(prefix, name)
-                    if os.path.isdir(tcldir):
-                        os.environ["TCL_LIBRARY"] = tcldir
-
-        # Compute TK_LIBRARY, knowing that it has the same version as Tcl
-        import _tkinter
-        ver = str(_tkinter.TCL_VERSION)
-        if "TK_LIBRARY" not in os.environ:
-            v = os.path.join(prefix, 'tk'+ver)
-            if os.path.exists(os.path.join(v, "tclIndex")):
-                os.environ['TK_LIBRARY'] = v
-
-        # We don't know the Tix version, so we must search the entire directory
-        if "TIX_LIBRARY" not in os.environ:
-            for name in os.listdir(prefix):
-                if name.startswith("tix"):
-                    tixdir = os.path.join(prefix, name)
-                    if os.path.isdir(tixdir):
-                        os.environ["TIX_LIBRARY"] = tixdir
-
-
 # Map native keys to vispy keys
 # e.keysym_num -> vispy
 KEYMAP = {
@@ -341,7 +253,6 @@ class ApplicationBackend(BaseApplicationBackend):
                 _tk_inst_owned = False
             else:
                 # Create our own top level Tk instance
-                _fix_tcl_lib()
                 _tk_inst = tk.Tk()
                 _tk_inst.withdraw()
                 _tk_inst_owned = True
--- a/vispy/app/backends/_tk.py
+++ b/vispy/app/backends/_tk.py
@@ -232,7 +232,7 @@ class CanvasBackend(OpenGLFrame, BaseCanvasBackend):
         kwargs.pop("fullscreen")
         kwargs.pop("context")
 
-        if p.parent is None:    
+        if p.parent is None:
             # Create native window and master
             self.top = _new_toplevel()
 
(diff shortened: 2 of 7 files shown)


vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,498 @@
...
+            self._vispy_canvas.app.process_events()
+
+    def _vispy_set_current(self):
+        # Make this the current context
+        self.tkMakeCurrent()
+
+    def _vispy_swap_buffers(self):
+        # Swap front and back buffer
+        self._vispy_canvas.set_current()
+        self.tkSwapBuffers()
```
djhoese, line 418: What happens if you comment out this line? I don't see a swap buffers call in the pyopengltk demo.py file so maybe it is auto-swapping?

Wosser1sProductions: Same thing, screen stays black.

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,498 @@
...
+        self.tkResize(e)
+        self._vispy_canvas.events.resize(size=(e.width, e.height))
+    
+    def _on_draw(self, e):
+        if self._vispy_canvas is None:
+            return
+        if not self._init:
+            self._initialize()
+        self._vispy_canvas.set_current()
+        self._vispy_canvas.events.draw(region=None)
```
Wosser1sProductions, line 376: I noticed that calling `self._vispy_canvas.events.draw(region=None)` just makes the screen black, nothing gets drawn and nothing can be drawn later. For example, when adding `GL.glClear(GL.GL_COLOR_BUFFER_BIT)` to clear the screen (to green on my system), it stays black. Removing the `events.draw` call does make the screen green.

So somehow, it's only clearing the screen to black, but then not draw to it.

Wosser1sProductions: I think I found it, calling `GL.glClear(GL.GL_COLOR_BUFFER_BIT)` first enables the OpenGL buffer for drawing.

Wosser1sProductions:
After some more fiddling, the basics seem to work now:

djhoese:
Nice job!

Wosser1sProductions:
I fixed the drawing and pausing of the update loop, but I'm hitting a circular event callback in the EventEmitter.
When Vispy requests a frame udpdate by calling `_vispy_update` in the backend, I can only redraw the frame by invoking `self._vispy_canvas.events.draw()` to get it to actually draw the update. But calling this, propagates eventually to a `CompoundVisual._subv_update` which will also call `_vispy_update`...

djhoese:
Hm yeah that's weird. I'm trying to compare it to the Qt backend but get even more confused since the `_vispy_update` just calls `self.update()`.

Wosser1sProductions:
> Hm yeah that's weird. I'm trying to compare it to the Qt backend but get even more confused since the `_vispy_update` just calls `self.update()`.

And `self.Refresh()` in wx actually calls [the top window's Refresh](http://openknights.sourceforge.net/neveredit/docs/private/wx._core.Window-class.html#Refresh) which probably causes a `wx.EVT_PAINT` event, which in turn calls `on_draw` in the backend that then calls `self._vispy_canvas.events.draw()`.

So I think I need to find a way to stop VisPy `EventEmitter` event propagation after I call draw.

djhoese:
I don't think so. I was looking at it again. The `self.update()` I was seeing in the Qt stuff is actually the *widget's* update on Qt. I don't think this actually adds events to the GUI event loop. Similar for Wx and .Refresh/.Update. I think these are typically the methods at the *end* of those events. Maybe the `update` methods mentioned here are what you need: https://stackoverflow.com/a/19477781

Wosser1sProductions:
Stopping events does not seem to do anything. 

So what happens is this:
#### Calling `vispy.app.process_events()` in the main program:

1. `_vispy_update()`
2. Underlying `tkExpose()` which calls `tkMakeCurrent(); redraw(); tkSwapBuffers()`
3. `redraw()`, which is where `_vispy_canvas.events.draw()` gets called
4. Return after draw call (within tkExpose) is complete

This is the normal behaviour and works as expected.

#### Pressing the mouse button:

1. `_on_mouse_button_press`
2. (Axis visuals update)
2. `_vispy_update` (see above)
4. `_vispy_canvas.events.draw`
5. (...)
6. `self._line.set_data(pos=self.pos, color=self.axis_color)` with eventually _RuntimeError: EventEmitter loop detected!_ and then again for
`self._ticks.set_data(pos=tick_pos, color=self.tick_color)` and again for
`self._text.text = list(labels)` and again
`self._line.set_data(pos=self.pos, color=self.axis_color)`
...

[With this traceback](https://pastebin.com/DXU6T6qS)

So the issue is maybe because pressing the mouse button causes the elements in the canvas to update, which then request a redraw with `_vispy_update` and maybe during the handling of that redraw, the previous event triggers another element to update, which also calls `_vispy_update` and triggers an event loop?

Wosser1sProductions:
> I don't think so. I was looking at it again. The `self.update()` I was seeing in the Qt stuff is actually the _widget's_ update on Qt. I don't think this actually adds events to the GUI event loop. Similar for Wx and .Refresh/.Update. I think these are typically the methods at the _end_ of those events. Maybe the `update` methods mentioned here are what you need: https://stackoverflow.com/a/19477781

Calling `_vispy_update` must redraw the canvas, as per the VisPy request, else nothing get updated and the canvas stays the same.
And you can only redraw the canvas by eventually calling `_vispy_canvas.events.draw` because that's what triggers OpenGL commands to be executed.

In Qt, invoking `self.update()` probably calls `self.paintGL()`, which the `CanvasBackendDesktop` class overrides and thus calls `_vispy_canvas.events.draw` to update the canvas. 

So I don't think the problem lies in the calling the drawing functions itself, also because just drawing is fine, it's the handling of the mouse events that cause an event loop in VisPy itself.

I am working with the PyPi version by the way, which is still `v0.6.4`. I will test with a repo clone next.

Wosser1sProductions:
The example I started from is this one: [demo/scene/scrolling_plots.py](https://github.com/vispy/vispy/blob/master/examples/demo/scene/scrolling_plots.py)

And since there is apparently a EventEmitter loop on the axis visuals, if I remove the `ax.link_view(self.view)` lines, everything behaves the same, but _without_ the EventEmitter loop. And of course without updating the ticks on the axes, since that is what that call connects.

Edit: Going deeper: `widgets.axis.link_view` connects an event callback to [`widgets.axis._view_changed`](https://github.com/vispy/vispy/blob/c59d40270684f1623730219125d426374e2fc0f0/vispy/scene/widgets/axis.py#L81). In that function, the new domain gets calculated and set:

```python
self.axis.domain = (p1[1], p2[1])
```

Now, `visuals.axis.AxisVisual.domain` has the `@updating_property` tag, which causes [this](https://github.com/vispy/vispy/blob/c59d40270684f1623730219125d426374e2fc0f0/vispy/visuals/visual.py#L751) to happen:

```python
if new value != old value:
    self._need_update = True
    self.update()
```

For some reason its this additional call to `update()` that causes the event loop. So the axis requests a domain update, which in turn updates the ticks and then causes the same component to update again from within the event callback chain...

Replacing [those lines](https://github.com/vispy/vispy/blob/c59d40270684f1623730219125d426374e2fc0f0/vispy/scene/widgets/axis.py#L86) with:

 ```python
if self.orientation in ('left', 'right'):
    self.axis._domain = (p1[1], p2[1])
else:
    self.axis._domain = (p1[0], p2[0])
self.axis._need_update = True
```

And thus circumventing the call to update, but still updating the visuals, because the event chain will already do that eventually anyway, seems to work. Strange though how this is not a problem with other backends.

Side note: The axes only update when clicking the view (also on glfw), so when starting and plotting data, the axes are always zrong until you click inside the view.

**Comment by djhoese** [comment]
I think the "side note" issue is only an issue on glfw. I'm 90% sure there is a bug report for it but don't have time to find it.

The `updating_property` stuff is pretty new, but is also not much different from the code that was there before it. This could be an issue in other backends and we just haven't found it. Would you be able to test with the pyqt5 backend?

**Comment by djhoese** [comment]
Maybe it would be best to start with a simpler example? Like the rotating cube which only has a timer update iirc.

**Comment by Wosser1sProductions** [comment]
> I think the "side note" issue is only an issue on glfw. I'm 90% sure there is a bug report for it but don't have time to find it.

Just tested with PyQt5/PySide2 and they also don't update the axes on init or update, only after pressing the view at least ones, so only after the event gets triggered.

> The `updating_property` stuff is pretty new, but is also not much different from the code that was there before it. This could be an issue in other backends and we just haven't found it. Would you be able to test with the pyqt5 backend?

Neither PyQt5, PySide2 or glfw trigger an EventEmitter loop, so it's definitely because of how I wrote it in Tkinter, still unsure what though.

Edit: Wait, so the only thing that changed is that I installed pyqt5 and pyside2 in the virtual env and I removed them again after testing, but it seems the EventEmitter loop I was running in before is not a problem anymore? It just works now. Maybe it was caused by some cached files that got updated by installing something else? Weird.

I guess I'll just move on and make the standalone window work now.

**Comment by Wosser1sProductions** [comment]
### **My last commit fixes all the above issues.** 

And I also tested window-only mode with some of the examples, all seem to work as expected compared to other backends.

Turns out the culprit was probably that merely calling _vispy_update, causes the redraw to happen too fast. So I added a delay for the update call:

```python
def _vispy_update(self):
    self.after(0, self._delayed_update, interval_ms)
```

Doing this seems to fix all the problems I had before, the EventEmitter callback loops are gone, the axes disappearing until the next update when you clicked on them and the poor performance when resizing the window by dragging.

I also included proper handling of creating a native window with `ApplicationBackend` and also creating multiple instances of the `CanvasBackend`, each in a new window. When the last one get closed, the program will exit cleanly. If there was already another Tkinter window created by the user, closing a VisPy created window won't close the application (as would be expected).

I think this is mergeable now, unless there are other things I need to include, like writing tests or examples?

djhoese:
It looks like at least some of the tests (see Travis CI builds) are already trying to use your backend but are failing. If you get those passing and make sure to add docstrings and add comments for the things you've added then I think we're on our way to a merge. Nice job!

Wosser1sProductions:
> Edit: You most likely have to add pyopengltk to the pip dependency list to the travis environments (see .travis.yaml) that test the backends then you should be good.

I have added those for the next commit.

> make sure to add docstrings and add comments for the things you've added

The other backends aren't documented in code. I have some docstrings with the gotchas and basic things that are different from other backends, but not for every function. Is there other documentation I should update? (Maybe saying that Tkinter backend is now supported...)

Wosser1sProductions:
Some notes on Travis:
1. On Travis, pytest fails immediately with `collecting ... Failed: unit failure (1)` and no further info
2. `test_meshio` fails somehow saying the module is not installed, but it's part of the deps? And that's the only reason Travis fails, because Tkinter does not even get tested.

When I run the trests locally (Windows) I have the following results:

```python
77 passed, 4 skipped, 691 deselected, 1 xfailed, 19 warnings in 68.04s (0:01:08)
```

But running `test_arrows` gives:
```
vispy\visuals\tests\test_arrows.py WARNING: Error drawing visual <vispy.visuals.line.line._GLLineVisual object at ...>
```
with a traceback coming from `backend._vispy_warmup` up to:
```
File "vispy\vispy\gloo\glir.py", line 1072, in activate
    gl.glUseProgram(self._handle)
```
Catching a `RuntimeError: OpenGL got errors (glUseProgram): GL_INVALID_VALUE`

I changed the deps to use `PIP_DEPENDENCIES="pyopengltk"` instead, since I first had it added to `CONDA_DEPENDENCIES` and its not available on Conda. Not sure what to do about the `meshio` error or the unit failure.

djhoese:
> The other backends aren't documented in code.

That doesn't mean we can't strive to be better :wink: 
The other part of this is that, so far, you are the only person who knows how this backend works. The maintainers of vispy are going to need as much help to solve bugs as possible. Anything you can do to help document non-obvious decisions would help.

Regarding the tests, they can be difficult. Sometimes different machines and different GPUs produce different results and failures. This is why some tests are skipped on Travis or Appveyor or are marked as expected failures because the expected results are only expected on travis.

I'll take a look now and see if I can make any guesses.

djhoese:
Oh and some tests will look like they are failing but still actually pass. For example, the test output might say it is testing a backend but that backend isn't available. That looks like a failure but is actually expected in most cases.

As for some of your failures, it looks like you have a lot of flake8 issues (code style):

```
Running flake8... 
vispy/app/backends/_tk.py:17:1: F401 '...config' imported but unused
vispy/app/backends/_tk.py:26:5: E722 do not use bare 'except'
vispy/app/backends/_tk.py:30:1: E722 do not use bare 'except'
vispy/app/backends/_tk.py:46:15: E401 multiple imports on one line
vispy/app/backends/_tk.py:70:40: E261 at least two spaces before inline comment
vispy/app/backends/_tk.py:73:40: E261 at least two spaces before inline comment
vispy/app/backends/_tk.py:81:45: E261 at least two spaces before inline comment
vispy/app/backends/_tk.py:94:37: E231 missing whitespace after ','
vispy/app/backends/_tk.py:105:49: E231 missing whitespace after ','
vispy/app/backends/_tk.py:121:49: E231 missing whitespace after ','
vispy/app/backends/_tk.py:153:8: E131 continuation line unaligned for hanging indent
vispy/app/backends/_tk.py:188:6: E231 missing whitespace after ':'
vispy/app/backends/_tk.py:188:9: E261 at least two spaces before inline comment
vispy/app/backends/_tk.py:189:6: E231 missing whitespace after ':'
vispy/app/backends/_tk.py:189:9: E261 at least two spaces before inline comment
vispy/app/backends/_tk.py:190:6: E231 missing whitespace after ':'
vispy/app/backends/_tk.py:190:9: E261 at least two spaces before inline comment
vispy/app/backends/_tk.py:248:9: E722 do not use bare 'except'
vispy/app/backends/_tk.py:248:15: E701 multiple statements on one line (colon)
vispy/app/backends/_tk.py:347:23: E701 multiple statements on one line (colon)
vispy/app/backends/_tk.py:348:22: E701 multiple statements on one line (colon)
vispy/app/backends/_tk.py:349:26: E701 multiple statements on one line (colon)
vispy/app/backends/_tk.py:366:19: E221 multiple spaces before operator
vispy/app/backends/_tk.py:372:9: F841 local variable 'c' is assigned to but never used
vispy/app/backends/_tk.py:386:44: E203 whitespace before ','
vispy/app/backends/_tk.py:390:40: E203 whitespace before ','
vispy/app/backends/_tk.py:393:39: E203 whitespace before ','
vispy/app/backends/_tk.py:394:39: E203 whitespace before ','
vispy/app/backends/_tk.py:395:39: E203 whitespace before ','
vispy/app/backends/_tk.py:396:39: E203 whitespace before ','
vispy/app/backends/_tk.py:399:39: E203 whitespace before ','
vispy/app/backends/_tk.py:405:5: E303 too many blank lines (2)
vispy/app/backends/_tk.py:455:17: E201 whitespace after '['
vispy/app/backends/_tk.py:455:61: E502 the backslash is redundant between brackets
vispy/app/backends/_tk.py:456:21: E127 continuation line over-indented for visual indent
vispy/app/backends/_tk.py:605:20: E701 multiple statements on one line (colon)
vispy/app/tests/test_app.py:143:72: E261 at least two spaces before inline comment
```

djhoese:
When you run the tests locally, are you sure it is running with the tkinter backend?

appveyor.yml
```diff
@@ -11,7 +11,7 @@ environment:
   matrix:
       - PYTHON_VERSION: "3.7"
         CONDA_DEPENDENCIES: "numpy scipy setuptools pytest coverage pytest-cov pytest-sugar pytest-faulthandler cython coveralls freetype-py numpydoc meshio"
-        PIP_DEPENDENCIES: ""
+        PIP_DEPENDENCIES: "pyopengltk"
```
djhoese, line 14: The testing on appveyor is pretty limited. I would just remove this and not test on Windows. It is difficult to fully test (accurately) the visualization stuff on Appveyor (or Windows CI). At least it has been in the past.

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,643 @@
...
+    from OpenGL import GL
+    from pyopengltk import OpenGLFrame
+except:
+    available, testable, why_not, which = False, False, "Import error", None
+    
+    class OpenGLFrame(object):
+        pass
+else:
+    available, testable, why_not = True, True, None
+    which = "Tkinter " + str(tk.TkVersion)
```
djhoese, line 37: What do you think about providing the tk version *and* the pyopengltk version?

Wosser1sProductions: I could add `pkg_resources.require("pyopengltk")[0].version`, or make a pull request to [pyopengltk](https://github.com/jonwright/pyopengltk) to add a `__version__` string in their `__init__.py` file.

I's assume the latter is preferred.

appveyor.yml
Wosser1sProductions: I get some the same errors as [shown on Appveyor](https://ci.appveyor.com/project/vispy/vispy/builds/35423539) locally on Windows, but the tests pass as I said previously. I'll remove the dependencies for Appveyor than though.

djhoese: Yeah I don't think we have the proper GL stuff setup on Appveyor. Errors like:

```
AttributeError: 'LooseVersion' object has no attribute 'version'
```

although not very helpful (there's a bug report for that) usually indicate OpenGL isn't properly installed or available in a way that vispy can find it.

vispy/app/backends/_tk.py
djhoese: Yeah, pyopengltk having a version string would be for the best. `pkg_resources` is known to have performance issues when used in large python environments (the more packages installed, the longer it takes). This was a major issue when vispy switched to a new way of describing its version number that included dynamically checking using pkg_resources. For an optional backend this may not be as big of a deal but it would add time to every startup of a Tk based vispy application.

Wosser1sProductions: I've added a new [PR to pyopengltk](https://github.com/jonwright/pyopengltk/pull/13) to include the version string.
In the meantime, I'll add this to the Tkinter backend:

```python
which_pyopengltk = pyopengltk.__version__ if hasattr(pyopengltk, "__verion__") else "???"
```

Wosser1sProductions:
> When you run the tests locally, are you sure it is running with the tkinter backend?

Yes, I removed the other backends, and I can see that every spawned window has the Tkinter icon.

Below is an abbreviated log:
```
(venv) >python -m pytest --tb=short -m vispy_app_test
================================================= test session starts =================================================
platform win32 -- Python 3.7.4, pytest-6.0.2, py-1.9.0, pluggy-0.13.1
rootdir: vispy, configfile: pytest.ini
collected 773 items / 691 deselected / 82 selected

vispy\app\tests\test_app.py ......
vispy\app\tests\test_backends.py .
vispy\app\tests\test_context.py .x
vispy\app\tests\test_qt.py s
vispy\app\tests\test_simultaneous.py .
vispy\gloo\gl\tests\test_basics.py ...s
vispy\gloo\gl\tests\test_functionality.py ...s
vispy\gloo\tests\test_glir.py ..
vispy\gloo\tests\test_program.py .
vispy\gloo\tests\test_use_gloo.py ....
vispy\gloo\tests\test_wrappers.py ..
vispy\plot\tests\test_plot.py ..
vispy\scene\cameras\tests\test_perspective.py .
vispy\scene\tests\test_node.py .
vispy\util\tests\test_config.py .
vispy\util\tests\test_vispy.py s
vispy\visuals\tests\test_arrows.py .....
 >>> WARNING: Error drawing visual <vispy.visuals.line.line._GLLineVisual object at ...>
 >>> RuntimeError: OpenGL got errors (glUseProgram): GL_INVALID_VALUE
 >>> ... and others
vispy\visuals\tests\test_axis.py ...
 >>> WARNING: Error drawing visual <vispy.visuals.line.line._GLLineVisual object at ...>
vispy\visuals\tests\test_collections.py .
 >>> WARNING: Error drawing visual <vispy.visuals.line.line._GLLineVisual object at ...>

[...] and same until the last one:
vispy\visuals\tests\test_windbarb.py .
 >>> WARNING: Error drawing visual <vispy.visuals.line.line._GLLineVisual object at ...>

---- 77 passed, 4 skipped, 691 deselected, 1 xfailed, 19 warnings in 62.10s (0:01:02) ----
```

So everything is fine up until `test_arrows.py`, then it starts giving the WARNING with the draw visual error, but the test itself still passes. I think it might have something to do with the test trying to use Tkinter in an odd way (that works fine with other backends but not Tkinter specifically).

Wosser1sProductions:
Only the Linux Xenial test with `PYTHON_VERSION=3.7 DEPS=full TEST=standard` fails on Travis, the rest all pass.
And the only output is still:
```
msg:
_VISPY_TESTING_APP=tkinter /home/travis/miniconda/envs/test/bin/python -m pytest --tb=short -m vispy_app_test --cov vispy --cov-report= -ra /home/travis/build/vispy/vispy/vispy
Test session starts (platform: linux, Python 3.7.8, pytest 6.1.0, pytest-sugar 0.9.4)
rootdir: /home/travis/build/vispy/vispy, configfile: pytest.ini
plugins: cov-2.10.1, sugar-0.9.4
collecting ... Failed: unit failure (1)
```

So I guess pytest is failing to collect all test cases?

And on OSX, `$ python -c "import vispy; print(vispy.sys_info())"` fails with
```
ImportError: cannot import name 'OpenGLFrame' from 'pyopengltk' (/Users/travis/miniconda/envs/test/lib/python3.7/site-packages/pyopengltk/__init__.py)
```
So pyopengltk is installed, but somehow not usable?

vispy/app/backends/_tk.py
djhoese: Looks good to me.

djhoese:
Judging by the `__init__` from pyopengltk I'm guessing it doesn't support OSX properly:

https://github.com/jonwright/pyopengltk/blob/master/pyopengltk/__init__.py

There is no import of the Frame for OSX (darwin).

Wosser1sProductions:
> Judging by the `__init__` from pyopengltk I'm guessing it doesn't support OSX properly:
> 
> https://github.com/jonwright/pyopengltk/blob/master/pyopengltk/__init__.py
> 
> There is no import of the Frame for OSX (darwin).
> 
> Edit: Ah it is commented out.

Now you mentioned it, they also have [an issue for this](https://github.com/jonwright/pyopengltk/issues/12), it still needs testing and tweaking apparently, I seem to have overlooked that part. I could remove the `PIP_DEPENDENCIES="pyopengltk"` from OSX from `.travis.yml` again to evade throwing that error OSX. (the tests can't be run anyway then)

What about the pytest collecting failure on Xenial?

djhoese:
The Xenial failure...I honestly don't know. I feel like the last time I saw a failure like that it was related to a segmentation fault. I think if it was an ImportError it would have shown that error. I could try recreating the test environment on my machine, but not sure I'll have the time this week.

Wosser1sProductions:
> The Xenial failure...I honestly don't know. I feel like the last time I saw a failure like that it was related to a segmentation fault. I think if it was an ImportError it would have shown that error. I could try recreating the test environment on my machine, but not sure I'll have the time this week.

I got VirtualBox working again and tried VisPy in an Ubuntu 16 machine (with my fork):

```python
>>> import vispy
>>> print(vispy.sys_info())
Platform: Linux-4.15.0-112-generic-x86_64-with-glibc2.17
Python:   3.8.6 (default, Oct  2 2020, 15:20:46)  [GCC 9.3.0]
NumPy:    1.19.2
Backend:  PyQt5
pyqt4:    None
pyqt5:    ('PyQt5', '5.15.1', '5.15.1')
pyside:   None
pyside2:  None
pyglet:   None
glfw:     None
sdl2:     None
wx:       None
egl:      EGL 1.4 (DRI2) Mesa Project: OpenGL OpenGL_ES
osmesa:   None
tkinter:  Tkinter 8.6 (with pyopengltk ???)
_test:    None

GL version:  '3.0 Mesa 18.0.5'
...
```

The EGL backend does not work (tests: `59 failed, 16 passed, 7 skipped, 308 deselected, 16 warnings`, with a lot of `RuntimeError: Config decorate is not supported by backend`, indicating decorating the spawned window with the default EGL backend does not work).

And after some fiddling, the Qt5 backend does mostly work: `1 failed, 73 passed, 7 skipped, 308 deselected, 1 xfailed, 22 warnings`.

Running my test file with PyQt5 as backend seems to run as intended, trying to use Tkinter though gives:

```
X Error of failed request:  BadDrawable (invalid Pixmap or Window parameter)
  Major opcode of failed request:  55 (X_CreateGC)
  Resource id in failed request:  0x460002d
  Serial number of failed request:  16
  Current serial number in output stream:  18
```

So maybe there is additional OpenGL setup/config required that other backends do but I am not aware of? Or perhaps something is off in the setup done inside [pyopengltk/linux.py](https://github.com/jonwright/pyopengltk/blob/5da68d59d95ac1659e1871f67ccd474761c57468/pyopengltk/linux.py#L19). I also opened an issue for this in [pyopengltk](https://github.com/jonwright/pyopengltk/issues/14).

djhoese:
Maybe it would be worth (either in your VM or in travis) running the pyopengltk demo in their repository? That way we know whether it is your backend and how it is or isn't setting up the environment or whether it is something in how the Travis environment is setup in general (and isn't compatible with what pyopengltk wants to do).

djhoese:
Might be as simple as trying to import the frame in an `if` statement in the travis tests (only run it when in the correct environments).

Wosser1sProductions:
I fixed it on my Ubuntu VM (was caused by trying to map the frame in a window before the window was visble):



Running tests:

```
$ _VISPY_TESTING_APP=tkinter python3 -m pytest --tb=short -m vispy_app_test
... (comfirmed it is actually using the Tkinter backend)
== 74 passed, 7 skipped, 308 deselected, 1 xfailed, 19 warnings ==
```

I'll push the changes and we'll see if that did anything on Travis. If not, we might need to tell Travis that Tkinter can't be used.

Wosser1sProductions:
Well, there we go:

```
_VISPY_TESTING_APP=tkinter /home/travis/miniconda/envs/test/bin/python -m pytest --tb=short -m vispy_app_test --cov vispy --cov-report= -ra /home/travis/build/vispy/vispy/vispy

...

Results (82.78s):
      77 passed
       1 xfailed
       4 skipped
     229 deselected
Passed
```

Wosser1sProductions:
Are there other steps to take or can this be merged now?

djhoese:
Sorry! Your last comments somehow got lost in my inbox and then I got swamped by some work stuff. I'll try to re-review this later today. Any updates from the pyopengltk maintainer(s) on adding a version string or OSX support? Does (or should) your backend raise a detailed error message about the backend not being available on OSX?

Wosser1sProductions:
> Any updates from the pyopengltk maintainer(s) on adding a version string or OSX support? 

It was [being worked on](https://github.com/jonwright/pyopengltk/pull/11), but I don't think it's coming any time soon...
And I can't try to create a fix, since I don't own or have access to any OSX devices.

> Does (or should) your backend raise a detailed error message about the backend not being available on OSX?

Where would I put this (if needed)? Just checking [this](https://github.com/Wosser1sProductions/vispy/blob/c4f65997f4fd07f115d5a363c5ed805acadd5186/vispy/app/backends/_tk.py#L27) and then setting the availability to False etc.?

djhoese:
Hm I was thinking maybe an additional if statement in that `except ImportError` part, but I guess in the end this is up to pyopengltk to display a decent error message. What you have is fine I suppose. I'm sure we'll get a bug report or two about it without a specific message, but that's fine.

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,781 @@
...
+    import sys
+    import os
+
+    _tk_on_linux = sys.platform.startswith('linux')
+    _tk_on_darwin = sys.platform == 'darwin'
+    _tk_on_windows = sys.platform.startswith('win')
+
+    try:
+        import tkinter as tk  # Python >= 3
+    except ModuleNotFoundError:
```
kmuehlbauer, line 32: @djhoese Haven't we removed all python2.7 compatibility code recently? Maybe we can just remove this here?

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,781 @@
...
+    available, testable, why_not = True, True, None
+    which_pyopengltk = getattr(pyopengltk, "__version__", "???")
+    which = "Tkinter " + str(tk.TkVersion) + " (with pyopengltk " + which_pyopengltk + ")"
+
+
+def _fix_tcl_lib():
+    """
+    It is possible that the Tcl library cannot be found when the Python environment
+    does not have the proper variables, so we force them here (mostly when running tests).
+    From: https://github.com/enthought/Python-2.7.3/blob/master/Lib/lib-tk/FixTk.py
```
kmuehlbauer, line 53: Is this also true for Python 3? Or is this just a Python 2 specific problem?

kmuehlbauer (commented):
@Wosser1sProductions @djhoese I've only briefly looked over this, since backends are not really  my area of knowledge. I've added two questions, regarding python2 stuff. Maybe it helps to clean up a bit of code. 

Hope this will go in smoothly. :+1:

vispy/app/backends/_tk.py
Wosser1sProductions: I wasn't sure if Python 2.x should still be supported, if not I'll remove the distiction.

vispy/app/backends/_tk.py
Wosser1sProductions: I have had this problem in the past with Python 2, in particular in conjunction with one-file PyInstaller applications. So it might still occur with Python3, albeit it being rare.

vispy/app/backends/_tk.py
djhoese: Good point. @Wosser1sProductions?

Wosser1sProductions: @djhoese So should I remove it then?

djhoese: Python 2 is no longer supported by vispy, so yeah remove any python 2 specific code. Thanks.

Wosser1sProductions:
Weird, I added the above comments, but now Travis fails the test with no backend due to not able to [fetch some test data](https://travis-ci.com/github/vispy/vispy/jobs/407978483#L7356):
```
E   RuntimeError: Error while fetching file https://raw.githubusercontent.com/vispy/demo-data/master/orig/crate.npz.
E   Dataset fetching aborted (<urlopen error [Errno 8] nodename nor servname provided, or not known>)
```

Probably just some fluke on Travis' part.

djhoese:
Hm that is weird. I've restarted the test...oh you just made a commit. That works too.

Wosser1sProductions:
> Hm that is weird. I've restarted the test...oh you just made a commit. That works too.

Yes I did an empty commit to retrigger CI. (`git commit --allow-empty`)

vispy/app/backends/_tk.py
djhoese: According to https://github.com/python-pillow/Pillow/pull/2455 this is only for Windows and only with Python 2.7. I think it can be removed.

djhoese (changes requested):
Nice job again. I had a few things I was hoping you could change and a few things I didn't understand (mostly from my limited experience with Tk).

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,790 @@
...
+                # There already is a tk.Tk() instance available
+                _tk_inst = tk._default_root
+                _tk_inst_owned = False
+            else:
+                # Create our own top level Tk instance
+                _fix_tcl_lib()
+                _tk_inst = tk.Tk()
+                _tk_inst.withdraw()
+                _tk_inst_owned = True
+        return _tk_inst
```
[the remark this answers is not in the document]
Wosser1sProductions: No there is only a global Tk() instance that every window uses as its parent if there is no other parent available.
This Tk() instance gets created the first time you create any tk related object, like a Toplevel but also just a Label or Button, because it's used to interface with the TCL interpreter that actually does the GUI stuff.

The thing is that if there is already an existing window that the OpenGL frame gets embedded in, you obviously don't want to close that one when closing the OpenGL frame, as that would end the application and that might be unintended.

On the other hand, if VisPy spawns one or more Tk windows, the first one would create the Tk instance and take it as parent, since it doesn't exist yet, and subsequent ones don't have to do that since they take the first one as parent and can just be created. If the user would now close the first window, all created windows would close, since that would destroy the base Tk() instance...
So to circumvent that, I resuse the Tk instance that is user-owned if available, or alternatively create my own (and hide it from view). Then when the last created window gets closed, [the global instance gets destroyed too](https://github.com/vispy/vispy/pull/1918/files/fb9889ebf78885b2d73f9e66f37f665a0727d9be#diff-db4b0a585243bf7475dda995d708b4d6bf0a7dd6b7cd813c5d5cef9ad4bacb47R288).

So to answer your question, Tk does not keep track, the very first thing that gets created, spaws the Tk instance, and when that gets destroyed, everything gets destroyed.

I don't know if it's a common pattern, in normal use you'd create a first window and that would own the instance. And when closing it, you'd expect your entire app to close with it. But since VisPy could spawn multiple windows, closing the first one would close everything.

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,790 @@
...
+_tk_toplevels = []      # References to created CanvasBackend Toplevels
+
+
+def _new_toplevel(self, *args, **kwargs):
+    """
+    Create and return a new withdrawn Toplevel.
+    Create a tk.Toplevel with the given args and kwargs,
+    minimize it and add it to the global list before returning.
+
+    :return: Return the created tk.Toplevel
```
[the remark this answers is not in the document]
Wosser1sProductions: I'll change them all then.

vispy/app/backends/_tk.py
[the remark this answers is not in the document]
Wosser1sProductions: If you want to get rid of `global` statements, I could just encapsulate them inside a static class.

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,790 @@
...
+                _fix_tcl_lib()
+                _tk_inst = tk.Tk()
+                _tk_inst.withdraw()
+                _tk_inst_owned = True
+        return _tk_inst
+
+
+# ------------------------------------------------------------------ canvas ---
+
+class Coord:
```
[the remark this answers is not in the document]
Wosser1sProductions: [Note that the wx backend also has this](https://github.com/vispy/vispy/blob/51f102bfb26fe8f6175f1fdf3dd89b3d193d27ae/vispy/app/backends/_wx.py#L195), but I'll remove it now.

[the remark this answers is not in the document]
djhoese: Hm yeah that is weird. I don't see a reason for them to exist since the creation and the use are both in the same module.

vispy/app/backends/_tk.py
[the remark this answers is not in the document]
djhoese: I'm fine with globals in this case then. Thanks for the explanation.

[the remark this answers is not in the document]
Wosser1sProductions: I already put them together with the classmethods `get_tk_instance`, `get_canvasses`, `new_toplevel` and `del_toplevel` in their own class `_TkInstanceManager`.

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,701 @@
...
+try:
+    import sys
+    import os
+
+    _tk_on_linux, _tk_on_darwin, _tk_on_windows = \
+        map(sys.platform.startswith, ("linux", "darwin", "win"))
+    _tk_pyopengltk_imported = False
+
+    import tkinter as tk
+    from OpenGL import GL
```
larsoner, line 31: This should use `gloo` instead with a nested import, e.g.:

https://github.com/vispy/vispy/blob/master/vispy/app/backends/_qt.py#L683

and then use `gloo` GL methods/constants/whatever

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,701 @@
...
+
+def _new_toplevel(self, *args, **kwargs):
+    """
+    Create and return a new withdrawn Toplevel.
+    Create a tk.Toplevel with the given args and kwargs,
+    minimize it and add it to the global list before returning.
+
+    :return: Return the created tk.Toplevel
+    """
+    global _tk_inst, _tk_toplevels
```
larsoner, line 172: I don't think you need this line since you don't assign to these variables
```suggestion
```

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,701 @@
...
+
+
+def _del_toplevel(fr=None):
+    """
+    Destroy the given Toplevel, and if it was the last one,
+    also destroy the global Tk instance if we created it.
+
+    :param fr: The CanvasBackend to destroy, defaults to None
+    """
+    global _tk_inst, _tk_inst_owned, _tk_toplevels
```
larsoner, line 186:
```suggestion
    global _tk_inst
```

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,701 @@
...
+    # If there are no Toplevels left, quit the mainloop.
+    if _tk_inst and not _tk_toplevels and _tk_inst_owned:
+        _tk_inst.quit()
+        _tk_inst.destroy()
+        _tk_inst = None
+
+
+class ApplicationBackend(BaseApplicationBackend):
+    def __init__(self):
+        BaseApplicationBackend.__init__(self)
```
larsoner, line 206:
```suggestion
```
Shouldn't be needed / this is a no-op wrapper

vispy/app/backends/_tk.py
```diff
@@ -0,0 +1,701 @@
...
+
+    def _vispy_run(self):
+        """Start the Tk.mainloop. This will block until all Tk windows are destroyed."""
+        self._vispy_get_native_app().mainloop()
+
+    def _vispy_quit(self):
+        """Destroy each created Toplevel by calling _vispy_close on it.
+        If there are no Toplevels left, also destroy the global Tk instance.
+        """
+        global _tk_toplevels
```
larsoner, line 238:
```suggestion
```

larsoner (commented):
Other than some minor gripes this looks reasonable to me

larsoner:
(Also I did not read the previous discussions, so let me know if some ground has already been covered or there is some previous discussion that could use my input)

vispy/app/backends/_tk.py
Wosser1sProductions: Globals are moved to their own class `_TkInstanceManager`.

vispy/app/backends/_tk.py
Wosser1sProductions: Globals are moved to their own class `_TkInstanceManager`.

vispy/app/backends/_tk.py
Wosser1sProductions: Globals are moved to their own class `_TkInstanceManager`.

vispy/app/backends/_tk.py
Wosser1sProductions: The backend OpenGL frame, `pyopengltk` explicitly uses the `OpenGL` module, so I also include it here (it's going to be included anyway). I only use it to clear the canvas initially (`GL.glClear` and `GL.glClearColor`).

Wosser1sProductions: The backend OpenGL frame, `pyopengltk` explicitly uses the `OpenGL` module, so I also include it here (it's going to be included anyway). I only use it to clear the canvas initially (`GL.glClear` and `GL.glClearColor`).

larsoner: Can you add a comment along these lines so that the next person who looks at the code understands why it's done this way?

Wosser1sProductions: Ok

djhoese (approved):
I think this has had enough work. Let's see what users think of it and see how it goes. Thanks for all this work.

malekcellier:
Hi!
I was wondering when a version with the tk support will be released ? As far as I can tell, it is not part of 0.6.6.

Cheers and happy new year!
Malek

Wosser1sProductions:
> I was wondering when a version with the tk support will be released ? As far as I can tell, it is not part of 0.6.6.

You can install the most recent version yourself using [these instructions](http://vispy.org/installation.html#via-github) on the VisPy website.

e.g. activate the environment where you want to use it, and run:
```
pip install git+https://github.com/vispy/vispy.git
```

djhoese:
This will be released in the 0.7 release. As usual, I have been completely swamped and have not had much time for VisPy to finish the various 0.7 things I had in mind. It'll happen eventually...

malekcellier:
Thanks for your response!

I completely understand.

I am really thankful such people like you exist! Let me know if there is anything I can do to help with that.

Cheers
Malek

Wosser1sProductions:
There is another problem I encountered. When using a Vispy canvas embedded within a Tkinter window, the app crashes when minimizing and then trying to restore it from the taskbar.
This does not happen if the Tkinter backend is used to create its own Toplevel.

I am not sure why it happens, I have already tried disabling drawing while the parent window is minimised, pausing Vispy events etc., but it keeps crashing or glitches out and then crashes. I'll update here if I find something else.

djhoese:
@Wosser1sProductions Do you get any kind of error message? If so, what is it? There was a Wx backend issue recently (don't have time to find it right now) that made me realize there are right and wrong ways to create the vispy "Application" object that may be impacting your work.

Wosser1sProductions:
Usually, the process (Python itself) just terminates without any Exception thrown or other messages printed. But I've had similar errors to this when testing:

```python
    (... calling code from my application ...)

  File "\vispy\app\backends\_tk.py", line 461, in _delayed_update
    self.tkExpose(None)
  File "\site-packages\pyopengltk\base.py", line 74, in tkExpose
    self._display()
  File "\site-packages\pyopengltk\base.py", line 98, in _display
    self.redraw()
  File "\vispy\app\backends\_tk.py", line 447, in redraw
  File "\vispy\util\event.py", line 454, in __call__
  File "\vispy\util\event.py", line 474, in _invoke_callback
    self, cb_event=(cb, event))
  << caught exception here: >>
  File "\vispy\util\event.py", line 470, in _invoke_callback
    cb(event)
  File "\vispy\scene\canvas.py", line 217, in on_draw
    self._draw_scene()
  File "\vispy\scene\canvas.py", line 266, in _draw_scene
    self.draw_visual(self.scene)
  File "\vispy\scene\canvas.py", line 304, in draw_visual
    node.draw()
  File "\vispy\scene\visuals.py", line 99, in draw
  File "\vispy\visuals\visual.py", line 596, in draw
    v.draw()
  File "\vispy\visuals\mesh.py", line 491, in draw
    Visual.draw(self, *args, **kwds)
  File "\vispy\visuals\visual.py", line 444, in draw
    self._vshare.index_buffer)
  File "\vispy\visuals\shaders\program.py", line 101, in draw
    Program.draw(self, *args, **kwargs)
  File "\vispy\gloo\program.py", line 534, in draw
    canvas.context.flush_commands()
  File "\vispy\gloo\context.py", line 176, in flush_commands
    self.glir.flush(self.shared.parser)
  File "\vispy\gloo\glir.py", line 572, in flush
    self._shared.flush(parser)
  File "\vispy\gloo\glir.py", line 494, in flush
    parser.parse(self._filter(self.clear(), parser))
  File "\vispy\gloo\glir.py", line 819, in parse
    self._parse(command)
  File "\vispy\gloo\glir.py", line 779, in _parse
    ob.draw(*args)
  File "\vispy\gloo\glir.py", line 1323, in draw
    gl.check_error('Check before draw')
  File "\vispy\gloo\gl\__init__.py", line 207, in check_error
    raise err
RuntimeError: OpenGL got errors (Check before draw): GL_INVALID_VALUE
```

This is on `vispy-0.6.5.dev280+g2dc8570b` by the way, since the Tkinter backend wasn't merged yet.

My scene contains a `visuals.Plane`, `visuals.XYZAxis` and some `visuals.Text` (without loading any data). If I never minimise the window everything works as expected.

PR #1919, opened 2020-09-23
Use oldest-supported-numpy in pyproject.toml

astrofrog:
I've switched the pyproject.toml file to use the [oldest-supported-numpy](https://github.com/scipy/oldest-supported-numpy) meta-package which is equivalent to specifying all the environment markers (see the README for that project) but means that the pyproject.toml file here doesn't need to be updated in future when a new version of Python is released since this just requires a new release of oldest-supported-numpy.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,11 +8,7 @@ requires = [
     # "setuptools_scm[toml]>=3.4",
     "setuptools_scm_git_archive",
     "Cython>=0.29.2",
-    "numpy; python_version=='2.7'",
-    "numpy==1.13.3; python_version=='3.5'",
-    "numpy==1.13.3; python_version=='3.6'",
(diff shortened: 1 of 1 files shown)


djhoese:
Cool! Any idea when Python 3.9 will be available/added to this? If I have to wait a week for a release of vispy 0.7 to get a Python 3.9 wheel I will.

djhoese:
Hhhmmm or maybe this won't matter for wheel building since we are using cibuildwheel which already has cp39.

astrofrog:
Python 3.9 will be added to that package as soon as possible, and the nice thing is that nothing will need to change in VisPy's pyproject.toml file for this to work.

djhoese:
Ah so that would help for people building from source. Ok.

djhoese (approved):

djhoese:
Thanks again!

PR #1920, opened 2020-09-25
Add 'texture_format' kwarg to ImageVisual for floating point textures

djhoese:
Similar to #1912 

This PR adds the `texture_format` keyword argument to the ImageVisual to allow for more complex data types to be stored internally on the GPU. This should allow for a greater bit depth and faster color limit changes.

One major thing I learned that will also need to be fixed in #1912 is what these internal format types actually look like in the GPU. I think the test originally created by @tlambert03 for the VolumeVisual has all zeros and fails to test this properly so we'll need to come up with something better. The main lesson is with normalized or not normalized formats. See https://www.khronos.org/opengl/wiki/Image_Format. Formats like `r8` (`GL_R8`) are normalized unsigned integers. This means that although the data is stored/sent as unsigned 8-bit integers, it appears in the shader as a [0, 1] floating point number. This means that if your color limits (clims) are in your data's original range, you'll be scaling the data wrong. The clims (or any other variable interacting with the texture data) need to be scaled/normalized in the same way the texture data is. I tried to do that in this PR for the ImageVisual.

Edit: The other part of the normalized stuff is that formats like `r32f` don't need to be normalized because they are already floats. Other formats supported by OpenGL 4 that vispy doesn't support yet are also not normalized (those ending in `ui` or `i`). There are also `_SNORM` formats which are normalized in the range of `[-1, 1]`. I'm not exactly sure how unsigned inputs are treated for this, but it isn't supported by vispy yet either so I just left a `XXX` in the code that it should be dealt with.

## TODO

- [ ] Add tests
- [ ] Remove all FIXMEs and XXXs (if they can be resolved/discussed)
- [ ] Come up with a better name for `normalize_value` on the `BaseTexture`


djhoese:
Oh I suppose I should have marked this as WIP, but I want feedback to feel free to provide it.

djhoese:
@rougier You gave your feedback on #1912, but I'd actually like to finish this PR first. If you have the time do you think you could look at this?

djhoese:
@jni @tlambert03 @sofroniewn I've run into an odd situation with this PR that I wanted to run by you. I already knew (from earlier work on this) that if you give a texture data of a different size, it will resize itself (creating a new texture object). What I figured out today is that it will also change the internalformat if it has to. This means that if you change from RGB to RGBA (or similar) and you don't explicitly tell the texture what to use, the texture internally decides what the new format should be. This is only really a problem if the original texture format was specified by the user so data normalization is meant to happen on the GPU. This was kind of a surprise to me, but I see a few options for how I could handle it:

1. Detect the change in internalformat and raise an exception that the internalformat must change to handle GPU-side normalization for this data.
2. Detect the change in internalformat, print warning, and then switch to CPU-side normalization.
3. Detect the change in internal format, check if the original `texture_format` was 'auto' and re-determine the texture_format. If it wasn't 'auto' then either 1 or 2 above.
4. Same as 4 except if texture_format was specified, try to find an equivalent internalformat that would also work. Like if it was rgb8 and now we have an RGBA data array, switch to rgba8.
5. Something else?

jni:
From memory, I *don't* think we care about this in napari — we'll generally create a new texture if the image size changes (@sofroniewn @tlambert03?), so we could keep track of the format ourselves and do this.

If you can do (1) in a backwards compatible way, then from our side as consumers of this API that would be best, because we can quickly figure out all the places where we can trip up and find the "right" approach. (3 -> 1) also is fine by me.

I am writing this from a very naive perspective so I'll *gladly* be overruled by @sofroniewn @tlambert03 and others.

sofroniewn:
Yes, I agree I don't believe we do this, so option 1 if easiest would be fine. Option 2 also seems reasonable. I see the appeal of option 3 and 4 though am now sure how much extra work they would be.

sofroniewn:
Yes, I agree I don't believe we do this, so option 1 if easiest would be fine. Option 2 also seems reasonable. I see the appeal of option 3 and 4 though am now sure how much extra work they would be.

djhoese:
I've been thinking it over a bit and I think I like 3 the most. I'd like to avoid "magic" if possible and I think 4 is a little bit too much "VisPy knows best and so VisPy will do things for you without telling you". Option 2 is a little too magical too in that if the user isn't paying attention it will do completely different things under the hood. Option 1 is the least magical but also the least useful (by itself at least). So I'll see what I can do later today if I do 3 + 1, but add an additional `texture_format` parameter to `set_data` that will default to reusing the same texture_format if possible (including auto) and raise an exception if it can't be reused. That way, if a user provides new data and says `texture_format='auto'` we (VisPy) know they mean "I know I might be changing the type of data I'm providing, but I still want GPU-side normalization".

...why do these things always take more text to explain than I think they will.

djhoese:
Oh no. I might hate myself. I think I'm going to refactor this to pull everything into a subclass of Texture2D (maybe a Mixin). The more and more I work on this it is very obvious that there is too much "how do we handle this data specific to this texture object" and needing to pull internal texture information into the ImageVisual. I don't (currently) want to move all of this logic into the regular BaseTexture/Texture2D classes as I like that those should remain "simple". After ImageVisual makes sense with this subclass and VolumeVisual (possibly with a 3D version) then we can promote things to a more public/global module or consider adding it to the base texture class.

To any devs who care, speak now.

djhoese:
It works! That actually wasn't that hard (the old implementation, not the new stuff I asked about). Now I should see if I can separate the texture subclasses into a CPU-scaled one and a GPU-scaled one. Damn this code was ugly.

djhoese:
Still working on this but I've hit some confusion that has wasted my time because I'm not sure how things are supposed to work. It looks like some ambiguity was introduced in the semi-recent #1844. The main question is should RGB/A data formats use color limits and gamma? It seems in vispy 0.6 it was relatively clear that RGBs don't get scaled at all regardless of data type (float, uint8, etc). In the master branch it is kind of "by accident" due to the color limits defaulting to (0, 1). For the napari folks, do you need clims/gamma for RGBs?

If RGBs are scaled, then that means I basically **have** to convert the data to 32-bit floats before sending them to the GPU (can't do linearly scale integers without losing some precision). I think this was happening implicitly with single band (L) image data during the scaling process, but it is also a less common use case (uint8s for single band data).

Note: In my own work it is very common to have RGB images made of 3 separate input arrays that have their own color limits and gammas so this will likely be a new separate Visual class in the future.

CC @sofroniew @jni @tlambert03 Does napari allow for color limits and gamma adjustments on RGB data?

tlambert03:
> CC @sofroniew @jni @tlambert03 Does napari allow for color limits and gamma adjustments on RGB data?

yeah, for better or worse, we do.  Definitely don't want to be wasting your time here though.  feel free to let us know how you want this to work for vispy, and we can discuss what needs to be done on the napari side if need be.  I'm sorry for any code ugliness.  It was a situation where I didn't want to make "major" changes to your objects, and so most of the ugliness was an attempt to patch things in optionally.

djhoese:
Ok, interesting. Does it actually work? It seems that depending on if you specify the clims before or after the ImageVisual is created, you might get different results. For RGBA images, do you expect gamma and the color limits to effect the Alpha channel? The thing that's odd about the code is that it isn't clear (mostly with the changes you and I made) what happens to RGB data when color limits are used. The "texture_limits" logic makes it even more confusing. I think I just need to make some hard decisions/rules about handling RGB data. I think in a lot of cases we could assume that RGB data doesn't need to be scaled on the CPU at all (whether it is floating point from 0-1 or uint8 from 0-255), but that also kind of defeats the purpose of the on-CPU scaling.

tlambert03:
> Does it actually work?

here's how it looks in napari:

https://user-images.githubusercontent.com/1609449/107889031-4ab36d00-6ede-11eb-9fc0-e4163f8f9722.mov


>  I think in a lot of cases we could assume that RGB data doesn't need to be scaled on the CPU at all (whether it is floating point from 0-1 or uint8 from 0-255)

I agree, it seems like a pretty unusual case to not have pre-scaled RGB data.

djhoese:
Ok, good. I think this is how the code worked it is just not super clear (at least to me) and it is possible some of my local changes in this PR have broken it so I'll make sure to fix it.

djhoese:
@tlambert03 Oh new question: Is it possible to set the clims outside of the RGB's range? I suppose if it is normalized then it shouldn't be possible to go outside the range, but VisPy should probably stop the user (exception) from doing it too, right?

tlambert03:
ah, interesting.  it is "possible"... but it doesn't do anything but appear to reset the "semantics" of the clims range.  definitely odd, and nothing we need to preserve here.  here's a demo:


https://user-images.githubusercontent.com/1609449/107891062-4db45a80-6eea-11eb-8323-5ad2008808e6.mov

djhoese:
@tlambert03 @jni @sofroniewn Ok, I think functionally this does everything I expect it to do. I'm sure there are corner cases that won't work but these are cases where I'm not sure what the user would be trying to accomplish so :man_shrugging: 

If there is any way you can do some basic functionality tests with napari that would be great. It would be great if you could test it with your existing set of keyword arguments (so it scales on the CPU) and when trying to pass `texture_format='auto'` which I assume would be all that's needed for the more performant version.

I still need to add documentation in the ImageVisual and the two texture helper classes I've made to clear up some of the behavior and the purposes of other parts.

tlambert03:
@djhoese, one more thought I had while playing with complex data: what would you think about making `_build_color_transform` a method on `ImageVisual` (or otherwise accessible, via a `color_transform` parameter or something)?  That's a pretty central bit of functionality I think we often find ourselves wanting to tinker with.

djhoese:
Yes! Absolutely. That was my plan for yesterday, but got caught up with writing notes of the sprint discussion. I hope to finish this today.

djhoese:
@tlambert03 Done. Do you think it should have a defined interface or just give it free rein to use `self`? It isn't exactly a defined interface right now so it is hard to say. Should I also add class attributes for all of the shader chunks? That seems not so pretty, but lets us do basically anything we want for subclasses.

tlambert03:
> Do you think it should have a defined interface or just give it free rein to use self? It isn't exactly a defined interface right now so it is hard to say. Should I also add class attributes for all of the shader chunks? That seems not so pretty, but lets us do basically anything we want for subclasses.

hmm. I think it's probably ok as is for now.  Just having the `_build_color_transform` method is already enough access for subclasses, (they can grab/copy the little clim/gamma shader chunks if needed).

If we _did_ try to expose the shaders, I feel like we'd have to try to impose a little API for those 4 parts of the `FunctionChain` which seem to be common to all color transforms:
1. prepare/convert data from texture:
   - for luminance, this is `vec4 -> float` (`_c2l_red`)
   - for RGB, this is `vec4 -> vec4` (`_null_transform`)
   - for complex data, this would be one of [these little `vec4 -> float` snippets](https://github.com/djhoese/vispy/compare/feature-image-texture-format...tlambert-forks:complex#diff-c26e206f620ac7424b6d2541c4727a67bf982841b9212dfbb6a6ce4ac1fb13a2R447-R469)
2. apply clim (type appropriate for step 1 output)
3. apply gamma (type appropriate for step 2 output)
4. optionally apply cmap `float -> vec4`, or no-op for already RGB data

.. and then provide some way to switch the chain based on data. but that all feels like something to consider in a followup PR

djhoese:
Good point. Now that we talked about primitive graphics yesterday I can't stop questioning everything for what should be an interface and what shouldn't.

djhoese:
Restarting tests as it seems the failures were on the Tk backend only.

tlambert03:
> Try wild guess and move requires application close to test

:joy:

djhoese:
I figure I can always force push and go back :stuck_out_tongue_winking_eye:

djhoese:
For more detail, the tests seem to be failing on the new 'tkinter' backend, but only for this PR which means it has to be this PR that is doing something different. My current assumption is something with the pytest parametrize and needing a new application. Especially since #1993 was filed. Commenting out the first test that failed just changes which test is failing and they aren't all ImageVisual related. Last resort I will comment out the tkinter backend from tests and we'll consider it experimental for the 0.7 release.

djhoese:
@tlambert03 @rougier @larsoner @kmuehlbauer I think this PR is ready for final review. You have all either shown interest or I trust you to review the PR. This PR got pretty large, but hopefully the general idea is understandable. Once this is merged I can start working more on #1912.

vispy/visuals/image.py
```diff
@@ -171,7 +518,28 @@ class ImageVisual(Visual):
             * 'hanning', 'hamming', 'hermite', 'kaiser', 'quadric', 'bicubic',
                 'catrom', 'mitchell', 'spline16', 'spline36', 'gaussian',
                 'bessel', 'sinc', 'lanczos', 'blackman'
-
+    texture_format: numpy.dtype | str | None
```
kmuehlbauer, line 521: It would be good to have examples showing the different behaviour, maybe a dedicated gallery section? Should be worked on after discussion of details.

vispy/visuals/image.py
```diff
@@ -199,28 +572,15 @@ def __init__(self, data=None, method='auto', grid=(1, 1),
...
-        self._interpolation_names.sort()
-        self._interpolation_names = tuple(self._interpolation_names)
-
-        # overwrite "nearest" and "bilinear" spatial-filters
-        # with  "hardware" interpolation _data_lookup_fn
-        self._interpolation_fun['nearest'] = Function(_texture_lookup)
-        self._interpolation_fun['bilinear'] = Function(_texture_lookup)
-
-        if interpolation not in self._interpolation_names:
+        interpolation_names, interpolation_fun = self._init_interpolation(
```
kmuehlbauer, line 575: :100:  for moving this into dedicated init-function.

kmuehlbauer (approved):
@djhoese :+1: I skimmed over the changes and added some minor remarks.

rougier (approved):

vispy/visuals/image.py
djhoese: What do you think the example could show? I could make an example (or update an existing one) that shows the same image with the old behavior and one with a floating point texture. :thinking: Although the main benefit of this PR is the performance. Otherwise, we'd be trying to show the quantization/precision loss of the old behavior which may not be that great of an example.

vispy/testing/_runners.py
```diff
@@ -389,6 +389,9 @@ def test(label='full', extra_arg_string='', coverage=False):
     if label not in known_types + backend_names:
         raise ValueError('label must be one of %s, or a backend name %s, '
                          'not \'%s\'' % (known_types, backend_names, label))
+    # remove troublesome backends
```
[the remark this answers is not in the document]
djhoese: Good call. Done #2009 and mentioned in latest commit.

vispy/gloo/texture.py
```diff
@@ -67,7 +67,7 @@ class BaseTexture(GLObject):
     _inv_internalformats = dict([
         (base + suffix, channels)
         for base, channels in [('r', 1), ('rg', 2), ('rgb', 3), ('rgba', 4)]
-        for suffix in ['8', '16', '16f', '32f']
+        for suffix in ['8', '16', '16f', '32f']  # , '8i', '8ui', '32i', '32ui']
```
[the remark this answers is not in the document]
djhoese: Done in latest commit. Thanks.

vispy/visuals/image.py
kmuehlbauer: Oh, maybe gallery is not the right place. What I had in mind was kind of how-to example.

djhoese: Example in `<vispy root>/examples/` or in the docstring itself? Or somewhere else? Maybe it comes in a future PR?

kmuehlbauer: Yeah, sure, future PR! That's what I meant with "after discussion of details". Sorry for not being clear enough. And it would well fit in examples-folder.

djhoese:
I'll take 2 approvals as this being good to go (since we don't have a defined procedure for this). Merging now. If anyone finds anything that needs to be changed in the next couple days comment here. Otherwise, file an issue or PR about the fix.

PR #1921, opened 2020-09-28
Fix EGL docstring copy/paste error

BioGeek:
eglTerminate had the docstring of eglInitialize. Now replaced with the docstring as used in https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglTerminate.xhtml


djhoese (approved):
Nice catch. Thanks!

PR #1928, opened 2020-09-30
Remove bundled decorator dependency

### LINKED ISSUE #1922: Replace bundled "decorator" dependency
We currently have a dependency on the "decorator" library, but include it as a bundled module here:

`vispy/ext/_bundled/decorator.py`

and then import it here:

`vispy/ext/decorator.py`

where the `decorator` decorator is only used once here:

https://github.com/vispy/vispy/blob/397c3122d5644692e6e1ae6d281b7b72d741fae6/vispy/visuals/transforms/_util.py#L85-L86

I think this dependency could be removed if `arg_to_vec4` was rewritten to use https://docs.python.org/3/library/functools.html#functools.wraps or a similar standard library utility.

@larsoner you probably know the most about this, any concerns? Otherwise I say we let a hacktoberfest participant handle this.

sgaist:
This refactors the `arg_to_vec4` decorator using Python's functools module and drops the bundled `decorator` module.

Closes #1922


djhoese (approved):
The tests haven't passed yet, but looking at the code... BEAUTIFUL!

Thanks!

djhoese:
Are you doing this for hacktoberfest? Does this count already?

djhoese:
Closes #1922

sgaist:
@djhoese Yes I am and AFAIK yes it does.

I updated the merge request message since it was missing the "closes".

PR #1931, opened 2020-10-05
Overhaul vispy website

djhoese:
See https://github.com/vispy/vispy-website/pull/29 for the real information and big changes. This PR, in addition to pointing to the newest `vispy-website` submodule, only includes various fixes needed to fix warnings and runtime issues when generating the docs now that they more completely cover the package's API.


larsoner (approved):

djhoese:
FYI @kmuehlbauer I've updated the github workflow to try to install vispy with `pip install -e .` which should be the "develop mode" of things. We previously had `python setup.py install` followed by `python setup.py develop` which was due to either the javascript building not going on with a develop install alone or the cython building not happening during a develop install. Let's see if this works now that pip has made some major improvements in the last couple years and is the preferred way of doing a development install.

...and everything fails.

djhoese:
FYI to @aaru143, in my last commit I added a `@pytest.mark.xfail` to `vispy/util/tests/test_docstring_parameters:test_docstring_parameters` because as part of this PR I made this testing much more strict by using a newer version of numpydoc docstring testing that @larsoner had pointed out. This PR is way too big to also include all the docstring fixes. As part of your work on cleaning up docstrings I would include removing this `xfail` decorator. If you then run this `test_docstring_parameters` test you should see all of the things that are currently failing (and therefore need to be fixed).

djhoese:
Ok @aaru143 I've now also commented out the docstring specific test in `.github/workflows/main.yml` and added a FIXME. This should also be re-enabled after fixing the docstrings.

PR #1933, opened 2020-10-05
Remove six dependency

### LINKED ISSUE #1926: Remove 'six' bundled dependency
The `six` package is bundled in vispy here: vispy/ext/_bundled/six.py
And imported here: vispy/ext/six.py

This should be removed and *all* usage of the `six` library removed/replaced with python 3 equivalent checks throughout vispy.

sgaist:
This merge request removes the bundled six library and updates the relevant parts to Python 3

Fixes #1926


vispy/app/application.py
```diff
@@ -106,7 +105,9 @@ def is_interactive(self):
 
         # IPython does not set sys.flags when -i is specified, so first
         # check it if it is already imported.
-        if '__IPYTHON__' not in dir(six.moves.builtins):
+        try:
+            get_ipython()
+        except NameError:
```
djhoese, line 110: Are you sure this behaves the same? Wouldn't this work:

```
import builtins
if hasattr(builtins, '__IPYTHON__'):
```

Maybe these are essentially the same?

vispy/testing/image_tester.py
```diff
@@ -49,9 +49,9 @@
 from subprocess import check_call, CalledProcessError
 import numpy as np
 
-from ..ext.six import string_types
-from ..ext.six.moves import http_client as httplib
-from ..ext.six.moves import urllib_parse as urllib
+import http.client as httplib
```
djhoese, line 52: From what I can tell, this module only uses HTTPConnection from http.client and urlencode from urllib.parse, mind just changing these to import those directly?

djhoese (changes requested):
Nice job! What a huge PR. Thanks for doing all this work. I had two small comments (one question, one change request). Otherwise this looks good.

vispy/app/application.py
sgaist: AFAIK, it should however you are correct, we can keep the original logic. I somehow messed up something in my installation when doing the checks to update the code.

vispy/testing/image_tester.py
sgaist: No problem with that

djhoese (approved):
Thanks!

vispy/util/fetching.py
```diff
@@ -12,8 +12,8 @@
 import shutil
 import time
 
-from ..ext.six.moves import urllib
-from ..ext.six import string_types
+import urllib
```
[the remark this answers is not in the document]
djhoese: I only see a `urllib.request` usage so doing `import urllib.request` should work.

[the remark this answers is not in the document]
sgaist: Indeed, that's the correct fix.

djhoese:
Looks like there are a couple examples that are using six too:

```
Example /basics/scene/flipped_axis.py failed (1):

  File "<string>", line 12, in <module>

  File "/home/travis/build/vispy/vispy/examples/basics/scene/flipped_axis.py", line 25, in <module>

    from vispy.ext.six import next

ModuleNotFoundError: No module named 'vispy.ext.six'
```

sgaist:
My search-fu was bad on these two... Fixed

djhoese:
I think you forgot to push those fixes :wink:

sgaist:
Almost, unlocking the key failed and I missed that 😅

djhoese:
Travis notification is stuck, but is passing. I think this all looks good. Thanks!

PR #1943, opened 2020-10-25
Removed Python 2.7 wrapper on GzipFile

### LINKED ISSUE #1923: Remove Python 2.7 wrapper around GzipFile
This module: https://github.com/vispy/vispy/blob/9dbe751dcdbaaa1d17d0a5885f15872b9b0741a8/vispy/ext/gzip_open.py

Is used in two places in vispy (one module) and only exists to be used by Python 2.7 environments which we no longer support. This module and it's usages should be replaced by `GzipFile`: https://docs.python.org/3/library/gzip.html#gzip.GzipFile

Usage 1:

https://github.com/vispy/vispy/blob/9dbe751dcdbaaa1d17d0a5885f15872b9b0741a8/vispy/io/wavefront.py#L70

Usage 2:

https://github.com/vispy/vispy/blob/9dbe751dcdbaaa1d17d0a5885f15872b9b0741a8/vispy/io/wavefront.py#L254

And of course the import in that module should be removed.

irajasyed:
Fixes #1923 

Changes:

- Changed usages of gzip_open with GzipFile 
- Removed the gzip_open extension file.


djhoese:
As mentioned in the issue, this task has been assigned to someone else. If they do not complete it in the next week then I'll review this one. Thanks for understanding.

irajasyed:
@djhoese i saw the assignee got removed themselves from the issue 6 days ago. That’s why i proceeded with the PR. 
:)

djhoese:
Ah I was looking at the issue on mobile and didn't see that change. Thanks.

djhoese (approved):
Looks good. Thanks!

PR #1954, opened 2020-11-27
Enable Python 3.9 wheels

astrofrog:
If this works, we could backport to maint/0.6 and do a quick 0.6.7 release which includes 3.9 wheels?


djhoese:
Ah I just saw this after I already made the direct change. I enabled python 3.9 builds and disabled all pypy builds. Thanks for getting it this far though. VisPy 0.6.6 is building now.

PR #1956, opened 2020-12-04
Modified data type in create_sphere()

desteemy:
Modified output data type in create_sphere() using 'ico' method. verts and faces are now np.float32 and np.uint32 dtype arrays. Before the change, they were created as np.float64 and np.int32. Passing faces to vispy.gloo.IndexBuffer without additional conversion resulted in error "Invalid dtype for IndexBuffer: dtype('int32')" in line 503 in function _prepare_data() from vispy.gloo.buffer.py.
verts dtype was changed to standardize with other methods of creating sphere - 'latitude' and 'cube'
Maybe the conversion of types in _prepare_data() should be improved but I don't know if that will lead to errors and unwanted behavior elsewhere.


djhoese:
Thanks for this. This makes sense given how the other methods work.

> Maybe the conversion of types in _prepare_data() 

What did you mean by this? Which `_prepare_data`?

Maybe you could add some tests for these methods?

djhoese:
Ah looks like there are actually already tests in `test_generation.py` and that's what is failing in the automated tests for this PR:

```
_________________________________ test_sphere __________________________________

    def test_sphere():
        """Test sphere function"""
        md = create_sphere(rows=10, cols=20, radius=10, method='latitude')
        radii = np.sqrt((md.get_vertices() ** 2).sum(axis=1))
        assert_allclose(radii, np.ones_like(radii) * 10)
        md = create_sphere(subdivisions=5, radius=10, method='ico')
        radii = np.sqrt((md.get_vertices() ** 2).sum(axis=1))
>       assert_allclose(radii, np.ones_like(radii) * 10)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 32 / 20472 (0.156%)
E       Max absolute difference: 1.9073486e-06
E       Max relative difference: 1.9073487e-07
E        x: array([10., 10., 10., ..., 10., 10., 10.], dtype=float32)
E        y: array([10., 10., 10., ..., 10., 10., 10.], dtype=float32)
```

Very interesting. Feel free to update the tests (change the precision or change the expected values) so they pass and then you an also add a check to make sure that `radii.dtype.type == np.float32` (I think that's the recommended way of doing dtype equality).

djhoese:
Linting was complaining about no blank line at the end of `test_generation.py`. I've fixed it so now let's see if the tests pass. Thanks for the updates.

desteemy:
Done. I have changed absolute tolerance value and added this extra type assertion. It's ready to merge.

PR #1968, opened 2020-12-21
Add totalRotationAngle and totalScaleFactor to touch (pinch) events.

### LINKED ISSUE #1964: Expose totalRotationAngle and totalScaleFactor in touch event ?
In [here](https://github.com/vispy/vispy/blob/7f3f6f60c8462bb8a3a8fa03344a2e6990b86eb2/vispy/app/backends/_qt.py#L509-L536), would it be ok to also expose the two above attributes; as well as maybe a reference to the native event (weak ref if you are concerned).

`total*` contain the value since the begin touch event; we cna of course accumulating the current delta when pinching/rotating but this accumulate error; and can be error prone as you also have to catch the begin and end events as well.

There might be reasons not to do that (have the same event struct across backend?) but I haven't found other code triggering `.touch()` events. And in that case I thin that a private fields `_native` that contain the raw even could be sufficient.

Note that there is also totalChangeFlags I believe; but I do not have any thoughts on this one, though I guess it makes sens.

Thanks.

Carreau:
Those two are supposed to hold the total scale change and rotation
change since the beginning of the touch event, and are easier to
manipulate than to accumulate the delta changes.

Note that testing locally on MacOS 10.12 using a macbook pro with a
touchpad the rotation does not seem to follow the description of the Qt
framework. That is to say:

 - event.rotation: total rotation since begin event
 - event.scale: Change of scale since last event, generally between 0.95 and 1.05.
 - totalRotationAngle: Always 0 for me
 - totalScaleFactor: Total change of scale change since evt.type=begin event.

I still include those two, as in particular it seem on some devices you
can "interrupt" and "restart" a gesture; especially for rotation, and
even if totalRotationAngle seem broken on my end; totalScaleFactor is
useful.

Closes #1964

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -517,13 +517,16 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
                 scale = gesture.scaleFactor()
                 last_scale = gesture.lastScaleFactor()
                 rotation = gesture.rotationAngle()
-                self._vispy_canvas.events.touch(type='pinch',
-                                                pos=(x, y),
-                                                last_pos=None,
-                                                scale=scale,
-                                                last_scale=last_scale,
-                                                rotation=rotation,
-                                                )
+                self._vispy_canvas.events.touch(
+                    type="pinch",
+                    pos=(x, y),
+                    last_pos=None,
+                    scale=scale,
+                    last_scale=last_scale,
+                    rotation=rotation,
+                    total_rotation_angle=gesture.totalRotationAngle(),
+                    total_scale_factor=gesture.totalScaleFactor(),
+                )
         # General touch event.
         elif (t == QtCore.QEvent.TouchUpdate):
             points = ev.touchPoints()


djhoese (commented):
Thanks for putting this together. A couple questions:

1. What are your thoughts on more pythonic names "total_rotation_angle" and "total_scale_factor"? Would those be useful if other backends started using them?
2. Is there a non-ridiculous way of testing this in the unit tests?

Carreau:
> 1. What are your thoughts on more pythonic names "total_rotation_angle" and "total_scale_factor"? Would those be useful if other backends started using them?

None, I I think consistency is good but usually don't work with Qt. I'm happy if you leave this PR sitting for some time; I'll be working more on napari next year and folks there might have a stronger opinion.

> Is there a non-ridiculous way of testing this in the unit tests?

I'll likely implement some handler for that in Napari next year; and will likely have conditional to check for those attributes; we can wait until this is done and working, to merge this PR, so that at least it's in the hands of users and maybe exercised in Napari test suite ?

jni:
I have a slight preference for "Pythonification" of variable names, with the exception of subclasses of Qt classes. Looking forward to working together @Carreau! 😊

Carreau:
rebase from 83a94bcc to update the naming and commit message

djhoese:
Thanks for the rename. If there aren't already tests that cover this event, then I am ok not adding a test in this small case because adding these new properties shouldn't break anything (famous last words). If we have tests that mimic mouse events then we may be able to easily add this, but I can't find anything in the existing tests.

kmuehlbauer:
@djhoese Is this good to go from your side (besides the tests?) I'll add it to the sprint project.

djhoese:
@kmuehlbauer Yes. Feel free to merge if you agree that it doesn't really need tests or not if you feel it does need them.

PR #1973, opened 2021-01-14
MNT: move CI to GitHub Actions

kmuehlbauer:
This PR is an effort to move CI to GitHub Actions. Starting with a simple lint, other jobs/steps are added one by one.

Current capabilities:

- General Checks
  -  [x] check OpenGL
- Lint and Stylechecks (use system python via pip on `ubuntu-latest`)
  - [x] lint with flake8
  - [x] check line endings LF vs. CR/LF (really needed nowadays?)
  - [x] check docstring parameters 
- Website (deps via pip)
  - [x] build
  - [x] upload implemented (but not tested)
- Test Harness (use conda-forge environment via micromamba)
  - [x] linux minimal tests
  - [x] linux full tests
  - [x] linux example tests
  - [x] linux osmesa tests 
  - [ ] macos tests (in follow-up PR)
  - [ ] windows tests (in follow-up PR)
- Build wheels and upload to pypi
  - [ ] pypi (in follow-up PR)


kmuehlbauer:
@dhoese 

Almost there.

It looks like it isn't that easy to recreate the build matrix as is efficiently. The test invocation is really not that easy to understand. Is the big cruft of boilerplate code really necessary? Anyway...

OK, I've currently ported 4 linux runs (minimal, full standard, full examples and osmesa). minimal and osmesa seem to work fine, the other two (full deps) seem to have some problems with the tests, although the version info looks fine. We might have to revisit the apt dependencies. Maybe you can have a look at the logs.

The job steps are now:

- checkout
- install apt-dependencies
- install micromamba
- install conda-environment
- install pip dependencies
- install vispy
- start xvfb as daemon
- version info
- run tests
- coveralls

djhoese:
Some initial impressions based on this and the limited experience I have on github actions so far:

1. Why not use conda environment YAML files instead of text? Then we can specify the channels and separate out conda dependencies and pip dependencies?
2. Why not use the conda-incubator: https://github.com/pytroll/satpy/blob/master/.github/workflows/ci.yaml#L30-L38. It even has experimental support for mamba: https://github.com/conda-incubator/setup-miniconda#example-6-mamba

kmuehlbauer:
@djhoese Yes, I considered that (conda/mamba), but it's really slow compared to micromamba. Especially in the stage of setting this up.

But sure, we can adapt this later.

djhoese:
Ah sounds good then. That's why I was asking "why", I just wanted to know the reasoning and double check that you were aware of the alternative. Semi-off topic: How hard is it to make an action like conda-incubator to use micromamba (to avoid all the installation step like you have it)? I suppose someone is probably working on that already.

kmuehlbauer:
I'm waiting for this every day :grinning:

kmuehlbauer:
@djhoese Now the linux part is almost done. Only the examples (running with PyQt5 backend have some weir error).

djhoese:
1. The `XDG` warnings should maybe just have an explicit environment variable definition to avoid the warning. It sounds like the container being used isn't properly configured like a "standard" desktop would be.
2. The shader compilation error is likely due to an old version of OpenGL running on the machines.
3. The shared context error sounds like it is an example that shouldn't be run or is using the wrong backend. I'm 90% sure I explicitly skip running this example (in the `_runners.py` module maybe?). Basically, I'm surprised it is being executed.

kmuehlbauer:
>     2. The shader compilation error is likely due to an old version of OpenGL running on the machines.
> 
>     3. The shared context error sounds like it is an example that shouldn't be run or is using the wrong backend. I'm 90% sure I explicitly skip running this example (in the `_runners.py` module maybe?). Basically, I'm surprised it is being executed.

Yes, these two need to be disabled in `_runners.py`. Not sure how to tackle the `XDG` warnings. I mean it's only a warning, why does it fail?

djhoese:
> Not sure how to tackle the XDG warnings. I mean it's only a warning, why does it fail?

We can set the variable it mentions to turn off the warning. We could even set it to the same path that it says in the warning. Not sure if it is the default or if vispy checks it specifically, but any stderr output is seen as a failure on our tests. It is useful in the cases that it is meant for and annoying in cases like this. Really this warning should be something more INFO-like but we can't control that. I vote for setting the environment variable in the workflow.

djhoese:
See shared_context shouldn't be run:

https://github.com/vispy/vispy/blob/55b2337fd45b725ca78010b73ca90ed317cb24e2/vispy/testing/_runners.py#L245-L250

djhoese:
Oh! Because it checks if we are running on travis!!!

kmuehlbauer:
I've added a check for `GITHUB_ACTIONS` to `_runners.py` and exported the `XDG` variable.

djhoese:
Success! So where are we at now:

1. This is all tests moved from travis to github actions, right?
2. What about the website generation (and upload)?
3. Should we move deployment to github actions based on a github release being made? Or stick with Azure (I prefer github actions but not sure when I would be able to do the work)?

kmuehlbauer:
> Success! So where are we at now:
> 
>     1. This is all tests moved from travis to github actions, right?

Not all, docstring tests are missing. I'll have a closer look tomorrow.

>     2. What about the website generation (and upload)?
> 
This is on the plate, will check tomorrow, too.

>     3. Should we move deployment to github actions based on a github release being made? Or stick with Azure (I prefer github actions but not sure when I would be able to do the work)?

This would be reasonable to do. I'll ping you when I need your assistance regarding secrets etc.

kmuehlbauer:
@djhoese @larsoner 
For website deployment I would suggest to create an organization secret (`VISPY_WEBSITE_TOKEN`) which should contain a valid access token to https://github.com/vispy/vispy.github.com.

Same would be needed for deployment to testpypi and pypi (`VISPY_TESTPYPI_TOKEN`, `VISPY_PYPI_TOKEN`).

kmuehlbauer:
@djhoese It seems that for GitHub Actions we need to build the wheels separately for the different os and upload the artifacts somewhere (need to look into this, how this works on GHA). Finally we would need to download the artifacts and deploy them to pypi. I would suggest to put this into another workflow which is just triggered by a release. WDYT?

kmuehlbauer:
For macos and windows which tests should we run (unit and examples)? I would put this in subsequent PullRequest.

djhoese:
RE: workflow only on releases, yes I think this is best.

RE: OSX/Windows, I think for appveyor we were running the bare minimum of unit tests. Looks like on travis OSX is running "standard" tests for full and minimal dependencies. I'm OK with this being a separate PR.

RE: The tokens, I can set those up later this weekend I think. I personally don't see a point of releases to test pypi. I saw someone contribute wheel building to pykdtree that seems to work fine: https://github.com/storpipfugl/pykdtree/blob/master/.github/workflows/deploy-wheels.yml

kmuehlbauer:
@djhoese Good, then I'll finish up the PR (rebasing, commenting, etc).

kmuehlbauer:
>  I suppose someone is probably working on that already
https://github.com/mamba-org/provision-with-micromamba

djhoese:
Oh nice. Switch to that?

kmuehlbauer:
Sure, will apply tomorrow with the other cleanups...

djhoese:
So I played with this micromamba thing on another package. Major limitation right now is that you can't specify the python version from the workflow/job so you have to have a separate environment file for each one. Not sure it is worth it for vispy. I ended up using it for my package because it is really simple so I only test it on python 3.9.

https://github.com/pytroll/donfig

kmuehlbauer:
> Major limitation right now is that you can't specify the python version from the workflow/job so you have to have a separate environment file for each one.

Yes, I was thinking the same. But we won't test on more than three Python versions I think, so this would be manageable.

kmuehlbauer:
OK, so here it what we have now:

- `OpenGL checks`
  This installs system OpenGL libraries and utils and runs `glxinfo` (currently runs on `ubuntu-latest`)
- `build website` 
  This installs necessary packages into system python (3.9) and builds the website. On push events it should upload the website. This runs  on `ubuntu-latest`.
- `lint and style checks`
  This installs necessary packages into system python and does:
    - check line endings
    - lint with flake8
    - test docstring parameters
  Runs on `ubuntu-latest`
- `vispy tests - linux` 
  This test matrix runs four jobs on `ubuntu-latest` all with `micromamba`-provisioned conda-forge environments (py3.7):
    - `DEPS`: min, `TESTS`: standard
    - `DEPS`: full, `TESTS`: standard
    - `DEPS`: full, `TESTS`: examples
    - `DEPS`: osmesa, `TESTS`: osmesa
 - `coveralls`
 This runs on `ubuntu-latest` and handles finalizing of corverage reports to coveralls.io for the parallel matrix tests. 

Notes:
- The test matrix is only running if the lint and style checks passed.
- Test coverage is recorded for `standard` and `osmesa` tests.
- The website uploads on push to master using a VISPY_WEBSITE_TOKEN secret which need to be defined in the vispy-org secrets.

djhoese:
Amazing!

Did you see any performance improvements switching to microconda?

No deployment to PyPI is currently done right? That's still all Azure (that's fine)?

djhoese:
Also could you update the badges on the README?

kmuehlbauer:
Re improvements:

The conda-environment creation is down to 40sec to 120sec (depending on the deps) from around 6 to 10 minutes (compared to latest travis-ci builds).

Re deplyoment:

will come in follow up PR as well es OSX and windows builds

Re badge:

sure, will update badges with another commit. Should the travis stuff be removed then?

djhoese:
> sure, will update badges with another commit. Should the travis stuff be removed then?

Yeah I think you can remove the travis configs too. Appveyor can be removed too.

kmuehlbauer:
> Yeah I think you can remove the travis configs too. Appveyor can be removed too.

We might need to have to unlink Appveyor in the repo somewhere. It's obviously not enough to remove the config.

Also, there are many mentions of travis/Appveyor in tests and examples. But I would treat those in a dedicated follow up PR.

kmuehlbauer:
@djhoese Feel free to merge if you are happy with the current status. I'll move on as outlined after merge.

djhoese:
I've deleted the appveyor project. Looks good to me. I'll work on a token after merging this.

Issue #1971, opened 2021-01-10
Add Support for PySide6


Tai-Zhou:
[PySide6](https://pypi.org/project/PySide6/) is released, maybe it can be added to the backend list.

djhoese:
From the small amount that I've read, my understanding was PySide5 -> Pyside6 is meant to be an easy transition. Any idea if this applies to the OpenGL interfaces?

I see no reason not to add it, but it probably won't be driven by me. My usual response to things like this...

Pull requests are welcome!

Kusefiru:
I would like to give it a try, if you are okay with it.

djhoese:
@Kusefiru Go for it. I'm not going to turn down help :grin:

Kusefiru:
Just a quick question, the name of the global variable `QT5_NEW_API` may not make much sense anymore since this also applies to Qt6.

Is it okay to rename it simply `QT_NEW_API` or is there a risk of breaking someone work ?

djhoese:
You should be ok renaming it, but double (triple?) check that the APIs are the same. Or create 2 new variables for a total of:

1. `QT5_NEW_API`
2. `QT6_NEW_API`
3. `QT5P_NEW_API` for "5 plus", meaning this API is in QT5 and 6. Or maybe `QT56_NEW_API` but that seems like it is easy to miss. Or `QT5_AND_6_NEW_API`. Or...just do `if QT5_NEW_API or QT6_NEW_API:` every time the current QT5 global is checked.

Kusefiru:
Okay, I'll use the last solution.

Kusefiru:
You may close this issue as well now ?

PR #1978, opened 2021-01-25
Add PySide6 backend

Kusefiru:
As discussed in #1971, I've added the support for PySide6.

There is no major changes compared to PySide2. Please tell me if I missed something.

I've been able to run the various examples and use the PySide6 backend along a PySide2 installation. Is there anything more I need to do to test the changes ?

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -279,7 +279,9 @@ class ApplicationBackend(BaseApplicationBackend):
 
     def _vispy_process_events(self):
         app = self._vispy_get_native_app()
-        app.flush()
+        # sendPostedEvents replaces flush which has been removed from Qt6.0+
+        # This should be compatible with Qt4.x and Qt5.x
+        app.sendPostedEvents()
         app.processEvents()
 
     def _vispy_run(self):
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -825,7 +825,8 @@ class CanvasBackendDesktop(QtBaseCanvasBackend, QGLWidget):
         if hasattr(self, 'devicePixelRatio'):
             # We take into account devicePixelRatio, which is non-unity on
             # e.g HiDPI displays.
-            ratio = self.devicePixelRatio()
+            # self.devicePixelRatio() is a float and should have been in Qt5 according to the documentation
+            ratio = int(self.devicePixelRatio())
             w = w * ratio
             h = h * ratio
         self._vispy_set_physical_size(w, h)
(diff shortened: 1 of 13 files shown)


djhoese:
@kmuehlbauer Should we try running unit tests against this? I suppose it would need to be available from conda-forge?

@Kusefiru Looks like there are some style/lint issues (see failing github action).

Tai-Zhou:
https://github.com/vispy/vispy/blob/8130a2229fcdf5e6a5f47190207d3fc948747d3a/vispy/app/backends/_pyside2.py#L17

Maybe `'PySide6'` should be added here?

kmuehlbauer:
@djhoese  On conda-forge `qt` is at version `5.12.9` and the pyside2 is at `5.13.2`.  I've no idea why the projects name changes every once in a while and what the differences are (`pyside`->`pyside2`->`pyside6`). If `pyside6` can work with the current conda-forge `qt` we might have the chance to install it alongside via `pip`.

Kusefiru:
@djhoese @Tai-Zhou I'll make the correction by the evening, thanks for your answers.

@kmuehlbauer PySide6 is based on the new Qt6. It was renamed that way to keep the same numbers between PySide and Qt and avoir confusing people later on, from what I understand.

Kusefiru:
I've submitted the corrections, all checks seem fine now !

vispy/app/backends/_qt.py
```diff
@@ -266,7 +279,9 @@ def _vispy_get_backend_name(self):
 
     def _vispy_process_events(self):
         app = self._vispy_get_native_app()
-        app.flush()
+        # sendPostedEvents replaces flush which has been removed from Qt6.0+
+        # This should be compatible with Qt4.x and Qt5.x
+        app.sendPostedEvents()
```
djhoese, line 284: The Qt4 docs are a little unclear about if flush is the exact same thing as sendPostedEvents (it seems like sendPostedEvents *only* handles events that were explicitly added with postEvent), but seeing as flush is completely deprecated in later APIs I approve of this. Thanks for catching this.

djhoese (approved):
So besides adding a pyside6 test environment I think this looks good. Thanks for all the work. I had one comment on the `sendPostedEvents` switch, but just commenting for comments sake.

Thanks again.

djhoese:
@kmuehlbauer If you have the time, I wouldn't mind a second review on this. If not that is fine too.

kmuehlbauer (approved):
LGTM

djhoese:
Alright great. Let's merge this and see what breaks :fearful:

PR #1980, opened 2021-02-04
Allow custom events (touch, pinch...) propagation

Carreau:
With this a custom canvas, with

    self._deliver_types.append('pinch')

And camera with for example

    def _viewbox_set(self, viewbox):
        super()._viewbox_set(viewbox)
        viewbox.events.add(pinch=Event)
        viewbox.events.pinch.connect(self.viewbox_pinch_event)

    def _viewbox_unset(self, viewbox):
        viewbox.events.pinch.disconnect(self.viewbox_pinch_event)
        super()._viewbox_unset(viewbox)

    def viewbox_pinch_event(self, event):
        if event.type == 'pinch':
            center = self._scene_transform.imap(event.pos)

Alternative is to reimplement most of the canvas process_mouse_event in
a _process_touch_event, or similar, while here we can simply in the
canvas:

   self.events.touch.connect(self._process_mouse_event)

----

I of course welcome suggestion of alternatives; as I'm still getting familiar with the code base. 

in particular this could be used to generalize :
```
if self._send_hover_events:
            deliver_types += ['mouse_move']
```
            

I would also understand why not; `touch` events do not always have the same structure as  mouse events, in particular they do not have `button` attribute. 

In napari this allowed me to implement panning and zooming using 2 finger gesture (scroll and pinch), without modifiers or click; which I find is a great improvement for user interface. I'm happy to also contribute (some of) those changes, directly to the pan/zoom camera is you think they are of interest. 

If course, 2 finger to pan is a "mouse_wheel" event, but touch/pinch will likely only trigger on compatible devices.

### DIFF
--- a/vispy/scene/canvas.py
+++ b/vispy/scene/canvas.py
@@ -125,6 +125,7 @@ class SceneCanvas(app.Canvas, Frozen):
         self._mouse_handler = None
         self.transforms = TransformSystem(canvas=self)
         self._bgcolor = Color(bgcolor).rgba
+        self._deliver_types = ['mouse_press', 'mouse_wheel']
         
         # Set to True to enable sending mouse events even when no button is
         # pressed. Disabled by default because it is very expensive. Also
@@ -334,7 +335,7 @@ class SceneCanvas(app.Canvas, Frozen):
 
     def _process_mouse_event(self, event):
         prof = Profiler()  # noqa
-        deliver_types = ['mouse_press', 'mouse_wheel']
+        deliver_types = self._deliver_types
         if self._send_hover_events:
             deliver_types += ['mouse_move']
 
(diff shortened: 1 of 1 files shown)


djhoese:
Hm this is very interesting. Thanks for starting the discussion and implementation. I'm not sure I completely follow all of your points, but let me bring up some of my own. Sorry if this was discussed in your other PR.

1. Does this require #1968 to be merged?
2. It would be great if this could be done in a way that applications currently using mouse_wheel (or other related mouse events) could automatically work on devices that can produce the events. But I guess this isn't the scope of this PR and is more a general "how does vispy handle gesture events from Qt", right?
3. Whatever way this ends up being implemented, I don't think this needs to be backwards incompatible and if we can avoid adding any additional `_` (private) attributes that are meant to be accessed outside the class that would be good (avoid using more `_send_hover_events`-like properties).
4. I'm not sure I followed your suggestions, but would there be any issue with always adding `mouse_move` and `pinch` to `deliver_types` when `_send_hover_events` is True? With your `lambda` in getattr would this end up being a no-op for pinch gestures if not implemented?

I honestly am not sure I follow all of the various options and use cases here. Whatever ends up being done it needs to be documented. The `master` branch here as a slightly better structure for the sphinx docs so we may be able to add a page here for details mouse event handling. Or one of the classes (Canvas/SceneCanvas?, MouseEvent?) could have this more fully explained. The less the user needs to know about the internals of the mouse event handling in the SceneCanvas, the better.

kevinjohncutler:
@Carreau how can I test your gesture implementation in napari? I tried installing your versions of vispy and napari, and while your vispy branch works fine with the main napari branch, your version of napari seems to be incompatible with your version of vispy.

PR #1983, opened 2021-02-10
Fixed multiple code style issues.

Aaru143:
- Fixed comment style in code.
- Removed unused variables.
- Removed semi-colons from python code.
- Fixed docstrings.


setup.py
```diff
@@ -122,23 +122,23 @@ def finalize_options(self):
...
             check_call([npmName, '--version'])
             return True
-        except:
+        except BaseException:
             return False
 
     def should_run_npm_install(self):
-        package_json = os.path.join(node_root, 'package.json')
-        node_modules_exists = os.path.exists(self.node_modules)
+        # package_json = os.path.join(node_root, 'package.json')
```
kmuehlbauer, line 140: @djhoese Did you had something special in mind with this? If not, we could strip the whole function and use `self.has_npm()` in the run-function below.

kmuehlbauer:
@Aaru143 Thanks for tackling this. There is only one minor issue which needs input from @djhoese.

@djhoese If we are on it, should we make use of some style tool, like black?

setup.py
djhoese: These lines were copied from the jupyter widget cookiecutter template. It is a bug in their template that I think has since been fixed. I think we need to update our usage anyway to be compatible with JLab 3.0+. I went over this change with @Aaru143 and I think it makes sense since they aren't being used.

djhoese:
@kmuehlbauer I think we had talked about black in the past, but decided against it because there are a lot of vispy classes/functions/methods with a lot of arguments. Black would make these definitions really long (one argument per line iirc). If we want to re-open that discussion I'm ok with that, but I think I'm leaning towards not using black. I could also be convinced pretty easily given the maintenance burden.

kmuehlbauer:
@djhoese Now that you say that, I remember. I really should take notes, since brain is not reliable it seems. Sorry for the noise.

kmuehlbauer (approved):

kmuehlbauer:
@Aaru143 I think this is ready to go, could you please rebase on or merge latest master to resolve the conflicts?

djhoese:
I think she had some more updates, but has been busy with midterms and stuff. I'll see if I can work with her this week to get this finalized. Even if not all things are fixed at least we can merge it so future conflicts aren't so difficult to resolve.

I know there are at least a couple code pieces that don't have docstrings. Part of me wants to ignore that rule as some of them are pretty redundant (ex. a property for `clims` with a docstring "Get color limits." isn't super useful or necessary). However, ignoring this rule opens the doors for people to leave off docstrings for code that really need it.

kmuehlbauer:
I fully agree. :+1:

djhoese:
Thanks @Aaru143 for making the extra updates. I've just pushed one more commit where I removed the extra ignored issue types you had added to `setup.cfg` and I ran `autopep8 -i --global-config setup.cfg --aggressive -r vispy` just to see what I would get. For me, `autopep8` fixed a couple more issues (mostly removing extra whitespace from blank lines).

I think I'm happy with this PR as-is as it seems to fix all code-specific style issues and the only remaining ones are docstring related. I think we can tackle that in the future.

And as I type this I notice it is mad at some of the examples. I'll fix those quick and then merge this.

PR #1999, opened 2021-02-27
ImageVisual support for complex-valued data.

tlambert03:
This is just a draft for now, for the sake of discussion.  It _includes_ #1920, which should be merged first.  (see [comparison of this PR against #1920](https://github.com/djhoese/vispy/compare/feature-image-texture-format...tlambert-forks:complex)).  The basic idea is to convert complex data into a 2-channel internal data array, with a modified `_build_color_transform` that knows how to convert those two channels to intensity (using a `complex_mode` such as "magnitude" or "phase") prior to standard colormap application.

Outstanding issues to discuss:
- `auto` clims are tricky here, since the `complex_mode` will dramatically change the "intensity" range (with "phase" always being somewhere from -π to π and "magnitude" being all over the map).  We can certainly apply the complex rendering mode in the CPU to glean the clims...
- _something_ needs to remember that the underlying data is complex, after the internal data has been changed to a real 2-channel array (assuming we stick with that approach).  In this current implementation, the `CPU/GPUScaledTexture2D` textures are "blind" to that fact, while the `ImageVisual` itself has a `_data_is_complex` and `complex_mode`.  But the texture subclasses probably also need to know this for the sake of clim calculation, etc...


tlambert03:
This has been fixed and updated... would be nice to have this included in 0.7 if you have a moment to take a look @djhoese 

It's minimal at the moment, which perhaps is a good place to leave it.  Could try to do more magic with clims, but I think best to leave that to the user for now

djhoese:
Lucky for you I just merged my #1912 which is now causing conflicts here.

tlambert03:
hah, no worries :)  will fix shortly

djhoese:
It seems the wheel building is failing as your `complex256` usage is only available on some platforms.

https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.clongdouble

tlambert03:
ah whoops!  removed

vispy/visuals/image.py
```diff
@@ -208,12 +232,20 @@ class ImageVisual(Visual):
...
+                 interpolation='nearest', texture_format=None, complex_mode="magnitude",
+                 **kwargs):
         """Initialize image properties, texture storage, and interpolation methods."""
         self._data = None
+        self._data_is_complex = False
+        if complex_mode not in self.COMPLEX_MODES:
+            raise ValueError(
+                "complex_mode must be one of %s" % ", ".join(self.COMPLEX_MODES)
+            )
+        self._complex_mode = complex_mode
```
djhoese, line 248: Can this be replaced with `self.complex_mode = complex_mode` (maybe a `self._complex_mode = None` first. That way the error message isn't repeated.

vispy/visuals/image.py
```diff
@@ -310,6 +342,16 @@ def set_data(self, image):
 
         """
         data = np.asarray(image)
+        if np.iscomplexobj(data):
+            self._data_is_complex = True
+            # even though we're only ever using the first and/or second channels,
+            # without the third empty "channel", it doesn't render correctly.
+            # could be a bug in rg32f rendering?
```
djhoese, line 349: Is this still relevant? I see you are using `r32f` mode in your GPU texture? Should it be `rg32f`?

djhoese (changes requested):
A couple inline requests, but I have an even bigger, perhaps annoying, suggestion: Does this functionality warrant a new `ComplexImageVisual` class or something like that that is a subclass of ImageVisual? My concerns come from the complication of completely documenting this new functionality in the docstring and how much special casing might be needed.

1. Should `texture_format` be allowed with this since in almost all cases of it being not `'auto'` would produce an incorrect result, right?
2. Is the CPU-based scaling allowed with this?
3. How does a user specify a meaningful clim?

Overall I'm just worried about every keyword argument for the ImageVisual having a "...except when complex data is provided, in which case none of this applies in the same way". In a subclass of the ImageVisual you could control or disallow a lot of these things. You could even go as far as using two r32f textures...hm maybe not that would need a few more changes in the shader.

Lastly, have any screenshots for what can be expected for the output? Just curious.

djhoese:
Thinking about it more...I have plans for implementing an RGB ImageVisual where each channel is a separate texture and input (I think that's how I wanted to implement it). We could try to do this in some generic way where a complex multi-array image visual could specify two bands (two textures) and a custom color transform.

Another idea, is there a way that you can cheat with the creation of the new numpy array (your vstack you're doing) by doing a `my_arr.view(dtype=[('r', np.float32), ('g', np.float32)])`. Hm, but that would require updating the texture classes to be OK with only 2 bands to pass to the texture. That shouldn't be too bad...I'd hope. At least not for the GPU texture.

vispy/visuals/image.py
tlambert03: my understanding of the texture was that even if you're using `rg32f`, the key should still be `r32f` (this is based  on your [comment here](https://github.com/vispy/vispy/blob/df09b3079d71b7d1942481a051b900f145c4cfee/vispy/visuals/_scalable_textures.py#L321) and the num_channels processing [here](https://github.com/vispy/vispy/blob/df09b3079d71b7d1942481a051b900f145c4cfee/vispy/visuals/_scalable_textures.py#L354).  

The very strange thing here is that, if I try to convert the complex data to `np.stack([data.real, data.imag], axis=-1)`, _without_ that extra "empty" third channel (i.e. a true `rg32f` texture), then for some reason, when I sample the data, everything is just the first `r` channel, even if I specifically sample the `g` channel.  I dug into this one for a _long_ time, trying to figure out where that second channel information was being lost, but never found it.  Realized eventually that just making it 3 channel worked.

maybe you know?  I kinda wonder whether `rg32f` actually works at all?  or whether some 1-or-3 channel assumption is being made lower in the code?

tlambert03:
> We could try to do this in some generic way where a complex multi-array image visual could specify two bands (two textures) and a custom color transform.

That would be fine with me too.  That's essentially all this is here, but just a single texture and a custom color transform.  I'm really hoping to use complex stuff in napari though, and would love to not have to keep vendoring the image visual.  Do you think the RGB stuff might happen before 0.7?  or might that be a ways down the road?  (that said, I definitely understand not wanting to add a new public feature that will just be moved later...).  Let me know what you think.

> Hm, but that would require updating the texture classes to be OK with only 2 bands to pass to the texture.

ah... maybe this is related to why my `vstack` wasn't working with only two channels?

vispy/visuals/image.py
djhoese: :thinking: Good point. You're right...at least that makes sense from the way it is described and implemented. I don't see anything obviously wrong in my code that necessarily assumes something about the data being RGB/A that would effect the storage, the `.r` and the `.g` in the shader (only the clims).

tlambert03:
> Should texture_format be allowed with this since in almost all cases of it being not 'auto' would produce an incorrect result, right?

I agree.  would be more cautious to make `texture_format` an error when providing complex data.

> Overall I'm just worried about every keyword argument for the ImageVisual having a "...except when complex data is provided,

.... buuut this also makes a lot of sense.
I can definitely pull all this out into a new subclass if you'd like.  I agree it's the most "not like the others".  perhaps it's just best to throw an error on complex data and suggest using the other Visual?

> Is the CPU-based scaling allowed with this?

tbh, I need to play with that

> How does a user specify a meaningful clim?

Very manually :) the challenge is of course that the rendering mode will affect the clim.  and technically, if you wanted to do an "auto" clim, you'd need to first do the full complex rendering on the CPU to get the min/max.  And that will then change if you change the mode from (e.g.) magnitude to phase.  As I said, I've avoided doing any of that magic here.  Personally, I think it's ok to expect that someone visualizing complex images set their own clims.

>  have any screenshots for what can be expected for the output? Just curious.

sure, just a minute

djhoese:
This is all stuff I didn't necessarily test or think about while doing the CPU/GPU texture subclasses. As far as the RGB visual, yes this will likely be further down the road since it is not widely requested and is only needed by one of my projects (that has very little funding and very little of my time goes towards it). I think the best approach would be for you to have a new custom visual for this complex "stuff" (leave it in `image.py` if you'd like but make sure to add it to the `vispy/scene/visuals.py`). My thinking is that we could always change the internals of how it is implemented if htat works in the future. But now that I think about it, this shouldn't be needed. Maybe I'll pull this PR down locally and play with it and see if I can figure out what is going wrong `.r` and `.g`-wise.

tlambert03:
>  I think the best approach would be for you to have a new custom visual for this complex "stuff" (leave it in image.py if you'd like but make sure to add it to the vispy/scene/visuals.py). My thinking is that we could always change the internals of how it is implemented if htat works in the future.

yeah, that sounds good to me!  will refactor

vispy/visuals/image.py
tlambert03: yeah, me neither! :joy:  your code seems very agnostic... which is why I was so surprised.  but, if I (for instance) manually go in to the shader code and show just the `.r` channel or just the `.g` channel, they were always the same (until I uploaded 3-channel data to the texture).  But all of the data on the CPU was correct (visualized using matplotlib at various places in the `set_data` processing chain)

djhoese:
I guess the only benefit of doing a multi-band ImageVisual would be that you should theoretically be able to set clims for each band. That should, again theoretically, make setting clims for a complex visual based on that much easier to understand.

tlambert03:
> That should, again theoretically, make setting clims for a complex visual based on that much easier to understand.

sort of... but many of the rendering modes combine the two channels.  For instance, the `phase` rendering mode will always have a min/max of `π/π` ... and knowledge of the real/imag input bands won't affect that.  so, it's tough :/

djhoese:
This may be the reason you don't see what you expect in `.g`:

https://github.com/vispy/vispy/blob/df09b3079d71b7d1942481a051b900f145c4cfee/vispy/gloo/texture.py#L69

It may be showing up in `.a` (the alpha band).

tlambert03:
yaaas.  that looks like it!  thank you!  i'll dig there a bit

djhoese:
So adding this to the texture creation seems to work:

```
format='rg' if np.iscomplexobj(data) else None,
```

or I should say it **wants** to work. There are two issues:

1. I want to use the `_data_is_complex` but that isn't set until later.
2. The other issue is that the `data` being provided to the texture is the original 2D array so the texture thinks it only has 1 band so then it fails because `format='rg'` doesn't match the number of bands.

And I was right about the alpha limitations. If you use alpha you get a different result than the original/expected result. I tested this by changing your example to use `complex_mode='phase'`.

tlambert03:
those are fantastic tips. thanks a lot.  I can likely clean things up a bit from there.  It's working nicely as a separate class, so we can at least keep a lot of this odd logic on its own.

djhoese:
And I don't think I've given you a clear answer regarding 0.7: If you can get this ready for 0.7 (let's say next week some time?) then absolutely no problem with it being added. It will need tests though. I also am trying to promise you (and other napari folks...and myself) that I will release feature/minor versions of vispy more quickly. I just get stuck in this trap of "oh we've been talking about X and Y being added, we can't release until those things are done".

tlambert03:
good by me!  on all points.  thanks

tlambert03:
here's some example images:

```python

...
camera = skimage.data.camera().astype("float32")
img_data = fftshift(fftn(camera))

# complex mag
image = scene.visuals.ComplexImage(
    img_data,
    clim=(3000, 5000000),
    gamma=0.2,
    parent=view.scene,
    complex_mode="magnitude",
)

```
Untitled

```python
# complex phase
image = scene.visuals.ComplexImage(
    img_data,
    clim=(-np.pi, np.pi),
    parent=view.scene,
    complex_mode="phase",
)
```

Untitled 2

vispy/visuals/image_complex.py
```diff
@@ -0,0 +1,96 @@
...
+            raise ValueError(
+                "complex_mode must be one of %s" % ", ".join(self.COMPLEX_MODES)
+            )
+        if self._complex_mode != value:
+            self._complex_mode = value
+            self._need_colortransform_update = True
+            self.update()
+
+    def _build_color_transform(self):
+        if self.complex_mode:
```
djhoese, line 65: It is impossible for this to be False right? So this could be removed?

vispy/visuals/image_complex.py
```diff
@@ -0,0 +1,96 @@
...
+                fgamma,
+                Function(self.cmap.glsl_map),
+            ]
+            fun = FunctionChain(None, chain)
+            fclim["clim"] = self._texture.clim_normalized
+            fgamma["gamma"] = self.gamma
+            return fun
+        return super()._build_color_transform()
+
+    @ImageVisual.clim.setter
```
djhoese, line 80: Does this actually work? This doesn't overwrite the setter for the clim property in the ImageVisual... :confused:

vispy/visuals/image_complex.py
```diff
@@ -0,0 +1,96 @@
...
+    "magnitude": "float cplx2float(vec4 data) { return length(data); }",
+    "phase": "float cplx2float(vec4 data) { return atan(data.g, data.r); }",
+}
+# aliases
+COMPLEX_TRANSFORMS["angle"] = COMPLEX_TRANSFORMS["phase"]
+COMPLEX_TRANSFORMS["abs"] = COMPLEX_TRANSFORMS["magnitude"]
+COMPLEX_TRANSFORMS["imag"] = COMPLEX_TRANSFORMS["imaginary"]
+
+
+class ComplexImageVisual(ImageVisual):
```
djhoese, line 20: Docstring please. You could reference the ImageVisual directly at some point:

```
:class:`~vispy.visuals.ImageVisual`
```

vispy/visuals/_scalable_textures.py
```diff
@@ -324,6 +324,8 @@ class GPUScaledTextureMixin(_ScaledTextureMixin):
         np.float64: 'r32f',
         np.uint8: 'r8',  # uint8 normalized
         np.uint16: 'r16',  # uint16 normalized
+        np.complex64: 'r32f',
+        np.complex128: 'r32f',
```
djhoese, line 328: Based on some feedback from @almarklein on my PR that added this functionality, I'm tempted to say these should be moved to the ComplexImageVisual. You can pass `r32f` directly as `texture_format` to `ImageVisual.__init__` and it should "just work". Having this in these base classes should include all the custom functionality for converting the numpy arrays to the proper storage solution. And if we're doing that then we might as well have a separate implementation of the Mixin class specifically for complex data...which is just complicated and is maybe easier to just put stuff in the ComplexImageVisual.

However, if you are planning on doing a ComplexVolumeVisual then maybe a separate base class for these mixins does make sense. But maybe that can wait until that PR.

djhoese (changes requested):
Looks pretty good as a separate module. Thanks for doing that. I have a couple requests. Otherwise I just want to express the general odd feeling for "ComplexImageVisual" as a name...but I can't think of something better at the moment...

"ComplexNumberImageVisual" lol

vispy/visuals/image_complex.py
tlambert03: well, with that line... the `ComplexImageVisual` should technically be possible to present both complex and non-complex data.  (i.e. it's effectively the same as the more class that we had before the separation).  If someone were to call `set_data` with non-complex data, then `self._data_is_complex` would go back to False, and then the `complex_mode` property would return `False`... I _think_ it might mean that napari could simply use the complex version and support all uses cases (I haven't thoroughly tested that, and will write some tests to see... but that was the motivation)

vispy/visuals/_scalable_textures.py
tlambert03: sure, I'll give that a shot.  volume visual probably won't be very soon.

vispy/visuals/image_complex.py
tlambert03: docstrings and tests coming still...

vispy/visuals/image_complex.py
```diff
@@ -0,0 +1,96 @@
...
+        super().__init__(data=data, **kwargs)
+
+    def set_data(self, image):
+        data = np.asarray(image)
+        self._data_is_complex = np.iscomplexobj(data)
+        if self._data_is_complex:
+            #  Turn the texture into an rg32f texture
+            # where r = 'real' and g = 'imag'
+            self._texture._format = "rg"
+            data = np.stack([data.real, data.imag], axis=-1)
```
almarklein, line 44: I think we can do this without data-copies, because the complex and imaginary part are aligned similar to the color planes in an RGB image. I think this should work: 
```py
data = data.astype(np.complex64, copy=False)  # turn complex128 into complex64 if needed
data = data.view(dtype=np.float32).reshape((data.shape + (2, ))) 
```

vispy/visuals/image_complex.py
almarklein: It looks like this is the way, see e.g. https://stackoverflow.com/a/7020271/2271927


<details>
<summary>I also checked it, just in case</summary>

    class Foo:        
        def __init__(self):
            self.v = 0
        
        @property
        def val(self):
            return self.v
        
        @val.setter
        def val(self, v):
            self.v = v
    
    
    class Bar(Foo):        
        @Foo.val.setter
        def val(self, v):
            self.v = v + 1
        
    
    f = Foo()
    print(f.v)
    f.val = 42
    print(f.v)
    
    b = Bar()
    print(b.v)
    b.val = 42
    print(b.v)

Yields:

     >>> (executing cell "" (line 22 of "tmp.py"))
    0
    42
    0
    43
</details>

vispy/visuals/image_complex.py
```diff
@@ -0,0 +1,96 @@
...
+from .shaders import Function, FunctionChain
+
+# In a complex Image, the texture will be rg32f, where:
+# data.r contains the real component
+# data.g contains the imaginary component
+COMPLEX_TRANSFORMS = {
+    "real": "float cplx2float(vec4 data) { return data.r; }",
+    "imaginary": "float cplx2float(vec4 data) { return data.g; }",
+    "magnitude": "float cplx2float(vec4 data) { return length(data); }",
+    "phase": "float cplx2float(vec4 data) { return atan(data.g, data.r); }",
```
almarklein, line 12: Maybe also an RG-mode, where the real and imag plane are shown as two colors?

almarklein (changes requested):
A few minor comments.

Users can always create an regular `ImageVisual` with the real/imag/mag/fase of their data. So the main benefit seems to be the ability to easily switch between the modes. Can we maybe emphasize that by e.g. making the example go through the different modes as you click on the image?

djhoese:
It seems like there are a lot of suggestions here so I'm going to say we move this to the 0.8 release.

tlambert03:
that's fine, I've been swamped this week.  Thanks for your help

djhoese:
@tlambert03 What's your current availability on this? Should we work towards merging this as is (I think there are still some suggestions that need to be addressed) or should we work on getting my #2156 PR working and merged and make this complex image visual based on that? What's your preference?

tlambert03:
don't have a strong preferences. you can tell me. If part of your motivation for #2156 is to support multi-texture images like this, then I'm happy to try to refactor this to use that pattern.   Definitely still need to incorporate the suggestions here either way.

djhoese:
How about you do the suggestions that are here already, we'll re-review and make sure to compare what would need to change (especially public interface-wise) if it was based on the multichannel visual, then merge this. Then we can merge the MultiChannelImageVisual. Then we can make a new PR rewriting this to use the MultiChannel. *If* switching to multichannel for this complex visual would require changes to the public interface then maybe the refactor should be done here. If switching to multichannel ends up being like two changed lines and removal of duplicated code then maybe this PR can take that on.

djhoese:
I think I'm going to move this PR and my multichannel one to the 0.9 milestone. Hoping to make a 0.8.0 release this week with the new volume visual changes.

tlambert03:
sounds good, thanks for checking in! :)

djhoese:
I don't have a lot of time right now to make sure, but I think #2156 is ready. If you could review it @tlambert03 that would be great. I also understand if you don't want to do that until you've attempt to port this PR's features to use it.

I am OK merging this PR if/when you consider it done so napari devs can play with it and we can have a baseline for it working as-is and then in a near-future PR port it to use my multi-channel stuff. This would mean my PR wouldn't get merged until your new PR proves that it works as we expect for subclassing as (at least to me) that is a major portion of its purpose.

roflmaostc:
Sorry for bumping again, but is there some progress 😅

tlambert03:
i'm so sorry @roflmaostc, I let it get buried again under all the other stuff going on 😬.  thank you for your gentle nudge.  I really do want to have this in as well, so i'll try to get something in this week.

djhoese:
@tlambert03 I don't remember where you're located, but VisPy is doing a sprint at the PyCascades conference this weekend. Let me know if there is anything you want me to look at. Otherwise, I have other ideas for the sprint.

vispy/visuals/_scalable_textures.py
tlambert03: Unfortunately, it's not as easy as just passing "r32f" or "rg32f" to the `ImageVisual.__init__`.   This is what I get if I change "auto" to "r32f":
```python
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/examples/basics/scene/complex_image.py", line 37, in <module>
    image = scene.visuals.ComplexImage(img_data, parent=view.scene, complex_mode="phase")
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/scene/visuals.py", line 131, in __init__
    subclass.__init__(self, *args, **kwargs)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/visuals/image_complex.py", line 48, in __init__
    super().__init__(data=data, **kwargs)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/visuals/image.py", line 291, in __init__
    self.set_data(data)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/visuals/image_complex.py", line 62, in set_data
    return super().set_data(data)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/visuals/image.py", line 344, in set_data
    self._texture.check_data_format(data)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/visuals/_scalable_textures.py", line 429, in check_data_format
    raise ValueError("Data being set would cause a format change "
ValueError: Data being set would cause a format change in the texture. This is only allowed when 'texture_format' is set to 'auto'.

```

and this is what I get i get if I use 'rg32f' instead:

```python
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/examples/basics/scene/complex_image.py", line 37, in <module>
    image = scene.visuals.ComplexImage(img_data, parent=view.scene, complex_mode="phase")
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/scene/visuals.py", line 131, in __init__
    subclass.__init__(self, *args, **kwargs)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/visuals/image_complex.py", line 48, in __init__
    super().__init__(data=data, **kwargs)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/visuals/image.py", line 266, in __init__
    self._texture = self._init_texture(data, texture_format)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/visuals/image.py", line 321, in _init_texture
    tex = GPUScaledTexture2D(
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/visuals/_scalable_textures.py", line 73, in __init__
    super().__init__(data, **texture_kwargs)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/gloo/texture.py", line 512, in __init__
    BaseTexture.__init__(self, data, format, resizable, interpolation,
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/gloo/texture.py", line 133, in __init__
    self._resize(data.shape, format, internalformat)
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/gloo/texture.py", line 300, in _resize
    internalformat = self._check_internalformat_change(internalformat, shape[-1])
  File "/Users/talley/Dropbox (HMS)/Python/forks/vispy/vispy/gloo/texture.py", line 288, in _check_internalformat_change
    raise ValueError('Internalformat does not match with given shape.')
ValueError: Internalformat does not match with given shape.
```

I tried to find a way to remove these two keys from this dict here, and put all the logic in the ComplexImageVisual, but couldn't figure it out without getting down to these `_scalable_texture` methods.  Let me know if this is a blocker.

tlambert03:
Hey @djhoese, got this working again, with very minor updates from merging main.  If you're ok with it, I think my preference would be to go with this roughly as is for now.  In terms of API changes, it's pretty much just opt in (i.e., one needs to specifically select `ComplexImageVisual`, otherwise nothing is affected), and if we were to eventually allow `ImageVisual` itself to work with complex dtypes, we could easily just point this to ImageVisual and add a deprecation.

The code itself is pretty minimal.  The only thing I couldn't figure out how to do was to avoid adding the two new keys to `GPUScaledTextureMixin._texture_dtype_format` (lemme know if you have a thought there).

If you're game, I'll add a couple basic tests and we can get it in to play with

tlambert03:
Tests added.  Ready for review

djhoese:
Thanks @tlambert03. I will get to this as soon as I can. Having trouble keeping up with emails this week. :fearful:

tlambert03:
No worries/hurry.  Took me months to get to it! 😅

vispy/visuals/image_complex.py
tlambert03: I do like this, but would prefer a new PR.  It kinda requires dual clims (and ideally dual cmaps) for a proper implementation.

vispy/visuals/image_complex.py
tlambert03: Added a test to confirm that complexImage can also be used for non complex data

examples/basics/scene/complex_image.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# vispy: gallery 30
+# -----------------------------------------------------------------------------
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
+Simple use of SceneCanvas to display an Image.
+"""
```
djhoese, line 9: Could you move this to `examples/scene/` so it shows up in the gallery? You'd also have to format the docstring a little differently. For example:

https://github.com/vispy/vispy/blob/49b0c499238750d95806ab75a4706e2eb5c2a84a/examples/scene/colored_line.py#L1-L11

It is basically RST so a title and then any description you want below it.

djhoese: Also didn't this used to have key bindings to cycling through some of the settings?

vispy/visuals/_scalable_textures.py
djhoese: I just had a realization/idea about this. Are these errors regarding shape changes and stuff all during `__init__` when no initial data is provided to the parent `ImageVisual`? Would it be possible to pass some representative data along with forcing the internalformat to `r32f`/`rg32f`?

vispy/visuals/image_complex.py
djhoese: Thanks.

djhoese (changes requested):
I have a small request on the example, one large request on the example (key bindings), and one new idea for the internalformat stuff. Thanks for working on this again and sorry about taking so long to get to this. It has been repeatedly on the end of my TODO list for weeks with deadlines beating it out of the running.

djhoese:
@tlambert03 I thought this would work, but it produces a different image when running the original example. Thoughts?

```diff
Index: vispy/visuals/image_complex.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/vispy/visuals/image_complex.py b/vispy/visuals/image_complex.py
--- a/vispy/visuals/image_complex.py	(revision 39c59e3bf7bf56c4c342a0761f10c748dc3cc4ef)
+++ b/vispy/visuals/image_complex.py	(date 1649901934590)
@@ -50,7 +50,9 @@
         if kwargs.get("clim", "auto") == "auto" and self._data_is_complex:
             kwargs["clim"] = self._calc_complex_clim(data)
 
-        kwargs["texture_format"] = "auto"
+        kwargs["texture_format"] = "r32f" if self._data_is_complex else "r32f"
+        if self._data_is_complex:
+            data = self._convert_complex_to_float_view(data)
         super().__init__(data=data, **kwargs)
 
     def set_data(self, image):
@@ -60,13 +62,18 @@
             #  Turn the texture into an rg32f texture
             # where r = 'real' and g = 'imag'
             self._texture._format = "rg"
-            # turn complex128 into complex64 if needed
-            data = data.astype(np.complex64, copy=False)
-            data = data.view(dtype=np.float32).reshape((data.shape + (2, )))
+            data = self._convert_complex_to_float_view(data)
         else:
             self._texture._format = None
         return super().set_data(data)
 
+    @staticmethod
+    def _convert_complex_to_float_view(complex_arr):
+        # turn complex128 into complex64 if needed
+        complex64_arr = complex_arr.astype(np.complex64, copy=False)
+        float_view_arr = complex64_arr.view(dtype=np.float32).reshape((complex64_arr.shape + (2, )))
+        return float_view_arr
+
     @property
     def complex_mode(self):
         return self._data_is_complex and self._complex_mode
Index: vispy/visuals/_scalable_textures.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/vispy/visuals/_scalable_textures.py b/vispy/visuals/_scalable_textures.py
--- a/vispy/visuals/_scalable_textures.py	(revision 39c59e3bf7bf56c4c342a0761f10c748dc3cc4ef)
+++ b/vispy/visuals/_scalable_textures.py	(date 1649901322672)
@@ -366,8 +366,8 @@
         np.float64: 'r32f',
         np.uint8: 'r8',  # uint8 normalized
         np.uint16: 'r16',  # uint16 normalized
-        np.complex64: 'r32f',
-        np.complex128: 'r32f',
+        # np.complex64: 'r32f',
+        # np.complex128: 'r32f',
         # np.int8: 'r8',  # not supported, there are no signed-integer norm formats
         # np.int16: 'r16',
         # np.uint32: 'r32ui',  # not supported, no normal formats for 32bit ints
```

Produces:

image

Compared to without these changes:

image

The main complication I think is the `_init_texture` method call in the `ImageVisual.__init__`. It is the only other portion of the code that uses `data` and it is setting up the texture. The created texture needs to know it will have N numbers of bands. In your current version that would mean 1 band (but complex data) and then later changing it to "rg" which I think it the not allowed part.

djhoese:
Ah I figured it out. The data was being interpreted as a normal float32 array inside `set_data` being called from `__init__`. So...how do I avoid a possible data copy and avoid ugly code...let's see...

djhoese:
Ok I got it all figured out and have pushed a commit to your branch. The last thing I don't like about what we've had to do here is the `self._texture._format = "rg"` when `set_data` is called with complex data. This isn't either of our fault, but rather the limited interface of the base Texture classes. They have no formal way of specifying the `format` of the texture so for these 2-band data arrays it assumes `luminance_alpha` data. This should probably be added, but in a future PR. It is a pretty big change.

I think the only other thing I'd like is the update to the example title and some key bindings if it isn't too much work.

CC @brisvag @alisterburt this is the main PR I'm waiting for before a release.

djhoese (approved):

djhoese:
Ok, the example has been updated. The requests by @almarklein and me to have a key event to cycle through the different complex modes. I'd prefer if @tlambert03 can give the "OK" on this and then I'll merge it and make a release.

tlambert03 (commented):
Fantastic.  Thanks for improving the example and figuring out how to cleanup the leak into `_scalable_textures`.  Pulled and tried, this is good to go by me!

djhoese:
For the record here is what the gallery page looks like:

image

And the actual gallery page when hovering over the example:

image

tlambert03:
I see, that's your real goal here 😂 ... yes it's not in napari yet.  It won't be hard to integrate, however there may be some hesitation to add it before we also implement it for volumes (#2314) lest you get an error every time you hit the 3D button

roflmaostc:
Oh yeah, sorry for that confusion :rofl:  

But thanks for the great work so far!

PR #2002, opened 2021-03-01
Optimize SurfacePlotVisual when only color is updated

### LINKED ISSUE #2001: SurfacePlot color change slower with set_data versus mesh_data.set_vertex_colors
Sample code below: a SurfacePlot whose color is updated when "O" is pressed using `mesh_data.set_vertex_colors`.

Problem: the `mesh_data.set_vertex_colors` works if called before adding the surface to the view, but not in later calls.

Temporary fix (thanks @djhoese ): add `surf.mesh_data_changed()` after changing the color.

PS: using `surf.set_data(colors=c)` instead of `surf.mesh_data.set_vertex_colors(c); surf.mesh_data_changed()` also works, but for some reason is much slower

```
import vispy, vispy.scene
import numpy as np

shape =  (200, 200)
Z = np.random.random(shape)*5
B = np.random.random(shape)

##


canvas = vispy.scene.SceneCanvas(keys='interactive', show=True)

surf = vispy.scene.visuals.SurfacePlot(z=Z, color=(1, 1, 1, 1))

# Add color
def setColor(C):
	C = np.array(C)
	c = vispy.color.get_colormap("hsl").map(C/np.max(C)).reshape(C.shape + (-1,))
	c = c.flatten().tolist()
	c=list(map(lambda x,y,z,w:(x,y,z,w), c[0::4],c[1::4],c[2::4],c[3::4]))
	surf.mesh_data.set_vertex_colors(c)
setColor(B)

# cycle colors on key press
col_cycle_i = 0
def cycle_colors(event):
	key = event.key.name
	if key != "O": return
	global col_cycle_i
	col_cycle_i = (col_cycle_i+1)%2
	print('Setting new coli', col_cycle_i)
	if col_cycle_i == 0:
		setColor(B)
	elif col_cycle_i == 1:
		setColor(Z)
		
canvas.events.key_press.connect(cycle_colors)
view.add(surf)
axis = vispy.scene.visuals.XYZAxis(parent=view.scene)
view.camera = 'fly'
```

djhoese:
Closes #2001 

See #2001 for details. The SurfacePlotVisual was updating the vertices of the MeshData every time `surf_plot.set_data` was called even if the vertices weren't being updated. This triggered the MeshData to reset and recompute all normals which was adding a very noticeable delay when changing only the color. This PR refactors the SurfacePlotVisual `set_data` method to distinguish between updating faces, vertices, or color and only updating that component when needed.

CC @hippalectryon-0


djhoese:
The tests on this were failing because of issues that were fixed in #2004. This has been merged now and the tests should pass. @kmuehlbauer have time to do a proper review and merge?

kmuehlbauer:
@djhoese I'll have a look into this tomorrow.

vispy/visuals/surface_plot.py
```diff
@@ -94,64 +77,106 @@ def set_data(self, x=None, y=None, z=None, colors=None):
...
             self.__vertices[..., 2] = self._z
-            update_mesh = True
+            update_vertices = True
+        return update_faces, update_vertices
 
-        if colors is not None:
-            self.__meshdata.set_vertex_colors(colors)
-            update_mesh = True
+    def _prepare_mesh_colors(self, colors):
+        if colors is None:
```
kmuehlbauer, line 122: @djhoese We could use `if colors is not None:` but this is just some style thing.

```python
if colors is not None:
    colors = np.asarray(colors)
    if colors.ndim == 3:
        # convert (width, height, 4) to (num_verts, 4)
        vert_shape = self.__vertices.shape
        num_vertices = vert_shape[0] * vert_shape[1]
        colors = colors.reshape(num_vertices, 3)
return colors
```

kmuehlbauer (approved):
Only a minor style thing, which is a can be changed but not a must.

vispy/visuals/surface_plot.py
djhoese: Interesting. I disagree with this :wink: 

I like that the rest of the method isn't at an extra indentation level.

kmuehlbauer: Point taken! Let's move on :grinning: .

Issue #2019, opened 2021-05-04
Axis with flipped camera bug (PanZoomCamera)


FeBe95:
## Issue
Any flipped camera (mostly used for image viewing) results in bugged linked axes. When moving/panning the view the axis moves in the opposite direction, the labels behave strangely as well.
The bug is reproducible with every implementation of the PanZoomCamera (e.g. `view.camera = "panzoom"`, `PanZoomCamera()`, or via a `Fig()`)

## Example code
```python
import vispy.plot as vp
from vispy.io import load_data_file, read_png


fig = vp.Fig(size=(600, 500))
ax = fig[0, 0]

img_data = read_png(load_data_file("mona_lisa/mona_lisa_sm.png"))
image = ax.image(img_data)

ax.camera.flip = (False, True, False)  # flips y-axis

fig.show(run=True)
```

## Example result
![Animation](https://user-images.githubusercontent.com/7470739/117049017-f0bf8800-ad13-11eb-8e14-b16da93f16a9.gif)

## System details
OS: Windows 10
Python: Python 3.9.4
Vispy: 0.6.6

djhoese:
My excuse for this is that I'm guessing the `.camera` is actually supposed to be something the user has public access to.

My solution for it though...I'm really not sure. I'm sure it needs to be handled differently somewhere in here:

https://github.com/vispy/vispy/blob/e33aae23847e8e5f43248ba9d014cc4aea211a90/vispy/scene/widgets/axis.py#L81-L88

But I'm really not sure and not an expert on how the plotting stuff was/is supposed to work. Sorry.

Kusefiru:
Hey,

I noticed the same issue. This is one of the reasons I would like to work on improving the axis widgets (as I said earlier on the gitter chat).
I'll work on it as soon as I can (if someone want to contribute in the mean time, please do so).

Kusefiru:
I've track downed the issue, which was because of this line : 

https://github.com/vispy/vispy/blob/e33aae23847e8e5f43248ba9d014cc4aea211a90/vispy/visuals/axis.py#L415

It flips the ticks positon if the axis is flipped, but it's not what we exactly want.

Say the `major_frac` (major ticks position) we get are `[0.15, 0.35, 0.55, 0.75, 0.95]`
Flipping it, we get `[0.95, 0.75, 0.55, 0.35, 0.15]`
But actually, the position of our ticks should be : `[0.85, 0.65, 0.45, 0.25, 0.05]` !

[...]

PR #2022, opened 2021-05-12
Fix axis labeling issue when flipped

Kusefiru:
Related to issue #2019

I've also removed some commented code, I don't know if I should have.

### DIFF
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -417,8 +417,8 @@ class Ticker(object):
                                     (minor_frac < 1.0001)]
             # Flip ticks coordinates if necessary :
             if flip:
-                major_frac = 1-major_frac
-                minor_frac = 1-minor_frac
+                major_frac = 1 - major_frac
+                minor_frac = 1 - minor_frac
         elif self.axis.scale_type == 'logarithmic':
             return NotImplementedError
         elif self.axis.scale_type == 'power':
--- a/vispy/visuals/axis.py
+++ b/vispy/visuals/axis.py
@@ -392,8 +392,6 @@ class Ticker(object):
             length = self.axis.pos[1] - self.axis.pos[0]  # in logical coords
             n_inches = np.sqrt(np.sum(length ** 2)) / transforms.dpi
 
-            # major = np.linspace(domain[0], domain[1], num=11)
-            # major = MaxNLocator(10).tick_values(*domain)
             major = _get_ticks_talbot(domain[0], domain[1], n_inches, 2)
 
             labels = ['%g' % x for x in major]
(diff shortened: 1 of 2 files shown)


vispy/visuals/axis.py
```diff
@@ -412,12 +410,15 @@ def _get_tick_frac_labels(self):
...
-            major_frac = major_frac[::-1] if flip else major_frac
             use_mask = (major_frac > -0.0001) & (major_frac < 1.0001)
             major_frac = major_frac[use_mask]
             labels = [l for li, l in enumerate(labels) if use_mask[li]]
             minor_frac = minor_frac[(minor_frac > -0.0001) &
                                     (minor_frac < 1.0001)]
+            # Flip ticks coordinates if necessary :
+            if flip:
+                major_frac = 1-major_frac
+                minor_frac = 1-minor_frac
```
djhoese, line 421:
```suggestion
                major_frac = 1 - major_frac
                minor_frac = 1 - minor_frac
```

djhoese (changes requested):
Thanks for doing this fix as a smaller pull request. Much easier to review. I've made a small style suggestion inline. Otherwise, do you think you could add a test to `vispy/visuals/tests/test_axis.py`? I'm not sure if there is a property that would be easy to check or some other easy way to add a test. From what I can tell the `flip` is never actually tested in the existing tests.

Kusefiru:
> Otherwise, do you think you could add a test to vispy/visuals/tests/test_axis.py? I'm not sure if there is a property that would be easy to check or some other easy way to add a test. From what I can tell the flip is never actually tested in the existing tests.

I'm not really sure about how this could be tested. Maybe from `self._ticks.pos` ?

I can give it a look, if you are okay with waiting a bit.

djhoese:
Yeah, no rush. It'd be nice to include this fix in 0.7, but I have a lot of work to do on another PR before that's ready.

djhoese:
@Kusefiru What's your availability to add tests for this?

Kusefiru:
I added two tests, tell me if they are enough or if I missed anything.

vispy/visuals/tests/test_axis.py
```diff
@@ -64,4 +64,57 @@ def test_rotation_angle():
...
+
+    canvas = scene.SceneCanvas(keys=None, size=(800, 600), show=True)
+    view = canvas.central_widget.add_view()
+    view.camera = scene.cameras.PanZoomCamera(parent=view.scene)
+
+    axis1 = visuals.Axis(pos=[[-1.0, 0], [1.0, 0]], domain=(0., 1.25), parent=view.scene)
+
+    canvas.draw_visual(axis1)
+    # Get a nice array of x ticks positions
+    x_ticks_positions = axis1._ticks.pos[:, ::2][::2].flatten()
```
djhoese, line 106: Why do you take every other (`::2`) point?

Kusefiru: Ticks positions are given for each tick line, so each tick has two (x, y) positions. As the axis is x-oriented, for each line, both x positions are equals so we can remove half of them.

As we are not interested in y positions either (again, we are testing an x-oriented axis), we can again remove half of the points.

djhoese: Got it. Sounds good. Seeing as your comment is from 8 hours ago, do you plan on making that commit with the change?

djhoese (approved):

Kusefiru:
Changes done, the request can be accepted if there's nothing more to add.

djhoese:
You rock! Glad such a basic feature is now fixed. Thank you.

PR #2025, opened 2021-05-18
Add ability to show only a mesh wireframe with the wireframe filter

asnt:
Follow-up on #1689

```python
wireframe_filter.wireframe_only = True
```
wireframe_no_interior

```python
wireframe_filter.wireframe_only = False
```
wireframe_with_interior

PR #2026, opened 2021-05-19
Add the ability to make the wireframe transparent with the wireframe filter

asnt:
That is, this draws only the interior of the faces.

Named the attribute `faces_only` to match `wireframe_only` previously defined. Maybe there is a better-suited naming, `transparent`?

Follow-up on #2025


djhoese (approved):
This looks good to me. Kind of sad that the example had to get more complicated, but I think you did a really good job of making it understandable. Thanks.

Issue #1998, opened 2021-02-19
why high-light point in darkness?


yxdragon:
![image](https://user-images.githubusercontent.com/24822467/108518606-b85b0100-7303-11eb-8cfc-0f11137e2a4b.png)

[https://forum.image.sc/t/some-doubt-when-using-vispy-high-light-point-in-darkness/49107](https://forum.image.sc/t/some-doubt-when-using-vispy-high-light-point-in-darkness/49107)

djhoese:
Could you please provide the full script you used to produce this image and maybe some system information (`python -c "import vispy; print(vispy.sys_info())"`?

yxdragon:
@djhoese 
```python
from vispy import scene
from vispy.visuals.transforms import STTransform

canvas = scene.SceneCanvas(keys='interactive', bgcolor='white', size=(800, 600), show=True)
view = canvas.central_widget.add_view()

sphere1 = scene.visuals.Sphere(radius=1, method='ico', color=(1,1,0), shading='smooth', parent=view.scene)

cam = scene.TurntableCamera(elevation=30, azimuth=30)
view.camera = cam

if __name__ == '__main__':
    canvas.app.run()
```

vispy.sys_info:
Platform: Windows-10-10.0.21301-SP0
Python:   3.6.8 (tags/v3.6.8:3c6b436a57, Dec 24 2018, 00:16:47) [MSC v.1916 64 bit (AMD64)]
NumPy:    1.16.2
Backend:  PyQt5
pyqt4:    None
pyqt5:    ('PyQt5', '5.15.1', '5.15.1')
pyside:   None
pyside2:  None
pyglet:   None
glfw:     None
sdl2:     None
wx:       wxPython 4.1.1
egl:      None
osmesa:   None
_test:    None

GL version:  '4.5.0 - Build 24.20.100.6299'
MAX_TEXTURE_SIZE: 16384


I think it is a bug, because even I use vispy.geometry.isosurface to generate a ball from 3d array, the high-light point is also wrong.
(But when I use skimage’s marching_cubes_lewiner, the result is right)

So it seems that, vispy’s geometry algrithsm give a wrong face index sequence? (so that count a invert normal vector)

djhoese:
So good news (?) is I get the same result. I don't know much about how this lighting is implemented, but interestingly if I change the camera with `view.camera = 'panzoom'` it looks like the bright spot is in the right location:

[...]

PR #2028, opened 2021-05-20
Fix various issues with shading in the MeshVisual

asnt:
Some fixes to the recent shading filter for correct/consistent behaviour.

### DIFF
--- a/examples/basics/scene/mesh_shading.py
+++ b/examples/basics/scene/mesh_shading.py
@@ -24,10 +24,12 @@ view.camera.depth_value = 1e3
 mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1))
 view.add(mesh)
 
-shading_filter = ShadingFilter(shininess=args.shininess)
-mesh.attach(shading_filter)
 wireframe_filter = WireframeFilter(width=args.wireframe_width)
+shading_filter = ShadingFilter(shininess=args.shininess)
+# Note: The wireframe filter is attached before the shading filter otherwise
+# the wireframe is not shaded.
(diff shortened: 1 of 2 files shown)


djhoese:
@asnt Should this PR be merged or should I wait for #2029 to be finished (currently a WIP)?

asnt:
> @asnt Should this PR be merged or should I wait for #2029 to be finished (currently a WIP)?

@djhoese This one is ready. (I'll try to finish #2029 soon.)

djhoese:
Just to be sure, this is for "correct/consistent" results, but these changes change the overall way the shading used to work, right?

Any idea if this PR would fix #1998?

asnt:
> Just to be sure, this is for "correct/consistent" results, but these changes change the overall way the shading used to work, right?

0c3b75e and 9036567 fix this implementation.
549a705 fixes an issue that probably existed before this filter.

The main difference with the shading before this filter is the shininess parameter. In the specular reflection, the shininness `alpha` is now the exponent `(...)^alpha` to follow the standard equations in [[1]]. Before this filter, the exponentiation was `(...)^(1/alpha)`, which is not standard, as far as I can tell. To me, this creates less confusion.

[1]: https://en.wikipedia.org/wiki/Phong_reflection_model

> 
> Any idea if this PR would fix #1998?

Indeed, partly. 9036567 fixes the specular highlight appearing on the non-illuminated side. Otherwise, I am suspecting an issue with the normals in the icosahedron-based sphere. (The recent MeshNormals visuals can help identify that.) I'll have a closer look at it when possible.

djhoese:
Thanks @asnt as always.

PR #2029, opened 2021-05-20
Identify and expose Phong shading parameters in mesh ShadingFilter

asnt:
Expose the Phong model coefficients with reasonable defaults and allowing for fine-grained control:
- `shading='flat'`
- `light_dir=(10, 5, -5)`
- `ambient_light=(1, 1, 1, 0)`
- `diffuse_light=(1, 1, 1, 1)`
- `specular_light=(1, 1, 1, .25)`
- `ambient_coefficient=(1, 1, 1, 1)`
- `diffuse_coefficient=(1, 1, 1, 1)`
- `specular_coefficient=(1, 1, 1, 1)`
- `shininess=100`

The light and reflection coefficients are RGBA where A controls the intensity. In most cases, one will only adjust the intensity (e.g. dimming the (white) light and modulating the strength of the reflection).

TODO:
- [x] Implement the Phong reflection model in [[1]].
- [x] Allow a single scalar to specify the intensity of a reflection coefficient. Reflected color defaults to white. Example: `ambient_coefficient=0.1` is transformed into `ambient_coefficient=(1.0, 1.0, 1.0, 0.1)`.
- [x] Add tests.
- [x] Add inline examples.
- ~~[RFC @djhoese ] Require light position, `light_pos`, as input parameter instead of light direction, `light_dir`. This will allow different types of lights in the future. Deprecate `light_dir`.~~ EDIT: Forget that. `light_dir` seems needed in any case. `light_pos` would be an additional parameter, for example for a point light.

Ideas, maybe for the future:
- Rename to PhongShadingFilter?  (ShadingFilter could be an alias for it if we introduce other shading models (e.g. BRDF).
- Maybe a list of basic Phong materials would be convenient
```python
standard_matrials = {
    "plastic_like": dict(ambient_coefficient=(...), diffuse_coefficient=(...), specular_coefficient=(...),...),
    "metal_like": dict(...),
    ...
}
```
- For the API, maybe something like this:
```python
material = standard_materials["plastic_like"]
filter = PhongShadingFilter(..., **material)
mesh = MeshVisual(..., material=material)
```

[1]: https://en.wikipedia.org/wiki/Phong_reflection_model

### DIFF
--- a/examples/basics/scene/mesh_shading.py
+++ b/examples/basics/scene/mesh_shading.py
@@ -21,16 +21,18 @@ view = canvas.central_widget.add_view()
 view.camera = 'arcball'
 view.camera.depth_value = 1e3
 
-# Keep the shading off here to manually manage the shading filter below for
-# demonstration purposes. Otherwise, with `shading=flat` or `shading=smooth`,
-# the Mesh class would take care of managing the filter.
-mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1), shading=None)
+# Create a colored `MeshVisual`.
+mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1))
 view.add(mesh)
 
+# Use filters to affect the rendering of the mesh.
 wireframe_filter = WireframeFilter(width=args.wireframe_width)
+# Note: For convenience, this `ShadingFilter` would be created automatically by
+# the `MeshVisual with, e.g. `mesh = MeshVisual(..., shading='smooth')`. It is
+# created manually here for demonstration purposes.
 shading_filter = ShadingFilter(shininess=args.shininess)
-# Note: The wireframe filter is attached before the shading filter otherwise
-# the wireframe is not shaded.
+# The wireframe filter is attached before the shading filter otherwise the
+# wireframe is not shaded.
 mesh.attach(wireframe_filter)
 mesh.attach(shading_filter)
 
--- a/examples/basics/scene/mesh_shading.py
+++ b/examples/basics/scene/mesh_shading.py
@@ -24,10 +24,12 @@ view.camera.depth_value = 1e3
 mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1))
 view.add(mesh)
 
-shading_filter = ShadingFilter(shininess=args.shininess)
-mesh.attach(shading_filter)
 wireframe_filter = WireframeFilter(width=args.wireframe_width)
+shading_filter = ShadingFilter(shininess=args.shininess)
+# Note: The wireframe filter is attached before the shading filter otherwise
+# the wireframe is not shaded.
 mesh.attach(wireframe_filter)
+mesh.attach(shading_filter)
 
 
 def attach_headlight(view):
--- a/examples/basics/scene/mesh_shading.py
+++ b/examples/basics/scene/mesh_shading.py
@@ -21,7 +21,10 @@ view = canvas.central_widget.add_view()
 view.camera = 'arcball'
 view.camera.depth_value = 1e3
 
-mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1))
+# Keep the shading off here to manually manage the shading filter below for
+# demonstration purposes. Otherwise, with `shading=flat` or `shading=smooth`,
+# the Mesh class would take care of managing the filter.
+mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1), shading=None)
 view.add(mesh)
 
 wireframe_filter = WireframeFilter(width=args.wireframe_width)
--- a/vispy/visuals/filters/mesh.py
+++ b/vispy/visuals/filters/mesh.py
@@ -219,7 +219,7 @@ void shade() {
 
 
 class ShadingFilter(Filter):
-    """Filter to apply shading to a mesh.
+    """Apply lighting to a mesh with the Phong reflection model.
 
     To disable shading, either detach (ex. ``mesh.detach(filter_obj)``) or
     set the shading type to ``None`` (ex. ``filter_obj.shading = None``).
@@ -227,6 +227,33 @@ class ShadingFilter(Filter):
     The shading filter should be attached after the other filters that modify
     the colors to be shaded. For example, to include the wireframe in the
     shading, the shading filter must come before the wireframe filter.
+
+    The light and reflection parameters are defined as RGBA colors with the
+    alpha channel controlling the intensity of the light or reflection. The
+    default intensity is 1.
+
+    Parameters
+    ----------
+    mode : str
+        Lighting mode: None, 'flat' or 'smooth'. If None, the lighting is
+        disabled.
+    light_dir : array-like
+        Direction of the light. Assuming a directional light.
+    ambient_light : str or tuple or Color
+        Color and intensity of the ambient light
+    diffuse_light : str or tuple or Color
+        Color and intensity of the diffuse light
+    specular_light : str or tuple or Color
+        Color and intensity of the specular light
+    ambient_coefficient : str or tuple or Color
+        Color and intensity of the ambient reflection coefficient (Ka)
+    diffuse_coefficient : str or tuple or Color
+        Color and intensity of the diffuse reflection coefficient (Kd)
+    specular_coefficient : str or tuple or Color
+        Color and intensity of the specular reflection coefficient (Ks)
+    shininess : float
+        The higher the shininess, the more localized the specular highlight.
+        Must be greater of equal to zero.
 
     Examples
     --------
--- a/vispy/visuals/filters/mesh.py
+++ b/vispy/visuals/filters/mesh.py
@@ -205,7 +205,10 @@ void shade() {
     }
     vec3 specular = specular_factor * specular_coeff * specular_light_intensity;
 
+    // XXX(asnt): Not sure if there is a physically more correct way of
+    // blending the base color with the lighting.
     vec3 color = base_color * (ambient + diffuse) + specular;
+    //vec3 color = base_color * (ambient + diffuse + specular);
     gl_FragColor.rgb = color;
 }
 """  # noqa
--- a/vispy/visuals/filters/mesh.py
+++ b/vispy/visuals/filters/mesh.py
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 # Copyright (c) Vispy Development Team. All Rights Reserved.
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
+import numbers
+
 import numpy as np
+
 from vispy.gloo import Texture2D, VertexBuffer
 from vispy.visuals.shaders import Function, Varying
 from vispy.visuals.filters import Filter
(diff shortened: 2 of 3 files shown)


examples/basics/scene/mesh_shading.py
```diff
@@ -24,10 +24,12 @@
 mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1))
 view.add(mesh)
 
-shading_filter = ShadingFilter(shininess=args.shininess)
-mesh.attach(shading_filter)
 wireframe_filter = WireframeFilter(width=args.wireframe_width)
+shading_filter = ShadingFilter(shininess=args.shininess)
+# Note: The wireframe filter is attached before the shading filter otherwise
+# the wireframe is not shaded.
```
djhoese, line 30: Would you mind adding this information to both the filter classes docstrings?

asnt: Sure. I started documenting.

vispy/visuals/filters/mesh.py
```diff
@@ -309,15 +312,20 @@ def __init__(self, shading='flat', light_dir=(10, 5, -5),
...
-        self._ambient_coefficient = Color(ambient_coefficient)
-        self._diffuse_coefficient = Color(diffuse_coefficient)
-        self._specular_coefficient = Color(specular_coefficient)
+
+        def _as_rgba(intensity_or_color):
+            if isinstance(intensity_or_color, numbers.Number):
+                intensity = intensity_or_color
+                return Color((1, 1, 1, intensity))
+            color = intensity_or_color
+            return Color(color)
```
asnt, line 321: @djhoese What do you think about this? This wraps the `vispy.color.color_array.Color` class for this specific use case and I'm not sure it fits in `Color` directly because the color of a scalar, e.g. `Color(0.3)`, doesn't make sense. Maybe a new `ColorAndIntensity` class would reflect the intent better. Should I keep it this way here?

Also, I'm thinking of moving `_as_rgba()` in the global scope of this submodule to be able to unit-test it.

djhoese: Could this be replaced with `as_vec4`:

https://github.com/vispy/vispy/blob/acafcb07a0c9947c319e1611d72cf7045670a8e7/vispy/visuals/transforms/_util.py#L50

Also, whatever method you choose you need to use in all the property `.setter`s too, right? I'm not sure this use case is unique enough to require a new class or even a new utility. This seems like something that should already exist in vispy (as_vec4?), but obviously if it doesn't then yeah we need it. There may be a different way of structuring this though too. Would doing something like `Color(as_vec4(intensity_or_color, default=(1, 1, 1, 1)))` work?

asnt: `as_vec4()` doesn't in this case as it fills the default vec4 from left to right.

asnt: Yes, it should be in the `.setter`s too. Thanks!

asnt: @djhoese I've settled for keeping the new `_as_rgba()`. Let me know if you have further thougs/preferences.

vispy/visuals/filters/mesh.py
```diff
@@ -121,104 +124,258 @@ def _attach(self, visual):
...
 
+    The shading filter should be attached after the other filters that modify
+    the colors to be shaded. For example, to include the wireframe in the
+    shading, the shading filter must come before the wireframe filter.
+
+    Parameters
+    ----------
+    shading : str
+        Shading mode: None, 'flat' or 'smooth'. If None, the shading is
+        disabled.
```
asnt, line 259: @djhoese I wonder if this parameter/property should not be renamed `mode` (or similar) and take values in `('flat', 'smooth')` and not `None`, which is currently used to disable the filter. Then, we could use an `enabled` property as in `WireframeFilter`. What do you think?

asnt: I had the impression that `ShadingFilter.shading` was redundant and didn't carry the effect of the parameter clearly but not sure that `ShadingFilter.mode` is better.
Maybe we can keep `shading` as it is and add an `enabled` property anyway for clarity and consistency with `WireframeFilter`? Then the filter would only be applied if `enabled and shading is not None`.

asnt (commented):
@djhoese I left a question/comment regarding the `shading` parameter and a potential `enabled` property.

vispy/visuals/filters/mesh.py
djhoese: Oh naming. What a fun set of decisions to make.

@kmuehlbauer @larsoner do you have any feelings about these names? I agree consistency between the different mesh filters to use `.enabled` would be nice. Regardless of how the filter is done though we probably want backwards compatibility with the MeshVisual's shading. So if the filter is different the property on the MeshVisual still needs to map to the same options to not break compatibility. `.mode` or `.method` or something else is fine by me.

asnt:
@djhoese It feels ready on my side. If not hearing another opinion on the naming above, I am OK with the current `shading` and `enabled` properties. Let me know if you have further comments. Thanks for the reviews so far ;-)

asnt:
@djhoese I think we could add tests for the expected behaviour, e.g. "if the light/coefficient is red, expect the output to have more red" and things like that. That would make quite a bunch of them but can probably be automated a bit. To me, that seems like a good check to have. What do you think?

asnt:
> That sounds like a good idea. pytest's parametrize would be good to combine a lot of these options into a single test probably. You could draw the base image and then change a single light in each parametrized test or change a couple of them and then compare to the original output.

@djhoese Tentative test added. Do you see anything else?

djhoese:
Busy day today. I'll try to review this sometime tonight. Assuming I don't find anything worth changing is this ready to be merged? Backwards compatible-ish?

asnt:
> Busy day today. I'll try to review this sometime tonight. 

No rush.

> Assuming I don't find anything worth changing is this ready to be merged?

Yes, I think it's in a reasonable shape now.

> Backwards compatible-ish?

1. Compared to the MeshVisual before this filter, configuring the shading is not backwards compatible because we cannot pass the parameters to the MeshVisual constructor (with the exception of `shading`).

2. Regarding the naming of the parameters (compared to the first iteration of the filter #1463 and also to the original MeshVisual):
  - `light_color` has been split into `ambient_light`, `diffuse_light` and `specular_light`;
  - `(ambient|diffuse|specular)_color` has been renamed to `(ambient|diffuse|specular)_color`.

So dropping `light_color` is not backwards compatible (but if 1. above is kept, this changed can possibly make sense).
The second renaming is not strictly necessary, but I found `_coefficient` clearer to relate to the standard [shading equations](https://en.wikipedia.org/wiki/Phong_reflection_model). Note: We could also add aliases `ka, kd, ks` for these coefficients as these are commonly used.

examples/basics/scene/mesh_shading.py
```diff
@@ -21,13 +21,18 @@
 view.camera = 'arcball'
 view.camera.depth_value = 1e3
 
-mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1))
+# Keep the shading off here to manually manage the shading filter below for
+# demonstration purposes. Otherwise, with `shading=flat` or `shading=smooth`,
+# the Mesh class would take care of managing the filter.
+mesh = Mesh(vertices, faces, color=(.5, .7, .5, 1), shading=None)
```
djhoese, line 27: Isn't the default `shading=None` in which case the MeshVisual never creates the internal filter?

vispy/visuals/filters/mesh.py
```diff
@@ -121,106 +124,296 @@ def _attach(self, visual):
...
+        intensity = intensity_or_color
+        return Color(default_rgb, alpha=intensity)
+    color = intensity_or_color
+    return Color(color)
+
+
 class ShadingFilter(Filter):
-    """Filter to apply shading to a mesh.
+    """Apply shading to a :class:`~vispy.visuals.mesh.MeshVisual` using the
+    Phong reflection model.
```
djhoese, line 249:
```suggestion
    """Apply shading to a :class:`~vispy.visuals.mesh.MeshVisual` using the Phong reflection model.
```

vispy/visuals/filters/mesh.py
djhoese: I think keeping it as `shading` is for the best. At least for now. Looking at the MeshVisual, it already has a `mode` property which may get confusing. And I think shading should keep all three options for now. In the future if we want to remove shading from the MeshVisual and make users have to add the filter for any shading then we can change what the shading parameter accepts.

djhoese (changes requested):
I had a couple comments/questions and one small style change. I figured I'd leave the style change to you in case you need to make another commit you can just throw it in with that commit. This looks really good from what I understand. I think I agree with just about every decision you've made. The biggest downside to all these options is that it is pretty complicated to understand all of them. I think this is a really good improvement though and should make anyone dealing with lighting happy.

djhoese:
Oh and about the materials in the future, I could see people wanting to use a single filter and cycle through the various "presets" of parameters. I think adding a `shading_filter.use_preset_material('plastic_like')` or something could be a good option. But maybe after this is merged or maybe even in a later release.

@almarklein @rougier You guys are more familiar with these types of algorithms so let me know if you have any strong feelings about this PR.

examples/basics/scene/mesh_shading.py
asnt: Yes, indeed. My intention was to be explicit but the comment more or less explains it. I'll change this and clarify the comment.

asnt:
> Oh and about the materials in the future, I could see people wanting to use a single filter and cycle through the various "presets" of parameters. I think adding a `shading_filter.use_preset_material('plastic_like')` or something could be a good option. But maybe after this is merged or maybe even in a later release.

Sounds good. Or even simpler: `shading_filter.material = 'plastic'`? And perhaps accepting a `dict` with custom values for custom materials.

examples/basics/scene/mesh_shading.py
asnt: I had a try at it in 389d998.

djhoese:
>  Or even simpler: shading_filter.material = 'plastic'?

Yeah I avoid the property because I thought it might be confusing for the user that they set a material but there isn't actually a "material" in the Filter/shader at all.

almarklein:
I think that as something that is applied on the visual, the filter should only specify the light coefficients, i.e. the properties intrinsic to the object/material. The light direction, colors, and intensities, should be global to the scene, IMO. Ideally via e.g. Light visuals. That way, multiple objects in the scene will get the same lighting applied to them, but react to it differently depending on their material properties (light coefficients).

What I propose entails quite a bit more work, so maybe the changes proposed in this PR are a good intermediate solution for the time being?

djhoese:
> What I propose entails quite a bit more work, so maybe the changes proposed in this PR are a good intermediate solution for the time being?

Yes, exactly. VisPy has very little support for proper global lighting. A lot of visuals don't even handle the z-dimension in an expected way, so lighting support for *all* visuals is going to take a bit of work. So @almarklein are you ok merging this as-is then?

almarklein:
> So @almarklein are you ok merging this as-is then?

Yes, it's still good progress. We may need to break this new API later, but we'll manage :)

djhoese (approved):
Beautiful!

PR #2031, opened 2021-05-22
Add visual for displaying mesh normals

asnt:
A subclass of LineVisual. Can display vertex and face normals.

Vertex normals (width = 2):
vertex_normals

Face normals (width = 1):
face_normals


djhoese:
Awesome! Just curious, based on your previous recent work with Meshes, could this be implemented as a Filter?

If you're done with this PR I'm OK merging it. Let me know if you want to still do some work on it.

asnt:
> Awesome! Just curious, based on your previous recent work with Meshes, could this be implemented as a Filter?

Not with this implementation, I think, because the drawing mode is "line". Otherwise, I'm not sure because we'd still need an additional geometry for the normals. Or maybe something is possible with a geometry shader (I have no experience, not sure): create a geometry on the fly for the normal vectors, provided we can distinguish that from the surface geometry in the subsequent fragment program?

> If you're done with this PR I'm OK merging it. Let me know if you want to still do some work on it.

The base functionality is ready. I just made two small enhancements for clarity.

examples/basics/scene/mesh_normals.py
```diff
@@ -0,0 +1,45 @@
+from vispy import app, scene
```
djhoese, line 1: Could you add a license header like:

https://github.com/vispy/vispy/blob/4aad67def21b539ea1b899a43196abd188788bd1/examples/basics/scene/image.py#L1-L9

djhoese: Looks like the other new mesh examples don't have it either. If you can add it to those too that would be appreciated.

vispy/visuals/mesh_normals.py
```diff
@@ -0,0 +1,61 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+
```
djhoese, line 6: I'm not sure why the linter doesn't complain about this, but since I've asked for other edits could you remove this blank line.

djhoese (changes requested):
Small license header and style requests. Then we can merge this.

vispy/visuals/mesh_normals.py
asnt: For info, there is this blank line in quite a few other files in the same directory.

examples/basics/scene/mesh_normals.py
asnt: > Could you add a license header like:
> 
> https://github.com/vispy/vispy/blob/4aad67def21b539ea1b899a43196abd188788bd1/examples/basics/scene/image.py#L1-L9

For info, what does the number after the gallery mean?

asnt: > Looks like the other new mesh examples don't have it either. If you can add it to those too that would be appreciated.

I'll do another PR.

djhoese: The gallery number is what frame to save to an image for the gallery. These are generated with the `make gallery` command and I try to run it on larger releases to update the gallery, but since the gallery does not show all of the examples (and we haven't had a release in a while) I haven't needed to run it for a while. See here:

https://github.com/vispy/vispy/blob/4aad67def21b539ea1b899a43196abd188788bd1/make/make.py#L280-L347

vispy/visuals/mesh_normals.py
djhoese: Yeah...I'm trying to standardize things and just haven't gotten around to adding a auto-formatter like `yapf` or `black`. We also have so many "wrong" docstrings that I had to disable to docstring testing because the updated version of the test was failing for a lot of vispy docstrings.

examples/basics/scene/mesh_normals.py
asnt: Thanks.

djhoese (approved):
Looks good. Thanks again.

PR #2032, opened 2021-05-22
Fix URL to contributor guide

asnt:
This is the one working for me.


djhoese (approved):
Thanks for catching this. I think I made this file before I had actually written/merged the contributor's guide. I think in the future we'll want this `en/latest` path to work, but that will take some work. Thanks again.

PR #2034, opened 2021-05-23
Add headers and license to mesh filters examples

asnt:
As discussed in #2031.

### DIFF
--- a/examples/basics/scene/mesh_shading.py
+++ b/examples/basics/scene/mesh_shading.py
@@ -4,9 +4,7 @@
 # Copyright (c) Vispy Development Team. All Rights Reserved.
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


examples/basics/scene/mesh_shading.py
```diff
@@ -1,3 +1,13 @@
+# -*- coding: utf-8 -*-
+# vispy: gallery 30
+# -----------------------------------------------------------------------------
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""Show how to use mesh filters for shading (lighting) a mesh and displaying a
+wireframe.
+"""
```
djhoese, line 9:
```suggestion
"""Show mesh filter usage for shading (lighting) a mesh and displaying a wireframe."""
```

I think we have the docstring tests turned off (or maybe not checking examples), but I think module docstrings should be one line.

PR #2036, opened 2021-05-24
Adapt mesh examples for gallery image generation

asnt:
The command line parsing in the examples was failing because it was receiving
the arguments from `python make image gallery`.

### DIFF
--- a/examples/basics/scene/mesh_shading.py
+++ b/examples/basics/scene/mesh_shading.py
@@ -19,7 +19,7 @@ default_mesh = load_data_file('orig/triceratops.obj.gz')
 parser.add_argument('--mesh', default=default_mesh)
 parser.add_argument('--shininess', default=100)
 parser.add_argument('--wireframe-width', default=1)
-args = parser.parse_args()
+args, _ = parser.parse_known_args()
 
 vertices, faces, normals, texcoords = read_mesh(args.mesh)
 
@@ -97,6 +97,7 @@ def on_key_press(event):
 
 
 canvas.show()
+
 
 if __name__ == "__main__":
     app.run()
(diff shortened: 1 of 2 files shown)


djhoese:
It's possible to only parse known arguments:

https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_known_args

Maybe that would help? It would be too bad if the examples weren't allowed to have argument parsing. It adds so much flexibility to them.

djhoese:
Or could we stop the make.py stuff from passing unknown command line arguments? Hm considering vispy has "global" command line arguments it might be best to selectively parse only the arguments we know about in the example scripts. Example: https://github.com/vispy/vispy/blob/c09edb662f3b090d61439d65c597d03dd2540ee9/vispy/app/canvas.py#L219-L220

asnt:
> It's possible to only parse known arguments:
> 
> https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_known_args
> 
> Maybe that would help? It would be too bad if the examples weren't allowed to have argument parsing. It adds so much flexibility to them.

I'm fine with keeping the CLI. It seems like `parse_known_args()` would suffice in this case. I'll do the change.

djhoese:
So you tested this and you were able to generate gallery images for it? Ready to merge?

Thanks for doing all this.

asnt:
> So you tested this and you were able to generate gallery images for it? Ready to merge?

The images are generated correctly in `_images/gallery/`, that's all I checked. Ready to merge.

NB: The default orientation of the meshes still needs to be changed for the shapes to appear upright. For another PR.

> Thanks for doing all this.

Thanks for your support.

PR #2038, opened 2021-05-26
Fix website deploy commit message references

djhoese:
I noticed while working on another project and copying from VisPy that the commit messages weren't actually rendering the environment variables I was trying to use. I think this PR should fix these (currently testing with the geoxarray project).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,7 +97,7 @@ jobs:
         cname: vispy.org
         allow_empty_commit: true
         external_repository: vispy/vispy.github.com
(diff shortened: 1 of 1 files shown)


djhoese:
It works: https://github.com/geoxarray/geoxarray.github.io/

kmuehlbauer:
Great! Thanks for fixing.

djhoese:
> Great! Thanks for fixing.

Will it ever actually benefit us to have this information in the commit message...probably not. But it's there.

kmuehlbauer:
@djhoese Now that we have it there we won't ever need it :grinning:

PR #2043, opened 2021-05-30
Add more options to control the length of normals in MeshNormals visual

asnt:
Use cases:

1. To get normals that are about the length across a triangle of the mesh:
```python
MeshNormals(..., length_method='median_edge', length_scale=1.0)
MeshNormals(...)  # equivalent to the previous line (default values)
```
2. To get normals that are about 10% the size of the mesh:
```python
MeshNormals(..., length_method='max_extent', length_scale=0.1)
```


djhoese:
Would it be possible for you to add a test?

Also, what about some examples like the ones you listed in this PR's description but in an `Examples` section in the docstring. Like:

```
    **kwargs : dict, optional
                Extra arguments to define the appearance of lines. Refer to
                :class:`~vispy.visuals.line.LineVisual`.

    Examples
    -------------
    Normals at about the length of a triangle:

    ```
    MeshNormals(..., length_method='median_edge', length_scale=1.0)
    ```

    For normals at about 10% the size of the mesh:

```

And so on. Thoughts?

asnt:
> Would it be possible for you to add a test?

I will look at that next.

> Also, what about some examples like the ones you listed in this PR's description but in an `Examples` section in the docstring. 

Good idea. I had a try.

> And so on. Thoughts?

I was thinking about making it more convenient to use this with a MeshVisual. Currently
```python
mesh = MeshVisual(...)
normals = MeshNormals(mesh.mesh_data, ...)
```
Two ideas:
1. `normals = MeshNormals(visual=mesh)`
2. `mesh = MeshVisual(..., normals='vertex')`

With option 1, it is obvious where to configure the normals, at construction time or through `normals`.
With option 2, it is not obvious where to go to configure the normals and we'd need to expose `mesh.normals`, for example.

asnt:
@djhoese In the tests, I set up a camera instead of defining the scene transformations manually. It seems simpler to maintain in my opinion. Is there maybe another reason for the manual transformations instead?

asnt:
The error is when initializing the `vispy-tests` environment with micromamba:
```bash
/usr/bin/bash -c source /home/runner/.bash_profile && micromamba list
ERROR   Cannot activate, prefix does not exist at: /home/runner/micromamba/envs/vispy-tests
Error: The process '/usr/bin/bash' failed with exit code 1
```
I don't see at first how this could be linked to the last commit.

kmuehlbauer:
> I don't see at first how this could be linked to the last commit.

@asnt Looks like a spurious thing. Maybe just rerun it will help?

asnt:
> @asnt Looks like a spurious thing. Maybe just rerun it will help?

Thanks @kmuehlbauer . I don't think I can relaunch the tests. Let's see with the new commit.

asnt:
@djhoese The tests now cover all parameters. For `length_method`, the check is for it not to raise. Not sure how better to test it.
Any further suggestion?

djhoese (changes requested):
This looks really good. Nice job on the tests. One request: could you have the first Example in your docstring show you creating the Mesh visual and passing the meshdata instance to the normals visual?

vispy/visuals/mesh_normals.py
```diff
@@ -15,19 +15,68 @@ class MeshNormalsVisual(LineVisual):
...
+    Create a :class:`~vispy.visuals.mesh.MeshVisual` on which to display
+    the normals and get the :class:`~vispy.geometry.meshdata.MeshData`:
+
+    >>> mesh = MeshVisual(vertices=vertices, faces=faces, ...)
+    >>> meshdata = mesh.mesh_data
+
+    Create a visual for the mesh normals:
+
+    >>> normals = MeshNormalsVisual(meshdata)
+    >>> normals.parent = mesh
```
djhoese, line 46: I don't think `.parent` is available on the `XVisual` version of the class. Only the `vispy.scene.visuals.MeshNormals` version of the class has that which is added by the `Node` parent class. I could be wrong. Assuming a person puts the same transforms on both visuals, is the `.parent` needed?

asnt: That's correct, there's no `.parent` here. Thanks.

djhoese (approved):

PR #2047, opened 2021-06-01
Add attenuated MIP (attenuated_mip) rendering mode to `VolumeVisual`

alisterburt:
This PR upstreams the attenuated maximum intensity projection rendering mode from napari/napari#846 (by @sofroniewn) into vispy

### DIFF
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -762,6 +762,11 @@ class VolumeVisual(Visual):
               maximum value that was encountered.
             * minip: minimum intensity projection. Cast a ray and display the
               minimum value that was encountered.
+            * attenuated_mip: attenuated maximum intensity projection. Cast a
+              ray and display the maximum value encountered. Values are
+              attenuated as the ray moves deeper into the volume.
+            * minip: minimum intensity projection. Cast a ray and display the
+              minimum value that was encountered.
             * additive: voxel colors are added along the view ray until
               the result is saturated.
             * iso: isosurface. Cast a ray until a certain threshold is
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -814,7 +814,8 @@ class VolumeVisual(Visual):
     @attenuation.setter
     def attenuation(self, value):
         self._attenuation = float(value)
-        self.shared_program['u_attenuation'] = self._attenuation
+        if 'u_attenuation' in self.shared_program:
+            self.shared_program['u_attenuation'] = self._attenuation
         self.update()
 
     @property
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -783,7 +783,7 @@ class VolumeVisual(Visual):
         # Get rid of specific variables - they may become invalid
         if 'u_threshold' in self.shared_program:
             self.shared_program['u_threshold'] = None
-        elif 'u_attenuation' in self.shared_program:
+        if 'u_attenuation' in self.shared_program:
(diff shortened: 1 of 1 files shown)


vispy/visuals/volume.py
```diff
@@ -749,6 +779,17 @@ def threshold(self, value):
...
 
+    @property
+    def attenuation(self):
+        """The attenuation rate to apply for the attenuated mip render method."""
+        return self._attenuation
+
+    @attenuation.setter
+    def attenuation(self, value):
+        self._attenuation = float(value)
+        self.shared_program['u_attenuation'] = self._attenuation
```
djhoese, line 790: This should probably have the `if 'u_attenuation' ...` like threshold's setter has.

vispy/visuals/volume.py
```diff
@@ -728,6 +756,8 @@ def method(self, method):
         # Get rid of specific variables - they may become invalid
         if 'u_threshold' in self.shared_program:
             self.shared_program['u_threshold'] = None
+        elif 'u_attenuation' in self.shared_program:
```
djhoese, line 759: Can this maybe be made an `if` instead of `elif` for future flexibility? Otherwise this assumes that two shaders will never use both parameters. I know, nit-pick, but still.

djhoese (changes requested):
Just some small changes, but otherwise looks pretty good.

sofroniewn:
This is looking great @alisterburt! It will let us remove our vendored copy of the volume visual from napari

vispy/visuals/volume.py
```diff
@@ -707,8 +737,6 @@ def method(self):
               the result is opaque.
             * mip: maxiumum intensity projection. Cast a ray and display the
               maximum value that was encountered.
-            * minip: minimum intensity projection. Cast a ray and display the
-              minimum value that was encountered.
```
almarklein, line 711: Why are these lines removed?

almarklein (commented):
Nice!

vispy/visuals/volume.py
alisterburt: whoops - merge shenanigans, will fix!

alisterburt:
thanks for the reviews @djhoese, @almarklein and @sofroniewn - much appreciated

I've updated to incorporate suggested changes, rebased onto master and fixed docstrings, assuming all tests pass I think we're good to go!

vispy/visuals/volume.py
```diff
@@ -731,6 +762,11 @@ def method(self):
               maximum value that was encountered.
             * minip: minimum intensity projection. Cast a ray and display the
               minimum value that was encountered.
+            * attenuated_mip: attenuated maximum intensity projection. Cast a
+              ray and display the maximum value encountered. Values are
+              attenuated as the ray moves deeper into the volume.
+            * minip: minimum intensity projection. Cast a ray and display the
+              minimum value that was encountered.
             * additive: voxel colors are added along the view ray until
```
djhoese, line 770:
```suggestion
            * attenuated_mip: attenuated maximum intensity projection. Cast a
              ray and display the maximum value encountered. Values are
              attenuated as the ray moves deeper into the volume.
            * additive: voxel colors are added along the view ray until
```

djhoese (approved):
I fixed one small thing (minip was listed twice in the docstring). Looks good now. It bothers me a little that `_attenuation` isn't directly declared in the `__init__` but threshold works the same way so :shrug: we'll ignore it for now. Thanks.

alisterburt:
thanks @djhoese for handling this so smoothly!

PR #2055, opened 2021-06-08
Add average intensity projection (average) rendering mode to `VolumeVisual`

alisterburt:
In an ongoing effort to upstream changes to the `VolumeVisual` made in napari here is the average intensity projection volume rendering introduced in napari/napari#1871

This addition will be covered by tests @djhoese added in #1912


djhoese (approved):
LGTM. Thanks.

PR #2057, opened 2021-06-09
Add setter for colorbar label text

djhoese:
Let's try this again. This is a replacement for #2056 which after trying to rebase and merge just got too ugly with commit history. I instead copied the current state of #1292 into a new branch and committed with @macrocosme as the author of the commit.

Updates .label property so it can be provided a TextVisual or a new string to update the existing TextVisual of a ColorbarVisual.

This is a replacement for #1292 so that I can actually make commits to it (compared to the original fork's master branch).

I'll need to reapply the style fixes from `main`.

### DIFF
--- a/vispy/scene/widgets/colorbar.py
+++ b/vispy/scene/widgets/colorbar.py
@@ -65,7 +65,6 @@ class ColorBarWidget(Widget):
     axis_ratio : float
         ratio of minor axis to major axis
     """
-
     def __init__(self, cmap, orientation,
                  label="", label_color='black', clim=("", ""),
                  border_width=0.0, border_color="black",
@@ -78,7 +77,7 @@ class ColorBarWidget(Widget):
 
         self._colorbar = ColorBarVisual(size=dummy_size, cmap=cmap,
                                         orientation=orientation,
-                                        label_str=label, clim=clim,
+                                        label_text=label, clim=clim,
                                         label_color=label_color,
                                         border_width=border_width,
                                         border_color=border_color, **kwargs)
@@ -103,7 +102,8 @@ class ColorBarWidget(Widget):
         self._colorbar.size = self._calc_size()
 
     def _calc_size(self):
-        """Calculate a size"""
+        """Calculate a size
+        """
         (total_halfx, total_halfy) = (self.rect.right, self.rect.top)
         if self._colorbar.orientation in ["bottom", "top"]:
             (total_major_axis, total_minor_axis) = (total_halfx, total_halfy)
@@ -155,7 +155,8 @@ class ColorBarWidget(Widget):
 
     @property
     def border_color(self):
-        """The color of the border around the ColorBar in pixels"""
+        """ The color of the border around the ColorBar in pixels
+        """
         return self._colorbar.border_color
 
     @border_color.setter
@@ -164,7 +165,8 @@ class ColorBarWidget(Widget):
 
     @property
     def border_width(self):
-        """The width of the border around the ColorBar in pixels"""
+        """ The width of the border around the ColorBar in pixels
+        """
         return self._colorbar.border_width
 
     @border_width.setter
--- a/vispy/scene/widgets/colorbar.py
+++ b/vispy/scene/widgets/colorbar.py
@@ -78,7 +78,7 @@ class ColorBarWidget(Widget):
 
         self._colorbar = ColorBarVisual(size=dummy_size, cmap=cmap,
                                         orientation=orientation,
-                                        label_text=label, clim=clim,
+                                        label=label, clim=clim,
                                         label_color=label_color,
                                         border_width=border_width,
                                         border_color=border_color, **kwargs)
--- a/vispy/scene/widgets/colorbar.py
+++ b/vispy/scene/widgets/colorbar.py
@@ -65,6 +65,7 @@ class ColorBarWidget(Widget):
     axis_ratio : float
         ratio of minor axis to major axis
     """
+
     def __init__(self, cmap, orientation,
                  label="", label_color='black', clim=("", ""),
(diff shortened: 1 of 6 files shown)


djhoese:
@kmuehlbauer This should be ready to go if you have time to review it.

kmuehlbauer:
> @kmuehlbauer This should be ready to go if you have time to review it.

I'm already afk, but will have a look tomorrow morning.

vispy/visuals/colorbar.py
```diff
@@ -255,7 +253,7 @@ class ColorBarVisual(CompoundVisual):
     pos : tuple (x, y)
         Position where the colorbar is to be placed with
         respect to the center of the colorbar
-    label_str : str
+    label : str
```
kmuehlbauer, line 256: I'm wondering if we should allow here `TextVisual` too, as we do below in the `label.setter`?

kmuehlbauer (approved):
I have just one remark. Otherwise LGTM.

vispy/scene/widgets/colorbar.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, cmap, orientation,
 
         self._colorbar = ColorBarVisual(size=dummy_size, cmap=cmap,
                                         orientation=orientation,
-                                        label_str=label, clim=clim,
+                                        label_text=label, clim=clim,
```
almarklein, line 81: This should be just `label`, right?

djhoese: Yeah...looks like we need another test.

djhoese:
@kmuehlbauer @almarklein Ok the last set of commits do a few things:

1. Fix the colorbar widget so it passes label properly
2. Add tests for the colorbar widget (there weren't any)
3. Add the ability to pass a TextVisual to the ColorbarVisual for the label (and add a test)
4. Remove `**kwargs` from the ColorbarVisual. I'm not really sure why this is here (maybe some of the tests will end up failing), but I'd confidently say we don't want `**kwargs` unless it is really really necessary.

djhoese:
Ok, tests pass now. This should be ready for re-review and merge.

vispy/visuals/colorbar.py
```diff
@@ -280,28 +284,27 @@ class ColorBarVisual(CompoundVisual):
...
         self._cmap = get_colormap(cmap)
         self._clim = clim
         self._pos = pos
         self._size = size
         self._orientation = orientation
 
-        self._label = TextVisual(self._label_str, color=self._label_color)
+        if not isinstance(label, TextVisual):
+            label = TextVisual(label, color=label_color)
+        self._label = label
```
kmuehlbauer, line 301: Do we need to override a given TextVisual's color with the given `label_color`? So if the user gives a `TextVisual` which is in `green` and also gives `label_color="red"`. Or should we just make clear in the docstring, that `label_color` is only used when `label` is a string?

kmuehlbauer (commented):
Just one more remark on the `label_color`.

vispy/visuals/colorbar.py
almarklein: > Or should we just make clear in the docstring, that label_color is only used when label is a string?

That's what happens now, right? :)

almarklein: or do you mean the ticks below? Maybe these should use `label.color` instead?

vispy/visuals/colorbar.py
```diff
@@ -255,9 +253,10 @@ class ColorBarVisual(CompoundVisual):
...
         Position where the colorbar is to be placed with
         respect to the center of the colorbar
-    label_str : str
+    label : str | vispy.visuals.TextVisual
         The label that is to be drawn with the colorbar
         that provides information about the colorbar.
+        If a TextVisual object then 'label_color' is ignored.
     label_color : str | vispy.color.Color
         The color of the labels. This can either be a
         str as the color's name or an actual instace of a vipy.color.Color
```
almarklein, line 262: This also determines the color of the ticks.

vispy/visuals/colorbar.py
almarklein: If so, we should also update the tick colors when the label setter is given a label object.

almarklein (changes requested):
Only comment is to mention ticks in docs for `label_color`.

vispy/visuals/colorbar.py
kmuehlbauer: So many options... We might need to add setters for all that stuff...

djhoese: I added a note in the `label` docstring that color is not used if it is a TextVisual. I think every Visual in VisPy needs to make their properties more available. Not sure this PR is going to be the place though. Might be something good for a hacktoberfest issue.

vispy/visuals/colorbar.py
djhoese: Ooohhh. :thinking: My added statement isn't exactly wrong though as far as the color of the label goes.

vispy/visuals/colorbar.py
```diff
@@ -255,9 +253,10 @@ class ColorBarVisual(CompoundVisual):
     pos : tuple (x, y)
         Position where the colorbar is to be placed with
         respect to the center of the colorbar
-    label_str : str
+    label : str | vispy.visuals.TextVisual
         The label that is to be drawn with the colorbar
         that provides information about the colorbar.
+        If a TextVisual object then 'label_color' is ignored.
     label_color : str | vispy.color.Color
         The color of the labels. This can either be a
```
djhoese, line 261:
```suggestion
        The color of the label and tick labels. This can either be a
```

almarklein (approved):

PR #2058, opened 2021-06-09
Dont force selection of gl2 backend

### LINKED ISSUE #1109: Vispy always tries to import gl2 by default
At https://github.com/vispy/vispy/blob/master/vispy/gloo/gl/__init__.py#L235-L238, vispy always tries to import gl2 and call use_gl with 'gl2' as the default argument. This fails on Android (and I assume other places) because only gles is available. I fixed it for that by patching to import es2 by default instead.

I'd like to change it so that Vispy works without modification on Android, but I'm not sure how since gloo.gl does need to import something as a default. One option would be to try the current method, then if it fails catch the Exception and try es2 instead, but I'm not sure if there are downsides to that.

Any ideas?

almarklein:
Closes #1109

### DIFF
--- a/vispy/gloo/gl/__init__.py
+++ b/vispy/gloo/gl/__init__.py
@@ -11,7 +11,7 @@ visualizations using Angle, WebGL, or other forms of remote rendering.
 This is in part possible by the widespread availability of OpenGL ES 2.0.
 
 All functions that this API provides accept and return Python arguments
-(no ctypes is required); strings are real strings and you can pass 
+(no ctypes is required); strings are real strings and you can pass
 data as numpy arrays. In general the input arguments are not checked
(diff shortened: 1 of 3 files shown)


djhoese (approved):
I think this makes sense to me and looks good. It kind of makes me think that the GL backends should have a value in them for this `.__name__.split(".")[-1]` operation but I suppose it is only used in two places so maybe not worth it.

djhoese:
Feel free to merge this if you're happy with it.

almarklein:
> I think this makes sense to me and looks good. It kind of makes me think that the GL backends should have a value in them for `this .__name__.split(".")[-1]` operation but I suppose it is only used in two places so maybe not worth it.

Exactly what I was thinking :)

PR #2066, opened 2021-06-11
Change builtin colormaps to all be instances

### LINKED ISSUE #1487: Some colormaps in the `get_colormaps` dictionary are not Colormap instances
If I recall correctly this was added by @larsoner. In `vispy/color/colormap.py` there is a `_colormaps` dictionary that can be accessed via the public `get_colormaps()`. However, as a user, I expect this dictionary to map colormap names to colormap instances, some of these values though are colormap classes or functions. For example, `cubehelix -> CubeHelixColormap` and `single_hue -> _SingleHue`. This is "worked around" by having the `get_colormap` function create these colormaps with optional kwargs passed to the class.

I'm not a huge fan of this so unless someone can argue for this behavior I would like to propose that we replace these non-instance colormaps with basic instances or remove them from the colormap dictionary returned from `get_colormaps`.

djhoese:
Closes #1487 

Some of the colormaps available through `vispy.color.colormap.get_colormap` were classes that were created when `get_colormap` was called with optional args, kwargs provided. This has always smelled bad to me code-wise. In this PR I create instances of the colormaps as replacements and deprecate ones that are redundant.

One consideration: should I just remove all of these instances, list them as deprecated, and then rename the classes to not be prefixed with `_`? So `_SingleHue` becomes `SingleHue` and if a user wants to create this colormap they can import it, create it, and pass it directly.

Edit: I have not updated the tests which seem to test some of these specifically.

### DIFF
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -1084,12 +1084,12 @@ _colormaps = dict(
     GrBu_d=Diverging(255, 133, 0.75, 0.6, "dark"),
     RdBu=Diverging(220, 20, 0.75, 0.5),
 
-    cubehelix=CubeHelixColormap,
-    single_hue=SingleHue,
-    hsl=HSL,
-    husl=HSLuv,
-    diverging=Diverging,
-    RdYeBuCy=RedYellowBlueCyan,
+    cubehelix=CubeHelixColormap(),
+    single_hue=SingleHue(),
+    hsl=HSL(),
+    husl=HSLuv(),
+    diverging=Diverging(),
+    RdYeBuCy=RedYellowBlueCyan(),
 )
 
 
@@ -1129,7 +1129,6 @@ def get_colormap(name, *args, **kwargs):
         warnings.warn(f"Colormap '{name}' has been deprecated. "
                       f"Please import and create 'vispy.color.colormap.{cls.__name__}' "
                       "directly instead.", DeprecationWarning)
-        return cls()
 
     if isinstance(name, BaseColormap):
         return name
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -763,7 +763,7 @@ class _Diverging(Colormap):
 
 
 class _RedYellowBlueCyan(Colormap):
-    """A colormap which is goes red-yellow positive and blue-cyan negative
+    """A colormap which goes red-yellow positive and blue-cyan negative
 
     Parameters
     ---------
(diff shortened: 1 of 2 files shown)


almarklein:
What exactly is the concern with the current situation? Is it the potential performance issue for creating new instances on every call to `get_colormap`?  I agree that having classes in the dict is not super-pretty, but that's private code that users will never see or touch, so maybe not so important.

If we're making this change, we'd have to think of another way for users to get these parameterized colormaps, right? Which means breaking the API, and possibly making it more complex. Just wondering if that's worth it ...

djhoese:
The overall complaint is code smell. I think when I originally made #1487 this was a simpler issue and I didn't see that `get_colormap` took args/kwargs to make the colormaps. However, `get_colormap` is also used in many visuals' `cmap.setter` property where it is impossible to pass args/kwargs. So there is no benefit to having this convenience when it can't even be used in the typical way that colormaps are set in vispy. That leaves the only other way you could do this which is to import the colormap class and create an instance of the colormap.

vispy/color/colormap.py
```diff
@@ -1085,48 +1085,56 @@ def __init__(self, limits=(0.33, 0.66, 1.0)):
...
+    diverging=_Diverging(),
+    RdYeBuCy=_RedYellowBlueCyan(),
 )
 
 
-def get_colormap(name, *args, **kwargs):
-    """Obtain a colormap
-    Some colormaps can have additional configuration parameters. Refer to
-    their corresponding documentation for more information.
+def get_colormap(name):
```
almarklein, line 1098: Just wondering whether we should have args and kwargs here and raise a DeprecationError or something.

I have no clue how many users would actually use it, but it if anyone would, it would be hard for them to see how they must now adjust their code.

almarklein (commented):
Fair enough :)  Only one comment to maybe help users that are actually using the args/kwargs, in case there are any.

vispy/color/colormap.py
djhoese: Yeah I can see what "feels" best. I think I'm going to go ahead and remove the instances of the classes that I created from the classes that were in the dictionary. I'll see how much information I can add to the versionchanged section in the get_colormap docstring.

djhoese:
Ok so I had to do this a little differently than I expected because of some errors I was seeing in napari (FYI @sofroniewn @jni @tlambert03 and whoever else). This PR deprecates any colormap that is a class instead of an actual instance. If I completely remove these colormap entries then vispy can't even start it seems (I have no idea why this is happening because I'm not importing napari) because napari is trying to use and translate the colormap entries:

https://github.com/napari/napari/blob/d26d84a3d9460763a3a8c29454a75845239a7c50/napari/utils/colormaps/colormap_utils.py#L65

So I've kept them in, but added a deprecation warning and still have it create a colormap instances. This is what the deprecation looks like:

```
In [1]: from vispy.color import get_colormap

In [2]: import warnings; warnings.simplefilter('always')

In [3]: get_colormap('husl')
/home/davidh/repos/git/vispy/vispy/color/colormap.py:1132: DeprecationWarning: Colormap 'husl' has been deprecated. Please import and create 'vispy.color.colormap.HSLuv' directly instead.
  "directly instead.", DeprecationWarning)
Out[3]: <vispy.color.colormap.HSLuv at 0x7f33de110990>
```

I'm not sure if this is better or if I should just let it KeyError with a deprecation warning.

tlambert03:
seems like a good change to me

sofroniewn:
> If I completely remove these colormap entries then vispy can't even start it seems (I have no idea why this is happening because I'm not importing napari) because napari is trying to use and translate the colormap entries:

That's good to know. Thanks for those deprecations, I think can remove those colormaps from napari/ recreate them as we see fit. Shouldn't be a problem

tlambert03:
I ❤️ cubehelix.  so I hope we can add it back! :)

djhoese:
@tlambert03 How do you currently create it? In vispy it is a class, so if it is created in napari then it must have some default arguments that are passed to make an instance of it.

tlambert03:
yeah, default args, lemme see if I can find them

tlambert03:
ah, sorry ... looks like our default one is vendored from matplotlib not vispy

vispy/color/colormap.py
```diff
@@ -1080,53 +1079,72 @@ def __init__(self, limits=(0.33, 0.66, 1.0)):
...
 
-    # Configurable colormaps
     cubehelix=CubeHelixColormap,
-    single_hue=_SingleHue,
-    hsl=_HSL,
-    husl=_HUSL,
-    hsluv=_HSLuv,
-    diverging=_Diverging,
-    RdYeBuCy=_RedYellowBlueCyan,
+    single_hue=SingleHue,
```
almarklein, line 1088: What about adding the default `SingleHue` instance here (and same for classes below). That way we don't break things in terms of keys, but still remove the classes, which was the original goal.

almarklein:
I thought `_colormaps` is but an implementation detail, but I guess it is not:
```py
def get_colormaps():
    """Return the list of colormap names."""
    return _colormaps.copy()
```

vispy/color/colormap.py
djhoese: I think that's what I had originally. I guess that would still work if I update the deprecation warning to use `.__class__.__name__`. Then it also effects anyone using the dictionary originally. I'd still like the deprecation warning even though they would not be instances since they are probably not what people want/expect.

almarklein (approved):
LGTM!

PR #2067, opened 2021-06-12
When using the wx backend with sys_info() the OpenGL version was alwa…

rklevin:
…ys an empty string.

The change below solved this problem.

In util/config.py in sys_info()
original:
canvas = Canvas('Test', (10, 10), show=False, app=app)
replacement:
canvas = Canvas('Test', (10, 10), show=True, app=app)

The "show=False" prevented the following line:
canvas._backend._vispy_set_current()
from setting the context. See _vispy_set_current() in _wx.py.

### DIFF
--- a/vispy/util/config.py
+++ b/vispy/util/config.py
@@ -426,7 +426,7 @@ def sys_info(fname=None, overwrite=False):
             out += '{0:<9} {1}\n'.format(backend + ':', which)
         out += '\n'
         # We need an OpenGL context to get GL info
-        canvas = Canvas('Test', (10, 10), show=False, app=app)
+        canvas = Canvas('Test', (10, 10), show=True, app=app)
         canvas._backend._vispy_set_current()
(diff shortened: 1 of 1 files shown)


djhoese:
Thanks for working on a fix for this. However, we can't have `show=True` in this function as it would mean that a window has to be created and shown whenever someone gets sys_info. `sys_info()` has to be available even without a display and without requiring the user to close a window that will pop up. This sounds like a Wx-only issue and should be a wx-only fix.

djhoese:
I do not use Wx myself, but do you @rklevin have any other ideas for how this can be fixed for the Wx backend specifically? Otherwise, should we close this until someone comes up with a solution?

I could be forgetting about some issues regarding this, but I'm not sure any other users have run into this so this could also be something specific about your environment/system or your version of Wx.

djhoese:
Well the original author of this PR has deleted their account and we can't really use these changes as I've described above. Going to close this.

PR #2069, opened 2021-06-13
Fix errors in geometry generation and improve mesh normal calculations

yxdragon:
1. upspeed vertex normal counting with np.add.at
2. modify geometry. create_grid_mesh, create_plane, upspeed and solve the normal dismatch problem.
(then the create_box and create_sphere is ok from create_plane)

### DIFF
--- a/vispy/geometry/generation.py
+++ b/vispy/geometry/generation.py
@@ -121,28 +121,33 @@ def create_plane(width=1, height=1, width_segments=1, height_segments=1, directi
         from http://git.io/vU1Fh
     """
     w, h = width_segments + 1, height_segments + 1
-    vts = np.mgrid[-0.5:0.5:h*1j, -0.5:0.5:w*1j, :1]
-    vts = vts.T.reshape(-1,3).astype(np.float32)
-    normals = np.repeat([np.array([0,0,1], np.float32)], len(vts), axis=0)
-    texture = vts[:,:2] + 0.5
+    vts = np.mgrid[-0.5:0.5:w*1j, -0.5:0.5:h*1j, :1]
+    vts = vts.T.reshape(-1, 3).astype(np.float32)
 
-    k = np.array([width, height, 1], dtype=np.float32)
-    o = np.array([0.5, 0.5, 0], dtype=np.float32)
-    lut = np.array({'x':[2,0,1], 'y':[1,2,0], 'z':[0,1,2]}[direction[1]])
-    vts[:] = vts[:,lut];  normals[:] = normals[:,lut]
-    if direction[0]=='-':  vts[:,lut>0] *= -1; normals[:,lut>0] *= 1
-    colors = np.hstack([vts+o[lut], np.ones((len(vts), 1), dtype=np.float32)])
+    normals = np.repeat([np.array([0, 0, 1], np.float32)], len(vts), axis=0)
+    texture = vts[:, :2] + 0.5
+
+    scale = np.array([width, height, 1], dtype=np.float32)
+    offset = np.array([0.5, 0.5, 0], dtype=np.float32)
+
+    lut = np.array({'x': [2, 0, 1], 'y': [1, 2, 0], 'z': [0, 1, 2]}[direction[1]])
+    vts[:], normals[:] = vts[:, lut], normals[:, lut]
+    if direction[0]=='-': vts[:, lut>0] *= -1
+    if direction[0]=='-': normals[:, lut>0] *= -1
+    colors = np.hstack([vts+offset[lut], np.ones((len(vts), 1), dtype=np.float32)])
     
     face_cell = np.array([[0, 1, 1+w, 0, 1+w, w]], dtype=np.uint32)
     line_cell = np.array([[0, 1, 1, 1+w, 1+w, w, w, 0]], dtype=np.uint32)
-    dif_idx = np.arange(0,w*h-w,w)[:,None] + np.arange(0,w-1,1)
-    faces = dif_idx.reshape(-1,1) + face_cell
-    lines = dif_idx.reshape(-1,1) + line_cell
+    dif_idx = np.arange(0, w*h-w, w)[:, None] + np.arange(0, w-1, 1)
 
-    mesh = np.hstack([np.multiply(vts, k[lut], out=vts), texture, normals, colors])
+    faces = dif_idx.reshape(-1, 1) + face_cell
+    lines = dif_idx.reshape(-1, 1) + line_cell
+
+    mesh = np.hstack([np.multiply(vts, scale[lut], out=vts), texture, normals, colors])
     mesh.dtype = np.dtype([('position', np.float32, 3), ('texcoord', np.float32, 2), 
                            ('normal', np.float32, 3), ('color', np.float32, 4)])
-    return mesh.ravel(), faces.reshape(-1,3), lines.reshape(-1,2)
+
+    return mesh.ravel(), faces.reshape(-1, 3), lines.reshape(-1, 2)
 
 
 def create_box(width=1, height=1, depth=1, width_segments=1, height_segments=1,
@@ -540,6 +545,7 @@ def create_arrow(rows, cols, radius=0.1, length=1.0,
 
     return MeshData(vertices=verts, faces=faces)
 
+
 def create_grid_mesh(xs, ys, zs):
     """Generate vertices and indices for an implicitly connected mesh.
 
@@ -568,6 +574,6 @@ def create_grid_mesh(xs, ys, zs):
     h, w = xs.shape
     vts = np.array([xs, ys, zs], dtype=np.float32)
     face_cell = np.array([[0, 1, 1+w, 0, 1+w, w]], dtype=np.uint32)
-    fase_index = np.arange(0,w*h-w,w)[:,None] + np.arange(0,w-1,1)
-    faces = fase_index.reshape(-1,1) + face_cell
-    return vts.reshape(3,-1).T, faces.reshape(-1,3)
+    fase_index = np.arange(0, w*h-w, w)[:,None] + np.arange(0, w-1, 1)
+    faces = fase_index.reshape(-1, 1) + face_cell
+    return vts.reshape(3, -1).T, faces.reshape(-1, 3)
--- a/vispy/geometry/generation.py
+++ b/vispy/geometry/generation.py
@@ -132,8 +132,8 @@ def create_plane(width=1, height=1, width_segments=1, height_segments=1, directi
 
     lut = np.array({'x': [2, 0, 1], 'y': [1, 2, 0], 'z': [0, 1, 2]}[direction[1]])
     vts[:], normals[:] = vts[:, lut], normals[:, lut]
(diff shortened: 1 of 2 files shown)


djhoese:
Looks like the linter has some code style issues. Would you mind fixing them? You can see them here:

https://github.com/vispy/vispy/pull/2069/checks?check_run_id=2815252935#step:7:54

Let me know if you have any questions.

@asnt After those are fixed, do you have time to review at least the meshdata.py changes?

vispy/geometry/generation.py
```diff
@@ -121,80 +120,29 @@ def create_plane(width=1, height=1, width_segments=1, height_segments=1,
...
-                         ('color', np.float32, 4)])
-
-    vertices['position'] = positions
-    vertices['texcoord'] = texcoords
-    vertices['normal'] = normals
-    vertices['color'] = colors
-
-    return vertices, faces, outline
+    w, h = width_segments + 1, height_segments + 1
+    vts = np.mgrid[-0.5:0.5:h*1j, -0.5:0.5:w*1j, :1]
```
asnt, line 124: I would name this explicitly: `vertices`.

vispy/geometry/generation.py
```diff
@@ -121,80 +120,29 @@ def create_plane(width=1, height=1, width_segments=1, height_segments=1,
...
+    vts = vts.T.reshape(-1,3).astype(np.float32)
+    normals = np.repeat([np.array([0,0,1], np.float32)], len(vts), axis=0)
+    texture = vts[:,:2] + 0.5
+
+    k = np.array([width, height, 1], dtype=np.float32)
+    o = np.array([0.5, 0.5, 0], dtype=np.float32)
```
asnt, line 130: I would suggest spelling out full words instead of `k` and `o`.

asnt:
Thanks @yxdragon. Since there was the question of style, I don't know what VisPy devs prefer, but I would suggest using explicit names where possible to aid in understanding the operations,. I tried to highlight to examples as inline comments.

> @asnt After those are fixed, do you have time to review at least the meshdata.py changes?

@djhoese Yes, sure. And I can help review the rest as well.

djhoese:
@yxdragon Do you have time to fix the style issues? You can push new commits to the same branch you used and github will automatically update the pull request.

Also, a note for the future, you may want to create a new branch whenever you plan on changing the vispy code and making a pull request. You used your `main` branch which means it is now out of sync with the upstream vispy branch. This can make it harder to keep your work organized and handle some "fun" things with git.

yxdragon:
@djhoese I have time, I am looking for a ide which can check style and mark it.

djhoese:
Pycharm is my go-to IDE, but you could also use `flake8` to manually check it. Otherwise VSCode is another popular IDE.

djhoese:
Thanks for fixing some of these. There are a few remaining:

```
vispy/geometry/generation.py:135:20: E225 missing whitespace around operator
vispy/geometry/generation.py:135:25: E701 multiple statements on one line (colon)
vispy/geometry/generation.py:135:37: E225 missing whitespace around operator
vispy/geometry/generation.py:136:20: E225 missing whitespace around operator
vispy/geometry/generation.py:136:25: E701 multiple statements on one line (colon)
vispy/geometry/generation.py:136:41: E225 missing whitespace around operator
vispy/geometry/generation.py:577:42: E231 missing whitespace after ','
```

vispy/geometry/generation.py
```diff
@@ -121,80 +120,34 @@ def create_plane(width=1, height=1, width_segments=1, height_segments=1,
...
-    texcoords = np.zeros(x_grid1 * y_grid1 * 2)
+    scale = np.array([width, height, 1], dtype=np.float32)
+    offset = np.array([0.5, 0.5, 0], dtype=np.float32)
 
-    y = np.arange(y_grid1) * height / y_grid - height / 2
-    x = np.arange(x_grid1) * width / x_grid - width / 2
+    lut = np.array({'x': [2, 0, 1], 'y': [1, 2, 0], 'z': [0, 1, 2]}[direction[1]])
+    vts[:], normals[:] = vts[:, lut], normals[:, lut]
+    vts[:, lut>0] *= {'+': 1, '-': -1}[direction[0]]
+    normals[:, lut>0] *= {'+': 1, '-': -1}[direction[0]]
```
djhoese, line 136:
```
vispy/geometry/generation.py:135:15: E225 missing whitespace around operator
vispy/geometry/generation.py:136:19: E225 missing whitespace around operator
```

```suggestion
    vts[:, lut > 0] *= {'+': 1, '-': -1}[direction[0]]
    normals[:, lut > 0] *= {'+': 1, '-': -1}[direction[0]]
```

djhoese: I think this should fix it.

djhoese:
@asnt looks like the style issues have all been fixed. Mind re-reviewing?

@yxdragon Do you think you could add some tests to https://github.com/vispy/vispy/blob/74c63f12f8e7a30954a19878f51b956487b69a5a/vispy/geometry/tests/test_generation.py that would fail without your fixes?

asnt:
> @asnt looks like the style issues have all been fixed. Mind re-reviewing?

I'll give it a try as soon as possible.

asnt (commented):
@djhoese @yxdragon I have inlined reviews for the orientation of the faces. I cannot look at the normals for now. As explained in the comments, there are many changes here. Maybe we should focus more?

djhoese:
If someone can write the tests for this PR that thoroughly check that everything is correct, I (or someone else) can do the checks that the simple fixes also pass the tests. If the rest of the changes in this PR are optimizations then I agree that we should be careful. Step 1: Add tests.

yxdragon:
I think the best test is show a box with old and new method. I render it with lines to show the structure is same as the old version. then I render it with faces to show the normals is right now.
```python
from vispy import scene
from vispy.geometry import create_box

vts, faces, ls = create_box(1,2,3, 1,2,3)
box = scene.visuals.Mesh(vertices=vts['position'], faces=ls, mode='lines', parent=view.scene)
# scene.visuals.MeshNormals(meshdata=box._meshdata, parent=view.scene)
view.camera = scene.TurntableCamera(elevation=30, azimuth=30)

if __name__ == '__main__':
    canvas.app.run()
```
the old method:
image
the new method:
image

then visualize it with face, and add a meshnormals.
the old method:
image
the new method:
image

djhoese:
What would be simplest and best from a test point of view is to call the geometry methods and check a few specific indexes in the vertices, faces, and other return values. That way we don't have to render anything to check the image output.

yxdragon:
So, is there a standard true data set? if there is, I can write a test to generate the standard data? If there is'n, I trust the render result.
the create_box has generated +xyz, -xyz 6 planes, I make a box in 1,2,3 width, height, depth, in 1,2,3 segments. render it in line and face mode, and show the normal vector. It looks right, I think it is the best test.

djhoese:
> vts, faces, ls = create_box(1,2,3, 1,2,3)

If you do this and check the output for the values you expect (at least for the geometry changes, not the mesh changes you've made) then there should be no need to render any visualization. You can do that yourself, "manually", to make sure the values returned by the function(s) look correct, but then you can just write the test to compare against these known good values.

yxdragon:
**the fact is: even we do a lot of test "manually", We also found some bugs or errors from visualization result.**
So why not use some typical visualization demo as test?

If we need to check it by data, not visualization. I think the true data must be provided by others, not the author. because the author would fall into the "inertial thinking".

djhoese:
Write the tests for what is being fixed. The visualization is harder to write, harder to test, is slower, and is overall less accurate as we'll be testing a large portion of other components of vispy, the GPU on the system it is being run on, and the OpenGL drivers. If we compare to an exact image then we have to store that image on vispy's test data repository so it can be compared against. If we compare to an expected rendered image result (like an 8-bit unsigned integer of what the rendered result will look like) then it is confusing as to what exactly is being tested and what exactly we are looking for ("we know the test is supposed to produce this resulting array, but we don't know what specifically about the image is being expected"). This also couples the test to the MeshVisual. Changes to the MeshVisual now require changes to this test when at the end of the day, as far as these geometry function tests are concerned, we don't care what the MeshVisual does with them, we care that they are producing valid and correct numbers. If that isn't possible without using MeshData because of the normals calculations it is doing or without actually drawing the Mesh then fine. For the tests for these geometry generation functions, we are testing the function, not the visual. If we need to add more tests so we have tests for the visual **and** for the numbers returned by these functions, then go for it.

> I think the true data must be provided by others, not the author. because the author would fall into the "inertial thinking".

That is kind of what we're doing here. A user (you, @asnt, and others) have pointed out that there is an issue. We take your example code and write a test for it. The test fails with the current code, but passes with the fixes. Now we know the code/implementation is **better** (maybe not perfect). We can either add more test data cases until we are more confident that we've caught all the edge cases or we wait for users to report more issues in the future. In the future if we get more reports, then we add more tests for any new bugs that show up. How does 

I'm not sure how checking the visualization prevents this scenario from happening. We can fix what we know is wrong and check that it is right or at least better. Whether we check that the visualization is right or the numbers that lead to the visualization are right doesn't really matter as far as making sure things are fixed. The point is that visualization tests are harder to write and less accurate for something like this and in my opinion not necessary here. If you want to add a test where you compare the visualization fine, but please also add a test that checks a few of the generated numbers to make sure they make sense.

yxdragon:
ok, I know. data test is for the CI. (I did not understand it befor), but it is hard to build the true data set.
And visualization demo is a good way to show it's right (at least not obviously wrong) for human directly.
So there is no conflict between data test and visualization test.

almarklein:
As for the shading, in the fragment shader, we should be able to flip normals towards the view direction (in `filters//mesh.py`) using:
```glsl
normal = -faceforward(-normal, normal, eye_vec);
```

Related issue:  #2087

djhoese:
@yxdragon Do you have time to add tests for this?

yxdragon:
sorry, a little busy last week, and did not see github. I would review it soon.

yxdragon:
what should I do next?

djhoese:
I'm not sure actually. @asnt started adding tests in #2098 but he hasn't had time to finish it. His PR is meant to only include the bug fixes I think and not necessarily the optimizations. If you have the time @yxdragon maybe you could review #2098. I'm not sure I followed the discussion on this stuff so I'm not sure if I can help move this forward. We may need to wait until @asnt has some free time to work on vispy again.

djhoese:
It is my understanding that #2434 (and #2433) have now implemented this functionality and include tests. All people involved, please review this and let me know if I'm missing something. My understanding is that there were optimizations and bug fixes that needed to be implemented, but it isn't clear to me if the bug fixes made it into #2434. @asnt feel free to comment here or in #2098.

asnt:
@djhoese AFAICT, the remaining part from this PR is the fixes for the generation of geometries, i.e. point 2 in the original description of this PR:

> 2. modify geometry. create_grid_mesh, create_plane, upspeed and solve the normal dismatch problem.
(then the create_box and create_sphere is ok from create_plane)

I would like to have a look at this in the coming future. I'll reuse the individual suggested commits if feasible.  From my point of view, we can keep this PR closed.

Issue #2064, opened 2021-06-10
question about volume render 's coordinate axis


yxdragon:
![image](https://user-images.githubusercontent.com/24822467/121571519-b6a17e80-ca55-11eb-8aa1-33fe7e90aadb.png)

I draw a volume, a isosurface, and a box by a same data. (the demo data stent)

I found the volume does not match with box and isosurface. 
(and the volume can not auto center and auto fit the scene)

So I transpose the image befor volume render. img.transpose(2,1,0), then it matched
![image](https://user-images.githubusercontent.com/24822467/121572926-409e1700-ca57-11eb-9508-1afe0710689a.png)
but can not auto center.

then modify volume's _compute_bounds method
```python
def _compute_bounds(self, axis, view):
        return 0, self._vol_shape[::-1][axis]
```
![image](https://user-images.githubusercontent.com/24822467/121574117-85767d80-ca58-11eb-9067-0a4ff1debe5e.png)
now it is shown in the center. every thing is OK.

djhoese:
It is hard to know if this is the right or best fix without seeing your actual code used to generate the original visualization. Can you provide a small example that doesn't use any external application like ImagePy?

yxdragon:
```python
from vispy import scene
import numpy as np
import scipy.ndimage as ndimg
from vispy.visuals.transforms import STTransform
from vispy.geometry import create_sphere

canvas = scene.SceneCanvas(keys='interactive', bgcolor='white', size=(800, 600), show=True)
view = canvas.central_widget.add_view()

imgs = np.zeros((100,200,300), dtype=np.float32)
imgs.ravel()[np.random.randint(0,100*200*300,1000)]=1
imgs = ndimg.gaussian_filter(imgs, 3)
scene.visuals.Volume(imgs, cmap='hot', parent=view.scene)
scene.visuals.Isosurface(imgs, level=0.001, shading='smooth', parent=view.scene)
#sphere1._mesh.mesh_data_changed()
cam = scene.TurntableCamera(elevation=30, azimuth=30)
view.camera = cam

if __name__ == '__main__':
    canvas.app.run()
```
![image](https://user-images.githubusercontent.com/24822467/121627168-710ea100-caa9-11eb-80d7-6100a7771acc.png)

[...]

PR #2070, opened 2021-06-14
Fix VolumeVisual bounds representing the wrong axis

djhoese:
This is a partial fix for #2064 (CC @yxdragon).

The issue is that the VolumeVisual assumes data is in the shape of `(z, y, x)`, but VisPy seems to assume that axis (0, 1, 2) -> (x, y, z). We can see this in the `ImageVisual` which passes data as `(y, x)` and in its `_compute_bounds` method it uses `self.size` which is the reversed shape of the array. So the bounds are computed accurately. For the VolumeVisual, we need to take these values in reverse as well (but aren't currently). This PR does this by using in a way so if a VolumeVisual ever has more than 3 dimensions, the last 3 dimensions are used. I'm not sure that is a bad assumption or not bug :shrug: 

Also, I've added a `self._is_zyx` property to the `VolumeVisual` so that in the future the rest of #2064 can be resolved in which a user can pass data with `(x, y, z)` (matching `MeshVisual`/`Isosurface`) and still get the correct bounds.

Running this code:

```python
from vispy import scene
import numpy as np
import scipy.ndimage as ndimg

imgs = np.zeros((100,200,300), dtype=np.float32)
imgs.ravel()[np.random.randint(0,100*200*300,1000)]=1
imgs = ndimg.gaussian_filter(imgs, 3)
scene.visuals.Volume(imgs, cmap='hot', parent=view.scene)

if __name__ == '__main__':
    canvas.app.run()
```

Without this PR produces (not centered on the Visual):

image

With this PR produces (centered on the Visual):

image

### DIFF
--- a/vispy/visuals/tests/test_volume.py
+++ b/vispy/visuals/tests/test_volume.py
@@ -47,6 +47,18 @@ def test_volume():
 
 
 @requires_pyopengl()
+def test_volume_bounds():
+    vol = np.zeros((20, 30, 40), 'float32')
+    vol[8:16, 8:16, :] = 1.0
+
+    # Create
+    V = scene.visuals.Volume(vol)
+    assert V._compute_bounds(0, V) == (0, 40)  # x
+    assert V._compute_bounds(1, V) == (0, 30)  # y
+    assert V._compute_bounds(2, V) == (0, 20)  # z
+
+
+@requires_pyopengl()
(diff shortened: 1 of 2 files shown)


almarklein (approved):
Awesome! Weird how this bug has been unnoticed for so long ...

PR #2072, opened 2021-06-15
Set default log level to warning

### LINKED ISSUE #1790: default logging_level
https://github.com/vispy/vispy/blob/57ebceb0f4f01938cc84e398829bcafc23fb5d0a/vispy/util/config.py#L74

In developing a workaround for https://github.com/vispy/vispy/issues/1787 I found that the logging level was set to Info.

Is the desired behaviour to log Info???
Would it be possible to leave it at the previous state if the user doesn't specify it in their config?

almarklein:
Closes #1790 ... from dec 2019 :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ node_modules/
 vispy/static/
 vispy/version.py
 .vscode
+vispy/visuals/text/_sdf_cpu*.pyd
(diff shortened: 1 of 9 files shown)


djhoese:
One thing brought up by @hmaarrfk was:

> Would it be possible to leave it at the previous state if the user doesn't specify it in their config?

Does this mean that vispy is changing the level even if the user has already set it?

Also, this PR should probably update the existing `INFO` messages I mentioned in #1790 related to the backends (failed backend loading and switching backend) to be warnings (either `warnings.warn` or log warning level).

djhoese (approved):

PR #2074, opened 2021-06-15
Rendering arbitrary planes from 3D textures in the `VolumeVisual`

alisterburt:
Here is a work-in-progress implementation of arbitrary plane rendering in the `VolumeVisual`.

https://user-images.githubusercontent.com/7307488/122108319-f8467680-ce13-11eb-8889-5fab58fd9fd3.mp4

there is an example you can play with called `volume_plane_playground.py`

## Current behaviour

Currently, the `VolumeVisual` casts a ray through a volume and performs some computation on the values encountered. These different calculations are implemented as 'snippets', inserted into the fragment shader template by string formatting. The start and endpoints of this raycasting are the intersections of the view ray with the bounding box of the volume.

## How to render an arbitrary plane?

To render arbitrary planes with a 'thickness', I intersect the view ray with a plane in the volume and sample the texture above and below that plane (along the plane normal vector)

key points:

- A plane in the fragment shader is defined by a position in the volume `u_plane_position`, a normal vector `u_plane_normal` and a thickness `u_plane_thickness`
- the setup for raycasting in the fragment shader can now be templated, name is `{raycasting_setup}`
- I add a template for this alternative raycasting mode

other things modified to make this work:

- expose new shader variables on the visual: `mode`, `plane_position`, `plane_normal` & `plane_thickness`
- FRAG_DICT keys are now a tuple of (`raycasting_mode`, `rendering_method`)
- FRAG_DICT is now contained in `VolumeVisual.fragment_shaders` and contains shaders for all rendering methods in both raycasting modes

## Things I would explicitly like some help with 😅 

- depth

Currently the depth buffer is not set explicitly in the `VolumeVisual`. Based on [this page](https://learnopengl.com/Advanced-OpenGL/Depth-testing) I think the value is constant across all non-discarded fragments (depicted - `gl_FragCoord.z`).
image
Can/should I set `gl_FragCoord.z`, or should I [set the `gl_FragDepth` buffer directly](https://www.labri.fr/perso/nrougier/python-opengl/#spheres)? (slow?) I'm not sure how to properly project the z position in data/texture coordinates into what is expected here by other visuals in the same scene (my motivation is consistency with other visuals e.g. markers, I want the possibility of having some markers in front and some behind the plane)

- invalid variables

https://github.com/vispy/vispy/blob/91972307cf336674aad58198fb26b9e46f8f9ca1/vispy/visuals/volume.py#L788-L792
the section of code here makes me think I should add some lines for my plane variables but I don't really understand why they're needed/what they're doing

I would definitely appreciate any and all general advice here as this is my first real attempt to contribute to vispy and I'm very much playing with openGL for the first time, thanks for taking the time to read!


asnt:
@alisterburt That looks great! I cannot comment on the algorithm as I'm not familiar with volume rendering but I can try to give a few hints for opengl/vispy.

> Can/should I set `gl_FragCoord.z`, or should I [set the `gl_FragDepth` buffer directly](https://www.labri.fr/perso/nrougier/python-opengl/#spheres)? (slow?)

I think you have to set `gl_FragDepth`, as you suggest, as `gl_FragCoord` is read-only.
[Fragment outputs](https://www.khronos.org/opengl/wiki/Fragment_Shader/Defined_Outputs)
[gl_FragDepth](https:/www.khronos.org/registry/OpenGL-Refpages/es3.1/html/gl_FragDepth.xhtml)
[gl_FragCoord](https://www.khronos.org/registry/OpenGL-Refpages/es3.1/html/gl_FragCoord.xhtml)

Writing to `gl_FragDepth` disables early depth testing, which otherwise improves performance by skipping the fragments that are not visible (see [depth test](https://www.khronos.org/opengl/wiki/Depth_Test)). But, is depth testing enabled at all in the volume visual? If not, then it shouldn't matter. (From a visual in VisPy, depth test is usually disabled with `self.set_gl_state(depth_test=False)`.) But in any case, I think this shouldn't be the first worry, the performance might still be acceptable as a first solution.

> I'm not sure how to properly project the z position in data/texture coordinates into what is expected here by other visuals in the same scene (my motivation is consistency with other visuals e.g. markers, I want the possibility of having some markers in front and some behind the plane)

Could [this comment](https://github.com/vispy/vispy/blob/91972307cf336674aad58198fb26b9e46f8f9ca1/vispy/visuals/volume.py#L264) mentioned in another issue recently be an indication? @almarklein 

> * invalid variables
>   https://github.com/vispy/vispy/blob/91972307cf336674aad58198fb26b9e46f8f9ca1/vispy/visuals/volume.py#L788-L792
>       
>   the section of code here makes me think I should add some lines for my plane variables but I don't really understand why they're needed/what they're doing

It seems like clearing variables that are specific to some rendering methods (and associated glsl snippets) when switching between methods. I would try to go on without worrying about it, and come back to it in a second step if needed. If you get errors or warnings about missing/unset/invalid variables in the shaders, that might be the place to fix them.

almarklein:
You indeed would need to write to gl_FragDepth. I think the main performance cost is that the shader will now have to write to both the color buffer and the depth buffer, but it should not matter too much, and correctness is more important, I think.

The depth testing does not have to be enabled for the volume visual to work, it should be enabled for the visuals that draw *afterwards* so that they take into account the depth set by the volume.

> Could this comment mentioned in another issue recently be an indication?

That code was from visvis, so we'd need to adapt it for Vispy. Something with the position of the start-plane that the ray originates from, offet by `n` steps :)

almarklein:
Very nice addition. We have something similar [in pygfx](https://github.com/pygfx/pygfx/blob/main/pygfx/materials/_volume.py#L53), but we simply sample - I like the idea of defining a slice thinness to apply some regularization!

The code is looking good to me so far. Some comments on a higher level:
* Instead of extending the functionality of the VolumeVisual, would it make sense to call this e.g. a VolumeSlice visual?
* We could define the plane as a normal plus a position, or as a plane with abcd, which is basically the normal plus a scalar value to shift away from the origin.
* Thinking of the thickness measure, you could also use a Gaussian distribution to combine values of multiple slices ... but that may be a fancy feature for later.
* Question: is the slice sampled in the view direction or perpendicular to the slice?

alisterburt:
Thanks for your input @asnt and @almarklein - the links and explanations are super useful, nice to see the other implementation @almarklein !

@almarklein - I currently have the intersection of the `view_ray` and the plane in both data coordinates and texture coordinates here
https://github.com/vispy/vispy/blob/bad240d70e9cc2bc9c19bdbdf4e0dd476c172764/vispy/visuals/volume.py#L312-L316

I'm not sure which transform I need to apply to this to get the position within the whole scene (world coordinates?), which is what should be used to determine the relative depth of different objects? Maybe the inverse of the view transform

I'm also not sure about what to use to do the normalisation from 0-1 to be consistent with other visuals, maybe [equation 2 in this link](https://learnopengl.com/Advanced-OpenGL/Depth-testing)

> Instead of extending the functionality of the VolumeVisual, would it make sense to call this e.g. a VolumeSlice visual?

I had the same thought - the reason I went this way is that there are cases where you may want to switch between volumetric and plane rendering, this would probably mean that switching between them would need data reupload to the GPU? I'm happy to implement as a separate visual if you feel strongly that having a separate visual is better - we don't want a huge dynamically generated mess

> We could define the plane as a normal plus a position, or as a plane with abcd, which is basically the normal plus a scalar value to shift away from the origin.

Interesting - I didn't think about that. Longer term I have some other things I would like to do with the volume visual where having a position (without a normal) would be useful (localised rendering around a point) - for that reason I would prefer to keep the variables separate but again, happy to accommodate changes 🙂 

> Thinking of the thickness measure, you could also use a Gaussian distribution to combine values of multiple slices ... but that may be a fancy feature for later.

Great idea, love it, a good thing to keep in mind moving forward. I would have to change the raycasting for this a little bit but it shouldn't be too hard - I will leave it for a followup PR

> Question: is the slice sampled in the view direction or perpendicular to the slice?

The texture is sampled perpendicular to the slice, starting from `u_plane_thickness` / 2 away along the plane normal. The step vector then steps along the plane normal until it reaches the same distance on the other side of the slice (what is rendered on the slice does not depend on the view direction) - it happens here
https://github.com/alisterburt/vispy/blob/bad240d70e9cc2bc9c19bdbdf4e0dd476c172764/vispy/visuals/volume.py#L338-L342

Thanks for taking a look!

almarklein:
> I'm not sure which transform I need to apply to this to get the position within the whole scene (world coordinates?), which is what should be used to determine the relative depth of different objects? Maybe the inverse of the view transform

I was looking at the [pygfx shader implementation](https://github.com/pygfx/pygfx/blob/main/pygfx/renderers/wgpu/volumerender.py#L86) for inspiration, but there we do the plane intersection logic in the vertex shader, so the plane is already at the proper depth :P 

Ok, I don't know the solution directly, but the general idea is this: since the ray is always in the view-direction, the depth to write (a value between -1..1 (or is it 0..1?)), that value is based on the depth of the starting plane, offset by a value directly proportional to the number of steps. Maybe in the vertex shader we can prepare a varying containing the base-depth and the per-step depth or something?  If you prefer, this is something that I could take on in a separate PR.

almarklein:
>I had the same thought - the reason I went this way is that there are cases where you may want to switch between volumetric and plane rendering, this would probably mean that switching between them would need data reupload to the GPU?

Right ... would be nice if we were able to shader data between visuals, but we can't, right?

> Interesting - I didn't think about that. Longer term I have some other things I would like to do with the volume visual where having a position (without a normal) would be useful (localised rendering around a point) - for that reason I would prefer to keep the variables separate but again, happy to accommodate changes 🙂

Makes sense!

alisterburt:
> I was looking at the [pygfx shader implementation](https://github.com/pygfx/pygfx/blob/main/pygfx/renderers/wgpu/volumerender.py#L86) for inspiration, but there we do the plane intersection logic in the vertex shader, so the plane is already at the proper depth :P

smart! 

> Ok, I don't know the solution directly, but the general idea is this: since the ray is always in the view-direction, the depth to write (a value between -1..1 (or is it 0..1?)), that value is based on the depth of the starting plane, offset by a value directly proportional to the number of steps. Maybe in the vertex shader we can prepare a varying containing the base-depth and the per-step depth or something? If you prefer, this is something that I could take on in a separate PR.

This seems like a great approach and I definitely wouldn't have thought of it - I would have to spend time building a bit more understanding (how varyings are interpolated is a bit fuzzy) to attempt it so would be happy if you took this on 👍🏼 no rush from my side of course. I think your 'per-step depth' would come from the scale factor in my plane intersection function

https://github.com/alisterburt/vispy/blob/bad240d70e9cc2bc9c19bdbdf4e0dd476c172764/vispy/visuals/volume.py#L210-L225

So I might need to expose that in the fragment shader for you to be able to use it with this

djhoese:
I tried updating your original description so github would render the section titles correctly...it doesn't like your markdown for some reason.

Anyway, regarding unsetting those attributes for the shader, I believe this is necessary to avoid complaints/errors/warnings from certain GLSL compilers. I've seen some systems not complain about certain things, but others do complain. VisPy also has some specific checks for uniforms that are set but unused in the shader.

https://github.com/vispy/vispy/blob/df09b3079d71b7d1942481a051b900f145c4cfee/vispy/gloo/glir.py#L1204-L1205

So bottom line is that this is an effort to avoid logging these messages or causing other related errors.

Also note that I merged my #1912 PR today so it is causing some conflicts in this PR. Most of it is likely related to the keyword arguments that I've deprecated in `__init__`.

almarklein:
> > I was looking at the pygfx shader implementation for inspiration, but there we do the plane intersection logic in the vertex shader, so the plane is already at the proper depth :P
>
> smart!

Not so sure. I mostly did it this way because we have not implemented volume rendering yet, so it did not occur to me to try it the way you did. The vertex shader is quite complex (calculating intersections with the 6 planes of the cube). Your approach might actually be simpler.

alisterburt:
superseded by #2149

PR #2079, opened 2021-06-17
Fix PanZoomCamera 'center' property not updating view

djhoese:
Replaces #1751. I tried to rebase that PR on to `main` and git (or rather pycharm) warned me that it wasn't going to end well. I instead created a new branch off of `main` and cherry-picked it since it was a single commit. I've then modified how the `DEFAULT_RECT_TUPLE` was defined as flake8 was pointing out that it isn't actually available in the internal method scope.

Now to write a test... :fearful:

### DIFF
--- a/vispy/scene/cameras/panzoom.py
+++ b/vispy/scene/cameras/panzoom.py
@@ -43,9 +43,11 @@ class PanZoomCamera(BaseCamera):
 
     """
 
+    DEFAULT_RECT_TUPLE = (0, 0, 1, 1)
+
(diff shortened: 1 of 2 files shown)


djhoese:
Well it's good I'm adding a test, because it turns out the `.center` setter never actually caused the camera to update the transforms. If it uses the existing `self.rect` and then later does `self.rect = rect` that assignment is seen as a no-op because they are already equal and no view_change is sent out. Will try to triple check my work and get these tests added before the end of the day.

almarklein (approved):

PR #2040, opened 2021-05-28
Prevent translucent window with QOpenGLWidget

asnt:
Fix #1889

Fixes it for me on Linux with wayland.

PR #2082, opened 2021-06-18
Close some GL leaks

almarklein:
* Make flake8 ignore `N` in case the pendantic "pep8-naming" is installed.
* Check for imports of `OpenGL`:
  * Removes use of `OpenGL` in `visuals/line/line.py`.
  * Remove use of `OpenGL` in  `backends/_tk.py`.
  * No other usages.
* Check for usage of `gl` in visuals and above:
  * It is used in path collections to disable the depth buffer. Seems hard to fix, so added a comment for now.
  * No other usages.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [flake8]
 max-line-length = 120
-ignore = E226,W291,W293,W503,F999,E305,F405,W504
+ignore = E226,W291,W293,W503,F999,E305,F405,W504,N
 exclude =
     glfw.py,_proxy.py,_es2.py,_gl2.py,_pyopengl2.py,
     _constants.py,png.py,ipy_inputhook.py,
--- a/vispy/visuals/collections/agg_fast_path_collection.py
+++ b/vispy/visuals/collections/agg_fast_path_collection.py
@@ -13,6 +13,7 @@ on miter joins which may result in some glitches on screen.
 """
 import numpy as np
 from ... import glsl
+from ... import gloo
 from ...gloo import gl
 from . collection import Collection
 from ..transforms import NullTransform
@@ -213,7 +214,6 @@ class AggFastPathCollection(Collection):
 
     def draw(self, mode="triangle_strip"):
         """Draw collection"""
-        # Would call context.set_depth_mask() here, but we have no access
-        gl.glDepthMask(0)
+        gloo.set_depth_mask(0)
         Collection.draw(self, mode)
-        gl.glDepthMask(1)
+        gloo.set_depth_mask(1)
--- a/vispy/visuals/collections/agg_fast_path_collection.py
+++ b/vispy/visuals/collections/agg_fast_path_collection.py
@@ -76,7 +76,7 @@ class AggFastPathCollection(Collection):
             vertex = glsl.get('collections/agg-fast-path.vert')
         if transform is None:
             transform = NullTransform()
-        self.transform = transform        
+        self.transform = transform
         if fragment is None:
             fragment = glsl.get('collections/agg-fast-path.frag')
 
@@ -213,6 +213,7 @@ class AggFastPathCollection(Collection):
 
     def draw(self, mode="triangle_strip"):
         """Draw collection"""
-        gl.glDepthMask(gl.GL_FALSE)
+        # Would call context.set_depth_mask() here, but we have no access
+        gl.glDepthMask(0)
         Collection.draw(self, mode)
-        gl.glDepthMask(gl.GL_TRUE)
+        gl.glDepthMask(1)
--- a/vispy/visuals/collections/agg_fast_path_collection.py
+++ b/vispy/visuals/collections/agg_fast_path_collection.py
@@ -14,7 +14,6 @@ on miter joins which may result in some glitches on screen.
 import numpy as np
 from ... import glsl
 from ... import gloo
-from ...gloo import gl
 from . collection import Collection
 from ..transforms import NullTransform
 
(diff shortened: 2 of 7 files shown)


almarklein:
@djhoese I think you mentioned leakage of `gloo` usage. Do you mean in the ones in scene?

* `scene/canvas.py` to obtain an FBO for taking screenshots.
* `scene/widgets/console.py` to get a VBO.

vispy/visuals/line/line.py
```diff
@@ -340,25 +341,11 @@ def _prepare_draw(self, view):
...
-                pass
-
-        # Turn on line smooth and/or line width
-        if GL:
-            if self._parent._antialias:
-                GL.glEnable(GL.GL_LINE_SMOOTH)
-            else:
-                GL.glDisable(GL.GL_LINE_SMOOTH)
+        if "webgl" not in default_app_module.default_app.backend_name:
+            self.update_gl_state(line_smooth=bool(self._parent._antialias))
```
djhoese, line 345: I honestly didn't even know the `update_` variant of this function existed. I've only ever seen `set_gl_state` used. Cool.

vispy/visuals/collections/agg_fast_path_collection.py
```diff
@@ -213,6 +213,7 @@ def bake(self, P, key='curr', closed=False, itemsize=None):
 
     def draw(self, mode="triangle_strip"):
         """Draw collection"""
-        gl.glDepthMask(gl.GL_FALSE)
+        # Would call context.set_depth_mask() here, but we have no access
```
djhoese, line 216: Do we not have access to it because this isn't a "real" Visual with access to those `set_gl_state`-like methods?

djhoese: Don't we have access (or can't we get it) by using the gloo functions that is in the `wrappers.py` you updated in this PR?

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
 max-line-length = 120
-ignore = E226,W291,W293,W503,F999,E305,F405,W504
+ignore = E226,W291,W293,W503,F999,E305,F405,W504,N
```
djhoese, line 3: I didn't know pep8-naming existed. Glad it can be ignored.

djhoese (approved):
I have a couple questions and comments, but otherwise this looks pretty good as-is. As far as the scene-based leakage, yes those seem relevant. I think some of these can be fixed by moving their usage lower into maybe the "raw" (I guess this would be the gloo-specific one in the future) Canvas and naming the method something generic enough to get the same result. For example, we don't currently have a "screenshot" method on the canvas (or some equivalent), but this would be nice to have and if named generically enough then any graphics backend could implement it.

Actually, I was hoping to get that into 0.7. I guess I'll have to make a PR today.

vispy/visuals/collections/agg_fast_path_collection.py
almarklein: I thought we could not access it, but indeed we can  just call `gloo.set_depth_mask()` :)

almarklein:
> As far as the scene-based leakage, yes those seem relevant. I think some of these can be fixed by moving their usage lower into maybe the "raw" (I guess this would be the gloo-specific one in the future) Canvas and naming the method something generic enough to get the same result. 

I just looked at that. There are some scene-specific calls in `SceneCanvas.render()` though. We could still factor out the gloo-specifics, but there are still functions like `push_fbo` (and maybe others) that do not explicitly create/use gloo objects, but they do make assumptions about some stuff being gloo objects (FBO's in this example). Maybe we can wait with that until we have a better idea what a new API will look like and how it will interact with Vispy. What do you think?

djhoese:
> I just looked at that. There are some scene-specific calls in SceneCanvas.render() though. We could still factor out the gloo-specifics, but there are still functions like push_fbo (and maybe others) that do not explicitly create/use gloo objects, but they do make assumptions about some stuff being gloo objects (FBO's in this example). Maybe we can wait with that until we have a better idea what a new API will look like and how it will interact with Vispy. What do you think?

I don't think the FBO stuff has to be modified/extracted itself, just defining the purpose of the method. For example, the `.render` method can be the defined interface for any graphics backend Canvas class. I was going to work on a new `screenshot` method PR, but realized I didn't like there was a `.render` method and then a `.screenshot` method. The `.render` draws to the new FBO, while the `.screenshot` is based on the `_screenshot` utility which uses `glReadPixels`. You can see some issues in `.render` for certain visualizations: https://github.com/napari/napari/issues/1566

I was going to try to do this for 0.7, but maybe this screenshot discussion needs to wait for 0.8. Or maybe I update `.render` to accept another keyword argument to use `glReadPixels` as an alternative.

almarklein:
The FBO approach uses `read_pixels` too, but on a custom framebuffer instead of the one attached to the screen. I guess in the latter case the alpha channel is "filled". Feels a bit similar to #2040. 

I wonder if the weird transparency would be gone if one simply dropped the alpha channel?

djhoese:
Theoretically `.render` should behave the same (same end result) as `_screenshot`.

Do you mean dropping the alpha channel for the whole result? In the napari example (the original issue screenshots), the original visualization doesn't even have any fully transparent pieces as far as I can tell and remember. Also see this comment: https://github.com/napari/napari/issues/1566#issuecomment-690748249. The problem was that it seemed like the transparency of the foreground layer was setting the transparency for the whole buffer instead of the completely opaque background image's alpha channel being used.

djhoese (approved):
LGTM. Feel free to merge if you are satisfied with it @almarklein

almarklein:
Merged. We should probably continue that screenshot discussion on another issue :D  but ...

> [...]  it seemed like the transparency of the foreground layer was setting the transparency for the whole buffer

that is very strange indeed. Maybe look into the blending rules...

PR #2084, opened 2021-06-19
Document OpenGL state presets

asnt:
- Clarify which attributes of the state are modified by each preset in the base `vispy.gloo.wrapper.set_state()` function.
- Add a link to `set_state()` in the higher-level `(Compound)Visual.(set|update)_gl_state()`.

### DIFF
--- a/vispy/gloo/wrappers.py
+++ b/vispy/gloo/wrappers.py
@@ -433,7 +433,7 @@ class BaseGlooFunctions(object):
         Returns
         -------
         presets : dict
-            The dictionary of presets usable with ``set_options``.
+            The dictionary of presets usable with :meth:`~set_state`.
         """
         return deepcopy(_gl_presets)
 
--- a/vispy/gloo/wrappers.py
+++ b/vispy/gloo/wrappers.py
@@ -22,14 +22,18 @@ __all__ = ('set_viewport', 'set_depth_range', 'set_front_face',  # noqa
            'get_state_presets', 'set_state', 'finish', 'flush',  # noqa
            'read_pixels', 'set_hint',  # noqa
            'get_gl_configuration', '_check_valid',
+           'GL_PRESETS',
            'GlooFunctions', 'global_gloo_functions', )
 
 _setters = [s[4:] for s in __all__
             if s.startswith('set_') and s != 'set_state']
 
 # NOTE: If these are updated to have things beyond glEnable/glBlendFunc
-# calls, set_preset_state will need to be updated to deal with it.
-_gl_presets = {
+# calls, set_state will need to be updated to deal with it.
+#: Some OpenGL state presets for common use cases.
+#:
+#: To be used in :func:`.set_state`.
+GL_PRESETS = {
     'opaque': dict(
         depth_test=True,
         cull_face=False,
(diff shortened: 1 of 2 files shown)


djhoese:
Looks good. Any idea if it is possible to do the API links to a "constant" in a module and have documentation/docstring for that? That way they are less likely to get out of sync.

Thanks for doing this by the way.

vispy/gloo/wrappers.py
```diff
@@ -433,12 +433,18 @@ def get_state_presets(self):
         Returns
         -------
         presets : dict
-            The dictionary of presets usable with ``set_options``.
+            The dictionary of presets usable with :meth:`~set_state`.
```
djhoese, line 436: Did you try building the docs? Does this actually work? I would have assumed you needed `~vispy.gloo.wrappers.BaseGlooFunctions.set_state`

asnt: For non-explicit names, Sphinx looks up in the current module. See [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-roles) after the description of the roles.

But in this case '~' is not needed and it must be `:func:` instead of `:meth:` because the methods of `BaseGlooFunctions` are injected in and only visible in the global scope later on.

asnt: And ```:func:`.set_state````referrs explicitly to this module's objects.

asnt:
> Looks good. Any idea if it is possible to do the API links to a "constant" in a module and have documentation/docstring for that? That way they are less likely to get out of sync.

I agree with this but I haven't managed yet.
I think that we have to make `_gl_presets` visible (not private or explicitly say sphinx to include it).
[This](https://stackoverflow.com/q/40748886) suggests we might need an [`:autodata:`](https://www.sphinx-doc.org/en/1.4.8/ext/autodoc.html#directive-autodata) along with the default `:automodule:` generated by `sphinx-apidoc`.
[This](https://stackoverflow.com/q/40748886) suggests that it could be possible to modify the apidoc templates to add `:autodata:`.
[Default apidoc templates](https://github.com/sphinx-doc/sphinx/tree/4.x/sphinx/templates/apidoc).

djhoese:
Yeah if we end up needing to do something that either effects all modules (always show data elements) or requires custom templates...maybe it is best to just keep it like you have it.

With that in mind, is this PR ready?

djhoese:
Do you have a screenshot of what the final sphinx page looks like for this?

asnt:
@djhoese In 82e1364, dict of presets is documented by sphinx at the module level. As far as I can tell from my tests, it has to be non-private and added to `__all__`. I commited to have a preview of it. Let me know if that sounds OK.

asnt:
> Do you have a screenshot of what the final sphinx page looks like for this?

The constant `GL_PRESETS` is at the top and the `set_state()` is highlighted somewhere in the middle:

image

djhoese:
Cool. If it isn't possible to have sphinx cleanup the formatting of that dictionary, maybe you could list the possible presets in the "docstring" for it? Not what they map to, but just the names that the user can enter.

vispy/gloo/wrappers.py
```diff
@@ -428,29 +432,29 @@ def set_sample_coverage(self, value=1.0, invert=False):
...
     def set_state(self, preset=None, **kwargs):
-        """Set OpenGL rendering state, optionally using a preset
+        """Set the OpenGL rendering state, optionally using a preset.
 
         Parameters
         ----------
-        preset : str | None
-            Can be one of ('opaque', 'translucent', 'additive') to use
-            use reasonable defaults for these typical use cases.
+        preset : {None, 'opaque', 'translucent', 'additive'}
```
djhoese, line 449: Sorry, I meant in the docstring for the `GL_PRESETS` constant. I think technically everything after the `: ` in a numpy-style docstring like this has to be a type (`str | None`).

asnt:
> Cool. If it isn't possible to have sphinx cleanup the formatting of that dictionary, maybe you could list the possible presets in the "docstring" for it? Not what they map to, but just the names that the user can enter.

@djhoese Indeed, that makes it clearer. Added.

That should be it for this PR regarding the scope of the changes, I think. Let me know if you have other remarks.

vispy/gloo/wrappers.py
asnt: Ah, OK. I think I got it now. I'll do that.

And you suggest resetting to `preset : str | None` in `set_state()` then?

asnt: For the docstrings, I don't know. I was following [this source](https://numpydoc.readthedocs.io/en/latest/format.html#parameters), which says

```
When a parameter can only assume one of a fixed set of values, those values can be listed in braces, with the default appearing first:

order : {'C', 'F', 'A'}
   Description of `order`.
```

djhoese: You're right. I wasn't sure, but that example is exactly like our case so :+1:

djhoese (approved):

PR #2085, opened 2021-06-19
Improve infinity/NaN handling in VolumeVisual and ImageVisual clim calculations

almarklein:
* The logic in the shader is simplified and has no branches (in the clim computation).
* Deal with nan and inf in data. Uses @larsoner's idea to stay performant by default at the cost of being a bit extra less performant when non-finite values are present :)
* Correct behavior when clim range is zero.

### DIFF
--- a/vispy/plot/plotwidget.py
+++ b/vispy/plot/plotwidget.py
@@ -202,7 +202,7 @@ class PlotWidget(scene.Widget):
         fg_color : Color or None
             Sets the plot foreground color if specified.
         kwargs : keyword arguments.
-            More args to pass to visual.Volume.
+            More args to pass to :class:`~vispy.visuals.image.Image`.
 
         Returns
         -------
@@ -412,7 +412,7 @@ class PlotWidget(scene.Widget):
         cmap : str
             The colormap to use.
         kwargs : keyword arguments.
-            More args to pass to visual.Volume.
+            More args to pass to :class:`~vispy.visuals.volume.Volume`.
 
         Returns
         -------
--- a/vispy/plot/plotwidget.py
+++ b/vispy/plot/plotwidget.py
@@ -187,7 +187,7 @@ class PlotWidget(scene.Widget):
         self.view.camera.set_range()
         return hist
 
-    def image(self, data, cmap='cubehelix', clim='auto', fg_color=None):
+    def image(self, data, cmap='cubehelix', clim='auto', fg_color=None, **kwargs):
         """Show an image
 
         Parameters
@@ -201,6 +201,8 @@ class PlotWidget(scene.Widget):
             min and max values.
         fg_color : Color or None
             Sets the plot foreground color if specified.
+        kwargs : keyword arguments.
+            More args to pass to visual.Volume.
 
         Returns
         -------
@@ -212,7 +214,7 @@ class PlotWidget(scene.Widget):
         The colormap is only used if the image pixels are scalars.
         """
         self._configure_2d(fg_color)
-        image = scene.Image(data, cmap=cmap, clim=clim)
+        image = scene.Image(data, cmap=cmap, clim=clim, **kwargs)
         self.view.add(image)
         self.view.camera.aspect = 1
         self.view.camera.set_range()
@@ -390,7 +392,7 @@ class PlotWidget(scene.Widget):
         return spec
 
     def volume(self, vol, clim=None, method='mip', threshold=None,
-               cmap='grays'):
+               cmap='grays', **kwargs):
         """Show a 3D volume
 
         Parameters
@@ -409,6 +411,8 @@ class PlotWidget(scene.Widget):
             the mean of the given volume is used.
         cmap : str
             The colormap to use.
+        kwargs : keyword arguments.
+            More args to pass to visual.Volume.
 
         Returns
         -------
@@ -420,7 +424,7 @@ class PlotWidget(scene.Widget):
         Volume
         """
         self._configure_3d()
-        volume = scene.Volume(vol, clim, method, threshold, cmap=cmap)
+        volume = scene.Volume(vol, clim, method, threshold, cmap=cmap, **kwargs)
         self.view.add(volume)
         self.view.camera.set_range()
         return volume
--- a/vispy/visuals/_scalable_textures.py
+++ b/vispy/visuals/_scalable_textures.py
@@ -10,7 +10,7 @@ from vispy.gloo.texture import should_cast_to_f32
 
 
 def get_default_clim_from_dtype(dtype):
-    """Get min and max color limits based on the rsange of the dtype."""
+    """Get min and max color limits based on the range of the dtype."""
     # assume floating point data is pre-normalized to 0 and 1
     if np.issubdtype(dtype, np.floating):
         return 0, 1
@@ -22,11 +22,11 @@ def get_default_clim_from_dtype(dtype):
 
 
 def get_default_clim_from_data(data):
-    """ Compute a reasonable clim from the min and max, taking nans into account.
+    """Compute a reasonable clim from the min and max, taking nans into account.
+    
     If there are no non-finite values (nan, inf, -inf) this is as fast as it can be.
     Otherwise, this functions is about 3x slower.
     """
-
     # Fast
     min_value = data.min()
     max_value = data.max()
--- a/vispy/visuals/_scalable_textures.py
+++ b/vispy/visuals/_scalable_textures.py
@@ -9,7 +9,7 @@ from vispy.gloo import Texture2D, Texture3D
 from vispy.gloo.texture import should_cast_to_f32
 
 
-def get_default_clim_from_dtype( dtype):
+def get_default_clim_from_dtype(dtype):
     """Get min and max color limits based on the rsange of the dtype."""
     # assume floating point data is pre-normalized to 0 and 1
     if np.issubdtype(dtype, np.floating):
@@ -41,8 +41,6 @@ def get_default_clim_from_data(data):
 
     return min_value, max_value
 
-
-##
 
 class _ScaledTextureMixin:
     """Mixin class to make a texture aware of color limits.
@@ -368,7 +366,6 @@ class GPUScaledTextureMixin(_ScaledTextureMixin):
         clim_min = self.normalize_value(self.clim[0], self._data_dtype)
         clim_max = self.normalize_value(self.clim[1], self._data_dtype)
         return clim_min, clim_max
-
 
     def _handle_auto_texture_format(self, texture_format, data):
         if isinstance(texture_format, str) and texture_format == 'auto':
--- a/vispy/visuals/_scalable_textures.py
+++ b/vispy/visuals/_scalable_textures.py
@@ -9,6 +9,41 @@ from vispy.gloo import Texture2D, Texture3D
 from vispy.gloo.texture import should_cast_to_f32
 
 
+def get_default_clim_from_dtype( dtype):
+    """Get min and max color limits based on the rsange of the dtype."""
+    # assume floating point data is pre-normalized to 0 and 1
+    if np.issubdtype(dtype, np.floating):
+        return 0, 1
+    # assume integer RGBs fill the whole data space
+    dtype_info = np.iinfo(dtype)
+    dmin = dtype_info.min
+    dmax = dtype_info.max
+    return dmin, dmax
+
+
+def get_default_clim_from_data(data):
+    """ Compute a reasonable clim from the min and max, taking nans into account.
+    If there are no non-finite values (nan, inf, -inf) this is as fast as it can be.
+    Otherwise, this functions is about 3x slower.
+    """
+
+    # Fast
+    min_value = data.min()
+    max_value = data.max()
+
+    # Need more work? The nan-functions are slower
+    min_finite = np.isfinite(min_value)
+    max_finite = np.isfinite(max_value)
+    if not (min_finite and max_finite):
+        finite_data = data[np.isfinite(data)]
+        min_value = finite_data.min()
+        max_value = finite_data.max()
+
+    return min_value, max_value
+
+
+##
+
 class _ScaledTextureMixin:
     """Mixin class to make a texture aware of color limits.
 
@@ -78,13 +113,6 @@ class _ScaledTextureMixin:
     @property
     def clim_normalized(self):
         """Normalize current clims to match texture data inside the shader.
-
-        Scaling only happens on the GPU so we only normalize
-        the color limits when needed (for unsigned normalized integer
-        internal formats). Otherwise, for internal formats that are not
-        normalized such as floating point (ex. r32f) we can leave the ``clim``
-        as is.
-
         """
         # if the internalformat of the texture is normalized we need to
         # also normalize the clims so they match in-shader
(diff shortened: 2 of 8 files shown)


vispy/visuals/_scalable_textures.py
```diff
@@ -359,18 +397,20 @@ def _get_texture_format_for_data(self, data, internalformat):
             num_channels = self._data_num_channels(data)
             texture_format = self._handle_auto_texture_format(internalformat, data)
             texture_format = self._get_gl_tex_format(texture_format, num_channels)
+        else:
+            raise NotImplementedError("wut?")
```
almarklein, line 401: This seems to be an oversight. Any idea what to do here @djhoese? Maybe internfalformat should simply not be None?

vispy/visuals/_scalable_textures.py
```diff
@@ -9,6 +9,39 @@
...
+    return dmin, dmax
+
+
+def get_default_clim_from_data(data):
+    """ Compute a reasonable clim from the min and max, taking nans into account.
+    If there are no non-finite values (nan, inf, -inf) this is as fast as it can be.
+    Otherwise, this functions is about 3x slower.
+    """
+
+    # Fast
```
djhoese, line 30:
```suggestion
def get_default_clim_from_data(data):
    """Compute a reasonable clim from the min and max, taking nans into account.
    
    If there are no non-finite values (nan, inf, -inf) this is as fast as it can be.
    Otherwise, this functions is about 3x slower.
    """
    # Fast
```

I believe we have docstring checking explicitly turned off due to all the violations currently in VisPy, but this is how most docstring linters will prefer this to be.

vispy/visuals/_scalable_textures.py
```diff
@@ -258,9 +273,11 @@ def _scale_data_on_cpu(data, clim, copy=True):
         elif not copy and not np.issubdtype(data.dtype, np.floating):
             raise ValueError("Data must be of floating type for no copying to occur.")
 
-        if clim[0] != clim[1]:
+        if clim[0] == clim[1]:
+            pass  # Does not matter - clim range will be inf
```
djhoese, line 277: This if block seems...odd. I understand having it be explicit that this is why you don't care about it, but can't we just have the comment without the extra `if` statement. It makes the code harder to read in my opinion.

vispy/visuals/_scalable_textures.py
```diff
@@ -273,12 +290,11 @@ def scale_and_set_data(self, data, offset=None, copy=True):
         is_auto = isinstance(clim, str) and clim == 'auto'
         if data.ndim == self._ndim or data.shape[self._ndim] == 1:
             if is_auto:
-                clim = np.min(data), np.max(data)
-            clim = (np.float32(clim[0]), np.float32(clim[1]))
```
djhoese, line 277: Why isn't this float conversion needed anymore?

vispy/visuals/_scalable_textures.py
djhoese: Yeah I guess this is a result of this code being extracted from the original visual modules/logic. Logically, this texture subclass should never be used if `texture_format` (passed to this class as `internalformat` is `None`. So maybe start the method with an `if internalformat is None: raise ValueError(...)`?

vispy/visuals/_scalable_textures.py
```diff
@@ -359,18 +394,20 @@ def _get_texture_format_for_data(self, data, internalformat):
...
     def _compute_clim(self, data):
         clim = self._clim
         is_auto = isinstance(clim, str) and clim == 'auto'
-        if data.ndim == 2 or data.shape[2] == 1:
+        if data.ndim == self._ndim or data.shape[2] == 1:
             if is_auto:
-                clim = np.min(data), np.max(data)
-            clim = (np.float32(clim[0]), np.float32(clim[1]))
+                clim = get_default_clim_from_data(data)
+            # clim = (np.float32(clim[0]), np.float32(clim[1]))
```
djhoese, line 407: Similar question to the CPU-based chunk of code. Shouldn't this be here? Maybe I took this from the old code and it isn't needed anymore? Or maybe it is to avoid integer operations in any of the math.

vispy/visuals/image.py
```diff
@@ -276,7 +269,7 @@ def __init__(self, data=None, method='auto', grid=(1, 1),
         # self._build_interpolation()
         self._data_lookup_fn = None
 
-        self.clim = clim
+        self.clim = clim or "auto"  # None -> "auto"
```
djhoese, line 272: The default in the `__init__` definition is `'auto'` so if this is provided as `None` then it must have been set explicitly by the user. The docstring does not say it can be `None`. I'd say either update the docstring or undo this change. I don't feel strongly about it.

vispy/visuals/tests/test_volume.py
```diff
@@ -29,7 +29,7 @@ def test_volume():
 
     # Clim
     V.set_data(vol, (0.5, 0.8))
-    assert V.clim == (np.float32(0.5), np.float32(0.8))
+    assert V.clim == (0.5, 0.8)
```
djhoese, line 32: At least locally, this failed for me and is why I changed it...I think. If it works now with your changes then great.

vispy/visuals/image.py
```diff
@@ -89,17 +89,10 @@
...
-            data = clamp(data, $clim.y, $clim.x);
-        }} else {{
-            // clims are the same, show minimum colormap value
-            return 0.0;
-        }}
-        data = data - $clim.x;
-        data = data / ($clim.y - $clim.x);
-        return max(data, 0);
+
+        data = clamp(data, $clim.x, $clim.y);
```
djhoese, line 93: How does this work for flipped color limits? For example, a floating point texture with values between `0.0` and `100.0` and I specify `clim=(80.0, 20.0)` because I want the colormap flipped?

djhoese (changes requested):
Lots of questions. A couple requests. It looks pretty good though. Part of me still dislikes that this is needed as it is likely (my bias of course) a rare data case but penalizes all other, more common, data cases.

vispy/visuals/_scalable_textures.py
almarklein: Done!

vispy/plot/plotwidget.py
```diff
@@ -201,6 +201,8 @@ def image(self, data, cmap='cubehelix', clim='auto', fg_color=None):
             min and max values.
         fg_color : Color or None
             Sets the plot foreground color if specified.
+        kwargs : keyword arguments.
+            More args to pass to visual.Volume.
```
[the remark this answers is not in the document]
almarklein: Nice! Updated for both the image and volume.

vispy/visuals/_scalable_textures.py
almarklein: I agree. I first did have code in this block, but I later removed it. Should have cleaned it up. Fixed!

vispy/visuals/_scalable_textures.py
almarklein: I'm not sure why it was ever this way. It seems weird to use numpy types in normal Python properties, so I figured I'd try removing it.

vispy/visuals/image.py
almarklein: Good point, I had overlooked this use-case. Although one could argue that you should use a reversed colormap in this case, let's revive this feature :)

vispy/visuals/image.py
almarklein: I like the idea of "None means default" where it is applicable. I updated the docs of both Image and Volume to align them.

almarklein:
Interesting, `test_volume_clims_and_gamma` in `test_volume.py`, fails for me, also in current main ...

almarklein:
> Interesting, `test_volume_clims_and_gamma` in `test_volume.py`, fails for me, also in current main ...

Nevermind. I was on a high-res display. Works fine on a regular display.

vispy/visuals/image.py
djhoese: I agree. Thanks.

vispy/visuals/_scalable_textures.py
djhoese: I think it is less weird considering it is formatting them to be used on the GPU. At least that was my assumption when I saw it. I think something on the lower level gloo layers was complaining about ints being passed or 64-bit floats or something.

vispy/visuals/image.py
djhoese: From a GUI point of view though and for performance, this is much faster. Flipping the colormap requires manipulating the colormap instance assuming there is an array of colors to flip. Some colormaps are not an array of colors and are actual functions that get called so a simple inverse is hard to do. Even for arrays of colors, that means uploading new color data to the texture-backed colormap.

vispy/visuals/volume.py
```diff
@@ -120,16 +120,10 @@
...
-        data = clamp(data, clim.x, clim.y);
-    }} else if (clim.x > clim.y) {{
-        data = clamp(data, clim.y, clim.x);
-    }} else {{
-        return $cmap(0.0);
-    }}
-
     data = (data - clim.x) / (clim.y - clim.x);
-    return $cmap(pow(data, gamma));
+    data = clamp(data, 0.0, 1.0);
```
djhoese, line 124: It may be too early for me to be reviewing this code again, but is there any worry about data type and overflow with doing this clamp after the scaling? Like data that's outside the clim range?

vispy/visuals/_scalable_textures.py
almarklein: Yeah, but then this should be handled there, I'd think. I now updated to at least always be float (no ints). If you prefer putting back the `np.float32()` then I could do that though.

djhoese: If there are no warnings or errors from doing `float(x)` then I'm OK with this.

vispy/visuals/volume.py
almarklein: I wondered about this too, but since it's all floats, I don't think it should matter (e.g. no int overflow).

almarklein: mmm, could put it back up and do `clamp(data, min(clim.x, clim.y), max(clim.x, clim.y))` though, just in case ...

almarklein: done

vispy/visuals/_scalable_textures.py
```diff
@@ -355,23 +388,22 @@ def _get_gl_tex_format(self, texture_format, num_channels):
         return texture_format
 
     def _get_texture_format_for_data(self, data, internalformat):
-        if internalformat is not None:
-            num_channels = self._data_num_channels(data)
-            texture_format = self._handle_auto_texture_format(internalformat, data)
-            texture_format = self._get_gl_tex_format(texture_format, num_channels)
+        assert internalformat is not None
```
djhoese, line 391:
```suggestion
        if internalformat is None:
            raise ValueError("'internalformat' must be provided for GPU scaled textures.")
```

Some code quality services (we don't have any here right now) complain about using asserts outside of tests. Plus it might be nice to have an explicit error message.

vispy/visuals/tests/test_scalable_textures.py
```diff
@@ -0,0 +1,152 @@
...
+    st.set_clim((10, 10))
+    st.scale_and_set_data(ref_data.astype(np.float32))
+    assert st.clim == (10, 10)
+    assert st.clim_normalized == (0, np.inf)
+    # assert np.min(st._data) == 0 - does not matter
+
+
+def test_clim_handling_gpu():
+    return
+    ref_data = np.array([[10, 10, 5], [15, 25, 15]])
```
djhoese, line 119:
```suggestion
def test_clim_handling_gpu():
    ref_data = np.array([[10, 10, 5], [15, 25, 15]])
```

almarklein: oh, I apparently disabled this during dev and forgot to remove. Nice catch!

vispy/visuals/tests/test_scalable_textures.py
```diff
@@ -0,0 +1,152 @@
...
+
+    internalformat = "r32f"
+
+    def _get_texture_format_for_data(self, data, internalformat=None):
+        return None
+
+
+def test_default_clim():
+
+    ref_data = np.array([10, 5, 15, 25, 15])
```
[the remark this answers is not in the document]
almarklein: We need an auto-formatter ;)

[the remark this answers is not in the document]
djhoese: Yes, we should maybe take a poll at the next dev meeting. Do people hate `black` enough that we should use `yapf`. I'm starting to care less and less about specific final styling as long as some tool is doing it for me.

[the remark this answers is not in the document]
almarklein: > I'm starting to care less and less about specific final styling as long as some tool is doing it for me.

Indeed. I sometimes scream at Black for formatting things in weird ways, but then I remember myself that at least its consistent :)

djhoese (approved):
Looks good to me. Merge if you feel it is ready.

vispy/visuals/image.py
```diff
@@ -89,17 +89,9 @@
...
-        }} else {{
-            // clims are the same, show minimum colormap value
-            return 0.0;
-        }}
-        data = data - $clim.x;
-        data = data / ($clim.y - $clim.x);
-        return max(data, 0);
+        data = clamp(data, min($clim.x, $clim.y), max($clim.x, $clim.y));
+        data = (data - $clim.x) / ($clim.y - $clim.x);
+        return data;
```
djhoese, line 94: I just realized, this `min/max` stuff should probably be done in the below `_apply_clim` function too right (for RGBs)?

djhoese: Resolved in #2102.

djhoese:
I've been thinking about this more as I start using the new implementation of GPU scaling in my own application (that SIFT one that I demoed at the monthly meeting). My data actually has NaNs in it that I would like to map to the minimum clim. From what I could tell I think they were mapping to the maximum value. I ended up doing something like this in my version of the ImageVisual that has individual textures, clims, and gammas for each band of the RGB:

```glsl
color.r = !(color.r <= 0.0 || 0.0 <= color.r) ? min($clim_r.x, $clim_r.y) : color.r;
```

This is done in the part of the shader applying the clim (where I have access to `clim_r` already). If I don't do this then `clamp` + the linear scaling  I know this isn't much to go on, but @almarklein do you have any other ideas?

Basically what I'm saying is that we handle NaNs in the clim calculation but not the actual shader.

Side note: Looks like github's code blocks can be declared as `glsl` and it highlights properly.

almarklein:
Good point. It occurred to me earlier, but I forgot about it. In the volume renderer nans probably matter less, though this may depends on the type of rendering. We should check this. For images, I recall being mentioned that NaN's should become transparent. Would that work for your case?

djhoese:
I have two custom Visuals right now in my application. One is a single 2D image and one is a multi-band image where each RGB channel is its own r32f texture. In the RGB case I have it set the individual channel to the min clim value if that channel is NaN, but also if all channels are NaN I have alpha set to 0. This transparency behavior may be application specific. I guess setting the RGB value to the minimum (0) is the same as transparent in this case.

For a single image, in the old code at least, I have the shader `discard` the fragment if it is NaN. I think this goes along with what matplotlib does with NaNs is that it doesn't draw them (or at least they show up transparent). I could maybe see people wanting it to be the minimum value. The complication for my application specifically is that my texture is made up of tiles where the tiles might be only partially filled with real data. I don't want the non-filled part of the tile to be shown as the minimum color.

Any idea if there is a noticeable difference between discard and just setting `.a` to 0? I suppose for the most control possible the NaN/alpha handling should go in the clim method. This also makes it easier for users to customize in a subclass (the build_color_transform method).

Lastly, another thing I'm regretting is not putting the texture creation of the ImageVisual into a separate method. It would make it easier for me to subclass the ImageVisual so I don't have to replicate the `__init__`. I might make a PR for that while we ask napari folks to test things.

djhoese:
Looking at adding this to the code and I'm not really sure where the `.a` handling should happen. I guess I could do discard which would probably work, but otherwise the part of the application where we have the float value are not the same parts as we have the colormapping/vec4.

almarklein:
The main difference between `discart` and setting `.a` is that with `discart` the depth buffer is not written to.

PR #2088, opened 2021-06-21
Restore ambient light behavior as in v0.6.0

asnt:
With these changes, the ambient light is:
- on by default with intensity 0.3,
- applied on the base color of the mesh.

The restores the behaviour of the MeshVisual as before the ShadingFilter was introduced.


almarklein (approved):
LGTM!

Issue #2083, opened 2021-06-18
monitor the GPU memory, log the visual's and UI frame's life cycle


yxdragon:
```python
from vispy import scene
from vispy.geometry import create_box

canvas = scene.SceneCanvas(keys='interactive', bgcolor='white', size=(800, 600), show=True)
view = canvas.central_widget.add_view()

vts, faces, ls = create_box(1,2,3, 1,2,3)
box = scene.visuals.Mesh(vertices=vts['position'], faces=faces, shading='smooth', parent=view.scene)
box.parent = None
del box
view.camera = scene.TurntableCamera(elevation=30, azimuth=30)
```
add __del__ method in Mesh visual class.
I add a mesh, then set the parent None, and del it. but it is not deleted, until I close the window.
is it a bug? or some other way to release the mesh object?

I think we need to do some work to monitor the gpu memory usage, log how many mesh are alive, when to release. 
And also every UI frame work, to make sure the frame, panel, canvas is released after close.

May be it is a hard work, but at least monitor and found these bugs first.

djhoese:
You brought some of this up in #2065. Some of this may be specific to the Wx backend, we'd have to check.

As far as I can tell the Visuals as you've used them above, should be deleted properly and garbage collected. However, since Python will garbage collect whenever it wants to, it may be hard to know when that will happen.

yxdragon:
```python
from vispy import scene
from vispy.geometry import create_box
import sys, weakref

canvas = scene.SceneCanvas(keys='interactive', bgcolor='white', size=(800, 600), show=True)
view = canvas.central_widget.add_view()

vts, faces, ls = create_box(1,2,3, 1,2,3)
box = scene.visuals.Mesh(vertices=vts['position'], faces=faces, shading='smooth', parent=view.scene)

weakbox = weakref.ref(box)
print(sys.getrefcount(weakbox()))

box.parent = None
weakbox = weakref.ref(box)
print(sys.getrefcount(weakbox()))

del box
print(sys.getrefcount(weakbox()))

view.camera = scene.TurntableCamera(elevation=30, azimuth=30)


if __name__ == '__main__':
    canvas.app.run()
```

I remove the del overwrite, for I saw some blog say "overwrite del would affect the gc", then use sys.getrefcount. but after many test, I can not trust the visuals has been deleted.

And the ui backend's leak, I had tracked the bug, and found the matter. I think it is not specific for wx, but for all UI framework. (maybe exclude the web backend)

[...]

PR #2089, opened 2021-06-21
Fix SceneCanvas Node leaking reference to itself

djhoese:
This is a possible cleanup as discussed in #2083. This assumes a lot of stuff about VisPy and Python garbage collection that may not actually matter. I kind of ran wild with this idea that VisPy is adding a reference to a Visual that never gets collected. I tracked it down to the `self.scene_node` property of the Node class (used by SceneCanvas graph nodes). The behavior that I was trying to "fix" was that when you set a parent for a Node and then remove that parent you should end up with the same number of references for that Visual. This may not matter or be wrong because:

1. I don't think Python garbage collection cares about an object referencing itself.
2. I'm not sure if this will break certain cases of using VisPy where the `.scene_node` should be a "hard" reference...but I'm not sure why that would be useful.

CC @yxdragon

### DIFF
--- a/vispy/scene/node.py
+++ b/vispy/scene/node.py
@@ -284,12 +284,12 @@ class Node(object):
             p = self.parent
             while True:
                 if isinstance(p, SubScene) or p is None:
-                    self._scene_node = p
+                    self._scene_node = p and weakref.ref(p)
                     break
                 p = p.parent
             if self._scene_node is None:
-                self._scene_node = self
-        return self._scene_node
+                self._scene_node = weakref.ref(self)
+        return self._scene_node()
 
     @property
     def root_node(self):


djhoese:
```python
from vispy import scene
from vispy.geometry import create_box
from vispy import visuals
import sys, weakref
import gc

vts, faces, ls = create_box(1,2,3, 1,2,3)
box = scene.visuals.Mesh(vertices=vts['position'], faces=faces, shading='smooth', parent=view.scene)
#box = scene.visuals.Mesh(vertices=vts['position'], faces=faces, shading='smooth')
#box = visuals.MeshVisual(vertices=vts['position'], faces=faces, shading='smooth')
print("Ref count after creation: ", sys.getrefcount(box))

weakbox = weakref.ref(box)
gc.collect()
print("Ref count after weakref creation: ", sys.getrefcount(box))

if hasattr(box, "parent") and box.parent is None:
    box.parent = view.scene
    gc.collect()
    print("Ref count after setting parent: ", sys.getrefcount(box))

if isinstance(box, scene.visuals.Mesh) and box.parent is not None:
    box.parent = None
    gc.collect()
    print("Ref count after unsetting parent: ", sys.getrefcount(box))

del view
gc.collect()
print("Ref count after view delete:", sys.getrefcount(box))

del canvas
gc.collect()
print("Ref count after canvas delete:", sys.getrefcount(box))

del box
gc.collect()
if weakbox() is not None:
    print("Weak ref count after box delete:", sys.getrefcount(weakbox()))
```

Results in:

```
Ref count after creation:  7
Ref count after weakref creation:  7
Ref count after unsetting parent:  5
Ref count after view delete: 5
Ref count after canvas delete: 5
```

Where as before unsetting the parent would keep a total reference count of 6 for the rest of the application.

djhoese:
So I've been thinking about this more and based on https://devguide.python.org/garbage_collector/ I think a self-referencing object is considered unreachable and can't be properly garbage collected. However, when I run my above example script code with or without this PR and with setting `gc.set_debug(gc.DEBUG_UNCOLLECTABLE)` it doesn't show anything.

 https://docs.python.org/3/library/gc.html#gc.set_debug

So that tells me that python never considers these Visual instances as uncollectable. So it is possible that this PR doesn't do anything leak-age wise.

That said, I'm not sure this PR does anything wrong either. In my opinion `.scene_node` should be an "alias"/reference to another node that actually lives in the scene graph and is the child of another node (this relationship is a strong reference). So any alias type reference should be a weak reference to avoid circular references.

almarklein (approved):
Looks good to me. Should at least help prevent memory leaks. I double-checked that `._scene_node` is only used in this module, so this should not break anything :)

Issue #1979, opened 2021-01-26
Add documentation showing how to create video of screenshots


djhoese:
It comes up pretty often that users want to make videos of their vispy visualization. I've done this a couple times now with the `imageio` library and the `_screenshot` utility function (https://github.com/vispy/vispy/blob/master/vispy/gloo/util.py#L10). As part of resolving this issue we should:

1. Add a section to my new documentation site (hope to merge it soon) about how to do this. Not sure where it would fit best, but that can be discussed.
2. Rename `_screenshot` to `screenshot`, add deprecated function wrapper called `_screenshot`, and make `screenshot` easier to import (maybe `from vispy import screenshot`?).

djhoese:
@aaru143 this may be a good issue for you to work on after you've started figuring things out. My new website isn't merged yet, but you can check it out here: www.ssec.wisc.edu/~davidh/tmp/vispy

djhoese:
I was just thinking about this in terms of #2000 and drawing a line between gloo and higher level interfaces. Maybe we should add a screenshot method to the Canvas class?

djhoese:
I don't think I mention it above, but maybe we should add a `screenshot` method to the `Canvas`. @kmuehlbauer what do you think?

PR #2090, opened 2021-06-21
Fix alpha handling in 'translucent' Visuals and add 'alpha' keyword argument to Canvas.render

djhoese:
This is an attempt at providing a more clear and stable way for users to get a numpy array of the current visualization. Users often see the `Canvas.render` method, but are then surprised by some of the results. For example:

https://github.com/napari/napari/issues/1566#issuecomment-690748249

I think it may be more important for us to figure out a way around this. I've made this PR to discuss it and to provide a possible workaround with a `Canvas.screenshot` method as the `_screenshot` utility function is my usual go to response for users who run into `.render` issues. @almarklein and I discussed this a little here: https://github.com/vispy/vispy/pull/2082#issuecomment-864962729

I think in the best case `.render` would produce the expected results, but I'm not sure what the proper solution is. This is a partial solution to #1979.

### DIFF
--- a/vispy/gloo/wrappers.py
+++ b/vispy/gloo/wrappers.py
@@ -38,7 +38,7 @@ _gl_presets = {
         depth_test=True,
         cull_face=False,
         blend=True,
-        blend_func=('src_alpha', 'one_minus_src_alpha')),
+        blend_func=('src_alpha', 'one_minus_src_alpha', 'one', 'one_minus_src_alpha')),
     'additive': dict(
         depth_test=False,
         cull_face=False,
@@ -444,7 +444,7 @@ class BaseGlooFunctions(object):
         ----------
         preset : str | None
             Can be one of ('opaque', 'translucent', 'additive') to use
-            use reasonable defaults for these typical use cases.
+            reasonable defaults for these typical use cases.
         **kwargs : keyword arguments
             Other supplied keyword arguments will override any preset defaults.
             Options to be enabled or disabled should be supplied as booleans
--- a/vispy/gloo/wrappers.py
+++ b/vispy/gloo/wrappers.py
@@ -43,7 +43,7 @@ GL_PRESETS = {
         depth_test=True,
         cull_face=False,
         blend=True,
-        blend_func=('src_alpha', 'one_minus_src_alpha', 'one', 'one_minus_src_alpha')),
+        blend_func=('src_alpha', 'one_minus_src_alpha', 'zero', 'one')),
     'additive': dict(
         depth_test=False,
         cull_face=False,
--- a/vispy/app/canvas.py
+++ b/vispy/app/canvas.py
@@ -529,8 +529,10 @@ class Canvas(object):
         logger.debug('Context manager exit complete for %s' % (self,))
 
     def render(self):
-        """Render the canvas to an offscreen buffer and return the image
-        array.
+        """Render the canvas to an offscreen buffer and return the image array.
+
+        This method may not return the expected result in some cases when it
+        comes to transparency or more complex visualizations.
 
         Returns
         -------
@@ -550,6 +552,31 @@ class Canvas(object):
             return fbo.read()
         finally:
             fbo.deactivate()
+
+    def screenshot(self, viewport=None, alpha=True):
+        """Read the current rendered buffer from the GPU and return as an array.
+
+        This function will capture whatever was last drawn.
+        Note this uses the low-level ``glReadPixels`` and does not run using
+        the gloo GLIR queue.
+
+        Parameters
+        ----------
+        viewport : array-like | None
+            4-element list of x, y, w, h parameters. If None (default),
+            the current GL viewport will be queried and used.
+        alpha : bool
+            If True (default), the returned array has 4 elements (RGBA).
+            Otherwise, it has 3 (RGB)
+
+        Returns
+        -------
+        pixels : array
+            3D array of pixels in np.uint8 format
+
+        """
+        from vispy.gloo.wrappers import read_pixels
+        return read_pixels(viewport=viewport, alpha=alpha)
 
 
 # Event subclasses specific to the Canvas
--- a/vispy/app/canvas.py
+++ b/vispy/app/canvas.py
@@ -528,17 +528,27 @@ class Canvas(object):
         sleep(0.1)  # ensure window is really closed/destroyed
         logger.debug('Context manager exit complete for %s' % (self,))
 
-    def render(self):
+    def render(self, alpha=True):
         """Render the canvas to an offscreen buffer and return the image array.
-
-        This method may not return the expected result in some cases when it
-        comes to transparency or more complex visualizations.
 
         Returns
         -------
         image : array
             Numpy array of type ubyte and shape (h, w, 4). Index [0, 0] is the
             upper-left corner of the rendered region.
+        alpha : bool
+            If True (default) produce an RGBA array (M, N, 4). If False,
+            remove the Alpha channel and return the RGB array (M, N, 3).
+            This may be useful if blending of various elements requires a
+            solid background to produce the expected visualization.
+
+        Returns
+        -------
+        image : array
+            Numpy array of type ubyte and shape (h, w, 4). Index [0, 0] is the
+            upper-left corner of the rendered region. If ``alpha`` is ``False``,
+            then only 3 channels will be returned (RGB).
+
 
         """
         self.set_current()
@@ -549,34 +559,13 @@ class Canvas(object):
         try:
             fbo.activate()
             self.events.draw()
-            return fbo.read()
+            result = fbo.read()
         finally:
             fbo.deactivate()
 
-    def screenshot(self, viewport=None, alpha=True):
-        """Read the current rendered buffer from the GPU and return as an array.
-
-        This function will capture whatever was last drawn.
-        Note this uses the low-level ``glReadPixels`` and does not run using
-        the gloo GLIR queue.
-
-        Parameters
-        ----------
-        viewport : array-like | None
-            4-element list of x, y, w, h parameters. If None (default),
-            the current GL viewport will be queried and used.
-        alpha : bool
-            If True (default), the returned array has 4 elements (RGBA).
-            Otherwise, it has 3 (RGB)
-
-        Returns
-        -------
-        pixels : array
-            3D array of pixels in np.uint8 format
-
-        """
-        from vispy.gloo.wrappers import read_pixels
-        return read_pixels(viewport=viewport, alpha=alpha)
+        if not alpha:
+            result = result[..., :3]
+        return result
 
 
 # Event subclasses specific to the Canvas
--- a/vispy/app/tests/test_canvas.py
+++ b/vispy/app/tests/test_canvas.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+
+from vispy.app import Canvas
+from vispy.visuals import ImageVisual
+from vispy.testing import requires_application
+
+
+@requires_application()
+def test_run():
+    """Test app running"""
+    with Canvas(size=(100, 100), show=True, title='run') as c:
+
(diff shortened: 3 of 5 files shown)


almarklein:
Two remarks:
* Should a similar method be added to the `SceneCanvas`?
* The original issue suggests there may be something off with our blend factors. Might be worth looking into that (I could do that).

djhoese:
> Should a similar method be added to the SceneCanvas?

The SceneCanvas is a subclass of `Canvas` so it should be fine.

> Might be worth looking into that (I could do that).

Yes please, if you can.

asnt:
As mentioned/discussed by @almarklein @djhoese above and in various places (e.g. #2082), I think the only thing needed is to discard the alpha channel (or make it fully opaque, i.e. 255 or 1.0).
I tried to reduce this to a minimal example test: https://gist.github.com/asnt/73ae83a9481a7c8345ee0e84c4832cdb
Not sure if this captures all the specificity in napari/napari#1566.

djhoese:
For reference, the images produced by that script on my Ubuntu/PopOS system (PyQt5 backend):

### image-read-rgb

image-read-rgb

### image-read-rgba

image-read-rgba

### image-read-rgba1

image-read-rgba1

### image-render-rgb

image-render-rgb

### image-render-rgba

image-render-rgba

### image-render-rgba1

image-render-rgba1

djhoese:
Here's what it looks like on my image viewer for the rgba versions:

image

djhoese:
Ok, so the missing piece for me here is the effect that the blending functions have on this. I guess I would have assumed an additive blending as far as alpha is concerned. I'm not sure if that is possible with only the alpha (currently `blend_func=('src_alpha', 'one_minus_src_alpha')`).

Regardless, should the proper solution for this be to add an `alpha` keyword argument to both `SceneCanvas.render` and `Canvas.render` that defaults to `True` for RGBA result or can be set to `False` to produce an RGB result? I like this better than adding an FAQ or something else that users have to discover to get the result they expect.

asnt:
> Regardless, should the proper solution for this be to add an `alpha` keyword argument to both `SceneCanvas.render` and `Canvas.render` that defaults to `True` for RGBA result or can be set to `False` to produce an RGB result? I like this better than adding an FAQ or something else that users have to discover to get the result they expect.

Exactly. That would make `render()` and `read_pixels()` consistent.

asnt:
> Ok, so the missing piece for me here is the effect that the blending functions have on this. I guess I would have assumed an additive blending as far as alpha is concerned. I'm not sure if that is possible with only the alpha (currently `blend_func=('src_alpha', 'one_minus_src_alpha')`).

I think that with `blend_func=('src_alpha', 'one_minus_src_alpha')`, the alpha that lies in the framebuffer is [not used](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendFunc.xhtml#examples). That's why it can be discarded at the end. Only the alpha of the incoming fragment is used.
1. Clear the frame buffer `C0.rgba = (0, 0, 0, 1)`.
2. Get a new fragment with colour `C1.rgba = (r, g, b, 0.5)`.
3. Blend the fragment with the current frame buffer colour, i.e the source is `C1` and the destination `C0`,
    ```
    C2 = C1.a * C1 + (1 - C1.a) * C0
        = 0.5 * C1 + 0.5 * C0
        = (r/2, g/2, b/2, 0.75)
    ```
4. Store the blended colours into the destination, i.e. the frame buffer, `C0 <- C2`.
5. Repeat from 2 for all fragments.

Note that in step 3 only the source alpha `C1.a` matters, the destination alpha `C0.a` doesn't play a role in the blending. The frame buffer alpha `C0.a` gets overwritten at each blending, though, and becomes different from its initial opaque value only when translucent colours are blended over it. That's why the see-through effect in the windows and rendered images only happens inside translucent objects. At least that's my current understanding.

Some further references:
https://en.wikipedia.org/wiki/Alpha_compositing
https://www.realtimerendering.com/blog/gpus-prefer-premultiplication/

If the above is correct, we could then use separate blend functions ([`glBlendFuncSeparate`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendFuncSeparate.xhtml)) for the colour and the alpha channels, so that the alpha never gets modified in the framebuffer and we don't have these see-through side effects.
In `gloo`, `glBlendFuncSeparate`is exposed through the same [`blend_func/set_blend_func()`](https://github.com/vispy/vispy/blob/e515b00de7086527070b3e51e1133f8c1c5ca165/vispy/gloo/wrappers.py#L256) with two extra parameters: 
```
blend_func=(srgb='src_alpha', drgb='one_minus_src_alpha', salpha='zero', dalpha='one')
```

almarklein:
I was playing with this yesterday and found that indeed with `blend_func=('src_alpha', 'one_minus_src_alpha')` the alpha channel in the screenshot becomes smaller than 255, but I could not see why.

Thanks @asnt for the step-by-step calculation, because now I finally see how `src_alpha` multiplies with itself, so indeed with a source alpha of 0.5, a 0.75 gets written.

So instead we could use `blend_func=('src_alpha', 'one_minus_src_alpha', "one", "one"))` or `blend_func=('src_alpha', 'one_minus_src_alpha', "zero", "one"))`.

All that said, I think its good practice to not include the alpha channel in screenshots.

asnt:
> All that said, I think its good practice to not include the alpha channel in screenshots (by default).

I agree, only rgb makes sense with the current blending modes. However, `vispy.gloo.util._screenshots()`, `vispy.gloo.wrappers.read_pixels()` and `*Canvas.render()` currently return an RGBA by default, so it would not be backwards compatible. So maybe an alternative would be to keep RGBA the default, but fix the blending with the separate functions as above (if that is indeed confirmed to work).

almarklein:
Here's the code I used to validate this:
```py
import numpy as np
from vispy import scene
from vispy import app

canvas.size = 800, 600
canvas.show()

im1 = np.zeros((100, 100, 4)).astype(np.float32)
im1[:,:,0] = 1
im1[:,:,3] = 1

im2 = np.zeros((50, 50, 4)).astype(np.float32)
im2[:,:,1] = 1
im2[:,:,3] = 0.4

# Create the image
image1 = scene.visuals.Image(im1, parent=view.scene)
image2 = scene.visuals.Image(im2, parent=view.scene)
image2.transform = scene.STTransform(translate=(-20, -20, -1))

image1.set_gl_state(depth_test=True, blend=True,
image2.set_gl_state(depth_test=True, blend=True,
                    blend_func=('src_alpha', 'one_minus_src_alpha', "one", "one_minus_src_alpha"))
view.camera.set_range()
# view.camera.zoom(0.1, (250, 200))


app.process_events()

rgba = canvas.render()
print(rgba[:,:,3].min())


if __name__ == '__main__':
    app.run()
```

Maybe ` blend_func=('src_alpha', 'one_minus_src_alpha', "one", "one_minus_src_alpha")` is even better, in case some decides to clear the render target with (0,0,0,0).

djhoese:
Ok so what are we saying here? I think we agree that there is no `.screenshot` method needed as `.render` should be able to produce a valid result. With that in mind should we:

1. Update `.render` for both Canvas and SceneCanvas with a `alpha=True` keyword argument to control if the Alpha band is included in the result.
2. Update the `translucent` preset to include the two alpha blend functions
3. Both

I'm a little worried about the side effects of 2, but it also seems like the best option. Are there obvious cases where you wouldn't want it?

almarklein:
I don't think there are any (visible) side-effects of 2. Depending how people use `render()`, 1 might be a breaking change, but for most it probably is not. I think we should probably do both. @asnt what do you think?

djhoese:
The `alpha=True` in option 1 would be backwards compatible, but users would have to know about it to get the result they expect with a screenshot.

almarklein:
Ah right. I was thinking `alpha=False`, which would be easier for users, but not backwards compatible. Yeah lets do `True`.

asnt:
With

> Ok so what are we saying here? I think we agree that there is no `.screenshot` method needed as `.render` should be able to produce a valid result. With that in mind should we:
> 
>     1. Update `.render` for both Canvas and SceneCanvas with a `alpha=True` keyword argument to control if the Alpha band is included in the result.

I'd say, this is the minimum. This is api compatible and completes `.render()` in way consistent with the existing api (`read_pixels()`).

>     2. Update the `translucent` preset to include the two alpha blend functions
> 
>     3. Both
> 
> 
> I'm a little worried about the side effects of 2, but it also seems like the best option. Are there obvious cases where you wouldn't want it?

I would go for both, but as you said, we have to ensure that it doesn't break any other rendering and that we get the blending right.
1. The `blend_func` settings would be part of the 'translucent' preset. It seems to me that these settings cannot leak into another preset because both `opaque` and `additive` reset the blending explicitly. Moreover, if a user first does `set_gl_state('translucent')` (and possibly further customises the state), then the user knows that the current drawing is based on the 'translucent' presets. So no surprises.
2. I think that this compositing algorithm (see comment above https://github.com/vispy/vispy/pull/2090#issuecomment-866361668) assumes that the background is not transparent because of step 3. In this case, this seems to me the most explicit blend function: `blend_func=(srgb='src_alpha', drgb='one_minus_src_alpha', salpha='zero', dalpha='one')`.

djhoese:
Ok so these are both implemented now with tests. The title of the PR got a little long, but I wanted to make sure both concepts were mentioned. Re-reviews welcome.

vispy/gloo/wrappers.py
```diff
@@ -43,7 +43,7 @@
         depth_test=True,
         cull_face=False,
         blend=True,
-        blend_func=('src_alpha', 'one_minus_src_alpha')),
+        blend_func=('src_alpha', 'one_minus_src_alpha', 'one', 'one_minus_src_alpha')),
```
almarklein, line 46: I'm not sure whether we've agreed on this part, as I think @asnt preferred `('src_alpha', 'one_minus_src_alpha', 'zero', 'one')`.

My take is that `.., zero, one` assumes that the destination is initialized with alpha being 1, and that with `.., one,  one_minus_src_alpha` it works "as expected" whatever the destination is initiated with. However, I might be missing something.

almarklein (approved):
Looks good. Only one comment.

vispy/gloo/wrappers.py
djhoese: Ah ok, I saw his comment after I started implementing this and then went to lunch and forgot about it. @asnt what are your thoughts?

asnt: I think that the destination alpha channel doesn't bring any information we need for the rendering itself. That's why we can discard it when grabbing a screenshot. However, `('src_alpha', 'one_minus_src_alpha', 'zero', 'one')` is convenient because it ensures that the alpha stays in whatever state the user chose it to be initially (most of the time, opaque).

Any other choice of blending functions do not seem to me to bring new capabilities. And this is I think because of step 3 above, in which the initial background 'rgb' colour of the framebuffer always gets blended in, even with a `glClear(0, 0, 0, 0)`, with an `alpha=0`, some black gets mixed with the foreground objects, even if one intended not to have it in.

asnt:
@djhoese I cannot review in more details right now, but the ideas seem in place to me.

vispy/gloo/wrappers.py
djhoese:
```suggestion
        blend_func=('src_alpha', 'one_minus_src_alpha', 'zero', 'one')),
```

We are keeping in mind that this isn't only for screenshots, right? Are there any non-screenshot cases where this is a bad idea?

almarklein: Probably not. Most common blending functions depend on the source, not the target. And with this, the target's alpha simply stays the same, as @asnt pointed out.

djhoese:
FYI @sofroniewn @kne42 This *should* theoretically fix everything related to odd transparency issues in napari's screenshots like those mentioned here: https://github.com/napari/napari/issues/1566

@asnt I'm going to merge this since it seems like we're all on the same page. Feel free to comment here if you notice anything after the merge and I'll make another PR.

kne42:
Thanks, this should help a ton!

sofroniewn:
Wow, incredible work here!! The 0.7 release is going to be amazing 🚀

Issue #2087, opened 2021-06-21
Unexpected lighting behavior in the basics/visuals/tube.py example


almarklein:
The lighting does not change with the camera, but I think it should. May be part due to the geometry, see #2069, but I suspect the calculation of the light and view vector might be wrong too.

almarklein:
Before I dive in, the light is supposed to be coming from the view-direction, right?

almarklein:
Nevermind, somehow I overlooked that `ShadingFilter` has a `light_dir` argument.

asnt:
The light is fixed in space as you noticed. This was the behaviour before the ShadingFilter as well.

Maybe your issue is with the ambient light? I know realise that there was a default ambient light before the filter, that is now off by default in the filter. So the tube appears black at the back. Moreover, the current ambient component doesn't act on the base color, which it was before. I hesistated during the creation of the filter and finally removed it. Will do PR to see the change it brings.

almarklein:
I was triggered because I expected the light to follow the camera view. Because I saw nothing of a light direction in the example, I was confused. So maybe a spinoff question is: should we support camera-lights (e.g. by setting light-dir to `(0,0,0)`) and make it the default?

But also, after closer inspection, I'm pretty sure the specular reflection is off because of a faulty view_dir.

asnt:
> I was triggered because I expected the light to follow the camera view. Because I saw nothing of a light direction in the example, I was confused. So maybe a spinoff question is: should we support camera-lights (e.g. by setting light-dir to `(0,0,0)`) and make it the default?

I like the idea. It seems like a good default that make things more intuitive.

> But also, after closer inspection, I'm pretty sure the specular reflection is off because of a faulty view_dir.

Indeed. Undoing [this commit](https://github.com/vispy/vispy/commit/549a705a5217b43debb5b451ebe92f9a49553d39) seems to resolve this issue, but I proposed it to fix another issue.

@almarklein I am not sure to understand exactly the computation of the normals and the viewing direction, starting from [here](https://github.com/vispy/vispy/blob/e49191516046572bad1a1ac977e401f6eb8a7c99/vispy/visuals/filters/mesh.py#L130).

almarklein:
This is solved.

PR #2091, opened 2021-06-21
Improve specular light in phong shading

almarklein:
Addresses the main point in  #2087

* Make the depth step for calculating the view direction a bit smaller to avoid glitches.
* Fixes that the view direction was reversed. I don't think this caused any problems, but it did cause confusion while making changes. (I verified the direction by adding `light_vec = eye_vec`.)
* Changes the reflection model to the better Blinn-Phong: see https://learnopengl.com/Advanced-Lighting/Advanced-Lighting.
* Tweaked default light paramseters.

Dropping this here:
```
    // Note (AK): Below is a way to flip the surface normal so that it's always in
    // the view direction. However, this only works well when the light and view
    // vector are (about) the same. Otherwise, normals on faces that that are
    // near-parallel to the eye vector can flip, affecting how they're lit,
    // causing bright and dark artifacts at the edges of objects.
    normal = faceforward(-normal, normal, eye_vec);
```

### DIFF
--- a/vispy/visuals/filters/mesh.py
+++ b/vispy/visuals/filters/mesh.py
@@ -127,28 +127,27 @@ void prepare_shading() {
     vec4 pos_scene = $render2scene(gl_Position);
     v_pos_scene = pos_scene;
 
+    // Calculate normal
     vec4 normal_scene = $visual2scene(vec4($normal, 1.0));
     vec4 origin_scene = $visual2scene(vec4(0.0, 0.0, 0.0, 1.0));
     normal_scene /= normal_scene.w;
     origin_scene /= origin_scene.w;
-    vec3 normal = normalize(normal_scene.xyz - origin_scene.xyz);
-    v_normal_vec = normal;
+    v_normal_vec = normalize(normal_scene.xyz - origin_scene.xyz);
 
-    vec4 pos_front = $scene2doc(pos_scene);
-    pos_front.z += 1e-6;
+    // Calculate eye vector per-vertex (to account for perspective)
+    vec4 pos_doc = $scene2doc(pos_scene);
+    pos_doc /= pos_doc.w;
+    vec4 pos_front = pos_doc;
+    vec4 pos_back = pos_doc;
+    pos_front.z -= 1e-5;
+    pos_back.z += 1e-5;
     pos_front = $doc2scene(pos_front);
-    pos_front /= pos_front.w;
-
-    vec4 pos_back = $scene2doc(pos_scene);
-    pos_back.z -= 1e-6;
     pos_back = $doc2scene(pos_back);
-    pos_back /= pos_back.w;
+    // The eye vec eminates from the eye, pointing towards what is being viewed
+    v_eye_vec = normalize(pos_back.xyz / pos_back.w - pos_front.xyz / pos_front.w);
 
-    vec3 eye = normalize(pos_front.xyz - pos_back.xyz);
-    v_eye_vec = eye;
-
-    vec3 light = normalize($light_dir.xyz);
-    v_light_vec = light;
+    // Pass on light direction (the direction that the "photons" travel)
+    v_light_vec = normalize($light_dir.xyz);
 }
 """
 
@@ -181,9 +180,7 @@ void shade() {
         // Note(asnt): The normal calculated above always points in the
         // direction of the camera. Reintroduce the original orientation of the
         // face.
-        if (!gl_FrontFacing) {
-            normal = -normal;
-        }
+        normal = gl_FrontFacing ? normal : -normal;
     }
     normal = normalize(normal);
 
@@ -194,21 +191,24 @@ void shade() {
     vec3 ambient = ambient_coeff.rgb * ambient_coeff.a
                    * ambient_light.rgb * ambient_light.a;
 
-    // Diffuse illumination.
-    float diffuse_factor = dot(light_vec, normal);
+    // Diffuse illumination (light vec points towards the surface)
+    float diffuse_factor = dot(-light_vec, normal);
     diffuse_factor = max(diffuse_factor, 0.0);
     vec3 diffuse = diffuse_factor
                    * diffuse_coeff.rgb * diffuse_coeff.a
                    * diffuse_light.rgb * diffuse_light.a;
 
-    // Specular illumination.
+    // Specular illumination. Both light_vec and eye_vec point towards the surface.
     float specular_factor = 0.0;
     bool is_illuminated = diffuse_factor > 0.0;
     if (is_illuminated && shininess > 0.0) {
-        vec3 reflection = reflect(light_vec, normal);
-        specular_factor = dot(reflection, eye_vec);
-        specular_factor = max(specular_factor, 0.0);
-        specular_factor = pow(specular_factor, shininess);
+        // Phong reflection
+        vec3 reflection = reflect(-light_vec, normal);
+        float reflection_factor = max(dot(reflection, eye_vec), 0.0);
+        // Blinn-Phong reflection
+        //vec3 halfway = -normalize(light_vec + eye_vec);
+        //float reflection_factor clamp(dot(halfway, normal), 0.0, 1.0);
+        specular_factor = pow(reflection_factor, shininess);
     }
     vec3 specular = specular_factor
                     * specular_coeff.rgb * specular_coeff.a
@@ -216,7 +216,6 @@ void shade() {
 
     // Blend the base color and combine the illuminations.
     vec3 color = base_color * (ambient + diffuse) + specular;
-
     gl_FragColor.rgb = color;
 }
 """  # noqa
@@ -395,8 +394,8 @@ class ShadingFilter(Filter):
                  specular_coefficient=(1, 1, 1, 1),
                  shininess=100,
                  light_dir=(10, 5, -5),
-                 ambient_light=(1, 1, 1, .3),
-                 diffuse_light=(1, 1, 1, 1),
+                 ambient_light=(1, 1, 1, .25),
+                 diffuse_light=(1, 1, 1, 0.7),
                  specular_light=(1, 1, 1, .25),
                  enabled=True):
         self._shading = shading
--- a/vispy/visuals/filters/mesh.py
+++ b/vispy/visuals/filters/mesh.py
@@ -203,11 +203,11 @@ void shade() {
     bool is_illuminated = diffuse_factor > 0.0;
     if (is_illuminated && shininess > 0.0) {
         // Phong reflection
-        vec3 reflection = reflect(-light_vec, normal);
-        float reflection_factor = max(dot(reflection, eye_vec), 0.0);
+        //vec3 reflection = reflect(-light_vec, normal);
+        //float reflection_factor = max(dot(reflection, eye_vec), 0.0);
         // Blinn-Phong reflection
-        //vec3 halfway = -normalize(light_vec + eye_vec);
-        //float reflection_factor clamp(dot(halfway, normal), 0.0, 1.0);
+        vec3 halfway = -normalize(light_vec + eye_vec);
+        float reflection_factor = clamp(dot(halfway, normal), 0.0, 1.0);
         specular_factor = pow(reflection_factor, shininess);
     }
     vec3 specular = specular_factor
--- a/vispy/visuals/tests/test_mesh.py
+++ b/vispy/visuals/tests/test_mesh.py
@@ -68,11 +68,11 @@ def test_mesh_shading_filter(shading):
         if shading in ("flat", "smooth"):
             # there should be a gradient, not solid colors
             assert np.unique(rendered).size >= 28
-            # sphere/circle starts "dark" on the left and gets brighter
-            # then hits a bright spot and decreases after
-            invest_row = rendered[23].astype(np.float64)
+            # sphere/circle is "dark" on the right and gets brighter as you
+            # move to the left, then hits a bright spot and decreases after
+            invest_row = rendered[34].astype(np.float64)
             # overall, we should be increasing brightness up to a "bright spot"
-            assert (np.diff(invest_row[:29]) >= -1).all()
+            assert (np.diff(invest_row[34:60]) <= 0).all()
         else:
             assert np.unique(rendered).size == 2
 


almarklein:
I hoped to also introduce the `faceforward` call here, which makes that the shader is more forgiving about the orientation of surface normals. However, due to how the eye vector is calculated, this causes artifacts. I started changing the eye vector math, but then thought there might be a reason for it being this way.

djhoese:
CC @asnt

asnt:
This seems to fix the random glittering that was appearing on the tubes in `examples/basics/visual/tube.py` and I don't see a visual regression on the other mesh examples.

vispy/visuals/filters/mesh.py
```diff
@@ -134,17 +134,17 @@ def _attach(self, visual):
...
-    pos_front.z += 1e-6;
-    pos_front = $doc2scene(pos_front);
     pos_front /= pos_front.w;
-
-    vec4 pos_back = $scene2doc(pos_scene);
-    pos_back.z -= 1e-6;
+    vec4 pos_back = pos_front;
+    pos_front.z -= 1e-5;
+    pos_back.z += 1e-5;
+    pos_front = $doc2scene(pos_front);
```
asnt, line 144: So, the main change is here, right? The eye vector is created in normalised coordinates and is thus now independent of the size of the scene/frustum/models. Right?

vispy/visuals/filters/mesh.py
```diff
@@ -134,17 +134,17 @@ def _attach(self, visual):
     vec3 normal = normalize(normal_scene.xyz - origin_scene.xyz);
     v_normal_vec = normal;
 
+    // Calculate eye vector in a vertex-depending manner. I.e. consider the eye to
+    // occupy a rectangular plane.
```
asnt, line 138: I am not sure I understand what the plane is here.

vispy/visuals/filters/mesh.py
```diff
@@ -187,8 +190,11 @@ def _attach(self, visual):
     }
     normal = normalize(normal);
 
-    vec3 light_vec = normalize(v_light_vec);
-    vec3 eye_vec = normalize(v_eye_vec);
+    // Note (AK): Below is a way to flip the surface normal so that it's always in
+    // the view direction. However, the eye_vector is calculated as if the eye is
+    // a rectangular plane, not a point, causing it to "look around objects" a bit,
+    // and this gives rise to artifacts if we apply the below.
+    //normal = -faceforward(-normal, normal, eye_vec);
```
asnt, line 197: Ah ok so this is the plane.

I'm not sure we want to systematically orient normals towards the viewer. This might correct some issues with inconsistent face orientations of the input models (#2069), but in specific cases only. For example, for face culling to work, there is no other possibility than to input a geomertry with a consistent definition.

asnt (commented):
This seems to fix one issue with the 'glittering' effect in `examples/basics/visuals/tube.py` following #2087. That was the main point of this PR, right? Or is there something else that this PR solves that I didn't notice?

vispy/visuals/filters/mesh.py
almarklein: The main change is actually turning those `1e6` into `1e5`. There is a bit of re-ordering to do the math with less transforms .

vispy/visuals/filters/mesh.py
almarklein: This description is also not correct. To be honest I am not sure why the artifacts appear.

vispy/visuals/filters/mesh.py
asnt: I see, changing to `1e5` only suffices, but I still have no evidence that this is not specific to this example. I wonder if a perspective correction would not be a more general solution. Examples:

No perspective correction (`1e6` too small relative to the size of the viewing frustum):
```c
    // Fails: 'glittering' artefacts when not close enough to the tubes in `examples/basics/visuals/tube.py`.
    vec4 pos_eye = $scene2doc(pos_scene);

    vec4 pos_front = pos_eye;
    pos_front.z += 1e-6;
    pos_front = $doc2scene(pos_front);
    pos_front /= pos_front.w;

    vec4 pos_back = pos_eye;
    pos_back.z -= 1e-6;
    pos_back = $doc2scene(pos_back);
    pos_back /= pos_back.w;
```

No perspective correction (`1e5` is big enough):
```c
    // Works.
    vec4 pos_eye = $scene2doc(pos_scene);

    vec4 pos_front = pos_eye;
    pos_front.z += 1e-5;
    pos_front = $doc2scene(pos_front);
    pos_front /= pos_front.w;

    vec4 pos_back = pos_eye;
    pos_back.z -= 1e-5;
    pos_back = $doc2scene(pos_back);
    pos_back /= pos_back.w;
```

With perspective correction, variant 1:
```c
    // Works.
    vec4 pos_eye = $scene2doc(pos_scene);

    pos_eye /= pos_eye.w;

    vec4 pos_front = pos_eye;
    pos_front.z += 1e-6;
    pos_front = $doc2scene(pos_front);
    pos_front /= pos_front.w;

    vec4 pos_back = pos_eye;
    pos_back.z -= 1e-6;
    pos_back = $doc2scene(pos_back);
    pos_back /= pos_back.w;
```

With perspective correction, variant 2 (the math is equivalent to variant 1):
```c
    // Works.
    vec4 pos_eye = $scene2doc(pos_scene);

    vec4 pos_front = pos_eye;
    pos_front.z += 1e-6 * pos_front.w;
    pos_front = $doc2scene(pos_front);
    pos_front /= pos_front.w;

    vec4 pos_back = pos_eye;
    pos_back.z -= 1e-6 * pos_back.w;
    pos_back = $doc2scene(pos_back);
    pos_back /= pos_back.w;
```

I'll try to check that in actual example tests.

asnt: To summarise, I think this change makes things more correct and would be good to merge.

vispy/visuals/filters/mesh.py
```diff
@@ -205,19 +207,17 @@ def _attach(self, visual):
     float specular_factor = 0.0;
     bool is_illuminated = diffuse_factor > 0.0;
     if (is_illuminated && shininess > 0.0) {
-        vec3 reflection = reflect(light_vec, normal);
-        specular_factor = dot(reflection, eye_vec);
-        specular_factor = max(specular_factor, 0.0);
+        // Blinn-Phong reflection
+        vec3 halfway = normalize(light_vec + eye_vec);
+        specular_factor = clamp(dot(halfway, normal), 0.0, 1.0);
```
almarklein, line 212: Changed from Phong to Blinn-Phong. If I had not done that I'd only have to change `reflect(-light_vec, normal);` (which is what you see in online code examples).

almarklein:
This is ready from my end. I added notes to the top post.

djhoese:
As we work on the new CI failures, I think this should be included in 0.7, right? @asnt if you have time to re-review that'd be great. Let's include this one, the `render` alpha changes, and my CoC PR and then release this thing. ...maybe have some napari folks test it.

vispy/visuals/filters/mesh.py
```diff
@@ -127,28 +127,25 @@ def _attach(self, visual):
...
 
+    // Calculate eye vector per-vertex (account for perspective)
     vec4 pos_front = $scene2doc(pos_scene);
-    pos_front.z += 1e-6;
-    pos_front = $doc2scene(pos_front);
     pos_front /= pos_front.w;
-
-    vec4 pos_back = $scene2doc(pos_scene);
-    pos_back.z -= 1e-6;
+    vec4 pos_back = pos_front;
```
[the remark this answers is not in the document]
asnt: I think this is really the only change needed to fix the main point. Isn't it? (Besides, clarifying the definition of `$light_dir`.)

asnt (changes requested):
I have the impression that there more changes than needed to fix the main point ("glitches"/"glitter artefact").
Could we reduce the focus to the main point and keep the rest for separate PR's?

Except for one point: The remaining ambiguity here for me is the definition of [`MeshVisual.light_dir`](https://github.com/vispy/vispy/blob/2eea22fbb0d4eea82351aca2dddca1785d3db242/vispy/visuals/mesh.py#L482). Is it the direction
- to the light source?
- or away from the light source? (i.e. the direction of propagation of the rays).

Intuitively, since it is a directional light, I'd say it is the direction of propagation. Given, the docstring I'd say it is the direction of propagation, too. But given how the diffuse component is computed `dot(light_vec, normal)` (and given that `light_vec = light_dir`), I'd say that `light_dir` is used as if it was the direction to the light source.
AFAICT, this is the only parameters exposed publicly that affects the choicr of signs inside the shader. So, if we can clarify this for this PR, then we make the minimal amount of changes to account for it, i.e. change the signs of `light_dir/light_vec`appropriately wherever they appear. Then, in another PR, we clarify the internal implementation by consistently defining the signs of the other direction vectors.

For the rest, good points are made:
- OK for the new default light intensities, but this is per se a different issue.
- If think the Phong model works and is standard enough to keep as the default. Blinn-Phong is a nice addition, but I'd rather have as an alternative (it's known to be more efficient and I've read it renders the highlights better in some ways, but I have no evidence that it is that noticeable. And if it is, it deserves being an alternative instead.)

vispy/visuals/filters/mesh.py
almarklein: 1. As I understand it, Blinn-Phong avoids issues with Phong at certain angles, so it its is considered an improvement. I only know because I read this, so correct me if I'm wrong :)   but if this is the case it seems like a better choice then, right?
2. I see your point. I tested it by making the light_vec and eye_vec equal, but the light_vec might not mean what I thought it meant. Let's add these definitions as comments in the shader. I think I'm going to take a look at this tomorrow morning, it's too late for this now :)

vispy/visuals/filters/mesh.py
```diff
@@ -127,28 +127,25 @@ def _attach(self, visual):
...
+    pos_front.z -= 1e-5;
+    pos_back.z += 1e-5;
```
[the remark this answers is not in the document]
almarklein: +z is away from the camera in NDC. I think the same applied to doc as well, but maybe double-check...

vispy/visuals/filters/mesh.py
```diff
@@ -181,14 +181,16 @@ def _attach(self, visual):
...
     normal = normalize(normal);
 
-    vec3 light_vec = normalize(v_light_vec);
-    vec3 eye_vec = normalize(v_eye_vec);
+    // Note (AK): Below is a way to flip the surface normal so that it's always in
+    // the view direction. However, this only works well when the light and view
+    // vector are (about) the same. Otherwise, normals on faces that that are
+    // near-parallel to the eye vector can flip, affecting how they're lit,
+    // causing bright and dark artifacts at the edges of objects.
+    //normal = faceforward(-normal, normal, eye_vec);
```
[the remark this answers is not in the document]
almarklein: Yeah, I should remove it. The fact that this caused artifacts while I have this working fine elsewhere really annoyed me and. Now that I understand, I want to document/remember it. But this is not the place.

vispy/visuals/filters/mesh.py
```diff
@@ -395,8 +395,8 @@ def __init__(self, shading='flat',
                  specular_coefficient=(1, 1, 1, 1),
                  shininess=100,
                  light_dir=(10, 5, -5),
-                 ambient_light=(1, 1, 1, .3),
-                 diffuse_light=(1, 1, 1, 1),
+                 ambient_light=(1, 1, 1, .25),
+                 diffuse_light=(1, 1, 1, 0.7),
```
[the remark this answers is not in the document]
almarklein: No, the reason is that because it already saturates quite a bit with the diffuse light, the specular light is barely visible.

vispy/visuals/filters/mesh.py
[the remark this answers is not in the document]
almarklein: This fixes the glitches, yes. Some of the other changes further improve/tweak the (specular) lighting, so still apply to this PR, I guess.

almarklein:
Updated and squashed.

I have commented the direction of the vectors. For both light_vec and eye_vec I've defined them to point towards the surface. This is surely the most natural for the light source. For the eye_vec there is more room for discussion, but this felt more natural to me (e.g. thinking of ray-tracing in volume rendering). 

It seems @asnt and I disagree about Phong vs Blinn-Phong for the reflection math. I don't see a reason to chose (or even support) Phong if Blinn-Phong is considered better. But I also don't feel too strongly about it to let this hold up this PR.

almarklein:
Also updated the mesh shading test, since the light now comes from the other (correct) direction. Great feeling that that test broke by this change, actually.

vispy/visuals/filters/mesh.py
```diff
@@ -181,9 +180,7 @@ def _attach(self, visual):
         // Note(asnt): The normal calculated above always points in the
         // direction of the camera. Reintroduce the original orientation of the
         // face.
-        if (!gl_FrontFacing) {
-            normal = -normal;
-        }
+        normal = gl_FrontFacing ? normal : -normal;
```
djhoese, line 183: Just curious, does this perform differently in the shader or is this change purely style?

almarklein: You want to avoid branches in shaders, and a ternary operation is more trivial to convert into branch-free code by the compiler.

djhoese:
Ok so here's my summary of what I understand from the few comments I've read here:

1. @asnt implemented Phong, @almarklein implemented Blinn-Phong.
2. According to wikipedia Blinn-Phong produces a better result. This makes me agree with @almarklein that I'm not sure why we should have Phong if Blinn-Phong can also be implemented. And I'm not sure there is a reason to support both unless someone is specifically comparing the two results (not something we need to support).
3. There seems to be some confusion over conventions with things like light direction and other properties, but docstrings were added to help minimize the confusion. None of the implementation is necessarily wrong (it produces valid results), but maynot necessarily match certain conventions depending on who you talk to and what paper you read.
4. The current implementation in this PR is Phong with Blinn-Phong commented out, right @almarklein ?

Where do we go from here? I'm in favor of Blinn-Phong if it produces better results and is already implemented but commented out. That said, I don't know what I'm talking about. I'd like to get this merged and then request that some napari folks double check a few things with VisPy main (meshes, visuals, images, shading, etc) so that we can still release tomorrow (if napari folks have a ton of free time) or as late as early next week.

almarklein:
> The current implementation in this PR is Phong with Blinn-Phong commented out, right @almarklein ?

yes

djhoese:
@almarklein With the way Blinn-Phong is implemented here, does it still have the higher efficiency that wikipedia mentions? Do you have a screenshot of the difference these two implementations will make in VisPy?

If the difference with which one of these is implemented comes down to a few lines of code being commented/uncommented then (with no defined authority on the matter) vote that you should change to Blinn-Phong and merge this when it passes. If we have strong feelings from @asnt that Phong is better for VisPy then we can switch it later. Thoughts?

My kid is sick today so I won't be working much. If you're ok making the change and the tests pass then feel free to merge.

almarklein:
There are small differences, but it's hard to tell which one is "better" from these images. [This image](https://learnopengl.com/img/advanced-lighting/advanced_lighting_comparrison.png) from the post I referenced above shows the differences more specifically.

Phong:
image
Blinn-Phong:
image

almarklein:
The code is just 2 lines each. The difference in performance is probably negligible on today's hardware. I'll wait for @asnt and otherwise merge this evening.

djhoese:
> but it's hard to tell which one is "better" from these images.

Yeah, swapping between the two images quickly and I think I could convince myself that either of them is better or more "realistic" for some reason.

PR #2092, opened 2021-06-22
Prevent memory surge in two examples

### LINKED ISSUE #1871: memory surge after minimizing the vispy window
Hi, thank you very much for working on vispy. Recently, I found a bug related to memory consumption. It appears after minimizing the vispy window, with a memory surge rate of about 50 MB/second. This surge doesn't exist even if you bring other programs to the top level. The only way to trigger it, based on my experience so far, is to minimize the window by either clicking the "Minimize" icon on the window or the taskbar icon of the vispy window. 

Please try the two scripts provided in the Examples folder in the vispy github: Examples\demo\scene\scrolling_plots.py and Examples\demo\gloo\realtime_signals.py

Below is my environment:
OS: Windows 10 ver.1909
Python: 3.8.3
vispy: 0.6.4 or github master
pyqt5: 5.14.2

Thanks! Appreciate the help.

almarklein:
Closes #1871

The problem is that if updates are done, these are encoded as GLIR commands, which end up in a queue. You can imagine that this queue becomes a problem if it is not emptied. If a Qt window is minimized, it won't invoke a draw event, so the queue is not flushed, and it builds up.

The solution/workaround is to call `canvas.context.flush()` explicitly. There are multiple examples affected by this. Basically any example that does *any* update to a GPU buffer, texture or program. However, most these updates are so small that it does not show a noticable memory increase (no increase seen in the process manager). These two examples change so much data at such a high speed that it does matter.

@rossant here's a nasty downside of the GLIR queue :)  On the long run it could be worth examining if we can get rid of the queue completely, since all our GLIR backends are local and can be flushes immediately ...

### DIFF
--- a/examples/demo/gloo/realtime_signals.py
+++ b/examples/demo/gloo/realtime_signals.py
@@ -155,6 +155,7 @@ class Canvas(app.Canvas):
 
         self.program['a_position'].set_data(y.ravel().astype(np.float32))
         self.update()
+        self.context.flush()  # prevent memory leak when minimized
 
     def on_draw(self, event):
         gloo.clear()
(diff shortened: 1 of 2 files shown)


djhoese (approved):
Merge if you'd like

PR #2098, opened 2021-06-23
Extract out and test the computation of normals

asnt:
This does not add any new functionality, but tries to isolate the routines for unit testing for correctness.

This is in preparation for including the accelerated computation of vertex normals from #2069.

### DIFF
--- a/vispy/geometry/tests/test_meshdata.py
+++ b/vispy/geometry/tests/test_meshdata.py
@@ -10,25 +10,31 @@ from vispy.geometry.meshdata import MeshData
 
 
 def test_meshdata():
-    """Test meshdata Class
-    It's a unit square cut in two triangular element
-    """
-    square_vertices = np.array([[0, 0, 0], [1, 0, 0], [1, 1, 0], [0, 1, 0]],
-                               dtype=np.float64)
-    square_faces = np.array([[0, 1, 2], [0, 2, 3]], dtype=np.uint8)
-    square_normals = np.array([[0, 0, 1], [0, 0, 1], [0, 0, 1], [0, 0, 1]],
-                              dtype=np.float64)
-    square_edges = np.array([[0, 1], [0, 2], [0, 3], [1, 2], [2, 3]],
-                            dtype=np.uint8)
+    dtype_float = np.float64
+    dtype_int = np.uint8
+    # Three ajacent triangles forming the tip of a tetrahedron at the origin.
+    vertices = np.array([[0, 0, 0], [1, 0, 0], [0, 1, 0], [0, 0, 1]],
+                        dtype=dtype_float)
+    faces = np.array([[0, 1, 2], [0, 2, 3], [0, 3, 1]], dtype=dtype_int)
+    face_normals = np.array([[0, 0, 1], [1, 0, 0], [0, 1, 0]],
+                            dtype=dtype_float)
+    isqrt3 = 1 / np.sqrt(3)
+    isqrt2 = 1 / np.sqrt(2)
+    vertex_normals = np.array([
+        [isqrt3, isqrt3, isqrt3],
+        [0, isqrt2, isqrt2],
+        [isqrt2, 0, isqrt2],
+        [isqrt2, isqrt2, 0]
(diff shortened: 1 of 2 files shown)


vispy/geometry/tests/test_meshdata.py
```diff
@@ -31,4 +32,50 @@ def test_meshdata():
...
+        (0, 0, 0),
+        (1, 0, 0),
+        (0, 1, 0),
+        (0, 0, 1),
+    ], dtype=dtype)
+    faces = np.array([(0, 1, 2), (0, 2, 3)], dtype=int)
+
+    expected_normals = np.array([(0, 0, 1), (1, 0, 0)], dtype=dtype)
+
+    computed_normals1 = meshdata._compute_face_normals(vertices, faces)
```
djhoese, line 47: Could you use the public methods (`get_x_normals`) so we're testing the actual MeshData interfaces instead of the implementation of how it is done. Shouldn't they return the same thing?

asnt:
@djhoese Thanks for the initial review. Sure I can do that but I'll have to delay the work on this a bit.
Is that ok for a later relase than 0.7.0?
In any case, @anybody, if pressing, feel free to take over.

djhoese:
Yes, I would consider most of this stuff as a bug fix so 0.7.1 doesn't sound bad.

vispy/geometry/tests/test_meshdata.py
asnt: This is already tested in `test_meshdata()`.

asnt:
@djhoese What do you think of testing both the isolated functions for computing the normals and through the MeshData interface? In some cases, both perform the same computations, in others, MeshData does some computations before actually calling the normals routine.

djhoese:
> What do you think of testing both the isolated functions for computing the normals and through the MeshData interface? 

I assume by `isolated` you mean the `_` private methods, right? Does `MeshVisual` (or other Visual subclass) ever use the private method? If so, then those methods should be made public. I think clean coding-wise and general TDD strategy there is no point in testing the private methods. We could change the entire internals of MeshData but keep the public methods the same and the user wouldn't care. We, the developers, shouldn't test things that aren't user facing or don't effect the user. This is what I meant by testing the implementation.

https://medium.com/@agile_ed/tdd-the-whole-point-from-the-very-beginning-9d8a45a5f465

> The reason this is important (or even vital) is that by testing only the behaviour of something you don’t care how it is done (the implementation), only that it works correctly (the behaviour).

asnt:
> > What do you think of testing both the isolated functions for computing the normals and through the MeshData interface?
> 
> I assume by `isolated` you mean the `_` private methods, right? Does `MeshVisual` (or other Visual subclass) ever use the private method? If so, then those methods should be made public. I think clean coding-wise and general TDD strategy there is no point in testing the private methods. We could change the entire internals of MeshData but keep the public methods the same and the user wouldn't care. We, the developers, shouldn't test things that aren't user facing or don't effect the user. This is what I meant by testing the implementation.
> 
> https://medium.com/@agile_ed/tdd-the-whole-point-from-the-very-beginning-9d8a45a5f465
> 
> > The reason this is important (or even vital) is that by testing only the behaviour of something you don’t care how it is done (the implementation), only that it works correctly (the behaviour).

Thanks @djhoese . I think I get your point. I think every computation goes through MeshData internally. Let's keep it this way, and this can always be changed if necessary.

djhoese:
What's the state of this PR? Is it ready to be merged? What were your findings and what do you suggest for #2069?

djhoese:
As mentioned in the other related PRs, is this PR needed to resolve some of the bugs mentioned in #2069 or were those already included in #2434? Is this PR (or will it be) only refactoring for cleaner code?

asnt:
> As mentioned in the other related PRs, is this PR needed to resolve some of the bugs mentioned in #2069 or were those already included in #2434?

This PR is independent.

> Is this PR (or will it be) only refactoring for cleaner code?

Refactoring and additional testing.

asnt:
@djhoese Superseded by #2444. I'll let you close this PR if you agree.

PR #2099, opened 2021-06-23
Prevent panzoom viewbox become zero

almarklein:
See #2093. This fix prevents the `scatter_histogram.py` example from producing an error when one of its widgets gets zero height, which happens intermittendly. This does not solve the underlying problem of a zero-height widget being created in the first place, but it does prevent errors in case a scene is created with a zero-sized viewbox in other situations.

See  #2097 for (an attempt at) a proper solution.


almarklein:
Not sure if we should guard against the viewbox being zero here, even with #2100. Not all viewboxes are produced via a grid. Or is it the responsibility of the user to prevent the viewbox being zero?

djhoese:
I guess I'd say it is the responsibility of the user? I'm not against this being reopened and merged, but it should maybe have a test added if you plan on doing that.

almarklein:
Yeah, let's leave this for now.

PR #2105, opened 2021-06-25
Refactor ImageVisual for easier subclassing

djhoese:
This PR refactors the ImageVisual's creations of its Texture object. This is to help me in my own applications where I want to subclass the ImageVisual. This simple change gives me much more control over how the textures are created while not requiring me to replicate the whole `__init__`. There is probably larger refactor that could serve subclasses (like the one @tlambert03 is working on in #1999), but I'm not sure we're ready for that yet. I hope to find some time today to use this refactor in my application and make sure it works. Eventually the subclass I'm working on will probably be added to VisPy-core.

### DIFF
--- a/vispy/visuals/image.py
+++ b/vispy/visuals/image.py
@@ -229,6 +229,7 @@ class ImageVisual(Visual):
         self._need_vertex_update = True
         self._need_colortransform_update = True
         self._need_interpolation_update = True
+        self._texture = self._init_texture(data, texture_format)
         self._subdiv_position = VertexBuffer()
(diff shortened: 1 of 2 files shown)


almarklein (approved):

djhoese:
Now that some of the picking and other issues have fixes, I should finally have time to test this refactoring with my own application and make sure it makes sense. I noticed that `.gamma` is set in an odd place in `__init__` and would like to investigate it and include it in this PR.

djhoese:
This seems like a good basic set of refactorings. I was able to get a subclass into my application without having to copy anything unnecessary.

PR #2106, opened 2021-06-25
Add handling of NaNs in ImageVisual

djhoese:
This was discussed a little in #2085 and I know this current implementation isn't optimized (multiple nan checks and branches), but I wanted to start the discussion since this shows (I think) what we want the end result to be. For single channel images, the NaN pixels should be fully transparent (or discarded). For RGB images, if all pixels are NaNs then make them transparent, otherwise make them the lower clim limit.

Thoughts? Changes?

### DIFF
--- a/vispy/visuals/image.py
+++ b/vispy/visuals/image.py
@@ -134,6 +134,10 @@ class ImageVisual(Visual):
     ----------
     data : ndarray
         ImageVisual data. Can be shape (M, N), (M, N, 3), or (M, N, 4).
+        If floating point data is provided and contains NaNs, they will
+        be made transparent (discarded) for the single band data case.
+        For RGB data, NaNs will be mapped to the lowest ``clim`` value.
+        If the Alpha band is NaN it will be mapped to 0 (transparent).
     method : str
         Selects method of rendering image in case of non-linear transforms.
         Each method produces similar results, but may trade efficiency
--- a/vispy/visuals/image.py
+++ b/vispy/visuals/image.py
@@ -89,12 +89,26 @@ _texture_lookup = """
 
 _apply_clim_float = """
     float apply_clim(float data) {
+        if (!(color.r <= 0.0 || 0.0 <= color.r)) {
+            discard;
+        }
         data = clamp(data, min($clim.x, $clim.y), max($clim.x, $clim.y));
         data = (data - $clim.x) / ($clim.y - $clim.x);
         return data;
     }"""
 _apply_clim = """
     vec4 apply_clim(vec4 color) {
+        // If all the pixels are NaN make it completely transparent
+        // http://stackoverflow.com/questions/11810158/how-to-deal-with-nan-or-inf-in-opengl-es-2-0-shaders
+        if (
+            !(color.r <= 0.0 || 0.0 <= color.r) &&
+            !(color.g <= 0.0 || 0.0 <= color.g) &&
+            !(color.b <= 0.0 || 0.0 <= color.b)) {
+            color.a = 0;
+        }
+        color.r = !(color.r <= 0.0 || 0.0 <= color.r) ? min($clim.x, $clim.y) : color.r;
+        color.g = !(color.g <= 0.0 || 0.0 <= color.g) ? min($clim.x, $clim.y) : color.g;
+        color.b = !(color.b <= 0.0 || 0.0 <= color.b) ? min($clim.x, $clim.y) : color.b;
         color.rgb = clamp(color.rgb, min($clim.x, $clim.y), max($clim.x, $clim.y));
         color.rgb = (color.rgb - $clim.x) / ($clim.y - $clim.x);
         return max(color, 0.0);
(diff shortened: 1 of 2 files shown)


vispy/visuals/image.py
```diff
@@ -89,12 +89,26 @@
 
 _apply_clim_float = """
     float apply_clim(float data) {
+        if (!(color.r <= 0.0 || 0.0 <= color.r)) {
```
almarklein, line 92: Just curious why `(color.r == color.r)` would not work for isnan. I guess the compiler optimizes it away? Anyway, what you have here seems the most suggested solution.

almarklein:
> For RGB images, if all pixels are NaNs then make them transparent, otherwise make them the lower clim limit.

I guess this makes sense. Is this what MPL does too?

vispy/visuals/image.py
djhoese: Yes. I think that's what I discovered when I originally did this and is why I went with this stackoverflow solution.

djhoese:
It does the lower-limit thing if clims are equal. I'm actually not sure if MPL can handle floating point RGBs. I've only ever done 8-bit unsigned integers I think.

djhoese:
Just tested MPL and for a single band of NaNs for floating point input `imshow` will map to `0`. So when RGB are all NaN then the image is black (0, 0, 0). Interestingly, for single channel data that are colormapped the NaNs become completely transparent (white by default).

I'm realizing the RGB handling I put in this PR works better for my multi-band image visual I'm working on in my own application, but not here. I may just need to keep a custom shader for my stuff. I also realized that I'm overwriting `.a` in this PR when the user could have provided RGBA data. So proper solutions...

Another option would be to update the colormaps' shader code to handle NaN values and make them fully transparent. This is completely flexible, but most colormaps are based on code with three different shader implementations.

## Solutions

So for single band data NaNs need to be made transparent. I think the `discard` as I have it is the easiest way. Colormap changes are another option, but it isn't as flexible and may not cover all cases.

For RGB/RGBA data, I think I just need to remove my alpha `if` statement but add another ternary for a NaN check on the alpha. Not sure if OpenGL allows for alpha to be NaN given precision differences between `.rgb` and `.a`, but then the NaN check should pass.

@almarklein Thoughts? Other ideas?

djhoese:
Alright I just ended up doing it. I'll add a test tomorrow if you agree with this behavior.

djhoese:
I really need to stop making commits late at night and sleep deprived from the baby. Will add tests in an hour or so. @almarklein let me know what you think. @kmuehlbauer or @asnt interested in your opinion as well.

almarklein (approved):
Looks good! I agree that we should probably follow MPL in the RGB case. Setting alpha to zero if it's nan also makes sense.

djhoese:
@almarklein I'm finishing up the tests locally and something I've discovered is that, at least for single band float data, the gloo-level texture code (maybe OpenGL itself) converts the NaNs to 0s. So for CPU-based scaling textures it isn't possible to detect the NaNs in the shader and since the colormapping gets applied in the shader I think it has to be this way. So GPU scaled textures will be able to match MPL, but CPU scaled won't unless we convert the data further.

Maybe this will make more sense once I push the tests...

almarklein:
Right. This makes sense, because the float data is converted to uint8 right before it's uploaded to the GPU. I suspect that it is in this step where NaN becomes 0. For RGB this is fine, should be same behavior, I guess? For grayscale I guess we'll have to live/document with it.

almarklein (approved):
Nice and clear tests!

djhoese:
Great...a test that passes locally fails on CI.

djhoese:
@almarklein I'm kind of leaning towards skipping this test on CI. Other ideas?

almarklein:
> On the CI system it is opaque black which I'm guessing is how it interprets the discard statement...

Sounds like this could also be the result of NaN being interpreted as zero? I'm a bit worried that if CI produces this result, other systems might as well...

djhoese:
This isn't comprehensive/exhaustive, but I've used these types of NaN checks on R32F textures in my SIFT application for years. Now we don't have a ton of input data with NaNs in them, but the tiles I talked about would have NaNs on the edges and they haven't been an issue. It does seem however that the OpenGL spec doesn't define behavior for NaN handling:

https://stackoverflow.com/a/33885182/433202

Also, I guess I should have suggested an xfail on this test instead of a skip. Then at least the test is run.

djhoese:
To add to that, I think if a GPU/OpenGL drivers decide to return 0 for NaN data then that isn't the worst fallback. I could add a comment about that in the docstring of the ImageVisual.

almarklein:
>  I've used these types of NaN checks on R32F textures in my SIFT application for years.

Ok that's good to know.

> I think if a GPU/OpenGL drivers decide to return 0 for NaN data then that isn't the worst fallback.

That is very true :)

Ok, assume this is rather CI-specific and get this merged.

PR #2110, opened 2021-06-27
Fix inconsistent picking behavior regarding depth testing

### LINKED ISSUE #2107: Picking broken (visual_at)
While testing current main I've found that the picking example doesn't work anymore.

examples/demo/scene/picking.py

@djhoese just confirmed on gitter chat, that it doesn't work for him too.

I'll try to bisect this tomorrow, but if anyone is faster, please do so.

djhoese:
Closes #2107

See #2107 for the long debug process and discussion on this. Basically the MeshVisual used by the ViewBox (and all Widget objects) only has `cull_face` set, but depends heavily on the behavior/setting of depth testing. This PR forces depth testing to be off which seems to make this picking Mesh behave more consistently with ViewBox child Visuals.

TODO:

- [x] Add a test reflecting at least one of the cases in #2107.

### DIFF
--- a/vispy/scene/tests/test_canvas.py
+++ b/vispy/scene/tests/test_canvas.py
@@ -75,7 +75,7 @@ def test_picking_basic():
         view.camera = 'panzoom'
 
         x = np.linspace(0, 400, 100)
-        y = np.linspace(-144.1, -9.44, 100)
+        y = np.linspace(0, 200, 100)
         line = scene.Line(np.array((x, y)).T.astype(np.float32))
         line.interactive = True
         view.add(line)
@@ -83,6 +83,7 @@ def test_picking_basic():
 
         c.render()  # initial basic draw
         for _ in range(2):  # run picking twice to make sure it is repeatable
+            # get Visuals on a Canvas point that Line is drawn on
             picked_visuals = c.visuals_at((100, 25))
             assert len(picked_visuals) == 2
             assert any(isinstance(vis, scene.ViewBox) for vis in picked_visuals)
--- a/vispy/scene/tests/test_canvas.py
+++ b/vispy/scene/tests/test_canvas.py
@@ -60,3 +60,30 @@ def test_canvas_render(blend_func):
         else:
             # the alpha should have some transparency
             assert (rgba_result[..., 3] != 255).any()
+
+
+@requires_application()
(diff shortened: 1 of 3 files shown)


djhoese:
This test was actually pretty easy to write. Tests are great. Let's use more of those. :stuck_out_tongue_winking_eye:

djhoese:
Also note that the test fails without the Widget change included in this PR and passes with it. I also noticed a weird definition of the `.draw` method for the MeshVisual that as far as I can tell will never do anything.

vispy/scene/tests/test_canvas.py
```diff
@@ -60,3 +60,30 @@ def test_canvas_render(blend_func):
...
+    Based on https://github.com/vispy/vispy/issues/2107.
+
+    """
+    with TestingCanvas(size=(125, 125), show=True, title='run') as c:
+        view = c.central_widget.add_view()
+        view.margin = 5  # add empty space where there are no visuals
+        view.camera = 'panzoom'
+
+        x = np.linspace(0, 400, 100)
+        y = np.linspace(-144.1, -9.44, 100)
```
asnt, line 78: Do these specific numbers matter? I'm not sure I can relate this directly to the picking position `(100, 25)` below.

asnt (commented):
I have one comment where I don't undestand otherwise this looks good.

vispy/scene/tests/test_canvas.py
djhoese: Oh, right. These are based on the min/max values of the data used for the first line in that picking example. Theoretically they should be arbitrary. The visuals_at position is a canvas position, not a data/array coordinate. It was just found manually as a Canvas position that has the line drawn on it.

almarklein (approved):
LGTM. Nice work on that additional test!

Would be good to eventually prevent state from "leaking" in the first place.

almarklein:
>> Would be good to eventually prevent state from "leaking" in the first place.

> I think that's an OpenGL issue more than a VisPy issue, but yeah.

I think it's an OpenGL issue that we can deal with in Vispy :)

asnt (approved):
:+1:

tlambert03 (approved):
impressive sleuthing!

PR #2111, opened 2021-06-28
Change Visual GL state so it is only set if drawing

### LINKED ISSUE #2109: LineVisual line smooth and width set after GL state is configured
In #2082 @almarklein updated the line visual to call `self.update_gl_state` to change `line_smooth` and `line_width`. The problem is, as I just realized while debugging #2107, that `update_gl_state` doesn't have an immediate effect on GL state. Only when `Visual._configure_gl_state` does the lower-level `gloo.set_state` call get executed. That happens on every `Visual.draw` call *BUT* before `_prepare_draw` is called. `_prepare_draw` is when these line smooth/width changes happen.

https://github.com/vispy/vispy/blob/184d854c7542f4c6df21bb4c7db3dd5b340dd900/vispy/visuals/visual.py#L442-L444

djhoese:
Closes #2109 

As discussed in #2109, this moves when the GL state is applied/sent to the GPU. This is things like depth test, blend functions, etc. Previously this was done *before* `_prepare_draw` of the individual Visual was called. If `_prepare_draw` returns `False` then a `Visual` isn't drawn and this is wasted. This PR is currently just a test to see what happens if we do this change. In the best case this passes all tests and is a slight performance improvement. In the next best case it only effects the `TextVisual` which has some complicated rendering before "real" drawing. Worst case is there are settings that reset GPU state (like clear color?) and doing these after texture, buffer, and other data commands will cause issues. In this last case, this can't be merged.

### DIFF
--- a/vispy/visuals/visual.py
+++ b/vispy/visuals/visual.py
@@ -439,13 +439,14 @@ class Visual(BaseVisual):
     def draw(self):
         if not self.visible:
             return
-        self._configure_gl_state()
         if self._prepare_draw(view=self) is False:
             return
 
         if self._vshare.draw_mode is None:
             raise ValueError("_draw_mode has not been set for visual %r" %
                              self)
+
+        self._configure_gl_state()
         try:
             self._program.draw(self._vshare.draw_mode,
                                self._vshare.index_buffer)


almarklein (approved):
This makes sense. CI is green. If the examples look alight this should be good.

PR #2118, opened 2021-06-30
fix: the FOV in BaseCamera is conflicting with Base3DRotationCamera...

vitsensei:
…and possibly other camera classes as well

In Base3DRotationCamera, changing the FOV through mouse event will be handle by viewbox_mouse_event(). With this line: self.fov = min(180.0, max(0.0, fov)), it is possible to have self.fov = 180 when fov = 180. And this will create an exception since BaseCamera class does not allow fov >= 180.

### DIFF
--- a/vispy/scene/cameras/base_camera.py
+++ b/vispy/scene/cameras/base_camera.py
@@ -218,7 +218,7 @@ class BaseCamera(Node):
     @fov.setter
     def fov(self, fov):
         fov = float(fov)
-        if fov < 0 or fov >= 180:
+        if fov < 0 or fov > 180:
             raise ValueError("fov must be between 0 and 180.")
         self._fov = fov
         self.view_changed()


djhoese:
Thanks for this. Any idea if there is a code example where I can use one of the cameras and see the effects of this? Are you actually receiving this error? Should the `0` check also be `>=`? Does the `fov` need to be less than 180 for some of the math in the camera code?

djhoese:
Is there some way that a test can be written for this?

vitsensei:
Use this code:
```python
from PyQt5 import QtWidgets
from vispy import scene, color
import numpy as np

class Plotter3D(QtWidgets.QWidget):
    def __init__(self, parent):
        # Create a canvas to draw
        super().__init__(parent)

        canvas = scene.SceneCanvas(keys=None, bgcolor=color.Color(color="white"))
        canvas.create_native()
        canvas.native.setParent(self)

        layout = QtWidgets.QHBoxLayout()
        layout.setContentsMargins(0, 0, 0, 0)

        layout.addWidget(canvas.native)
        self.setLayout(layout)

        self.view = canvas.central_widget.add_view()
        self.view.camera = scene.cameras.TurntableCamera()

        # Add axes
        self.view.add(scene.visuals.XYZAxis())

        # Markers visual that store the point clouds
        self._scatter = scene.visuals.Markers()
        self.view.add(self._scatter)

    def plot(self, cloud):
        self._scatter.set_data(cloud[:, :3], edge_color=(1, 1, 1, 1), face_color=(0, 0, 1, 1), size=1)

    def clear_plot(self):
        self._scatter.set_data(np.ones((0, 3)))


if __name__ == '__main__':
    import sys

    app = QtWidgets.QApplication(sys.argv)

    plotter = Plotter3D(None)
    plotter.show()

    app.exec()
```

`Shift + Right mouse + Dragging` should let you to control the FOV. I'm not sure how to write a proper test for this...

`Should the 0 check also be >=? Does the fov need to be less than 180 for some of the math in the camera code?`
I don't see any weird effect when fov = 0 so I guess not? Either way, they just need to align with whatever other camera module is doing. 

Another way to deal with this is instead of throwing an exception in BaseCamera, we can reset the fov to something close to the FOV limit (0.1, 179.9). Whatever makes sense the most I guess.

almarklein:
A fov of 180 makes no sense, mathematically, so they should be avoided somehow. It's a bit hard decide whether to limit to e.g. 179 or 179.999, so my feeling is that the actual camera implementation should handle this. E.g. the `Base3DRotationCamera` both has the logic to set the fov on user interaction, and does the math with it.

This is similar to a fov of 0, which arguably makes a bit more sense, since the `PerspectiveCamera` (a base class of  `Base3DRotationCamera`) interprets going towards zero as becoming more like an orthographic camera. Still, a fov of zero would not work for the math, so it makes sure it does not reach zero:
https://github.com/vispy/vispy/blob/main/vispy/scene/cameras/perspective.py#L166

In summary, I agree with the proposed change. Ideally I'd like to see the fov limited to say 179.9 right where the zero is avoided (in the link above).

djhoese:
@almarklein I did a quick test and it looks like you'd also need to modify the limits on this check:

https://github.com/vispy/vispy/blob/acafcb07a0c9947c319e1611d72cf7045670a8e7/vispy/scene/cameras/perspective.py#L261

And the change you were thinking for 179.9 in the PerspectiveCamera also needs to be done in the Base3DRotationCamera which overrides that method:

https://github.com/vispy/vispy/blob/acafcb07a0c9947c319e1611d72cf7045670a8e7/vispy/scene/cameras/perspective.py#L306

Also I just realized @vitsensei that you're removing the `=` in the comparison :facepalm:. Sorry for the confusion. If I try the change in this PR and not the `179.999` changes suggested by me and @almarklein then when I get to FOV 180 in the example the visual disappears. I'd say that's unexpected.

I'm a little confused on where is best for these changes though. The BaseCamera currently has the fov property, the PerspectiveCamera doesn't allow for an fov of 0, but the Base3DRotationCamera does allow for an fov of 0. I also just realized the BaseCamera initializes fov to 0. This all just seems very inconsistent.

vitsensei:
@almarklein 
Pragmatically, it's quite likely the user won't have an use case for FOV = 0. So I guess to avoid complexity in the math, setting the limit to be `0.1 <= fov <= 179.9` is enough?

almarklein:
Well, the `Base3DRotationCamera` does suppot fov=0 (it'll use the math for an orthographic transform then). I had overlooked this earlier, thanks @djhoese for pointing this out. In other words, fov=0 *can* mean something useful in a camera, and therefore I think these 0/180 cases should be handled by the camera subclasses.

djhoese:
So @almarklein what you're saying is that we should merge this and see what breaks? And anything that does should be fixed in subclasses in separate PRs?

almarklein:
I'm saying that IMO the base class should have the range [0..180], and that we should handle the fov being exactly 0 or 180 in the subclasses. Can be as simple as `fov = min(self.fov, 179.9)`. Could be in this PR or another.

Issue #2104, opened 2021-06-25
Request for testers for VisPy 0.7


djhoese:
Hi everyone! We are getting ready to release VisPy 0.7. This is another **HUGE** release and includes changes anywhere from performance improvements, to obvious bug fixes, to near-total rewrites of some of the major Visuals like the ImageVisual, VolumeVisual, and MeshVisual. We're looking for anyone who can help us test this new version and let us know of any unexpected results or issues. We have only a few small changes remaining, but they shouldn't effect the basic testing needed for this release.

I'm hoping at least some of the napari folks will have time in the next day or two to do some minimal testing with napari.

CC @jni @sofroniewn @alisterburt @tlambert03 @kne42 

Thank you all for any help you can provide.

vitsensei:
Is this something that people who are not heavily involved in the project can help with?
I'm happy to help during this lock down!

djhoese:
@vitsensei All help is appreciated!

sofroniewn:
Both @alisterburt and I have down done some fairly extensive usage and testing in napari with vispy master (0.7) and no longer have any problems, so i think we'll be good for a seamless transition. I'm only using vispy master (0.7) locally now. If you make an `0.7.0rc0` release candidate available on PyPI we will test against that in our CI test against upstream release candidates. That would probably be best for us as then we'll really make sure all systems and all tests are passing. Thanks for all the hard work here @djhoese! I think we'll be able to trim quite some vendored napari code after the release too!!

kmuehlbauer:
Not sure if I suggested this before, but we could think about uploading every push to main to `testpypi`. So everyone has easy access to published commits via `testpypi`.

djhoese:
@kmuehlbauer does testpypi accept the version numbers we're generating with commit hashes in them? Does pip know what to do with that? PRs for the github workflow are welcome and something added to the contributor's guide would be nice.

kmuehlbauer:
@djhoese I'm doing this in one of the packages I maintain. Not with commit hashes, but that should work too. I'll get to this the next days.

djhoese:
One thing that will be a little awkward is that our setuptools_scm usage can only detect some tags...it seems. So even though we have a 0.6.6 release of VisPy, it only sees the 0.6.5 tag on my local system. Regardless, this would likely mean the version numbers going to test PyPI might be something like: `vispy-0.6.5.dev718+g79c8f2ab`.

[...]

PR #2121, opened 2021-06-30
Add CI job for uploading to test PyPI

djhoese:
As discussed in #2104 we should upload test packages to test PyPI. This is my attempt at getting the workflows setup for this. @kmuehlbauer how does this look?

CC @sofroniewn


kmuehlbauer (approved):
@djhoese LGTM!

PR #2134, opened 2021-07-07
Set stacklevel for colormap deprecation.

Carreau:
This allows to get proper file/line for deprecation in test suite, which
make it a bit easier to track where the errors come from.

Also update a literal usage of get_colormap('cubehelix') with the
corresponding new import.

### DIFF
--- a/vispy/color/colormap.py
+++ b/vispy/color/colormap.py
@@ -1132,9 +1132,13 @@ def get_colormap(name, *args, **kwargs):
     if name in _colormaps:  # vispy cmap
         cmap = _colormaps[name]
         if name in ("cubehelix", "single_hue", "hsl", "husl", "diverging", "RdYeBuCy"):
-            warnings.warn(f"Colormap '{name}' has been deprecated. "
-                          f"Please import and create 'vispy.color.colormap.{cmap.__class__.__name__}' "
-                          "directly instead.", DeprecationWarning)
+            warnings.warn(
+                f"Colormap '{name}' has been deprecated since vispy 0.7 . "
+                f"Please import and create 'vispy.color.colormap.{cmap.__class__.__name__}' "
+                "directly instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
 
     elif has_matplotlib():  # matplotlib cmap
         try:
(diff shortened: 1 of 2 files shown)


djhoese:
Thanks for this! Do you have an example of the difference that the stack level products? Just curious.

Carreau:
> Thanks for this! Do you have an example of the difference that the stack level products? Just curious.

Ah, hum, not on this machine, but roughly here is a deprecation warning in napari test suite:

```
/tmp/.tox/py38-linux-pyside/lib/python3.8/site-packages/vispy/color/colormap.py:1135: DeprecationWarning: Colormap 'cubehelix' has been deprecated. Please import and create 'vispy.color.colormap.CubeHelixColormap' directly instead.
      warnings.warn(f"Colormap '{name}' has been deprecated. "
```

And you get the same but roughly from memory:

```
/whatver/path/to/what/call/this/function.py:xxxx: DeprecationWarning: Colormap 'cubehelix' has been deprecated. Please import and create 'vispy.color.colormap.CubeHelixColormap' directly instead.
      warnings.warn(f"Colormap '{name}' has been deprecated. "
```

(which is how I found the "cubehelix" literal in mesh.py).

It also allow you to better filter warnings [like so](https://github.com/napari/napari/blob/804cebee7c0f60fcec1513f5908638c8330deaa2/pyproject.toml#L84-L95). After the 3rd colon you can give a module glob pattern, with no stacklevel the warning is seen in vispy, with the stacklevel=2 the module will vary depending on the caller, which seem more accurate.

almarklein (approved):

vispy/color/colormap.py
```diff
@@ -1132,9 +1132,13 @@ def get_colormap(name, *args, **kwargs):
     if name in _colormaps:  # vispy cmap
         cmap = _colormaps[name]
         if name in ("cubehelix", "single_hue", "hsl", "husl", "diverging", "RdYeBuCy"):
-            warnings.warn(f"Colormap '{name}' has been deprecated. "
-                          f"Please import and create 'vispy.color.colormap.{cmap.__class__.__name__}' "
-                          "directly instead.", DeprecationWarning)
+            warnings.warn(
+                f"Colormap '{name}' has been deprecated since vispy 0.7 . "
```
djhoese, line 1136:
```suggestion
                f"Colormap '{name}' has been deprecated since vispy 0.7. "
```

djhoese (approved):

djhoese:
Thanks again!

vispy/color/colormap.py
Carreau: +1, I never know how to treat version numbers at end of sentences.

PR #2135, opened 2021-07-08
Dont warn about missing networkx until it's actually used

### LINKED ISSUE #2132: Remove warning on import about missing networkx dependency
Currently, importing vispy (or some submodule) without networkx installed results in a warning about installing networkx, even if the user/downstream library has no intention of using the graph layout. The warning should be delayed and turned into an error, as noted by @djhoese [here](https://github.com/napari/napari/issues/2979#issuecomment-874159877).

xref  napari/napari#2979

almarklein:
Closes  #2132

cc @jni

### DIFF
--- a/vispy/visuals/graphs/layouts/networkx_layout.py
+++ b/vispy/visuals/graphs/layouts/networkx_layout.py
@@ -6,9 +6,6 @@ import numpy as np
 try:
     import networkx as nx
 except ModuleNotFoundError:
-    import warnings
-    warnings.warn(
-        "Networkx not found, please install network to use its layouts")
     nx = None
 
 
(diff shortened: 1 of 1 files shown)


kmuehlbauer (approved):

djhoese (commented):
Two thoughts:

1. Do we need to also have an `if nx is None` check on line 88 in the `adj` property?
2. I find it cleaner to do `if nx is None: raise ValueError(...)` and then you don't need an `else` branch and everything is de-dented one level less.

almarklein:
> I find it cleaner to do if nx is None: raise ValueError(...) and then you don't need an else branch and everything is de-dented one level less.

I completely agree. I moved the check up, outside the if-statement. This way this class can only be instantiated if `networkx` is available (and we don't have to add a check in the `.addj` property.

PR #1975, opened 2021-01-20
Add workaround for MacOS dlopen

rayg-ssec:
Adapted stackoverflow monkeypatch to ctypes.util.find_library for macOS 11 (Big Sur).
This gates off OpenGL.GL dlopen failure and uses a 2-entry lookup for quartz and objc exception cases. 
Should be backward compatible and is likely to work on both x86 and and m1 macOS.
Tested on x86 macOS 11.1 with `alpha-pyside2.py` test case.

PR #2137, opened 2021-07-10
Fix for Apple Silicon M1 (arm64)

psobolewskiPhD:
As requested by @djhoese here:
https://github.com/vispy/vispy/pull/1975#issuecomment-873084868

In order to have vispy run on Apple Silicon M1 (arm64) some functions marked as OBJC_ARM64_UNAVAILABLE need to be screen via arch
See also: 
https://github.com/pyglet/pyglet/pull/335/commits/5aa1870188b9675cdee090bbecb342661ccc06a6
This fix was originally proposed by @isuruf here:
https://github.com/vispy/vispy/issues/2013#issuecomment-819690863
And then expanded by @watjurk here:
https://github.com/vispy/vispy/issues/2013#issuecomment-819724112
However, the first two lines they also modified:
```
objc.class_getClassMethod.restype = c_void_p 
 objc.class_getClassMethod.argtypes = [c_void_p, c_void_p] 
```
seem to not need skipping on M1.

Using this fix, napari (python image viewer, etc.) runs just fine. 
Without, each of those functions results in "symbol not found" errors.

Disclaimer: I don't really understand this fully, but I've been using this fix for a week and everything seems fine in napari—I have no other vispy using software/code.

### DIFF
--- a/vispy/ext/cocoapy.py
+++ b/vispy/ext/cocoapy.py
@@ -176,8 +176,9 @@ objc.class_getIvarLayout.argtypes = [c_void_p]
 objc.class_getMethodImplementation.restype = c_void_p
 objc.class_getMethodImplementation.argtypes = [c_void_p, c_void_p]
 
-objc.class_getMethodImplementation_stret.restype = c_void_p
-objc.class_getMethodImplementation_stret.argtypes = [c_void_p, c_void_p]
+if platform.machine() != "arm64":
+    objc.class_getMethodImplementation_stret.restype = c_void_p
+    objc.class_getMethodImplementation_stret.argtypes = [c_void_p, c_void_p]
 
 objc.class_getName.restype = c_char_p
 objc.class_getName.argtypes = [c_void_p]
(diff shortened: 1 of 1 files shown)


almarklein (approved):

djhoese (approved):
Looks good to me. Thanks! Since I don't have a mac anymore I depend on others to make and test these kinds of fixes so thank you very much for doing this.

djhoese:
@kmuehlbauer Doesn't conda-forge build for arm64 now? Maybe we should have a test import of this module added to the conda-forge recipe when building for arm64. Theoretically if we add it to the conda-forge recipe now it should fail until we release this PR in vispy 0.7.1. Right?

kmuehlbauer:
@djhoese It should already be built for arm64. I think @hmaarrfk merged https://github.com/conda-forge/vispy-feedstock/pull/28 a while ago.

kmuehlbauer:
Nevermind, didn't read you comment properly. But yes, it should work the way you describe

PR #2140, opened 2021-07-13
Fix scalable textures clim_normalized when auto clims are used

### LINKED ISSUE #2139: Fix scalabe textures using clim_normalized when "auto" limits are used
If you create an ImageVisual with clims and then set clims to "auto" later, the ImageVisual fails. Here's an updated image test showing the use case

```python
@requires_application()
@pytest.mark.parametrize('is_3d', [True, False])
def test_image_auto_clim(is_3d):
    """Test image visual"""
    size = (100, 50)
    with TestingCanvas(size=size, bgcolor='w') as c:
        image = Image(cmap='grays', clim=[0, 1], parent=c.scene)
        shape = (size[1]-10, size[0]-10) + ((3,) if is_3d else ())
        np.random.seed(379823)
        data = np.random.rand(*shape)
        image.set_data(data)
        assert_image_approved(c.render(), "visuals/image%s.png" %
                              ("_rgb" if is_3d else "_mono"))

        # change to auto clims later
        image.clim = "auto"
        assert_image_approved(c.render(), "visuals/image%s.png" %
                              ("_rgb" if is_3d else "_mono"))
```

Here's the code in the ImageVisual that fails:

```python
    @clim.setter
    def clim(self, clim):
        if self._texture.set_clim(clim):
            self._need_texture_upload = True
        # shortcut so we don't have to rebuild the whole color transform
        if not self._need_colortransform_update:
            self.shared_program.frag['color_transform'][1]['clim'] = self._texture.clim_normalized
        self.update()
```

So because no other color transform changes have been done, we take the shortcut to assign `clim_normalized`. However, the data hasn't actually been set yet so auto clims haven't been calculated. So `clim_normalized` fails when it tries to split the clims into two elements (min, max).

I'm working on a PR with this updated test. My guess is that the easiest fix is to just update this colortransform if statement with an "auto" check...or maybe try/except with a new exception in clim_normalized. I'll make the PR...

djhoese:
Closes #2139 

I'm not super happy with the try/except in the ImageVisual, but also it seems like the most flexible and future-proof.

### DIFF
--- a/vispy/visuals/_scalable_textures.py
+++ b/vispy/visuals/_scalable_textures.py
@@ -122,9 +122,9 @@ class _ScaledTextureMixin:
 
         """
         if isinstance(self.clim, str) and self.clim == "auto":
-            raise RuntimeError("Can't return normalized color limits until "
-                               "data has been set. Call 'scale_and_set_data' "
-                               "first.")
+            raise RuntimeError("Can't return 'auto' normalized color limits "
+                               "until data has been set. Call "
+                               "'scale_and_set_data' first.")
         if self.clim[0] == self.clim[1]:
             return self.clim[0], np.inf
         # if the internalformat of the texture is normalized we need to
@@ -273,9 +273,9 @@ class CPUScaledTextureMixin(_ScaledTextureMixin):
 
         """
         if isinstance(self.clim, str) and self.clim == "auto":
-            raise RuntimeError("Can't return normalized color limits until "
-                               "data has been set. Call 'scale_and_set_data' "
-                               "first.")
+            raise RuntimeError("Can't return 'auto' normalized color limits "
+                               "until data has been set. Call "
+                               "'scale_and_set_data' first.")
 
         range_min, range_max = self._data_limits
         clim_min, clim_max = self.clim
--- a/vispy/visuals/_scalable_textures.py
+++ b/vispy/visuals/_scalable_textures.py
@@ -114,7 +114,19 @@ class _ScaledTextureMixin:
     @property
     def clim_normalized(self):
         """Normalize current clims to match texture data inside the shader.
+
+        If data is scaled on the CPU then the texture data will be in the range
+        0-1 in the _build_texture() method. Inside the fragment shader the
+        final contrast adjustment will be applied based on this normalized
+        ``clim``.
+
         """
+        if isinstance(self.clim, str) and self.clim == "auto":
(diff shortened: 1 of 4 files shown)


vispy/visuals/_scalable_textures.py
```diff
@@ -114,7 +114,19 @@ def set_clim(self, clim):
...
+        If data is scaled on the CPU then the texture data will be in the range
+        0-1 in the _build_texture() method. Inside the fragment shader the
+        final contrast adjustment will be applied based on this normalized
+        ``clim``.
+
         """
+        if isinstance(self.clim, str) and self.clim == "auto":
+            raise RuntimeError("Can't return normalized color limits until "
+                               "data has been set. Call 'scale_and_set_data' "
+                               "first.")
```
almarklein, line 127: Maybe include the word "auto" in this error message to make it clear that it's related to auto-clim?

vispy/visuals/image.py
```diff
@@ -348,11 +348,17 @@ def clim(self):
...
+        self._update_colortransform_clim()
         self.update()
 
+    def _update_colortransform_clim(self):
+        # shortcut so we don't have to rebuild the whole color transform
+        try:
+            if not self._need_colortransform_update:
+                self.shared_program.frag['color_transform'][1]['clim'] = self._texture.clim_normalized
+        except RuntimeError:
+            return
```
almarklein, line 360: Maybe we can be a bit more specific and only put the `self._texture.clim_normalized` part in the try-except. Right now a RunTimeError raised by the program is also caught.

almarklein (commented):
Nice catch!

djhoese:
I think I've addressed your comments @almarklein

vispy/visuals/image.py
```diff
@@ -348,11 +348,21 @@ def clim(self):
...
 
+    def _update_colortransform_clim(self):
+        if self._need_colortransform_update:
+            # we are going to rebuild anyway so just do it later
+            return
+        try:
+            norm_clims = self._texture.clim_normalized
+        except RuntimeError:
+            return
+        else:
```
djhoese, line 362: This `else` isn't really needed. I'm discussing this with some colleagues about other ways I could do this try/except/else. Not sure what is cleanest.

almarklein: Looks clean enough to me. Removing the `else` and dedenting the nextlines would also be fine.

almarklein (approved):

PR #2142, opened 2021-07-14
Jupyter widget based on jupyter_rfb

### LINKED ISSUE #134: Online server-side browser backend
This online web backend generates PNG images on the server, and streams binary images to the browser via WebSockets. In return, the client streams user actions to the server.

This server-side online backend may be useful for heavy visualizations based on huge amounts of data. In effect, it is more efficient to transfer small PNG images instead of huge datasets.

Assigned to @mfkaptan (GSoC 2014).

### LINKED ISSUE #2127: New Jupyter widget
I'm working on a new implementation of the Jupyter widget. Basically a revival of the "vnc appproach", where the rendering takes place at the server, png's are send to the browser, and the browser sends events to the kernel.

The idea is to make this a separate project/library, that can also be used by other projects. This should also make Vispy's build process easier, if we decided to drop the webgl-based widget.

I have some progress on the code, and would like to put it in a repo. Right now I've called it `jupyrerfb` where `fb` stands for "frame buffer" (and the last `r` could stand for remote), because it's essentially a remote framebuffer solution. I think "vnc" refers too much to the specific application rather than the technology. Any other ideas for a name?

cc @rossant

almarklein:
This implements a new backend based on `jupyter_rfb`. I'll be updating jupyter_rfb at the same time, so it aint easy to give this is a spin yet. This is basically the same as what was intended with the `ipynb_vnc` backend, but all the "hard parts" are delegated to an external library.

Includes a new simple way to create a (global) hidden window just for having a GL context, plus an approach to render offscreen using an FBO. The event system integrates with asyncio, because that's what Jupyter uses (via Tornado).

Closes #2127, closes #134, relates to all issues tagged with the jupyter-widget tag.

Todo:

* [x] Basic implementation
* [x] Can simply use canvas as cell output  
* [x] Events (mouse, key, scroll, ...)
* [x] Timers
* [x] Test multiple canvases
* [x] Test using a widget as subwidget
* [x] Tests
* [x] Examples
* [x] ~Use a global context~ -> no, we need to use one context per canvas to avoid interference.
* [x] Check for remaining todos that I left behind
* [x] Update and release jupyter_rfb.
* [x] More tests?
* [x] Look into `_ipython_display_`.

### DIFF
--- a/vispy/app/backends/_ipynb_rfb.py
+++ b/vispy/app/backends/_ipynb_rfb.py
@@ -1,7 +1,6 @@
 import asyncio
 
 from ..base import BaseApplicationBackend, BaseCanvasBackend, BaseTimerBackend
-# from ...gloo.context import set_current_canvas
 from ...app import Timer
 from ...util import keys
 from ._offscreen_util import OffscreenContext, FrameBufferHelper
--- a/vispy/app/backends/_ipynb_rfb.py
+++ b/vispy/app/backends/_ipynb_rfb.py
@@ -18,24 +18,6 @@ else:
     which = "jupyter_rfb"
 
 
-# try:
-#     # Explicitly use default (avoid using test-app)
-#     _app = Application('default')
-# except Exception:
-#     _msg = 'ipynb_rf backend relies on a proxy backend'
-#     available, testable, why_not, which = False, False, _msg, None
-# else:
-#     # Try importing jupyter_rfb
-#     # todo: give good error message if jupyter_rfb is not available
-#     try:
-#         from jupyter_rfb import RemoteFrameBuffer
-#     except Exception as exp:
-#         available, testable, why_not, which = False, False, str(exp), None
-#     else:
-#         available, testable, why_not = True, False, None
-#         which = _app.backend_module.which
-
-
 # -------------------------------------------------------------- capability ---
 
 capability = dict(
--- a/vispy/app/backends/_ipynb_rfb.py
+++ b/vispy/app/backends/_ipynb_rfb.py
@@ -4,10 +4,8 @@ from ..base import (BaseApplicationBackend, BaseCanvasBackend,
                     BaseTimerBackend)
 from ...app import Timer
 from ...util import keys
-from ._offscreen_util import OffscreenCanvasHelper
+from ._offscreen_util import GlobalOffscreenContext, FrameBufferHelper
 
-
-# todo: can create a global GL context helper here too
 
 try:
     from jupyter_rfb import RemoteFrameBuffer
@@ -41,27 +39,26 @@ else:
 # -------------------------------------------------------------- capability ---
 
 capability = dict(
-    title=True,  # But it only applies to the dummy window :P
+    title=False,
     size=True,
     position=False,
-    show=True, # todo: can show/hide window?? -> should probably be false
+    show=False,
     vsync=False,
     resizable=True,
     decorate=False,
     fullscreen=False,
-    context=True,  # via the proxy backend
-    multi_window=True,  # todo: right?
+    context=True,  # the offscreen global context
+    multi_window=True,
     scroll=True,
-    parent=False,
+    parent=False,  # todo: we can make this work
     always_on_top=False,
 )
 
 
 # ------------------------------------------------------- set_configuration ---
 
-# todo: call this?
-def _set_config(c):
-    _app.backend_module._set_config(c)
+# The configuration mostly applies to the framebuffer. So if we'd want
+# to implement some of that, we'd probably have to apply it to the FBO.
 
 
 # ------------------------------------------------------------- application ---
@@ -71,26 +68,20 @@ class ApplicationBackend(BaseApplicationBackend):
 
     def __init__(self):
         super().__init__()
-        # self._proxy_app_backend = _app._backend
 
     def _vispy_get_backend_name(self):
-        # proxyname = self._proxy_app_backend._vispy_get_backend_name()
-        # return f'ipynb_rfb (via {proxyname})'
-        return f'ipynb_rfb'
+        return 'ipynb_rfb'
 
     def _vispy_process_events(self):
-        # return self._proxy_app_backend._vispy_process_events()
         raise RuntimeError("Cannot process events while asyncio event-loop is running.")
 
     def _vispy_run(self):
         pass  # We're in IPython; don't enter a mainloop or we'll block!
 
     def _vispy_quit(self):
-        # return self._proxy_app_backend._vispy_quit()
         pass
 
     def _vispy_get_native_app(self):
-        # return self._proxy_app_backend._vispy_get_native_app()
         return asyncio
 
 
@@ -103,7 +94,8 @@ class CanvasBackend(BaseCanvasBackend, RemoteFrameBuffer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args)
         # Init
-        self._helper = OffscreenCanvasHelper()
+        self._context = GlobalOffscreenContext()
+        self._helper = FrameBufferHelper()
         self._loop = asyncio.get_event_loop()
         self._draw_pending = False
         self._logical_size = 1, 1
@@ -124,13 +116,12 @@ class CanvasBackend(BaseCanvasBackend, RemoteFrameBuffer):
             self._physical_size = int(w * r), int(h * r)
             self._loop.call_soon(self._emit_resize_event)
         elif type == "pointer_down":
-            with self._helper:
-                self._vispy_mouse_press(
-                    native=ev,
-                    pos=(ev["x"], ev["y"]),
-                    button=ev["button"],
-                    modifiers=self._modifiers(ev),
-                )
+            self._vispy_mouse_press(
+                native=ev,
+                pos=(ev["x"], ev["y"]),
+                button=ev["button"],
+                modifiers=self._modifiers(ev),
+            )
         elif type == "pointer_up":
             self._vispy_mouse_release(
                 native=ev,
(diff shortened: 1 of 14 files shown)


almarklein:
Arg. I was seeing strange behavior when multiple canvases were active at the same time. Turns out that in vispy we make some assumptions about the context being local to the canvas, and I was using one global GL context. Now that I changed that, the problems are gone.

vispy/gloo/glir.py
```diff
@@ -844,8 +844,7 @@ def _gl_initialize(self):
             else:
                 this_version = this_version[0]
 
-            this_version = LooseVersion(this_version)
-            if this_version < '2.1':
+            if this_version and LooseVersion(this_version) < '2.1':
```
almarklein, line 847: I ran into `this_version` being an empty string, causing `LooseVersion` to fail.

djhoese: This is usually a sign that OpenGL isn't being loaded correctly or isn't being found. At least in my experience. It has been on the radar for a long time to improve the error message that is produced, but if we silently let it through then it may confuse some users. In what cases is the GL version being an empty string not a sign of something being wrong?

almarklein:
@djhoese any idea what's up with the test? The `test_plot.py` fails consistently with a segfault. And on my machine it does too.

djhoese:
It isn't test_plot, it is test_perspective

vispy/app/backends/_ipynb_rfb.py
```diff
@@ -0,0 +1,287 @@
...
+
+# ------------------------------------------------------------------ canvas ---
+
+class CanvasBackend(BaseCanvasBackend, RemoteFrameBuffer):
+
+    _double_click_supported = True
+
+    def __init__(self, *args, **kwargs):
+        RemoteFrameBuffer.__init__(self)
+        BaseCanvasBackend.__init__(self, *args)
```
djhoese, line 77: Not that I necessarily have a problem with this, but couldn't you swap the order in the class definition and let `RemoteFrameBuffer`'s use of `super().__init__` handle this? Or maybe that's dumb to depend on that?

almarklein: Oh nice idea! I think I defined it in this order so that the Vispy-subclass appends to the chain of `ipywidget.DomWidget` classes, instead of fiddling it in at the top of the "mro-chain". But I'll have a look at least. 

I think generally you should be able to depend on each class to call `super().__init__()` but we break that rule in vispy (as found out the hard way) because for qt we call the `__init__` depending on qt4/qt5. Maybe I should try to clean that up so that each subclass can behave as expected.

almarklein: I'm looking into this, but I think I'll just clean this up properly, and replace the explicit `SomeBaseClass.__init__()` calls to `super().__init__()`, which will affect most backends. Let's do that in another PR.

almarklein:
Ready for review!

vispy/app/canvas.py
```diff
@@ -512,6 +514,16 @@ def __repr__(self):
                 % (self.__class__.__name__,
                    self.app.backend_name, hex(id(self))))
 
+    def _ipython_display_(self):
```
djhoese, line 517: As mentioned in the meeting, I think this is the old interface for this. There is likely a `_repr_*_` equivalent for newer jupyter versions. Maybe https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

vispy/app/backends/tests/test_rfb.py
```diff
@@ -0,0 +1,5 @@
+# This currenly only tests that the backend exists and can be imported ...
+
+def test_rfb():
```
djhoese, line 3: Let's spend 10 minutes seeing if there is an easy way to test this. Maybe ipywidgets has a test fixture that we can use to fake the notebook.

djhoese: This looks complicated but I did find this for how ipywidgets does their tests: https://ipywidgets.readthedocs.io/en/latest/dev_testing.html

djhoese:
Let's rename this backend to `jupyter_rfb` (or other better name, but not based on ipython).

vispy/app/canvas.py
almarklein: IPywidget only implements this in bleeding edge - the latest release still uses `_ipython_display_`. See [this commit](https://github.com/jupyter-widgets/ipywidgets/commit/7ae97acc96ce9f78bd43dbd0778cd7db36804966#) that adds it (merged after latest release).

almarklein: Let's proxy both.

vispy/app/backends/tests/test_rfb.py
almarklein: Another option is to fake it by calling `handle_event()` and `get_frame()`. Basically assuming that `jupyter_rfb` works as it should, and testing that the backend implementation does what it must do. Then we can bypass the need ffor a jupyterlab/notebook instance

djhoese: Works for me. `unittest.mock` for the win (or pytest mock)

vispy/app/canvas.py
djhoese: The pull request was merged January of 2020: https://github.com/jupyter-widgets/ipywidgets/pull/2021
The latest release is January of 2021: https://pypi.org/project/ipywidgets/

almarklein: That's ... odd. What you say is true, but the 7.6.3 release contains the version from *before* that commit. I double-checked by opening the release tarball.

almarklein: Ah well ... it should work regardless of the version now.

djhoese: They have a `7.x` branch that I think they cherry-pick fixes/merges from the master branch. The master branch is their dev/0.8 branch currently I think. That's my best guess.

djhoese:
Ran out of time today to actually test this. Will try to do a more thorough review and test tomorrow.

almarklein:
* Renamed the backend
* Made it the default when in the notebook
* More tests

djhoese:
Maybe after we merge this I'll make another PR to remove the current vispy.js based widget from installation and update the installation instructions to not mention the fancy building. :thinking: I guess that is a pretty hard deprecation, but it is much simpler. It would also effect wheel building and tests as I think I had to put extra effort into getting `npm` installed. Oh! I can remove the stupid `js` git submodule. Any objections to any of this @almarklein @kmuehlbauer @larsoner?

almarklein:
Also maybe the `vispy.ipython` module.

djhoese:
@almarklein Without reading any instructions I cloned the jupyter_rfb repos, did `pip install -e .` in my environment, checked out this PR's branch, and ran jupyter lab. When trying to run an example I get:

> Error displaying widget: model not found

Do I need to reinstall vispy too?

almarklein:
@djhoese See the readme of jupyter_rfb. With a development install you need to:
```
pip install -e .
jupyter nbextension install --py --symlink --overwrite --sys-prefix jupyter_rfb
jupyter nbextension enable --py --sys-prefix jupyter_rfb
```

vispy/app/backends/tests/test_rfb.py
```diff
@@ -0,0 +1,77 @@
...
+        gloo.clear()
+
+
+@requires_application()
+def test_rfb_canvas():
+
+    try:
+        import jupyter_rfb
+    except ImportError:
+        return  # only the "all deps" build have jupyter_rfb
```
djhoese, line 33: This should be a pytest.skip. You could probably even put it at the module level: https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-test-functions

almarklein: You're right. I think I was in a lazy mood.

djhoese:
Ah I just found those before you posted that comment. I'm still getting the same error even though it clearly built the jupyter widget this time (a lot more log output). I'll keep looking around the jupyter directories to see what's going on.

vispy/app/backends/tests/test_rfb.py
almarklein: fixed.

djhoese:
> Uncaught (in promise) Error: Module jupyter_rfb, semver range ^0.1.0 is not registered as a widget module

in the web console (Firefox, but I also tried Chrome). I disabled the old vispy extension and still got this. Maybe I should try updating my jupyter lab...

almarklein:
Could it be that you'e also installed jupyter_rfb earlier. Maybe the old extension is lingering around ...

djhoese:
I don't think I did install it earlier.

The `nbextension` commands only deal with the notebook extension. The jupyter lab extension is separate. Running the Rotating Cube notebook with `jupyter notebook` works fine. Running from jupyter lab does not. So...the question is how isthe mybinder stuff in jupyter_rfb different from the existing installation instructions :thinking:

djhoese:
Ran:

```
$ jupyter labextension develop jupyter_rfb --overwrite
```

Based on the cookiecutter instructions and still a model error.

djhoese:
Got it! Disabling the old vispy extension wasn't enough, I had to uninstall it. I discovered a discussion where someone pointed out an old package that was requiring jupyterlab widgets 3.0 compatibility or something like that and so two extensions were conflicting. One was pulling in old jupyterlab widgets and one new versions and they were conflicting. I do think the above labextension command needs to be added to the dev instructions for jupyter_rfb.

djhoese:
At for some reason now it doesn't work. Will look after this meeting.

djhoese:
So if I comment out this line:

https://github.com/vispy/vispy/blob/0d941e087968f81412cf782dd95344cd8af527e1/vispy/__init__.py#L32

then it works. There must be something hanging around even though I uninstalled the extension that is getting used when this code runs. Anyway, if you are getting these issues @almarklein then maybe we should merge this and then get going on removing the old widget as soon as possible. Or at least the setup.py and on-import side effects.

almarklein:
> WTF, I uncommented that line and added a print statement to see if it was being run and the example notebook still works.

Sounds like fun! :/

djhoese:
I still have issues every once and a while that when I start jupyter lab the extension isn't installed properly or something. I say we merge this and get started on pulling the webgl backend and all the auto-enable/install ipython/jupyter extension stuff out of the vispy package. I don't think there is a way to do this without completely removing the webgl backend (depends on vispy.js-based git submodule and widget).

PR #2144, opened 2021-07-16
Add filter keyword arguments to subclassed filters

### LINKED ISSUE #2143: Add ability to pass "fpos" as a parameter to the ColorFilter
Hi all,
I am currently trying to use the ```ColorFilter``` (https://github.com/vispy/vispy/blob/main/vispy/visuals/filters/color.py) in a project along with several other filters, which I need to be placed in a specific order. However, right now, ```fpos``` cannot be passed as a parameter to ```ColorFilter```, which is always using 8:
```
    def __init__(self, filter=(1., 1., 1., 1.)):
        super(ColorFilter, self).__init__(fcode=self.FRAG_SHADER, fpos=8)

        self.filter = filter
```

Is it possible to change this so the user can specify any position for this filter?

Thanks so much,
Clare

clarebcook:
This PR will allow users to pass keyword arguments to the subclassed filters in ```color.py```.

Closes #2143

### DIFF
--- a/vispy/visuals/filters/color.py
+++ b/vispy/visuals/filters/color.py
@@ -60,8 +60,8 @@ class IsolineFilter(Filter):
         }
     """
 
-    def __init__(self, level=2., width=2.0, antialias=1.0, color='black'):
-        super(IsolineFilter, self).__init__(fcode=self.FRAG_SHADER)
+    def __init__(self, level=2., width=2.0, antialias=1.0, color='black', **kwargs):
+        super(IsolineFilter, self).__init__(fcode=self.FRAG_SHADER, **kwargs)
 
         self.level = level
         self.width = width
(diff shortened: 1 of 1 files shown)


djhoese (approved):
I'm not a huge fan of `**kwargs` but I'm not sure there is an easy way around it here. Looks good. Thanks for doing this!

djhoese:
@almarklein @kmuehlbauer what's your opinion on a change like this being a bug fix or a feature? Most importantly should it go in 0.7.2 or 0.8.0?

kmuehlbauer:
No strong opinion, but I would tend to bug fix.

djhoese:
That works for me Kai, thanks. @clarebcook thanks again for fixing this. Let's merge it and plan to include it in a 0.7.2 release in the next week or two.

clarebcook:
Wonderful, thank you!

almarklein:
> No strong opinion, but I would tend to bug fix.

+1

PR #2148, opened 2021-07-17
Switch examples and website gallery to sphinx-gallery

### LINKED ISSUE #2122: Add sphinx-gallery Gallery to the website
This is a replacement for #2024 as @larsoner pointed out we should just use sphinx-gallery instead of something custom.

The VisPy website's gallery is rather "custom" at this point: https://vispy.org/gallery.html

It would be nice if we could have:

1. Examples that are actively and automatically tested.
2. Run/tested on either every pull request/push in GitHub Actions or at least on pushes to the `main` branch.
3. Automatically generated one or more pages showing what the examples look like (screenshot) and linked to the code.

See https://github.com/vispy/vispy/wiki/Releasing-new-version-of-VisPy-python-package#update-documentation for how gallery images are run right now.

djhoese:
Depends on https://github.com/vispy/vispy-website/pull/40. See that PR for details.

Closes #2122


djhoese:
The tests pass!

djhoese:
So I tried updating the CI to do two things:

1. Upload the website output so it can be downloaded and viewed for debugging.
2. Use caching to save the download test data to speed up CI

I have the website upload working fine. I can download the zip, extract, and view it in the browser.

The caching though is...difficult. The hashing functions in GH Actions can only hash files inside the git workspace. I've updated the CI jobs with a custom config to save demo data and test data to a relative `vispy_data/data` and `vispy_data/test_data` directory. As far as I can tell I've got that all correct (I'm going to do some more debugging in a bit) *BUT* the caching action never seems to pick up on the new files because it gets the initial cache hit before all the jobs are executed and assumes it doesn't need to update anything.

larsoner:
We cache a versioned testing dataset in GH actions in MNE outside the source directory I think, and it seems to work. 

https://github.com/mne-tools/mne-python/blob/c1557623904877274607b6b7af7706c759d50571/.github/workflows/linux_pip.yml#L54

djhoese:
Yes, sorry, the caching can handle files outside the git workspace but the `hashingFiles` function I'm using in generating the cache key requires the files to be in the workspace. In MNE it looks like you have an environment variable that takes the place of this.

djhoese:
@larsoner @almarklein @kmuehlbauer So everything in the CI works, but I had to be a little dumb about the caching and just have a hard coded variable in there that can be updated if we want the cache to reflect the actual downloaded demo/test data. It isn't even actually as much data (~19MB in the examples case) as I thought. I thought about caching the sphinx-gallery intermediate output but was worried that it might cause issues if things need to be fully regenerated. Additionally we currently do a `make clean` and `make html` for the sphinx docs so I'd have to remove that (not a big deal, just noting). Could be something good for a future PR.

As-is, I think this PR is "good enough" for merging. It only includes the examples from the `examples/basics/scene` directory in the gallery, but this is much better than the few that we had there before that I think were mostly gloo-based. I could add more for the other sections to this PR, but was thinking that this is getting a little large so want to see what you guys think.

djhoese:
Also note you can download the website from the "build website" job in github actions and view the site locally if you'd like.

larsoner:
> Could be something good for a future PR.

What we do in MNE is on `main` build everything, and on PRs only build examples that were changed. This decreases the doc build time quite a bit. But it looks like you're only at 8 minutes to build the whole thing anyway, which I consider very fast. (This is faster than the build-no-examples MNE build. Same with SciPy. And it's way faster than matplotlib.)

> Also note you can download the website from the "build website" job in github actions and view the site locally if you'd like.

We should switch to CircleCI for the doc build, you can just directly view the output as an artifact, e.g.:

https://github.com/mne-tools/mne-python/pull/9598 leads to https://31429-1301584-gh.circle-artifacts.com/0/dev/index.html

This has been incredibly useful as a maintainer of packages

.github/workflows/main.yml
```diff
@@ -68,25 +68,84 @@ jobs:
...
     runs-on: ubuntu-latest
+    defaults:
+      # use login shell to make use of .bash_profile /.bashrc
+      run:
+        shell: bash -l {0}
     steps:
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.6.0
+      with:
+        access_token: ${{ github.token }}
```
larsoner, line 79: For this I would use concurrency, I implemented it a couple of days ago and it cancels immediately on queue rather than requiring the job to actually start:

https://github.com/mne-tools/mne-python/pull/9592/files

.github/workflows/main.yml
```diff
@@ -68,25 +68,84 @@ jobs:
...
       with:
         fetch-depth: 0
         submodules: "recursive"
-    - name: Set up Python
-      uses: actions/setup-python@v2
+    - name: Cache VisPy Demo and Test Data
+      uses: actions/cache@v2
+      env:
+        cache-name: cache-vispy-data-website
+        test-version: 1
```
larsoner, line 88: I think hard coding values is fine for now to get things to work. But really 19 MB is trivial to download each time, so maybe we should just scrap it?

ci/requirements/linux_min_deps_conda.txt
```diff
@@ -10,4 +10,5 @@ pytest-sugar
 scipy
 networkx
 hsluv
-kiwisolver
\ No newline at end of file
+kiwisolver
+sphinx-gallery
```
larsoner, line 14: SG should not be in the minimal deps, why did you have to add it? If it's for the tests I would just add this to the test file:
```
try:
    import sphinx_gallery
except Exception:
    pytest.skip("skipping sphinx-gallery tests", allow_module_level=True)
```
you can do something similar with decorators if you want

djhoese:
> We should switch to CircleCI for the doc build, you can just directly view the output as an artifact, e.g.:

I am not against this in principle but:

1. I have zero experience with CircleCI so it won't be me doing it.
2. I don't usually need to see the website from CI jobs unless examples are being added to the gallery and even then in my own PRs I've probably already run it locally. Downloading a zip file isn't a huge deal to me.
3. My research shows this is a requested feature (view individual files on github actions) but requires a lot of work on the backend and frontend so probably not going to get it any time soon.
4. I'm worried about the complexity of sharing OpenGL friendly environments and configurations between Actions and CircleCI. If it just works, then great.

larsoner:
> I have zero experience with CircleCI so it won't be me doing it.

If you (or someone else) ever feels like learning, you can more or less just take what we have for MNE and trim it quite a bit:

https://github.com/mne-tools/mne-python/blob/main/.circleci/config.yml

Then set up the redirector action:

https://github.com/mne-tools/mne-python/blob/main/.github/workflows/circle_artifacts.yml

FWIW I also find the caching on CircleCI to be much easier to implement and use...

> I don't usually need to see the website from CI jobs unless examples are being added to the gallery and even then in my own PRs I've probably already run it locally. Downloading a zip file isn't a huge deal to me.

I think this is just because you haven't experienced how nice it is :) For example, anytime someone changes a docstring, I get to look at the rendered output. If they mess up the RST (not uncommon), building with `-nWT --keep-going` they will get some (often) cryptic RST error but looking at the output it's easier to see what went wrong, etc.
 
> I'm worried about the complexity of sharing OpenGL friendly environments and configurations between Actions and CircleCI. If it just works, then great.

FWIW we've been building 3D example on CircleCI + Xvfb for years with MNE. We have more problems on GitHub actions + Xvfb than we do with CircleCI.

.github/workflows/main.yml
djhoese: Can you find a link to the `github.event.number` item? I can't seem to find it. What does it equal and what does it mean? My worry is that if I push a commit to `main` and then push a tag that the tag job would stop the commit job. The website upload job only happens on pushes (are tags considered pushes too?).

ci/requirements/linux_min_deps_conda.txt
djhoese: Yes it was for the tests. I guess that's a good point. Just because it is a test doesn't mean it needs to be run in every environment.

.github/workflows/main.yml
djhoese: Oh maybe I should try reading the links right above the concurrency change. :facepalm: 

I added this config in my last commit but it doesn't seem to have done anything (old jobs still running).

larsoner: > I added this config in my last commit but it doesn't seem to have done anything (old jobs still running).

That's because the old jobs on 4a96a9f weren't part of the named concurrency group yet. If you push *two* quick commits (or had pushed one right after 1823718) then it should have cancelled the old one that had the group assigned

.github/workflows/main.yml
larsoner: > Any thoughts on caching the gallery directory generated by SG?

I wouldn't cache this. A sphinx build of 8 minutes (including setup!) is very fast. While Sphinx and SG try to be smart about what to rebuild it doesn't seem worth the hassle here

.github/workflows/main.yml
djhoese: Ah makes sense. Thanks.

.github/workflows/main.yml
djhoese: True. We may want to consider it in the future. Right now this isn't rendering a majority of vispy's examples.

larsoner: If speed is eventually a concern, rather than caching I'd go for what MNE does and just build examples that are changed in a given PR (it's very easy to make a trivial change to trigger this if you want to see one) and on `main` always build all from scratch (`main` commits are rare enough that it's not too bad).

larsoner:
@djhoese this seems like a step in the right direction so feel free to merge if you're happy! I can then make a quick follow-up PR to try to build on CircleCI if it would help

.github/workflows/main.yml
djhoese: The issue with not testing all examples on PRs is that vispy is lacking a lot of basic unit tests and the examples are the only places some pieces of code are used and tested. Also, there is a chance that someone changing a Visual interface wouldn't think to check for examples (or might miss one since there are so many) so we wouldn't detect a broken example until it was merged to main.

larsoner: In that case I wouldn't cache the generated directery *or* do anything special to limit what's executed, I'd just let them all build completely every time.

If there are a few examples that take a long time (> 30 s?) to build, you can add something like `_slow` to the filename for example, and not build just this select few on PRs (by using a negative lookahead filename_pattern) but always build everything on main commits (or on PRs where people put `[circle full]` in the commit title for example).

djhoese: Yeah, I suppose there is endless fanciness we could get into with this. Maybe I'll just leave it for now. I'll wait for others to review. I'm ok leaving the limited caching that I have in here now. It's something. Oh but I have some commented out code in there I'll remove that.

.github/workflows/main.yml
```diff
@@ -1,4 +1,9 @@
 name: CI
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency
+# https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pullrequestevent
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}
```
larsoner, line 5: You can add more `-${{ ... }}` to this for example with tags if you want, e.g. [`${{ github.event.type }}`](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#event-object-common-properties). That would probably be safer in case you push a tag and then quickly push a commit to `main` the main one won't kill the tag one (since push to main should have `push` type and tag should have `create` type, I think?).

djhoese: Yeah that seems reasonable.

larsoner:
I pushed an empty commit to test concurrency, looks like it worked! (Feel free to force push over it if you want, just wanted to make sure it did what it was supposed to.)

djhoese:
@almarklein @kmuehlbauer Let me know if you have any opinions on this. Otherwise maybe I'll plan merging it tomorrow or Friday.

kmuehlbauer:
@djhoese @larsoner I'm currently traveling (only on the phone). It looks like you got this working. Hope I can catch up later. :+1:

almarklein:
This looks good to me, to the extend that I can judge it :)  Out of curiosity, did you actually write tons of new examples, or where the previously defined elsewhere?

djhoese:
> Out of curiosity, did you actually write tons of new examples, or where the previously defined elsewhere?

Oh this is a very good point and something we should probably discuss. These are all examples, so far, from `examples/basics/scene/` that have been moved to `examples/scene/` to meet the requirements of sphinx-gallery. Sphinx-gallery only goes 2 levels deep into a directory structure to find examples. So the root if `examples/` and it will go into the newly created `examples/gloo/`, `examples/scene/` and `examples/plotting`. I **could** probably just softlink the examples from the original location to this new location, but we had also talked about reorganizing the examples so maybe this is the start of that?

Overall I think there are too many directories in the examples where it is hard to find what people want. This gallery will make that easier, but maybe the problem isn't the number of directories but how they are organized. Maybe instead of `examples/basics/scene/` it should be `examples/scene/basics/`? Or for sphinx-gallery's sake maybe `examples/scene-basics/`. The only other way to have more complex structures that SG can handle is to have multiple galleries. I could see that being helpful here so we could have a "gloo" gallery, a "scene" gallery, and a "plotting" gallery. I'm also realizing now that there are the regular "visuals" gallery that we should probably support too.

Hhmmm...I'm liking this more and more. Maybe it would be better to have individual galleries for each level. I think going forward we want to get rid of gloo as a primary interface which also means hiding non-scene Visual classes from users. So directing users towards gallerys for each interface may make it easier to separate these things out and then separate them on each gallery page by "layout", "visuals", "advanced", or whatever.

larsoner:
> Sphinx-gallery only goes 2 levels deep into a directory structure to find examples.

FWIW I'm surprised this is a limitation, and we can probably change this at the SG level if it would help.

> I could see that being helpful here so we could have a "gloo" gallery, a "scene" gallery, and a "plotting" gallery. I'm also realizing now that there are the regular "visuals" gallery that we should probably support too.

+1 for this, it seems like a nice way to organize things. A single monolithic gallery containing all of these would probably be a bit much

djhoese:
> FWIW I'm surprised this is a limitation, and we can probably change this at the SG level if it would help.

I think it makes sense from a SG point of view. Top level if the gallery page, second level are the sections. That's pretty clear once I realized what it was doing. Multiple levels I think would only lead to multi-page galleries (probably what I'll do here) which can get a lot more opinionated between projects. Or maybe a README in the lowest level that defines the section...nah, just confusing.

> +1 for this, it seems like a nice way to organize things. A single monolithic gallery containing all of these would probably be a bit much

Agreed. I'll see what I can do.

djhoese:
Ok I updated it so there are now three galleries. The main gallery page just has links to the individual pages in a table of contents. I'm not sure how this should be changed to be prettier. I also removed all the old example scripts that were moved and now link to their new location. I have not separated the scene examples into sub-categories. That may need another pass later on.

I did update the writing examples documentation to talk about how other projects could use the vispy custom scraper.

djhoese:
I tried to add some of the gloo examples and it turns out my scraper doesn't work because I use `Canvas.render`. For the SceneCanvas this is fine because it ignores loops. For the regular Canvas this fails because it itself triggers the draw event causing a loop. Switching to `_screenshot` fixes it, but last time I tried that that didn't draw things properly for the SceneCanvas.

While doing this, I realized that there is a vispy.visuals.filters.material module that has a non-existent import back from 2015 when filters were called components. The two components in here are a shading one which has obvious been replaced and another that adds grid lines to whatever you are drawing. Since this entire module isn't importable and I'm not sure how useful the grid contour component/filter is I'm going to remove it. No one was requesting their use for the last 6 years.  FYI @campagnola who was the original one who reorganized these. Here's the commit if we need it in the future:

https://github.com/vispy/vispy/commit/457657e51ad49b2848143557197180865bcfe4d4#diff-8219ede44966270faf0fac842996cab4bec4c8bf661d730f4e05702043489d99

larsoner:
> Switching to _screenshot fixes it, but last time I tried that that didn't draw things properly for the SceneCanvas.

+1 for triaging which method to used based on if there is an object of type SceneCanvas in the namespace (or not). I would expect SceneCanvas examples to use the scene way of capturing renders and gloo to use _screenshot

djhoese:
Ok I was wrong. Using `_screenshot` only seems to grab the cleared buffer (one solid color). It is definitely being called multiple times, but nothing. Fun.

djhoese:
`canvas.set_current()` seems to fix it. :+1:

djhoese:
I've made a maint/0.7 branch and will consider this the first change toward 0.8 even though our CIs currently upload to a single website deployment so users will see this right away. I've added plotting and gloo examples now (not all of them but everything from `examples/basics/`. This is a good starting point and will hopefully make it easier for the future.

Going to merge the website PR and then update this one and merge this.

PR #2149, opened 2021-07-17
Rendering arbitrary planes in the `VolumeVisual`

alisterburt:
https://user-images.githubusercontent.com/7307488/126048864-4c694229-3d41-4e3b-9362-b064d41cb54b.mp4

This PR supersedes #2074 to implement the rendering of arbitrary planes through volumetric data in the `VolumeVisual` - please see a detailed explanation of the rendering in #2074.

The reason for a new PR was that I wanted to incorporate changes from the current main branch of vispy and also #2117 (and brisvag/vispy#2) which improves the way alternative rendering modes are handled in the visual.


djhoese:
@alisterburt I know you mentioned wanting to talk about this, but I ran out of time. Let me know if you want to meet this week or let me know if this is ready for review and I'll try to make time for reviewing this week too.

alisterburt:
thanks for offering to meet @djhoese - I would love to but this week is tough and it would only be to discuss the depth issues in this PR (same as in #2074). 

For the use envisioned in napari those depth issues are (kind of) solved by setting the blending mode to 'additive' (see napari/napari#3023). Given that, I think this PR is ready for review - it might be worth looking at #2117 separately first because this PR includes that one!

djhoese:
@alisterburt Looks like the recent merges have caused some conflicts. Do you have any time to fix those?

alisterburt:
@djhoese conflicts fixed, should be ready to go!

vispy/visuals/volume.py
```diff
@@ -34,6 +34,8 @@
 coordinates).
 
 """
+from numpy.typing import ArrayLike
```
djhoese, line 37: Can you triple check, I'm pretty sure numpy typing was only added in a semi-recent version. I don't think we set a very strict minimum numpy version for vispy.

vispy/visuals/volume.py
```diff
@@ -275,6 +286,65 @@
...
+    if (intersection_tex.y > 1.0 )
+        discard;
+    if (intersection_tex.z > 1.0 )
+        discard;
+    if (intersection_tex.x < 0.0 )
+        discard;
+    if (intersection_tex.y < 0.0 )
+        discard;
+    if (intersection_tex.z < 0.0 )
+        discard;
```
djhoese, line 332: @almarklein You've been a real stickler on conditionals in shader code, any idea if there is a preferred way of handling the above? Could you do `intersection_tex.xyz < 0`? Or I suppose that is an AND when we need an OR.

examples/basics/visuals/volume_plane.py
```diff
@@ -0,0 +1,105 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+# vispy: gallery 2
+"""
+Example volume rendering with plane rendering
```
djhoese, line 8: Would you mind adding a `===` line underneath this and moving this example to `examples/scene`? I also assume this example starts by showing a plane rather than the whole volume (the gallery would capture the whole volume for the website which isn't as "fun").

Github is also showing that you did "something" with the doc submodule. Did you make changes to any files? Maybe you need to update the `doc` submodule:

```
cd doc
git pull  # assuming on upstream main
cd ..
git add doc
git commit
```

examples/basics/visuals/volume_plane.py
```diff
@@ -0,0 +1,105 @@
...
+    if event.button == 1 and event.is_dragging:
+        axis.transform.reset()
+
+        axis.transform.rotate(cam.roll, (0, 0, 1))
+        axis.transform.rotate(cam.elevation, (1, 0, 0))
+        axis.transform.rotate(cam.azimuth, (0, 1, 0))
+
+        axis.transform.scale((50, 50, 0.001))
+        axis.transform.translate((50., 50.))
+        axis.update()
```
djhoese, line 61: Is there no way to do this with the builtin functionality of the axis and/or camera?

djhoese (changes requested):
A couple questions, suggestions, and requests. This looks pretty good. I don't know much about the low level details so hopefully @almarklein can catch anything important.

vispy/visuals/volume.py
alisterburt: you're right, this is new in 1.20 - will remove!

vispy/visuals/volume.py
alisterburt: would love to hear if there's a better way of doing this - for context, I implemented as a series of `if` statements to allow for the earliest possible exit

examples/basics/visuals/volume_plane.py
alisterburt: good catch! Not sure why but that *something* always happens with my local setup in pycharm... should probably read into submodules a bit - will fix!

examples/basics/visuals/volume_plane.py
alisterburt: Probably! I actually ripped this out of the existing volume example, happy to try to update both in a followup PR?

examples/basics/visuals/volume_plane.py
alisterburt: and I'll make sure the init plane is an interesting one for the gallery :)

examples/basics/visuals/volume_plane.py
djhoese: I'm ok with that. I'd like it in this PR, but no reason to hold it up if you can't get it...and I just realized github says the code is outdated.

djhoese:
What should we expect for a screenshot when this example first loads?

alisterburt:
> What should we expect for a screenshot when this example first loads?

I set it to a y-slice through the stent example volume - is making animations for the gallery particularly complicated? Could you point me at the best example to pattern from?

I was playing with having the slice overlay a lower-opacity 3D view but in single screenshots it never looked that good... in an animation it looks great though!

examples/basics/visuals/volume_plane.py
alisterburt: Cool that's fair! I'm not immediately sure how to do it so it will take me some time but happy to spend the time to make the PR better

djhoese:
Screenshot?

For animations, read the "Writing Examples" doc in the developer's guide part of the website. However, that all depends on the example already producing an animation. Typically you'll see these types of examples use the spacebar as the key binding to start and stop a timer that triggers an update of the data on every trigger. Then in the example you specify `#vispy: gallery 5:55:5` to say capture every 5th frame from the 5th draw to the 55th draw (start:stop:step).

vispy/visuals/volume.py
almarklein: My gut says that many if-statements look scary :)  I would feel more confortable if we can convert this to a single `if` and a bunch of OR's and AND's. But also it's 2021 and I'm not sure how much this matters on modern hardware. And whether replacing many with (still) one helps.

@rougier what are your thoughts on this? Similar questions have popped up before. Is there a resource that we could use as a guideline when dealing with conditionals in shaders (in this day and age)?

vispy/visuals/volume.py
```diff
@@ -805,6 +941,45 @@ def relative_step_size(self, value):
         self._relative_step_size = value
         self.shared_program['u_relative_step_size'] = value
 
+    @property
+    def plane_position(self):
+        return self._plane_position
```
almarklein, line 946: These props could use a docstring :)

rougier:
Is it possible to adapt the colormap limits when slicing? From the movie, it's like the slice is much darker once sliced.

alisterburt:
@djhoese 
> Screenshot?

image

Thanks for the pointers on the animation! I'll have a go at that in the next few days

@rougier I think the 'relative darkness' of the plane is a product of the max intensity projection which (of course) takes the maximum value encountered as the ray moves through the volume. Switching to average for the slice (screenshot below) possibly gives more the effect you were expecting?

image

vispy/visuals/volume.py
```diff
@@ -753,15 +870,34 @@ def method(self, method):
...
         self.shared_program['texture2D_LUT'] = self.cmap.texture_lut() \
             if (hasattr(self.cmap, 'texture_lut')) else None
         self.update()
 
+    @property
+    def _raycasting_setup_snippet(self):
+        return RAYCASTING_MODE_DICT[self.raycasting_mode]
+
+    @property
+    def raycasting_mode(self):
```
djhoese, line 889: Could you also add a docstring to this property?

vispy/visuals/volume.py
djhoese: My laziness says this won't matter overall (it won't be noticeable to the majority of users). My developer gut says these should at least be combined for cleanliness so the `.x` comparisons happen in one if statement, the `.y` in one if, and the `.z` in the third and last if statement. @alisterburt would you mind updating this to do that.

djhoese (changes requested):
Two more small changes. So sorry for the delay (real life distractions).

vispy/visuals/volume.py
alisterburt: implemented an alternative suggestion from @kevinyamauchi to reduce the branching

```
    // discard if intersection not in texture
    
    float out_of_bounds = 0;

    out_of_bounds += float(intersection_tex.x > 1);
    out_of_bounds += float(intersection_tex.x < 0);
    out_of_bounds += float(intersection_tex.y > 1);
    out_of_bounds += float(intersection_tex.y < 0);
    out_of_bounds += float(intersection_tex.z > 1);
    out_of_bounds += float(intersection_tex.z < 0);
    
    if (out_of_bounds > 0) {
        discard;
    }
```

hope that's okay!

alisterburt:
@djhoese no worries about the delay, I'm only just getting back to this myself, thanks for your patience!

I've added an animation to the example, I'm not sure how it will render in the gallery but here's how it looks here

https://user-images.githubusercontent.com/7307488/128708173-0e7b839d-69ae-458a-b76c-3e84b90a8faa.mp4

I also reduced the branching in the shader based on a suggestion by @kevinyamauchi 🙂

djhoese:
@alisterburt given your other comment, would you mind merging `main` with your branch so the tests pass?

Do you think there is a preferred merging order between this PR and #2116?

brisvag:
We tried to keep things tidy, so *in theory* these 2 PRs should merge just fine no matter the order (there will probably be a couple of trivial conflicts).

One thing that might be worth noting is that the shader as is might not work *exactly* as one might expect if combining the clipping planes and the plane rendering. If you activate plane rendering with a non-1 thickness, the clipping planes will clip the volume *before* sampling, which will potentailly result in undersampled areas on the plane.

I doubt this is a problem in most cases, but it's worth writing down somewhere :)

djhoese:
> I doubt this is a problem in most cases, but it's worth writing down somewhere :)

Appreciated. Thanks.

djhoese:
@alisterburt Do you have time to merge with `main` and resolve the conflicts?

alisterburt:
done! thanks @djhoese - I merged via the github interface because of some github permissions errors when trying to push, hopefully the linting doesn't complain, if so I'll fix when I get home in a few hours

vispy/visuals/volume.py
```diff
@@ -575,11 +668,25 @@ def __init__(self, vol, clim="auto", method='mip', threshold=None,
...
         # Set params
+        self.raycasting_mode = raycasting_mode
         self.method = method
         self.relative_step_size = relative_step_size
         self.threshold = threshold if threshold is not None else vol.mean()
         self.attenuation = attenuation
+
+        # Set plane params
+        if plane_position is None:
+            self.plane_position = np.array(vol.shape) / 2
```
djhoese, line 679:
```suggestion
            self.plane_position = [x / 2 for x in vol.shape]
```

djhoese (changes requested):
One small style choice change requested and then could you move the example to `examples/scene` instead of the `examples/basics/scene` directory it is in now? Thank you.

alisterburt:
done and done!

djhoese (approved):

alisterburt:
🚀 🚀 🚀 

Pleasure working with you @djhoese !

PR #2152, opened 2021-07-21
Test PR to see if the CI segfault is in main

almarklein:


djhoese:
Test passes, wheel was an HTTP hiccup.

almarklein:
Bingo. It was the added call to `super.__init__()` causing the Qt widget's `__init__` to be called twice (or something like that).

PR #2162, opened 2021-08-04
Cleanup init in backends

almarklein:
I had a stab at doing clean `super().__init__(...)` in all the backend code, but that broke for Qt. Might be fixable but that would only complicate the code. So let's stick with the rule that the backend must call both the `BaseCanvasBackend.__init__` and the native widget's `__init__`.

* Most changes replace `*args` with just a `vispy_canvas` argument.
* Changed superclass order in a few places.
* ~Use `_process_backend_kwargs` in jupyter_rfb backend.~
* Also a tweak to flake8 config to ignore errors from certain flake8 extensions that one might have installed in the env.


setup.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
 max-line-length = 120
-ignore = E226,W291,W293,W503,F999,E305,F405,W504,N
+ignore = E226,W291,W293,W503,F999,E305,F405,W504,N,D,B
```
djhoese, line 3: Eventually we will want to include these. Personally I don't see why the `B` can't checked. That is meant to detect bugs (flake8-bugbear) and I haven't found any cases in my own work where it wasn't wrong about something.

almarklein: I'll remove the B and prep a separate pr to fix any issues it reports.

vispy/app/backends/_jupyter_rfb.py
```diff
@@ -28,7 +28,7 @@
     resizable=True,
     decorate=False,
     fullscreen=False,
-    context=True,  # the offscreen global context
+    context=True,  # Could work, but not implemented
```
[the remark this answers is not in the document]
almarklein: Yes it should :D  updated!

PR #2164, opened 2021-08-04
Remove 'VNC' backend

almarklein:
The VNC backed was based on the same idea as the new jupyter_rfb backend, but it was always experimental and disabled by default.


vispy/util/wrappers.py
```diff
@@ -40,10 +40,8 @@ def use(app=None, gl=None):
             * 'SDL2': use SDL v2 backend.
             * 'osmesa': Use OSMesa backend
         Additional backends:
-            * 'ipynb_webgl': run vispy from a Jupyter notebook (not fully
-               functional)
-            * 'ipynb_vnc': render in a Jupyter notebook via a VNC approach
-              (experimental)
+            * 'jupyter_rfb': show vispy canvases in Jupyter lab/notebook.
```
djhoese, line 43: Do you maybe want to mention that this is "via 'jupyter_rfb'" or something about it depending on that package?

almarklein: done!

Issue #1927, opened 2020-09-29
Investigate old ipython input hooks


djhoese:
I'm looking through the various "ext" libraries we use and things that are bundled and noticed `vispy/ext/ipy_inputhook.py` and its usage in `vispy/app/inputhook.py` which is describing IPython 3.x hooks. Ever since IPython 5.x this functionality isn't supported so I'm not sure if/how this is working. Either the bundled hook management is good enough to survive the changes in IPython core or nothing is actually working or it is accidentally working because IPython 5+ supports the Qt event loop (among others) by default. See:

https://ipython.readthedocs.io/en/stable/config/eventloops.html?highlight=inputhook

djhoese:
@Aaru143 This may be another good issue for you to look into. It does require some knowledge of ipython/jupyter but this should be easy-ish to test. For example, try removing the various hook "stuff" in `vispy/app/inputhook.py` and try running the `examples/tutorial/app/interactive.py` example.

I just did a quick test by running `ipython` then typing:

```
>>> %run examples/tutorial/app/interactive.py
```

It brought up a window but didn't do anything which tells me the Qt event loop isn't running. I then (in the same ipython session) ran `%gui qt` and it started running the example. You can tell because the background color cycles from black to white to black. If you close ipython and run it again but do `%gui qt` first then doing the `%run ...` also works. So it seems the `set_interactive` isn't actually doing anything.

I'd say a pull request removing both the `vispy/ext/ipy_inputhook.py` module, the `vispy/app/inputhook.py` module, and even the `tutorial/app/interactive.py` example would be good, but doing some tests to make sure would be good. I just did `%gui qt` in ipython followed by running the `tutorial/app/simple.py` example and it still worked. I'd be curious what you find @Aaru143.

PR #2167, opened 2021-08-07
Remove vispy.js submodule and ipynb_webgl backend

### LINKED ISSUE #2130: Dropping the webgl backend
When #2127 gets done, we might consider dropping the webgl backend. For one it would avoid the limitations of GL es 2.0, which we already violate in a few places (e.g. 3d textures, special texture formats). Further it may allow us to drop/simplify GLIR. Any other things that dropping the webgl backend would make easier?

On the other hand, do we expect that people who now use the webgfl backend be happy enough with the new Jupyter widget?

Thoughts?

djhoese:
Closes #2130

Now that the `jupyter_rfb` backend is available and performs so much better than the existing ipynb_webgl backend, this PR removes the `ipynb_webgl` and everything that was added for ipython or jupyter webgl-based. This includes the backend, the `js` git submodule, and all ipython webhooks which were old and not actually used by anything, and the setup.py trying to build the jupyter widget.

TODO:

- [x] Update notebook examples to use jupyter_rfb (or at least don't mention webgl)


djhoese:
Regarding `inputhook` see #1927

djhoese:
Now I'm torn on removing the input hooks completely. Or at least the example and the intended functionality. We could rewrite the inputhook with:

https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.hooks.html

But also as noted in #1927 for qt you can do `%gui qt` and that works just as well from my testing.

almarklein:
> [...] for qt you can do `%gui qt` and that works just as well from my testing.

My inclination is to remove the hooks then. Less code to maintain that can break. If this does cause a problem for users, we could consider bringing it back, but at least then we have a reference (since #1927 suggests that it is unclear whether the current implementation actually works).

almarklein:
image

These are the better PR's :)

almarklein (approved):

djhoese:
> My inclination is to remove the hooks then. Less code to maintain that can break. If this does cause a problem for users, we could consider bringing it back, but at least then we have a reference (since #1927 suggests that it is unclear whether the current implementation actually works).

Exactly. As I say in #1927, the manager and hook interfaces we are/were using are not supported by ipython any more and they were technically "vendored" in to vispy. I think that's the only reason they worked is that we essentially duplicated what has been deprecated in current ipython. So it may have worked a little bit, but isn't the "right" way of doing it. The main purpose of the hook is to have a generic event loop handler for vispy's backends by calling "process_events" on whatever backend happens to be used. I think for some/most of the backends (at least the ones supported by matplotlib) there is likely already a `%gui` hook for handling the event loop of that GUI framework.

@larsoner from what I could tell a lot of the interactive stuff was added/requested by you. What's your opinion on this?

djhoese:
Oh also as part of this I'd like to put vispy.js in archive mode so users know it isn't actively being developed and shouldn't be depended on. I'd have to look at what I need to do to provide the same indication in the npm package repository for vispy.js.

djhoese:
@almarklein should I also remove the ipynb_static backend? Or did you have plans to revive that in some way with jupyter_rfb?

almarklein:
> @almarklein should I also remove the ipynb_static backend? Or did you have plans to revive that in some way with jupyter_rfb?

I plan on having a look to revive it, making use of the offscreen utils should make it much leaner.

djhoese:
@almarklein I'm trying to update the "webgl_*" notebooks and am running into a few weird results. First, any idea why a print statement in a key press event isn't showing up anywhere in the notebook or in the jupyter lab terminal output?

almarklein:
> First, any idea why a print statement in a key press event isn't showing up anywhere in the notebook or in the jupyter lab terminal output?

I think this is how Jupyter works, print statements in a callback generally don't show up in cell output. I know, this sounds terrible, but it seems by design. See e.g. https://github.com/jupyter-widgets/ipywidgets/issues/2068 ... but now that we're talking about it, perhaps there is a way for us to provide a default behavior. I'll have a look.

almarklein:
Also see https://github.com/matplotlib/ipympl/issues/116 for a more similar issue. The fact that print statements (and errors!!) are not displayed anywhere seems like a huge design flaw to me. I'd think it would be possible to shove these to the last created output or something ...

djhoese:
Ah no big deal then. I figured out some of what was going wrong. I'm still having issues with the molecular viewer notebook which I just pushed changes for.

The way it is now shows black molecules rotating. The only difference between it and the `examples/demo/gloo/molecular_viewer.py` example is the last statement of the fragment shader:

```
    gl_FragColor.rgba = vec4(.15*v_color + .55*diffuse * v_color
                        + .35*specular * v_light, 1.0);
```

This is what is in the molecular_viewer.py example and works fine. In the notebook if I do this I get a black canvas. If I change it to what it was before this PR with `.rgb = ` and no `1.0` at the end, then it shows me the black molecules.

image

djhoese:
Ha they aren't black molecules, they are fully transparent. If I change my theme to light mode then the molecules are white.

djhoese:
Hm doing `    gl_FragColor.a = 1.0;` after the `.rgb = ` version seems to work. :confused:

djhoese:
I guess I got into some weird state. Now it works. :man_shrugging: 

@almarklein I *think* this is ready for merging, what do you think?

almarklein:
LGTM!

djhoese:
I have now archived the vispy.js github project and deprecated the npm package:

https://www.npmjs.com/package/vispy

By doing:

```bash
$ npm deprecate vispy "The VisPy Python library now depends on jupyter_rfb. The vispy.js library is no longer maintained."
```

PR #2169, opened 2021-08-11
Switch to cibuildwheel github action and skip 3.10 wheel building

djhoese:
Addresses the build issues identified and discussed in #2116 

Currently cibuildwheel is trying to build with Python 3.10. However, many libraries don't have wheels for 3.10 yet (including vispy) and this causes issues for CI. We depend on Pillow for testing our wheels and since Pillow doesn't currently have a 3.10 wheel it tries to build from source during cibuildwheel's test phase. This fails with some missing C dependencies.

This PR switches to using the new PyPA-based cibuildwheel github action and skips testing the wheel for Python 3.10.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -20,21 +20,14 @@ jobs:
       with:
         fetch-depth: 0
         submodules: 'recursive'
-    - name: Use MSBuild (Windows)
-      uses: microsoft/setup-msbuild@v1.0.2
-      if: matrix.os == 'windows-2019'
-    - uses: actions/setup-python@v2
-      name: Install Python
-    - name: Install cibuildwheel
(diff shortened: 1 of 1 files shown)


djhoese:
Looks like building fails on Windows and OSX because of numpy issues. Numpy wheels for 3.10 are not available so we might have to wait until at least then to build Python 3.10 wheels for vispy. @almarklein you ok with me skipping building 3.10 wheels for now?

djhoese:
Numpy prerelease wheels are available from here: https://github.com/numpy/numpy/issues/19630#issuecomment-896176442

But not sure we want to build wheels against pre-releases.

djhoese:
TODO: Re-enable wheel building for Python 3.10

PR #2172, opened 2021-08-14
Add PyQt6 backend

### LINKED ISSUE #2171: PyQt6 backend missing
Hi,

I am using VisPy in FlatCAM application (beta branch) and I have ported my app to PyQt6. But VisPy (nor matplotlib) do not work with PyQt6 at the moment therefore I tried to add this new backend for the VisPy project.

I am not experienced with VisPy (I only use a limited set of features) but what changes I made are enough for my project to run without errors.

My changes are here, perhaps they will help minimize the work to add this backend:
https://github.com/mars0001/vispy

Do you think that I should make a pull request with this?

Best regards,
Marius

mars0001:
Closes #2171 

Hi,

I have modified the VisPy project to work with PyQt6. I have used as a guide a previous accepted PR that added support for PySide6.
I have tested the modifications in my app (that uses VisPy) and it worked for me so I don't expect major issues.
Based of David's recommendation I have made this PR. Hope that it helps.

Best regards,
Marius


vispy/app/backends/_qt.py
```diff
@@ -125,55 +138,107 @@ def _check_imports(lib):
...
-    QtCore.Qt.Key_F11: keys.F11,
-    QtCore.Qt.Key_F12: keys.F12,
-
-    QtCore.Qt.Key_Space: keys.SPACE,
-    QtCore.Qt.Key_Enter: keys.ENTER,
-    QtCore.Qt.Key_Return: keys.ENTER,
-    QtCore.Qt.Key_Tab: keys.TAB,
-}
-BUTTONMAP = {0: 0, 1: 1, 2: 2, 4: 3, 8: 4, 16: 5}
+if qt_lib == 'pyqt6':
```
djhoese, line 141: Could this be simplified by doing something like `qt_keys = QtCore.Qt.Key if qt_lib == 'pyqt6' else QtCore.Qt` and then update the dictionary to use `qt_keys` instead?

vispy/app/backends/_qt.py
```diff
@@ -453,39 +527,59 @@ def sizeHint(self):
     def mousePressEvent(self, ev):
         if self._vispy_canvas is None:
             return
+        if hasattr(ev, 'pos'):
+            posx, posy = ev.pos().x(), ev.pos().y()
+        else:
+            # Compatibility for PySide6 / PyQt6
+            posx, posy = ev.position().x(), ev.position().y()
```
djhoese, line 534: It seems you do this a lot. Maybe you could make a helper function (near the top of the module) like `_get_event_xy(ev):` and put this `if/else` there and then reuse it everywhere you current have this if/else.

vispy/app/backends/_qt.py
```diff
@@ -525,13 +619,26 @@ def keyReleaseEvent(self, ev):
     def event(self, ev):
         out = super(QtBaseCanvasBackend, self).event(ev)
         t = ev.type()
+        if PYQT6_API:
```
djhoese, line 622: Similar to the keys map, maybe this could be simplified with a `qt_event_types = QtCore.QEvent.Type if PYQT6_API else QtCore.QEvent`.

vispy/app/backends/_qt.py
```diff
@@ -575,12 +682,20 @@ def _modifiers(self, event):
...
+                         [QtCore.Qt.KeyboardModifier.MetaModifier, keys.META]):
+                if q & qtmod:
+                    mod += (v,)
+        else:
+            for q, v in ([QtCore.Qt.ShiftModifier, keys.SHIFT],
+                         [QtCore.Qt.ControlModifier, keys.CONTROL],
+                         [QtCore.Qt.AltModifier, keys.ALT],
+                         [QtCore.Qt.MetaModifier, keys.META]):
+                if q & qtmod:
+                    mod += (v,)
```
djhoese, line 698: Same idea here.

djhoese (changes requested):
Thank you! This looks pretty good to me. I put some inline comments for changes I think you could make to shorten the code and make it overall easier to understand. Let me know if you have any other questions.

I'll have to look into how hard it would be to include tests for Qt6 in our automated testing.

Side note: @almarklein this is pretty sad that this many places are required to change to add a new backend.

djhoese:
:confused: Also it looks like your branch has a conflict with the `main` branches `_qt.py`. This suggests your fork is a little out of date. You'll need to merge the upstream vispy `main` branch and resolve conflicts. Let me know if you have any questions about this.

vispy/app/backends/_qt.py
mars0001: I agree, I thought about that but I didn't want to make too many changes and depart from what people are used to see. I know that I don't like seeing too many changes in a big project :)
Will do.

vispy/app/backends/_qt.py
mars0001: Done.

vispy/app/backends/_qt.py
mars0001: Done.

vispy/app/backends/_qt.py
mars0001: Done. Thank you for the suggestions!

kmuehlbauer:
@mars0001 I've activated CI runs for your PR. We have to do this only once. All of your  subsequent commits and PR, CI will kick in instantly.

As soon as the linter errors are fixed, unit tests will be run, too.

djhoese:
@mars0001 Just for simplicity, you can see the flake errors if you click "Details" next to the fail "lint and style checks" item, but I'll paste them here for you:

```
Running flake8... 
vispy/app/backends/_pyqt6.py:38:5: F401 'PyQt6.QtWidgets' imported but unused
vispy/app/tests/test_backends.py:44:64: E231 missing whitespace after ','
```

djhoese:
FYI @kmuehlbauer I had to approve this commit too.

djhoese:
@mars0001 Don't worry about the wheel failures. I'm working on it in #2169

kmuehlbauer:
> FYI @kmuehlbauer I had to approve this commit too.

Did GH change this behaviour? But also good chance that my brain failed me.

djhoese:
> Did GH change this behaviour? 

I've only ever known it to be this way. Maybe your brain did fail you :wink:

djhoese:
@mars0001 I was a little busy today, but do you have time to merge `main` with your branch? It should fix the tests.

mars0001:
Hi David (@djhoese),

I did the merge blindly, I did not had the time to check what are the changes and if the new package works with my app. I guess the tests should see to that but once I have the time to make my own checks I will let you know.

Best regards,
Marius

vispy/app/backends/_pyqt4.py
```diff
@@ -13,7 +13,7 @@
 
 try:
     # Make sure no conflicting libraries have been imported.
-    for lib in ['PySide', 'PyQt5', 'PySide2', 'PySide6']:
+    for lib in ['PySide', 'PyQt5', 'PyQt6', 'PySide2', 'PySide6']:
         lib += '.QtCore'
         if lib in sys.modules:
             raise RuntimeError("Refusing to import PyQt4 because %s is "
```
almarklein, line 19: This code is repeated in each Qt backend and can surely be refactored away. This could be fixed here, but maybe its easier if one of us (e.g. me) does this once this PR has merged.

vispy/app/backends/_pyqt6.py
```diff
@@ -0,0 +1,46 @@
...
+    has_uic = False
+
+    which = ('PyQt6', QtCore.PYQT_VERSION_STR, QtCore.PYQT_VERSION_STR)
+    # Remove _qt module to force an import even if it was already imported
+    sys.modules.pop(__name__.replace('_pyqt6', '_qt'), None)
+    # Import _qt. Keep a ref to the module object!
+    if backends.qt_lib is None:
+        backends.qt_lib = 'pyqt6'  # Signal to _qt what it should import
+        from . import _qt  # noqa
+        from ._qt import *  # noqa
```
almarklein, line 43: If @mars0001 is up for it ... I think the way to go (to clean up these  `sys.modules` checks) would be to add a `def _check_imported(current_lib)` in `_qt.py` that we call right here.

mars0001: @almarklein, Sure, once I get home from work I will do it!

djhoese: We could also in the long run switch to `qtpy` which @sofroniewn and maybe @tlambert03 have suggested in the past: https://pypi.org/project/QtPy/

almarklein: Wait, such a function already exists, and is already called in `_qt.py` itself. It looks to me like the import-checks in all qt backends (except the one one in `_qt.py` can be removed ...

vispy/app/backends/_qt.py
```diff
@@ -452,39 +494,43 @@ def sizeHint(self):
     def mousePressEvent(self, ev):
         if self._vispy_canvas is None:
             return
+        posx, posy = _get_event_xy(ev)
         self._vispy_mouse_press(
             native=ev,
-            pos=(ev.pos().x(), ev.pos().y()),
+            pos=(posx, posy),
```
almarklein, line 500: Nitpick: just use `pos=_get_event_xy(ev),`

almarklein: Also in other cases below.

vispy/app/backends/_pyqt6.py
tlambert03: > We could also in the long run switch to qtpy

which also doesn't support pyqt6 yet 😞 ...  there's a seemingly now stale thread there from Dec 2020 to add support for pyside6 and pyqt6. we have an independent compatibility layer ([superqt.qtcompat](https://github.com/napari/superqt/tree/main/superqt/qtcompat)) that has qt6 support

almarklein (changes requested):
Very nice, thanks! Some minor remarks. It looks like we can clean up the import-checks by simply removing them from the actual backends. Or am I missing a reason why these checks are duplicated, @djhoese ?

djhoese:
> Or am I missing a reason why these checks are duplicated, @djhoese ?

No idea. My only guess would be import order as the Qt backends have a weird way of doing dependencies and importing each other module. Or maybe that's just me getting confused.

mars0001:
So, @almarklein, @djhoese do we do you want me to remove this:
```
# Make sure no conflicting libraries have been imported.
    for lib in ['PySide', 'PyQt5', 'PyQt6', 'PySide2', 'PySide6']:
        lib += '.QtCore'
        if lib in sys.modules:
            raise RuntimeError("Refusing to import PyQt4 because %s is "
                               "already imported." % lib)
```
from all backends?

almarklein:
> from all backends?

Yes, from all files except `_qt.py`

djhoese:
@mars0001 I'd like to test some examples with this backend. Have you been installing PyQt6 from PyPI? I usually use conda-forge which doesn't provide the newest Qt versions.

mars0001:
@djhoese 
Hi David,
Yes, I've installed the latest version available on Pypi. It is 6.1.1.
image

djhoese (approved):

djhoese:
@mars0001 I'm trying to run this example locally and whenever I try to run an example I get this error:

```
$ python examples/scene/mesh_shading.py
  File "/home/davidh/repos/git/vispy/examples/scene/mesh_shading.py", line 119, in <module>
    app.run()
  File "/home/davidh/repos/git/vispy/vispy/app/_default_app.py", line 60, in run
    return default_app.run()
  File "/home/davidh/repos/git/vispy/vispy/app/application.py", line 163, in run
  File "/home/davidh/repos/git/vispy/vispy/app/backends/_qt.py", line 333, in _vispy_run
    return app.exec_()
AttributeError: 'QApplication' object has no attribute 'exec_'
```

Any ideas?

almarklein (approved):

djhoese:
Ah looks like `exec_` doesn't exist anymore in PyQt6:

https://stackoverflow.com/questions/22610720/should-i-use-app-exec-or-app-exec-in-my-pyqt-application#comment121343730_22614643

djhoese:
I'll push a fix in a second. Now this is making me think we need to add PyQt6 to the tests. Let me see how hard that would be

mars0001:
@djhoese,
It looks like there is a need for this change in a lot more places:
image

djhoese:
Those are all examples which right now are probably specific to PyQt5 (they directly import PyQt5). This is OK.

djhoese:
So it isn't pretty, but I've added a new test environment for new Qt versions (it just uses PyQt6 right now). :thinking: theoretically I should be able to add PySide6 to this too since it should use the same version of the C Qt, right? Well let's see how this goes.

djhoese:
@kmuehlbauer I think you did most of the CI stuff for this? Does my addition here make you :vomiting_face: ?

kmuehlbauer:
@djhoese I didn't follow your changes in depth. If it gets to complicated or unreadable we might think about a complete refactor of the CI test sections. :grimacing: Not sure if it was easy and readable before, :grinning:

djhoese:
Yes, definitely needs a refactor. We don't even test on Windows and OSX which would be nice to have if we can make it work.

djhoese:
:thinking: So matplotlib on conda-forge pulls in pyqt which for conda is pyqt 6. Matplotlib as far as I can tell doesn't support PyQt6 yet. So...I could try installing matplotlib-base and hope for the best or not install matplotlib at all and hope that the tests that require it are skipped when it isn't installed.

djhoese:
Ok, PyQt6 is now tested. It was much harder than I thought because of the conflicts with other conda-forge packages and the number of apt packages that were needed. I also noticed some warnings that could be easily fixed in the test output so I fixed those too. I'm going to let this sit over night, but @almarklein or @kmuehlbauer feel free to merge this if you think it is OK. Otherwise I will merge it when I start my day tomorrow and make a VisPy 0.8 (unless you disagree).

kmuehlbauer (approved):
This is LGTM. Thanks @mars0001 for taking the chance and bringing PyQt6 to `VisPy`. 

@djhoese thanks for going down that rabbit hole. I'll merge this now. So if everything works out nice, you'll have a head start for 0.8.0.

almarklein:
Those commit messages look like @djhoese having a great time with CI ;)  - I can relate.

mars0001:
I have to say that David shows a great determination in getting things done... Good job all!

PR #2179, opened 2021-08-20
Fix volume_plane.py example not having a toggle for the animation

### LINKED ISSUE #2178: Update volume_plane.py to allow timer to be stopped
I noticed the volume_plane.py has keyboard handling for moving the plane through the volume, but it doesn't allow you to stop the animation so these events don't actually mean much.

@alisterburt do you have time to fix this? Tradition is to use the spacebar (`" "`) to toggle the timer start/stop

djhoese:
Close #2178


alisterburt (approved):
perfecto, thanks for dealing with this! Looking forward to the 0.8.0 release 🚀

Issue #2183, opened 2021-08-23
time.clock() on python 3.10


Llewyllen:
in vispy/util/ptime.py

replace
sys.version < '3.3'

by
(3 > sys.version_info[0] or (3 == sys.version_info[0] and 3 > sys.version_info[1]))

as '3.10' is less than '3.3

djhoese:
We actually don't support Python <3.3 anyway so we could probably just remove that whole if branch. Thanks for pointing this out. Do you feel up to making a pull request for this?

tirkarthi:
This is a duplicate of https://github.com/vispy/vispy/issues/1659

Llewyllen:
it is not a duplicate, since it is a bug in the fix, which worked up to python 3.9

tirkarthi:
Sorry, I misread the issue.

PR #2185, opened 2021-08-24
Fix python version check for python > 3.9

Llewyllen:
since '3.10' is smaller than '3.3', sys.version < '3.3' returns True for python 3.10

### DIFF
--- a/vispy/util/ptime.py
+++ b/vispy/util/ptime.py
@@ -10,6 +10,7 @@ import time as systime
 # get a reference starting time - initial performance counter
 START_TIME = systime.time() - systime.perf_counter()
 
+
 def time():
     # return reference starting time + delta of performance counters
     return START_TIME + systime.perf_counter()
--- a/vispy/util/ptime.py
+++ b/vispy/util/ptime.py
@@ -4,38 +4,12 @@
 
 """
 ptime.py -  Precision time function made os-independent
-(should have been taken care of by python)
 """
 
-from __future__ import division
-
-import sys
 import time as systime
-START_TIME = None
-time = None
+# get a reference starting time - initial performance counter
+START_TIME = systime.time() - systime.perf_counter()
 
-
-def winTime():
-    """Return the current time in seconds with high precision
-
-    (Windows version, use Manager.time() to stay platform independent.)
-    """
-    return systime.clock() + START_TIME
-    # return systime.time()
-
-
-def unixTime():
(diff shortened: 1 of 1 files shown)


djhoese:
I mentioned in the related #2183 issue that we don't support Python versions older than 3.3 anymore. I think this can just be updated to be `time = unixTime`. Or even better `time = systime.time()` and get rid of the functions. Even better again, remove this entire module and use `time.time()` everywhere it is used.

@kmuehlbauer @almarklein let me know if I'm missing something here, but it looks like this workaround is no longer needed.

almarklein:
I don't think [time.time()](https://docs.python.org/3/library/time.html#time.time) is generally useful as a precision timer. The docs say:

    Note that even though the time is always returned as a floating point number, not all systems provide time 
    with a better precision than 1 second. While this function normally returns non-decreasing values, it can
    return a lower value than a previous call if the system clock has been set back between the two calls. 

There is [time.perf_counter()](https://docs.python.org/3/library/time.html#time.perf_counter) for precision (introduced in Python 3.3). That provides a platform-independent solution, but it's "zero" is arbitrary. So maybe we should do something like:
```py
START_TIME = time.time() - time.perf_counter()

def time():
    return time.perf_counter() + START_TIME
```

djhoese:
@Llewyllen What do you think about updating your pull request to do what @almarklein suggested? You could get rid of the two existing functions and the `time =` line and use what Almar suggested but change the usage to `systime`:

```python
START_TIME = systime.time() - systime.perf_counter()

def time():
    return systime.perf_counter() + START_TIME
```

vispy/util/ptime.py
```diff
@@ -32,7 +32,7 @@ def unixTime():
     return systime.time()
 
 
-if sys.platform.startswith('win') and sys.version < '3.3':
+if sys.platform.startswith('win') and (3 > sys.version_info[0] or (3 == sys.version_info[0] and 3 > sys.version_info[1])):
```
tirkarthi, line 35: I guess this same as writing `sys.version_info > (3, 3)`

djhoese: It is actually the opposite (less than 3.3), but doesn't matter overall as we don't support Python <3.7 so no point in checking this anymore.

almarklein (approved):

Llewyllen:
fixed the missing line for flake8 tests

PR #2188, opened 2021-08-24
Remove 'doc' submodule and embed directly in main repository

### LINKED ISSUE #2184: Move vispy-website submodule into main repository
We currently have the `https://github.com/vispy/vispy-website` (sphinx source) repository made available to this main vispy repository as a git submodule under the `doc` directory. There are pros and cons to this:

* Pro: I think we get separate size quotas/limits per repository. So by having it separate, I think the static files we may or may not add to the website repository to not impact how much we can have on this repository. However, we currently aren't using much space at all. I think GitHub prefers you stay under 1GB with a warning at 5GB (last I heard by reading a stackoverflow answer):

   image

   Note: I had to add myself individually as an Admin to each of those repositories for that to show up in https://github.com/settings/repositories. It doesn't show up if you only have access via a Team.

* Pro: Separate issue/pull request tracking. However it is pretty rare that an issue with the website doesn't involve something on this main repository.
* Con: Difficult for users, especially those unfamiliar with git, to clone the repository and get all of the components they need. Take for example #2138 where I had to explain the complicated process of "I know you want to work on the website but that actually requires get this repository and following complex instructions to make sure you also get the doc submodule".
* Con: Some tools don't work well with submodules. We have special commands in some of our github workflows to make sure the submodule is pulled. It works, but it is something extra. More importantly we can't use https://holzhaus.github.io/sphinx-multiversion/master/index.html because it doesn't work with submodules.
* Con: Maintenance. We have to always make sure that any changes to the vispy-website source gets reflected in a new commit here in the main repository when we update the git submodule.

djhoese:
Closes #2184 

I had two options:

1. Take the current content of `doc` and the entire thing as a single commit. Keeping no history of the original vispy-website in this main vispy repository.
2. Put in a lot of extra work (biased much :wink: ) to add the history of the doc directory to this repository.

I went with option 1. The `vispy-website` repository will stay around but I will archive it so if we're really curious the history will be there.


djhoese:
Oh I need to update the installation instructions. One sec...

djhoese:
So I'm also including changes to `make/make.py` here. The gallery is now auto-generated so we don't need the `make gallery` commands. The documentation is now very simple to generate with `make html` and `make clean` so no need to include these in a special in the python `make.py`. Removing all of these leaves a coverage html target and a `test` target. The test target is actually used in the CI so I can't remove that...yet.

almarklein:
You mentioned in #2184 that some images were not used anymore, and that others might be replaced by generated images. Have you looked at that? This PR contains quite a few images.

djhoese:
You are absolutely right. I've removed all the ones that didn't need to exist. We had a lot of "font awesome" stuff in there and it looks like it doesn't change the result if I remove those CSS files and fonts. The remaining images consist of:

1. 4 carousel images: In the future these could be replaced by gallery generated images *if* we knew they were sized appropriately and still looked good. The examples that generate the same images as the ones in the carousel have not been ported to sphinx-gallery and I'm not going to do that in this PR.
2. Images used in the "Modern OpenGL" page
3. Logos

I'm going to try to fix a few other docstring issues that sphinx is complaining about.

djhoese:
This is ready for re-review and merge. All unnecessary static files have been removed, but more importantly I have fixed all docstring issues to the point that I've added the `-W` flag to the sphinx generation. This makes it so that any warnings when generating the website will now cause an error and fail immediately. This means that we'll now know in a PR if someones changes are going to not render properly to sphinx (at least if they are going to warn about bad syntax). This is not a style thing, this is usually sphinx warning that it doesn't know how to interpret a docstring (bad section naming, bad indentation, etc) and the result is probably not going to render how the author wanted it.

djhoese:
@almarklein any additional thoughts on this?

ci/build_website.sh
```diff
@@ -9,10 +9,8 @@ export TRAVIS_TAG="main"
 cd doc
 
 make clean
-# TODO: Fail if warnings from generation
-#       make html SPHINXOPTS="-W"
 # TODO: Switch to sphinx-multiversion
-make html
+make html SPHINXOPTS="-W"
```
almarklein, line 13: I literally found out today (when working on docs for pygfx) about this `-W` option. I also learned about `--keep-going`, so that you'll see all warnings instead of only the first one.
```suggestion
make html SPHINXOPTS="-W" --keep-going
```

almarklein (approved):

ci/build_website.sh
djhoese:
```suggestion
make html SPHINXOPTS="-W --keep-going"
```

Cool! Flag needs to go in the quotes, but yeah.

almarklein: > Cool! Flag needs to go in the quotes, but yeah.

Oh good point!

djhoese:
Let's merge this and see what breaks!

PR #2190, opened 2021-08-25
set gl_FragDepth in volume visual iso rendering

kevinyamauchi:
Hello! In this PR I am proposing to set `gl_FragDepth` in the Volume visual. The depth is set using the position of the surface/point found for each fragment (e.g., the surface for isosurface render or the max intensity for max intensity projection). This is based on the commented out code snippet in the frag shader. I have added an example which shows that the rendered volume now properly "layers" over a mesh visual.

No depth (run on `main`). Note how the mesh pyramid is always in front of the blue volume.
volume_no_depth

With depth (this PR):
depth_with_perspective

Currently, I have implemented it for the isosurface rendering and I plan to also implement it for for MIP, MINIP, and attenuated MIP modes. I am less sure what to do for the translucent and additive modes since the rendered intensity value is not from a particular point. Perhaps for those we can set the depth as the mid point of the ray through the data volume? We could also just not set `gl_FragDepth` for those modes (i.e., leave it how it is currently on main).

When you have time, it would be great to get some feedback if this is a reasonable implementation. If so, I can proceed to add it for the other rendering modes.

A bit of context: this is my proposed fix for https://github.com/napari/napari/issues/2418.


vispy/visuals/volume.py
```diff
@@ -478,6 +485,7 @@
         vec4 color3 = vec4(0.0);  // final color
         vec3 dstep = 1.5 / u_shape;  // step to sample derivative
         gl_FragColor = vec4(0.0);
+        int surface_found = 0;
```
almarklein, line 488: Nitpick: maybe a bool is more clear for this purpose.

kevinyamauchi: Makes sense to me. Done!

djhoese (approved):
Nice job! I'm curious if @almarklein or @brisvag have any opinions on this? Do you see any possible issues with the new plane rendering and clipping functionality interacting with this?

Otherwise, I'm wondering if we need another example for this? I'm not sure if we have a defined policy for when examples should or should not be added, but in this case it feels like this is something most users would just expect (or at least hope) already works. It doesn't show anything unique about how to use the visual or how to structure a vispy application. @almarklein @kmuehlbauer thoughts?

examples/scene/volume_depth.py
```diff
@@ -0,0 +1,82 @@
...
+
+# Create the volume visual
+im_data = np.zeros((30, 30, 30))
+im_data[5:25, 5:25, 5:25] = 1
+blue_cmap = get_colormap('blues')
+volume_1 = scene.visuals.Volume(
+    im_data,
+    cmap=blue_cmap,
+    parent=view.scene,
+    threshold=0.5,
```
djhoese, line 34: This example is actually failing in the tests because of these warnings:

```
Example /scene/volume_depth.py failed (0):
/home/runner/work/vispy/vispy/vispy/gloo/texture.py:28: UserWarning: GPUs can't support floating point data with more than 32-bits, precision will be lost due to downcasting to 32-bit float.
  warnings.warn(F64_PRECISION_WARNING)
WARNING: Program has unset variables: {'u_threshold'}
```

The 64-bit float one should be easy to fix if you set `dtype=np.float32` in your `np.zeros` call. The other one about threshold, I'm not sure why threshold would be unset if the `iso` method uses it (does it?).

kevinyamauchi:
> Otherwise, I'm wondering if we need another example for this? 

I don't think it necessary to have the example. I was mostly including it so the patch would be easy to try out. I am happy to remove the example from the PR if you all think it is superfluous!

examples/scene/volume_depth.py
kevinyamauchi: Thanks for the tip. I've changed the data dtype to `np.float32` and that fixed the precision warning.

I'm a bit confused by the `u_threshold` warning. My understanding is that `u_threshold` is the threshold for finding the surface in ` iso` rendering mode. I tested the example on `main` and I get the same warning, so I think it's a pre-existing issue.

kevinyamauchi: If we end up removing the example, I can make an issue for the `u_threshold` warning.

djhoese: I think this is the result of some weird ordering. In the `Volume.__init__` method, the `self.method` is set first, then the `self.threshold` is set. And `self.update()` is called at the end of each method. Because you set the method to `iso` immediately, it doesn't have a threshold set and tries to draw immediately (I guess) and produces the warning. Looking at the code, even if you used the default `method` first then set it to `iso` later, I think you'd still get the warning no matter what you did because of the way things are set. It looks like it could definitely be improved, but don't worry about it in this PR if others think the example should get removed. If the example does stay then we'll have to think of something.

almarklein:
If we have an existing isosurface volume example, we can perhaps modify it to demonstrate that depth is handled correctly.

almarklein (approved):
Feels like a bugfix to me too.

djhoese:
Thanks @kevinyamauchi! Sorry for the delay on this.

PR #2205, opened 2021-08-31
Fix key event handling in Tk backend

matthiasverstraete:
character keys (letters/digits) are not properly detected by the tkinter backend.
Somehow, the keycode value is not the same as the ordinal of a character in tkinter.
My changes fix this by looking at the 'char' field of the event.

### DIFF
--- a/vispy/app/backends/_tk.py
+++ b/vispy/app/backends/_tk.py
@@ -364,6 +364,7 @@ class CanvasBackend(BaseCanvasBackend, OpenGLFrame):
 
         self._init = False
         self.is_destroyed = False
+        self._dynamic_keymap = {}
 
         OpenGLFrame.__init__(self, parent, **kwargs)
 
@@ -499,8 +500,13 @@ class CanvasBackend(BaseCanvasBackend, OpenGLFrame):
             return KEYMAP[e.keysym_num], ""
         # e.char, e.keycode, e.keysym, e.keysym_num
         if e.char:
+            self._dynamic_keymap[e.keycode] = e.char
             return keys.Key(e.char), e.char
-          
+
+        if e.keycode in self._dynamic_keymap:
+            char = self._dynamic_keymap[e.keycode]
+            return keys.Key(char), char
+
         warnings.warn("The key you typed is not supported by the tkinter backend."
                       "Please map your functionality to a different key")
         return None, None
--- a/vispy/app/backends/_tk.py
+++ b/vispy/app/backends/_tk.py
@@ -7,6 +7,7 @@
 from __future__ import division
 
 from time import sleep
+import warnings
 
 from ..base import (BaseApplicationBackend, BaseCanvasBackend,
                     BaseTimerBackend)
@@ -499,13 +500,10 @@ class CanvasBackend(BaseCanvasBackend, OpenGLFrame):
         # e.char, e.keycode, e.keysym, e.keysym_num
         if e.char:
             return keys.Key(e.char), e.char
-        key = e.keycode
-        if 97 <= key <= 122:
-            key -= 32
-        if key >= 32 and key <= 127:
-            return keys.Key(chr(key)), chr(key)
-        else:
-            return None, None
+          
+        warnings.warn("The key you typed is not supported by the tkinter backend."
+                      "Please map your functionality to a different key")
+        return None, None
 
     def _on_mouse_enter(self, e):
         """Event callback when the mouse enters the canvas.
--- a/vispy/app/backends/_tk.py
+++ b/vispy/app/backends/_tk.py
@@ -497,6 +497,8 @@ class CanvasBackend(BaseCanvasBackend, OpenGLFrame):
         if e.keysym_num in KEYMAP:
             return KEYMAP[e.keysym_num], ""
         # e.char, e.keycode, e.keysym, e.keysym_num
+        if e.char != "":
+            return keys.Key(e.char), e.char
         key = e.keycode
         if 97 <= key <= 122:
             key -= 32
--- a/vispy/app/backends/_tk.py
+++ b/vispy/app/backends/_tk.py
@@ -497,7 +497,7 @@ class CanvasBackend(BaseCanvasBackend, OpenGLFrame):
         if e.keysym_num in KEYMAP:
             return KEYMAP[e.keysym_num], ""
         # e.char, e.keycode, e.keysym, e.keysym_num
-        if e.char != "":
+        if e.char:
             return keys.Key(e.char), e.char
         key = e.keycode
         if 97 <= key <= 122:


djhoese:
Thanks for submitting this fix.

Correct me if I'm wrong, but the code below your new `if` statement will almost never be reached, right? According to this SO answer (https://stackoverflow.com/a/27535889/433202) `.keycode` isn't even consistent between platforms. Should this lower section just be removed? Does using `.char` as you have it do work in all cases?

matthiasverstraete:
> Thanks for submitting this fix.
> 
> Correct me if I'm wrong, but the code below your new `if` statement will almost never be reached, right? According to this SO answer (https://stackoverflow.com/a/27535889/433202) `.keycode` isn't even consistent between platforms. Should this lower section just be removed? Does using `.char` as you have it do work in all cases?

Indeed, my guess is also that this code will never be reached. However, I'm not a 100% sure, so I didn't dare removing it (I'm thinking of special unicode characters maybe?)

djhoese:
@ThenTech What do you think about this PR?

@matthiasverstraete You make a good point that non-ASCII or special characters may be an issue here but I don't think that code would handle them properly anyway. The first branch where it subtracts `32` seems arbitrary and likely based on ASCII characters. Based on that stackoverflow answer I linked to above it is likely specific to @ThenTech's machine when they originally created this backend.

ThenTech:
I haven't looked at the PR yet, but I think it's possible the Tkinter events don't implement anything other than ASCII, but I'll verify it on my local machines.

ThenTech:
```python
        if 97 <= key <= 122:
            key -= 32
```
This was indeed to make just the letters case-independent.

But if the character is a printable ASCII character, `e.char` should be enough.
I'm not sure what should be done about the characters however. On my win10 machine for example, pressing backktick gives:

```python
...
print(e.char, e.keycode, e.keysym, e.keysym_num)
 223 Multi_key 65312
```
with `e.char` being an empty string, so the function would return `None, None` regardless.

matthiasverstraete:
That's surprising, for me it gives `` ` 49 grave 96``. I suppose this is really dependent on the keyboard?
Which keyboard layout are you using?

vispy/app/backends/_tk.py
```diff
@@ -497,6 +497,8 @@ def _parse_keys(self, e):
         if e.keysym_num in KEYMAP:
             return KEYMAP[e.keysym_num], ""
         # e.char, e.keycode, e.keysym, e.keysym_num
+        if e.char != "":
```
almarklein, line 500:
```suggestion
        if e.char:
```

almarklein:
For me (Win10, ENG-US layout): `` ` 192 quoteleft 96``
For me (Win10, ENG-INTlayout): `` 192 Multi_key 65312``

matthiasverstraete:
So the 'char' field is also not a reliable way of getting the correct character...
The 'keycode' field seems to be the only constant, but it doesn't tell you which character it represents.

almarklein:
> The 'keycode' field seems to be the only constant, but it doesn't tell you which character it represents.

No, it does not seem to be the Unicode code point, unfortunately :/

ThenTech:
Yes, on my keyboard as well (UK Extended), tilde is not typed immediately because you can also use it to add an accent to another letter.

See this code in the Qt backend:
```python
    def _keyEvent(self, func, ev):
        # evaluates the keycode of qt, and transform to vispy key.
        key = int(ev.key())
        if key in KEYMAP:
            key = KEYMAP[key]
        elif key >= 32 and key <= 127:
            key = keys.Key(chr(key))
        else:
            key = None
        mod = self._modifiers(ev)
        func(native=ev, key=key, text=str(ev.text()), modifiers=mod)
```

It's very similar and I followed this for the Tkinter implementation. Pressing tilde in Qt gets detected as the `text` is set, but the `key` is also `None`, so it might be acceptable to leave the PR as it is and use only the `char` field. All the "regular" keys can be used already.

matthiasverstraete:
Just to be sure, do you still want me to change anything?
If not, how do I proceed to get this merged in?

djhoese:
So my understanding is that `.keycode` should never be used. That's what that SO answer points out and you can see in documentation like here: https://web.archive.org/web/20190515021108id_/http://infohost.nmt.edu/tcc/help/pubs/tkinter/web/key-names.html As you all have pointed out even using `keysym_num` isn't consistent.

As far as I can tell there is no equivalent to `.text()` from Qt in Tk for non-ASCII characters. So in that case maybe we need to have everything after this new `if` statement removed and replaced with a `RuntimeError` or something that says that Tk does not fully support non-ASCII and special characters? Also @matthiasverstraete note that @almarklein suggested a change to the if statement that I think should be included.

matthiasverstraete:
I've applied the suggestion from @almarklein (thanks for tip).
I agree that we should give a clear warning, but I would still keep the code as a 'best effort'.

djhoese:
> I agree that we should give a clear warning, but I would still keep the code as a 'best effort'.

You mean the `.keycode` code? But didn't the three of you prove that it isn't consistent and depends on keyboard layout? I'd rather error out than have a best guess that is wrong.

ThenTech:
Yes I agree with @djhoese that instead of returning `None`, an error would be clearer to the user, so that he can change the used key to something else.

djhoese:
Actually...doing an exception could mean crashing the application whenever the user (not the application developer) accidentally types a key that isn't recognized. Maybe just a UserWarning with `warnings.warn` and return some decent return value?

ThenTech:
Then I'd say to give the warning in the `return None, None` path, and then still return the `None, None` tuple, so the developer can check if the key is `not None`?

almarklein:
> Actually...doing an exception could mean crashing the application whenever the user (not the application developer) accidentally types a key that isn't recognized.

I don't think an exception here will crash the application, since the exception would be thrown in a key event callback (which typically catch and print exceptions). But yeah, actuallu returning a value gives the calling code a chance to react properly.

djhoese (approved):
Looks good to me!

matthiasverstraete:
I tried this and it seems that tkinter doesn't add the 'char' field on an KeyRelease event...
So this solution is not complete yet...

matthiasverstraete:
Nvm my proposal to use keysym as fallback. That way, a keypress and keyrelease event of the same key would lead to a different callback...
If I print the following in `parse_keys`: `print(e.char, e.keycode, e.keysym, e.keysym_num)` I get

```
b 56 b 98
 56 b 98
```
So on key press it knows it's a 'b', but on key release it doesn't...

ThenTech:
> So on key press it knows it's a 'b', but on key release it doesn't...

So using the `e.keycode` as I did initially should still remain as fallback then?

almarklein:
Thanks for diving into this mess :)

An alternative could be to keep a dict mapping keycode to char. You'd fill it on keypress and loopup the char on release...  Bit of a hack though, not sure if it's worth it. Either way, we have to accept that key events on tk are very limited.

djhoese:
> So using the e.keycode as I did initially should still remain as fallback then?

@ThenTech I thought we decided that `.keycode` was different for everyone depending on computer and layout.

matthiasverstraete:
I like your idea @almarklein. That would be the most robust solution.

matthiasverstraete:
I implemented a dictionary that stores the mapping from keycode to char.
It seems to work now for all cases.

almarklein (approved):

djhoese:
Thank you!

matthiasverstraete:
Thanks for the useful feedback!
I'm not familiar with the release schedule of vispy, but when can I expect this to be part of a release?

djhoese:
So maybe as early as this week? What do you think @almarklein? Am I missing anything?

almarklein:
No, not that I know. I plan to work on jupyter_rfb performance this week, but that won't affect vispy's releases.

djhoese:
Let me look tomorrow at whta we've merged so far and maybe I'll make a bug fix release if it is only this PR and other bug fixes.

matthiasverstraete:
Any update on this perhaps?

djhoese:
I've merged a couple PRs today, but there are still a few remaining before I make a feature release since they are basically done but I want @almarklein to check them one more time.

matthiasverstraete:
Ok, I guess I'll see it when a new release pops up. Thanks!

PR #2215, opened 2021-09-18
Use glfw wrapper from the 3rd party pyGLFW library

### LINKED ISSUE #2214: Replace glfw wrapper with pyglfw
At the moment, vispy has its own glfw wrapper that targets the glfw DLL, and it needs to know where you've installed GLFW.

We should probably change this to use `pip install glfw` (https://github.com/FlorianRhiem/pyGLFW)

_Originally posted by @almarklein in https://github.com/vispy/vispy/issues/2213#issuecomment-920763391_

almarklein:
Closes #2214

By making use of Florian Rhiem's [pyGLFW](https://github.com/FlorianRhiem/pyGLFW), which is a well-maintained wrapper for glfw, we don't have to maintain our own wrapper, plus its easier for users to use this backend because pyGLFW ships the DLL on all (common) platforms.

IIRC at the time that we first wrote the glfw backend, there was not a well-maintained wrapper yet, and binary wheels did not exist yet :)

This PR:
* Imports `glfw` (the library) instead of `vispy.ext.glfw`.
* Produces a warning with a hint to `pip install glfw` if `glfw` could not be imported.
* Fixes the glfw attributes, mostly removing prefixes and changing CamelCase function names to snake_case.
* Moves the `glfw.swap_interval()` down because that needs a valid gl context to be active.
* Deletes the `vispy/ext/glfw.py` module.

@rougier it looks like you wrote most of that low-level wrapper. Are you ok with this change?


rougier:
Yes, I just merged a PR for glumpy to use the pyGLFW which is much better.

Issue #2219, opened 2021-09-24
Explain vispy.visuals.transform using Row Major Notation in API Documentation


hershey890:
The vispy.visuals.transform package takes row-major matrices instead of the column major matrices typically used in OpenGL and computer graphics. It would be incredibly helpful if this could be documented in the [API Documentation](https://vispy.org/api/vispy.visuals.transforms.html) as I lost a few hours and I'm sure others have too. 

Also previously mentioned in #507 

Thanks!

djhoese:
@hershey890 Thanks for filing this. It sounds like you would be a great person to update the documentation to include this information as you know exactly what was missing and what would make it understandable. Any chance you'd be willing to submit a pull request with the updates?

hershey890:
@djhoese I'd be happy to. I'm having trouble finding where the API docs files for the website are stored as this [folder](https://github.com/vispy/vispy/tree/main/doc/api) is empty.

Also I've never contributed to anything before. Is the process just fork repo > make changes > push to my forked repo > create pull request once I'm done so it can be approved and merged

djhoese:
Ah yes, so the API sphinx files are generated on the fly. But that isn't an issue since the contents of those files are all based on the actual docstrings in the python code. So for `vispy.visuals.transforms` you'll want to look for the python function/class object here: https://github.com/vispy/vispy/tree/main/vispy/visuals/transforms. Edit the docstring and then since should update properly in the generated sphinx docs.

As for workflow, you are correct, but I'll make a couple suggestions.

1. Make sure to make a new branch and not to work on `main`. This makes it easier for future changes or if you end up working on multiple things at once or if you need to more easily pull in upstream changes.
2. You can create a PR as soon as you make your branch on your fork. Even if you aren't completely done this will let the vispy maintainers review your code earlier if we have a moment and lets us run the automated testing (which includes website generation) to make sure things work as expected. You can even create your pull request as a "draft" pull request and then say it is "Ready for Review" later to let the maintainers know you are done.

I will look into the Edit this page issue. Thanks.

djhoese:
I should also note that if you try to generate the documentation locally it will try to generate the sphinx gallery examples. To turn this off do:

```
cd doc
make html_dev-pattern PATTERN="xxx"
```

The "xxx" is just a filename pattern that won't match any examples so it won't generate anything.

djhoese:
@hershey890 any chance you have time to work on these docs updates?

PR #2220, opened 2021-09-24
Fix "Edit this page" links for API docs

djhoese:
See #2219 for where this issue was first discovered. The summary is that the sphinx site shows "Edit this page" buttons on the right which leads to a GitHub edit page for that particular `.rst` document in the repository. This doesn't work for our API docs as they are dynamically generated and don't live in the repository. This PR does two things:

1. Removes the old `linkcode` logic to point to the external source files. This was used when the docs and code repos were separate, but now that they are one the links should just work.
2. Adds some hacky custom URL stuff to the pydata-sphinx-theme context so that when the file being referenced is an API document then generate a different URL for it that points to the actual python source.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -15,6 +15,7 @@ from datetime import date
 import sys
 import os
 import re
+import warnings
 import vispy
 
 # If extensions (or modules to document with autodoc) are in another directory,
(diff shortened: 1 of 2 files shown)


djhoese:
I'm going to ask the pydata-sphinx-theme people if they have any ideas for doing this in a better way.

almarklein (approved):

PR #2237, opened 2021-10-20
Readd Python 3.10 to wheel building

djhoese:


djhoese:
Darn. Numpy doesn't have wheels for all systems for Python 3.10.

djhoese:
See https://github.com/scipy/oldest-supported-numpy/pull/27 where (once released) this should make these builds more consistent. I hope :crossed_fingers:

PR #2257, opened 2021-11-18
Use stacklevel in DeprecationWarnings.

Carreau:
That give the correct location most of the time for where the fix/update
should be instead of pointing into vispy codebase.

That make updates by consumer easier (and thus more likely),
And in turn should make you more confident about removing deprecated
things.

### DIFF
--- a/vispy/gloo/texture.py
+++ b/vispy/gloo/texture.py
@@ -105,9 +105,12 @@ class BaseTexture(GLObject):
         GLObject.__init__(self)
         if resizeable is not None:
             resizable = resizeable
-            warnings.warn('resizeable has been deprecated in favor of '
-                          'resizable and will be removed next release',
-                          DeprecationWarning)
+            warnings.warn(
+                "resizeable has been deprecated in favor of "
+                "resizable and will be removed next release",
+                DeprecationWarning,
+                stacklevel=2,
+            )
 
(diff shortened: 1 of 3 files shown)


djhoese:
Wow, awesome. Thank you for this. I always forget about `stacklevel` when dealing with this kind of stuff. @kmuehlbauer or @almarklein any issue with these? Their placement seems to make sense to me.

Carreau:
> Wow, awesome. Thank you for this. I always forget about `stacklevel` when dealing with this kind of stuff.

No worries, this is one of my pet peeves, with another that I didn't fixed here which is "since which version" things are deprecated. For example vispy/gloo/texture.py's seem to have been deprecated for 7 years since 0.4.  And honestly I don't get why default value for stacklevel is not 2 , it's almost always the right value (or a mandatory parameter).

kmuehlbauer:
@Carreau Thanks, this is really worth it and it is also requested in other packages I contribute.

We could think about adding deprecation tests which will fail, if not removed in a specific version. But that would mean to have at least a RC release to get these tests into effect.

almarklein (approved):
Awesome!

djhoese:
> with another that I didn't fixed here which is "since which version" things are deprecated. 

Pull requests welcome :smile:

PR #2264, opened 2021-11-28
Add typing to `create_visual_node`

tlambert03:
small change that makes it much easier to work with vispy classes in an IDE.

currently, since `create_visual_node` dynamically creates a type, you pretty much lose all type hinting and autocompletion in an IDE when working with visual nodes. 

Untitled3

With this change, you at least retain the methods and constructor on the corresponding Visual, (though, due to the dynamic type creation, it's still not possible to retain the merged attributes of both the `VisualNode` and the underlying `Visual`).

Untitled1

Untitled2

### DIFF
--- a/vispy/scene/visuals.py
+++ b/vispy/scene/visuals.py
@@ -18,6 +18,10 @@ import weakref
 from .. import visuals
 from .node import Node
 from ..visuals.filters import Alpha, PickingFilter
+from typing import TypeVar
+
+
+_T = TypeVar("_T")
 
 
 class VisualNode(Node):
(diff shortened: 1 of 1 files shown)


djhoese (approved):
This is a great idea. Thanks for submitting this. I'm fine with this being merged.

I think in the long run I would hope that we could get rid of the separate Scene and Visual versions of the visual classes. I would rather just have one Visual that could be used with a regular Canvas or a SceneCanvas. This may not be simple but I just want to throw the idea out there.

kmuehlbauer (approved):
This is a very convenient addition.

PR #2266, opened 2021-12-03
MNT: unpin mesalib 

kmuehlbauer:
There has been some issue with mesalib, which might be fixed by https://github.com/conda-forge/mesalib-feedstock/pull/40.

Just having a look, if this will also fix the issues here.


kmuehlbauer:
Unfortunately this didn't work. I'll leave it open for further testing purposes.

PR #2269, opened 2021-12-06
remove utf-8 encode in _vispy_set_title

Llewyllen:
the encoding is already done in glfw.set_window_title through a call to _to_char_p (at least since glfw 2.3.0 )

### DIFF
--- a/vispy/app/backends/_glfw.py
+++ b/vispy/app/backends/_glfw.py
@@ -320,7 +320,7 @@ class CanvasBackend(BaseCanvasBackend):
         if self._id is None:
             return
         # Set the window title. Has no effect for widgets
-        glfw.set_window_title(self._id, title.encode('utf-8'))
+        glfw.set_window_title(self._id, title)
 
     def _vispy_set_size(self, w, h):
         if self._id is None:


kmuehlbauer:
Thanks @Llewyllen much appreciated. Would you mind putting in a version check for backwards compatibility?

almarklein (approved):

almarklein:
I don't think a version check is necessary. The thing is that we used our own builtin glfw wrapper, and recently shifted to using https://github.com/FlorianRhiem/pyGLFW. I must have missed this bit.

djhoese:
Ok so this could be considered a bug fix then?

almarklein:
Looks like it is.
image

almarklein:
This only occurs when people change the title after the figure has been created. Which explains why this has not yet been reported.

kmuehlbauer:
> I don't think a version check is necessary. The thing is that we used our own builtin glfw wrapper, and recently shifted to using https://github.com/FlorianRhiem/pyGLFW. I must have missed this bit.

Ah, I haven't thought about that. Then we do not need backwards compatibility ;-) @Llewyllen.

kmuehlbauer (approved):

PR #2282, opened 2022-01-05
Feature: Add GL_MIN, GL_MAX modes to glBlendEquation

psobolewskiPhD:
These small changes add GL_MIN and GL_MAX modes for glBlendEquation (see: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendEquation.xhtml) which makes them available via `gloo.set_state`.

This permits, for example, implementing `minimum` blending in napari, as in this branch:
https://github.com/psobolewskiPhD/napari/tree/feature/min-blending

Note: 
1. The GL_MIN and GL_MAX macros in `gl2.h` are from:
http://podsvirov.github.io/osg/reference/openscenegraph/a01104.html
2. The constants in `_constants.py` are from:
http://pyopengl.sourceforge.net/documentation/pydoc/OpenGL.GL.VERSION.GL_1_4.html


psobolewskiPhD:
Based on the test failures I now realize why these weren't available before: they're part of GL ES 3, not 2. 
I was fooled by the Version support graph here:
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendEquation.xhtml
Makes this PR a no-go, correct?

djhoese:
No the test failures (at least the "full, standard" one) seems to be a test that is explicitly checking what constants exist. I think it would be OK to update those to expect these names...I think. I've spent 30 seconds look at this though so feel free to let me know if I'm missing something.

Also looks like the wheel builds need to remove Python 3.6. @vispy/core any problems with removing 3.6 and 3.7 (now that numpy no longer supports 3.7)?

psobolewskiPhD:
I'm sorry, I'm a total noob, who was just copy pasting :cry: 
(the blending in napari does work though!)
I maybe misunderstanding, but looking here:
https://github.com/vispy/vispy/runs/4719566402?check_suite_focus=true#step:11:781
I think it means this test:
https://github.com/vispy/vispy/blob/26256fdc2574259dd227022fbce0767cae4e244b/vispy/gloo/gl/tests/test_names.py#L41-L45

And this seems to imply ES 2.0 only. And looking at the called functions, I think I could fix the test by editing this section and adding GL_MIN and GL_MAX: 
https://github.com/vispy/vispy/blob/26256fdc2574259dd227022fbce0767cae4e244b/vispy/gloo/gl/tests/test_names.py#L144-L234
But that seems to defeat the purpose of the test?
Perhaps my error was to insert this into gl2.h ?

djhoese:
@almarklein @kmuehlbauer @larsoner @rossant Any opinions on this?

My opinion, which is based on passed discussion with the other maintainers on this, is that we don't care about keeping strict ES/2.0 compatibility. As long as it works on "most" modern systems then it is fine. However, I think you're right that editing the codegen `.h` files is probably not the right approach unless it is a last resort. I believe those files are copied from some past OpenGL version's header files. I'm not sure if there is another location in vispy where they should/could be manually updated.

djhoese:
Making a PR started this discussion so that's all fine. Others may have ideas for where the changes could go.

kmuehlbauer:
@psobolewskiPhD _constants.py is created within :
https://github.com/vispy/vispy/blob/main/codegen/createglapi.py

IIUC we would need to tamper/add the constants there...

larsoner:
According to memory (and `git blame`) Someone named @almarklein wrote most of that code :) So he probably knows the right way to make these changes...

almarklein:
Yeah, I think the best way to address this is to define a set of "auxillary constants" in the file that Kai [mentioned above](https://github.com/vispy/vispy/pull/2282#issuecomment-1006268746).

BTW, there is this, but it looks like this file is not used in the codegen.
https://github.com/vispy/vispy/blob/26256fdc2574259dd227022fbce0767cae4e244b/codegen/headers/gl2ext.h#L451-L455

djhoese:
So maybe codegen should get updated to include that `gl2ext.h` file in its generation? @psobolewskiPhD Do you want to take a shot at that?

almarklein:
> So maybe codegen should get updated to include that gl2ext.h file in its generation?

Well, that's not ideal, because then these extras all have `_EXT` suffixes. We could include our own `ext.h` if you want to do it really cleanly :)

psobolewskiPhD:
Thanks for the feedback!!
I saw those GL_MAX_EXT, but couldn't get them to work. 
Assumed they were related or something different.
I'll try to look at https://github.com/vispy/vispy/blob/main/codegen/createglapi.py
Not sure when I'll have time to make some progress, but...

BTW because as far as I can tell GL_MAX and GL_MIN are GL3, maybe should make a gl3.h?

djhoese:
I'd say an `ext.h` like @almarklein suggested would be good enough as long as it has comments/sections for what GL version the constant was introduced in (ex. one section for GL3, one for GL4, etc).

almarklein:
> maybe should make a gl3.h?

Well, that may give the false impression that we include *all* gl3 constants, so I like what David proposed.

psobolewskiPhD:
First, I *really* appreciate the discussion and advice.
IMO, this PR is DOA—it's just not correctly implemented.
I think it's easiest—for me anyways—to make a new branch starting from main vispy.
Shall I try giving that a go and then make a new PR if I get something functional? (closing this PR) or is there some other avenue?
Sorry about the stupid questions, but I'm *really a newbie* regarding PRs and such, let alone anything related to actual code. (Maybe I should list "googlefu" and "copypasta" on my CV tho...)

djhoese:
If you'd like to make a new branch and start a new PR that's fine. If you'd like to revert your changes in this current branch and continue with your new attempts at making this work that is fine also. Either way is fine as long as any new PRs on the subject reference this one in the comments.

psobolewskiPhD:
Thanks! I'm also a git newb—shocking amirite—so if I can figure out reverting I'll give it a whirl. Else I will make a new branch and if it works a new PR with links/reference to this one.
(I'll prolly not have too much time till friday to give it a whirl...)

djhoese:
To revert, you can search for examples using `git revert` or just make a new commit that undoes all the changes you did. Then push it to the same branch on github and the PR should be updated.

djhoese:
I see there were commits after my last comment, but must have missed the emails. Sorry. What is the state of this PR? Ready for re-review?

djhoese:
> I can try and take one more stab at it, maybe next week? But I'm not optimistic that my skills have sufficiently advanced from January.

Don't be too hard on yourself. I'm preparing to go on vacation so I may not have a ton of time to look at this, but will try after our next release.

djhoese:
Closed via #2320

PR #2284, opened 2022-01-20
Fix touch event handling issue for Qt6-based backends

mars0001:
The touchPoints() method is replaced by points() in the Qt6 framework. The pos() and lastPos() methods are replaced in Qt6 by position() and lastPosition().

Pyside6 only mention the deprecation of the pos() and lastPos() methods but PyQt6 does not have them at all.

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -5,7 +5,7 @@
 """
 Base code for the Qt backends. Note that this is *not* (anymore) a
 backend by itself! One has to explicitly use either PySide, PyQt4 or
-PySide2, PyQt5. Note that the automatic backend selection prefers
+PySide2, PyQt5 or PyQt6. Note that the automatic backend selection prefers
 a GUI toolkit that is already imported.
 
 The _pyside, _pyqt4, _pyside2, _pyqt5 and _pyside6 modules will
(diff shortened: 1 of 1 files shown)


djhoese:
Thanks! This makes me think we really need to drop Qt4 support. Hopefully that would clean up some of this special handling (I realize this is Qt5 vs 6). Anyway, we'll wait for the tests to pass and then I'm ok with this. It looks like we don't have a test for this yet. Any idea how one could be made? Would be nice to have, but I understand if it is too difficult.

mars0001:
Sorry David, I'm not a programmer by trade, just pulling my weight on a hobby I'm infected with :)

djhoese:
@kmuehlbauer @almarklein Any ideas on how a test could be written for this?

kmuehlbauer:
Sorry, I'm lacking in the field of UI backends.

almarklein:
I'm sure there's tool to write tests that run a qt gui, do some clicks, and check results, but setting this up and maintaining it probably costs more time than simply fixing what is broken, like this PR does. We don't have *that* much GUI-specific code.

> This makes me think we really need to drop Qt4 support.

Good idea. I don't think PyQt4 and Pyside are supported on the minimum Python version that vispy supports.

almarklein (approved):

djhoese:
Thanks @mars0001!

PR #2288, opened 2022-01-31
New Feature: "Axis Mapping" to allow dates on axis

Bliss3d:
This is a feature I was sorely missing - axis with date labels. 
It allows to map pretty much anything to the axis, but its mostly intended for dates.
It works pretty well, there are some (very small) rounding issues when it comes to very zoomed in seconds but I would say its very usable.
Also in this pull request is an update to my Bar plotter which works very nice now with all sorts of lists and arrays

Take a look for yourself: 
vispy4


vispy/visuals/axis.py
```diff
@@ -392,9 +458,133 @@ def _get_tick_frac_labels(self):
             length = self.axis.pos[1] - self.axis.pos[0]  # in logical coords
             n_inches = np.sqrt(np.sum(length ** 2)) / transforms.dpi
 
-            major = _get_ticks_talbot(domain[0], domain[1], n_inches, 2)
+            major = _get_ticks_talbot(domain[0], domain[1], n_inches, density)
+
+            labels = []
+
+            if self.axis.axis_mapping is not None:
```
djhoese, line 465: There is a lot added here under this if statement. Would it be possible for you to move this into a separate function/method and then further split each `if` branch into a separate function? It looks like there is at least some duplicated code that could be simplified that way.

vispy/visuals/axis.py
```diff
@@ -392,9 +458,133 @@ def _get_tick_frac_labels(self):
...
+                    if x >= 0:
+                        try:
+                            label = self.axis.axis_mapping[int(x)]
+                            label_plus_one = self.axis.axis_mapping[int(x) + 1]
+                        except IndexError:
+                            label = x
+                            label_plus_one = None
+                            print("Axis mapping ran out of data at index %s use update_axis_mapping() in AxisVisual "
+                                  "to extend the mapping range in Runtime or map a longer list/array in the first "
+                                  "place" % str(int(x)))
```
djhoese, line 477: It looks this should maybe be a `warnings.warn`? When does this condition get encountered? Will this be a common occurrence if people are panning/zooming on a plot?

djhoese (changes requested):
This looks really nice! I think there are a couple things that could be cleaned up and I've commented inline. Also it seems this PR includes your Bar chart visual stuff. Would it be possible to separate this new Axis functionality from the bar chart commits?

Bliss3d:
>  Would it be possible to separate this new Axis functionality from the bar chart commits?

I expected that...

In theorie yes i could split the commits however I pushed the Bar Graph commit to main of my fork (I didnt think about it...) so those changes are also in any new branch. Im not a git wizard so what would be the best way to proceed?

vispy/visuals/axis.py
Bliss3d: Good question I'm also not quite sure. zooming out can obviously cause this.
I was thinking about interpolating the dates once the mapping runs out of entries. i.e. self.axis.axis_mapping[last] - self.axis.axis_mapping[2nd_to_last]
then for x = last+ n -> self.axis.axis_mapping[last] + n * (self.axis.axis_mapping[last] - self.axis.axis_mapping[2nd_to_last])

This would only work if the mapping has a constant delta!

In case of people not mapping dates, something like strings I have no clue what to do - probebly a warnings.warn would be good

vispy/visuals/axis.py
Bliss3d: Yes working on it

djhoese:
> In theorie yes i could split the commits however I pushed the Bar Graph commit to main of my fork (I didnt think about it...) so those changes are also in any new branch. Im not a git wizard so what would be the best way to proceed?

Well it looks like this current branch has intermixed commits with the bar graph commits so it isn't super easy. You have the commit "Numerous improvements to Bar Plotting" in this PR, but it isn't in the bar graph PR so you'll probably want to fix that up too. So the best fix I can think of is:

```
git checkout main
git cherry-pick 1f647d65c598986c121c319dfb24d371fee4c174
git push
```

this updates your bar graph PR with that commit. Then to fix this PR do the following. I assume you have two remotes, one named `origin` which is your fork and one named `upstream` which is the vispy repository (if you don't have this then `git add remote upstream git@github.com:vispy/vispy.git`).

```
git fetch upstream
git checkout -b date_axis2 upstream/main
git cherry-pick 814267a7c931982367e48b066febac1517a219bc
git cherry-pick a26125b418f84ca8cf247c3e0abf2807f511e77a
git push -u origin date_axis2
```

Then you could either make a new PR off of date_axis2 or try force merging it into `date_axis`, but I'm not sure how happy git will be about that.

Bliss3d:
> > In theorie yes i could split the commits however I pushed the Bar Graph commit to main of my fork (I didnt think about it...) so those changes are also in any new branch. Im not a git wizard so what would be the best way to proceed?
> 
> Well it looks like this current branch has intermixed commits with the bar graph commits so it isn't super easy. You have the commit "Numerous improvements to Bar Plotting" in this PR, but it isn't in the bar graph PR so you'll probably want to fix that up too. So the best fix I can think of is:
> 
> ```
> git checkout main
> git cherry-pick 1f647d65c598986c121c319dfb24d371fee4c174
> git push
> ```
> 
> this updates your bar graph PR with that commit. Then to fix this PR do the following. I assume you have two remotes, one named `origin` which is your fork and one named `upstream` which is the vispy repository (if you don't have this then `git add remote upstream git@github.com:vispy/vispy.git`).
> 
> ```
> git fetch upstream
> git checkout -b date_axis2 upstream/main
> git cherry-pick 814267a7c931982367e48b066febac1517a219bc
> git cherry-pick a26125b418f84ca8cf247c3e0abf2807f511e77a
> git push -u origin date_axis2
> ```
> 
> Then you could either make a new PR off of date_axis2 or try force merging it into `date_axis`, but I'm not sure how happy git will be about that.

Okay thank you very much.
I guess you meant git remote add upstream git@github.com:vispy/vispy.git
but worked out fine - I'll fix up the if condition marathon and figure out if I can implement date interpolation if the mapping runs out of entries and create than a new PR

Best regards

Bliss3d:
@djhoese It now quite a bit shorter and optimized, however I dont see a way to break it up in more functions and not being kinda stupid
this is the new version, if you have any suggestions I"ll be happy to make changes
`            if self.axis.axis_mapping is not None:
                for x in major:
                    x = round(x, 10)
                    if x >= 0:
                        try:
                            label = self.axis.axis_mapping[int(x)]
                            label_plus_one = self.axis.axis_mapping[int(x) + 1]

                        except IndexError:
                            label = x
                            label_plus_one = None
                            disable_warming = False
                            if self.axis.axis_mapping_is_date:
                                last_date = self.axis.axis_mapping[len(self.axis.axis_mapping) - 1]
                                second_to_last_date = self.axis.axis_mapping[len(self.axis.axis_mapping) - 2]
                                date_integer_delta = last_date - second_to_last_date
                                for j in range((int(x) + 1) - (len(self.axis.axis_mapping) - 1)):
                                    self.axis.axis_mapping.append(last_date + ((j + 1) * date_integer_delta))
                                disable_warming = True
                                label = self.axis.axis_mapping[int(x)]
                                label_plus_one = self.axis.axis_mapping[int(x) + 1]

                            if not (disable_warming or self.axis.disable_warning):
                                warnings.warn("Axis mapping ran out of data at index " + str(int(x)) +
                                              "use update_axis_mapping() in AxisVisual to extend the mapping range in "
                                              "Runtime or map a longer list/array in the first place - As last resort "
                                              "you can disable this warning by setting disable_warning to True")
                            if not self.axis.disable_warning:
                                label = ' '

                        if x.is_integer():
                            if type(label) == datetime.datetime:
                                if self.axis.date_format_string is not None:
                                    date_string = str(label.strftime(self.axis.date_format_string))
                                    labels.append(
                                        (date_string + " " + str(x)) if self.axis.date_verbose else date_string)
                                else:
                                    labels.append(
                                        (str(label) + " " + str(x)) if self.axis.date_verbose else str(label))
                            else:
                                labels.append((str(label) + " " + str(x)) if self.axis.date_verbose else str(label))

                        else:
                            if type(label) == datetime.datetime and label_plus_one is not None:

                                a = label
                                b = label_plus_one

                                date = (b - (b - a) / round((1 - (x - int(x))) ** -1, 3))

                                if self.axis.rounding_seconds:
                                    date = _round_seconds(date)

                                if self.axis.date_format_string is not None:
                                    date_string = str(date.strftime(self.axis.date_format_string))
                                    labels.append(
                                        (date_string + " " + str(x)) if self.axis.date_verbose else date_string)
                                else:
                                    labels.append(
                                        (str(date) + " " + str(x)) if self.axis.date_verbose else str(date))

                            else:
                                labels.append('')
                    else:
                        labels.append("")
            else:
                labels = ['%g' % x for x in major]`

djhoese:
I'm reading the book "Clean Code" and there isn't much you can do to this that would be "stupid". For example the datetime handling:

```python
                            if self.axis.date_format_string is not None:
                                date_string = str(label.strftime(self.axis.date_format_string))
                                labels.append(
                                    (date_string + " " + str(x)) if self.axis.date_verbose else date_string)
                            else:
                                labels.append(
                                    (str(label) + " " + str(x)) if self.axis.date_verbose else str(label))
```

Could be put into a function to return the new label string. Then this chunk of code just gets replaced by something like:

```python
labels.append(_label_string_for_datetime(x, label, self.axis.date_format_string)
```

Or whatever else is needed. Also note that you should use `isinstance(label, datetime.datetime)` instead of `type(label) ==`.

Another possible code cleanup would be to pull out all the `labels.append` and only set one `label_string` and then at the end do `labels.append`. This will make it easier to extract the block of code that assigns to `label_string` into its own function too.

djhoese:
Closing as it is replaced by #2290

PR #2293, opened 2022-02-04
Fix TypeError for membership test of KeyboardModifiers

### LINKED ISSUE #2212: Python DeprecationWarning on PyQt 5.12.3
I don't have a simple reproducer yet, as I encounter this on a rather complex napari workflow, but I wanted to write it down in case others have encountered it. Under certain circumstances and *only* with older PyQt versions (5.12.3 in this case), I get the following warning emitted from VisPy:

```pytb
/Users/jni/conda/envs/all/lib/python3.9/site-packages/vispy/app/backends/_qt.py:627: DeprecationWarning: an integer is required (got type KeyboardModifiers).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.!
```

VisPy version is 0.8.1.

I was able to silence the warning by wrapping the relevant code block into a

```python
with warnings.catch_warnings():
    warnings.simplefilter('ignore')
    ...
```

context. But I don't know if that's the desired workaround. Suggestions are super welcome! Here is the offending code:

https://github.com/vispy/vispy/blob/5d58b832c853c45b44f5036b9e7c90d6e9f4f454/vispy/app/backends/_qt.py#L627-L628

Note: I can't update my PyQt version because that is the latest available on conda-forge and the pip version breaks a *different* part of my workflow. 😂

rayg-ssec:
KeyboardModifiers follows from C++ QFlags<> which is closer to set-like semantics,
even if its implementation may be a bit field. Swapping to `qtmod & q` lets it
perform properly as a result.

This was triggering in SIFT 1.2.3 on mouseMove events as
```
TypeError: unsupported operand type(s) for &: 'KeyboardModifier' and 'KeyboardModifiers'
```

Also a minor comment fix to a presumed-missing "Compat" comment since it follows from nearby kin.

Closes #2212

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -97,7 +97,7 @@ elif qt_lib == 'pyqt5':
         else:
             from PyQt5.QtOpenGL import QGLWidget, QGLFormat
     from PyQt5 import QtGui, QtCore, QtWidgets, QtTest
-    QWidget, QApplication = QtWidgets.QWidget, QtWidgets.QApplication  # 
+    QWidget, QApplication = QtWidgets.QWidget, QtWidgets.QApplication  # Compat
 elif qt_lib == 'pyqt6':
     _check_imports('PyQt6')
     if not USE_EGL:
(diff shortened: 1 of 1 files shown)


djhoese (approved):
Looks good! Thanks @rayg-ssec. @jni this should fix your issue/DeprecationWarning on Python 3.9.

Issue #2280, opened 2021-12-20
Possible optimization in shader compilation


djhoese:
While using `ipdb` to debug a users code on gitter I noticed this block of code in the low-level gloo Program:

https://github.com/vispy/vispy/blob/26256fdc2574259dd227022fbce0767cae4e244b/vispy/gloo/program.py#L238-L252

Nothing about this shader regular expression stuff is specific to the code being parsed from what I can tell. Additionally, the list of `kind`s that is iterated over is static *and* the regexes being used are compiled every time. This should all be done in the global scope **once** and then reused for all vispy Programs. This maybe won't make a huge difference in performance but it will reduce the extra work needed if/when shaders do require compilation.

PR #2297, opened 2022-02-06
Optimize shader regular expression compilation

PydPiper:
proposed update related to #2280

### DIFF
--- a/vispy/gloo/program.py
+++ b/vispy/gloo/program.py
@@ -42,13 +42,13 @@ from .preprocessor import preprocess
 REGEX_VAR = {}
 for kind in ('uniform', 'attribute', 'varying', 'const', 'in', 'out'):
     REGEX_VAR[kind] = re.compile(fr"\s*{kind}\s+"  # kind of variable
-                                  r"((highp|mediump|lowp)\s+)?"  # Precision (optional)
-                                  r"(?P<type>\w+)\s+"  # type
-                                  r"(?P<name>\w+)\s*"  # name
-                                  r"(\[(?P<size>\d+)\])?"  # size (optional)
-                                  r"(\s*\=\s*[0-9.]+)?"  # default value (optional)
-                                  r"\s*;"  # end
-                                  , flags=re.MULTILINE)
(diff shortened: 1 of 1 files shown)


djhoese (approved):
This looks good to me. @almarklein do you have any comments on this?

I'm not sure the easiest way to profile this to see how much of an improvement it is. I suppose we could just time how long it takes to generate those regular expressions and then we know that that much time is saved now when compiling shaders.

djhoese:
```
In [5]: import re

In [6]: %%time
   ...: REGEX_VAR = {}
   ...: for kind in ('uniform', 'attribute', 'varying', 'const', 'in', 'out'):
   ...:     REGEX_VAR[kind] = re.compile(fr"\s*{kind}\s+"  # kind of variable
   ...:                                  r"((highp|mediump|lowp)\s+)?"  # Precision (optional)
   ...:                                  r"(?P<type>\w+)\s+"  # type
   ...:                                  r"(?P<name>\w+)\s*"  # name
   ...:                                  r"(\[(?P<size>\d+)\])?"  # size (optional)
   ...:                                  r"(\s*\=\s*[0-9.]+)?"  # default value (optional)
   ...:                                  r"\s*;",  # end
   ...:                                  flags=re.MULTILINE)
   ...:
CPU times: user 4.48 ms, sys: 44 µs, total: 4.52 ms
Wall time: 4.48 ms
```

Dang! Milliseconds!

PydPiper:
you are also not doing string.replace() calls either but instead a 1 time f-string injection thats way faster

djhoese:
FYI I wanted to see what the difference would be if we were still using `.replace`. It is about 1ms more. However, I found out that `re.compile` seems to cache the regex so running the same `%%time` cell multiple times only takes ~20us. So maybe this isn't as big of an optimization as we thought, but still an improvement.

djhoese:
I don't see any objections from anyone so I'm going to merge this. Thanks again @PydPiper! The only other thing I could think of for optimizing this and saving vispy on import time is to move the dictionary to a class property on the Program class that gets created the first time a Program is created. I'm not sure it is that big of a deal right now though.

If any other maintainers have a problem with this PR comment here and I'll fix it.

Issue #2298, opened 2022-02-06
Create VSCode Remote-Containers development environment


shipshupe:
To get up and running quickly for local development, wondering if it'd be valuable to configure a remote-containers definition?

Overview of the tool (vscode specific extension)
https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers

What this would look like is a .devcontainer folder at the root level, defining a DockerFile and a runlevel devcontainer.json to instruct the VSCode extension to reopen in Container.

This would be based on Debian / Ubuntu, which with WSLG (at least in Windows 11) can now draw to the desktop and NVIDIA provides CUDA support for WSL2.

## Future state:

Dependencies required:
  X server (provided by Linux or WSLG)
  Docker-ce / container service running
  Host system with VSCode + Remote-Containers extension (in WSL2 on Windows 11, otherwise at root)
  CUDA Drivers (in WSL2 for Windows 11)

After that, fork/clone the repository and open in VSCode. 

It should open and prompt 'this looks like a development environment, do you want to build and open it?'

Click yes
At the end of the pull/build process, run `python examples/{xx}.py` and it should open a window.
Make a change, `python setup.py build/install`(?) and test
Push changes and PR

shipshupe:
Where we left off, WSLG in the container had an issue drawing out to the canvas, and it seemed very similar to https://github.com/vispy/vispy/issues/2068

I did try the suggestion to `pip install pyglet` and edit an example file to use that as the app engine, that worked!

results of `python -c "import vispy; print(vispy.sys_info())"` after that are below

Platform: Linux-5.10.60.1-microsoft-standard-WSL2-x86_64-with-glibc2.31
Python:   3.9.7 (default, Sep 16 2021, 13:09:58)  [GCC 7.5.0]
NumPy:    1.20.3
Backend:  PyQt5
pyqt4:    None
pyqt5:    ('PyQt5', '5.9.2', '5.9.7')
pyqt6:    None
pyside:   None
pyside2:  None
pyside6:  None
pyglet:   pyglet 1.5.21
glfw:     None
sdl2:     None
wx:       None
egl:      EGL 1.5 Mesa Project: OpenGL OpenGL_ES
osmesa:   None
tkinter:  None
jupyter_rfb: None
_test:    None

PR #2299, opened 2022-02-07
add remote-container definition

shipshupe:
WIP proof of concept for

https://github.com/vispy/vispy/issues/2298

Worked with @djhoese during PyCascades Sprint

TODO:
Still some things to debug but on Windows 11, will open a WSLG window

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,26 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.209.6/containers/python-3-anaconda/.devcontainer/base.Dockerfile
+
+FROM mcr.microsoft.com/vscode/devcontainers/anaconda:0-3
+
+# [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
+ARG NODE_VERSION="none"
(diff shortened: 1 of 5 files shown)


shipshupe:
Sounds great! I may have some time tomorrow about 8:30am PST or this weekend to look at it more. Would like to test with both a Linux and WSLG host OS and add some documentation. 

Thanks for entertaining this idea, have found it fun to explore the 'ship the dev environment' tooling that both remote-containers and Binder offer!

djhoese:
I won't be able to do tomorrow, but if you are able to spend time on it and can let me know how it goes then I could meet 11am-ish on Saturday if we need further testing/debug. I followed you on twitter so we can DM there if we need further scheduling. Otherwise, little updates here may be nice for tracking the progress if we can't get it all working right away.

shipshupe:
I've updated the pull request with findings that the blank canvas seems to be tied to PyQt5, changing this out allowed a sampling of example programs to run but seemed a bigger scope issue I don't have the experience base to tackle.

It seemed like this is a known incompatibility between PyQt5 and WSLG so will try again if I see progress there. 

That's it for now :)

shipshupe:
Yep! Sorry, had just commented in the pull request. Here on the issue is a better place, I can see now!

Can test that out and push a change, is there a configuration default option for vispy GUI framework?

djhoese:
We can force it in scripts (with extra code), but otherwise it prefers qt-based backends over others. So the answer is: no, not yet.

shipshupe:
No worries, can work around it. Thanks!

PR #2306, opened 2022-03-02
moved a debug line to where it works

olinickalls:
Moved line 96 into the Canvas class (on_mouse_press method) so it can now see the event object.

The next line
    # pos_scene = event.pos[:3]
would also fail in uncommented but as it changes the pos_scene variable it is not so straightforward to fix unless event is also sent to the method, changing the API.

### DIFF
--- a/examples/basics/visuals/line_draw.py
+++ b/examples/basics/visuals/line_draw.py
@@ -93,7 +93,6 @@ class EditLineVisual(scene.visuals.Line):
                               size=size, face_color=self.marker_colors)
 
     def on_mouse_press(self, pos_scene):
-        # self.print_mouse_event(event, 'Mouse press')
         # pos_scene = event.pos[:3]
 
         # find closest point to mouse and select it
@@ -168,6 +167,8 @@ class Canvas(scene.SceneCanvas):
         self.freeze()
 
     def on_mouse_press(self, event):
+        # self.line.print_mouse_event(event, 'Mouse press')
+
         tr = self.scene.node_transform(self.line)
         pos = tr.map(event.pos)
         self.line.on_mouse_press(pos)


djhoese:
So if the line of code was uncommented where it was, it failed? Sorry, not sure I understand what this line is doing and what it needs as input.

olinickalls:
Sorry- really tiny change here - I feel I'm making a meal out of it.  I do apologise if I am describing this badly as well- I am not a professional software engineer...

The line I moved was line 95, in `EditLineVisual.on_mouse_press()` method.
It needs the `event` object passed by the `on_mouse_press` _event_, although this seems to only be handled by the `Canvas.on_mouse_press()` re-implemented method (line 170).

original 
```
class EditLineVisual(scene.visuals.Line):

    <snip>

    def on_mouse_press(self, pos_scene):
        # self.print_mouse_event(event, 'Mouse press')
        # pos_scene = event.pos[:3]

        # find closest point to mouse and select it
        self.selected_point, self.selected_index = self.select_point(pos_scene)

        # if no point was clicked add a new one
        <snip>
```

Where it was (1st line of `EditLineVisual.on_mouse_press()`) the `event `object was inaccessible. Here's the stack trace with the line uncommented:

```
    << caught exception here: >>
    File "c:\Users\oli_n\Documents\pycode\vispy\.env\lib\site-packages\vispy\util\event.py", line 469, in _invoke_callback
      cb(event)
    File "c:\Users\oli_n\Documents\pycode\vispy\line_draw.py", line 173, in on_mouse_press
      self.line.on_mouse_press(pos)
    File "c:\Users\oli_n\Documents\pycode\vispy\line_draw.py", line 96, in on_mouse_press
      self.print_mouse_event(event, 'Mouse press')
  NameError: name 'event' is not defined
  ERROR: Invoking <bound method Canvas.on_mouse_press of <Canvas (PySide2) at 0x1d5ffd519d0>> for MouseEvent
```

The method actually handling the `on_mouse_press `event is `Canvas.on_mouse_press() ` which in-turn calls the `EditLineVisual.on_mouse_press()` method, and does not pass the complete `event`, just a transformation of `event.pos`:
```
    def __init__(self):
        <snip>
        self.line = EditLineVisual(pos=self.pos, color='w', width=3,
                                   antialias=True, method='gl')
    <snip>

        # self.line.print_mouse_event(event, 'Mouse press')  # <--- the moved line

        tr = self.scene.node_transform(self.line)
        pos = tr.map(event.pos)
        self.line.on_mouse_press(pos)

```

By moving the line into the `Canvas.on_mouse_press()` method it is in the correct scope to use the `event` object.

djhoese:
Ah thanks. I missed that the event *position* was passed instead of the actual event. I was confused why you didn't just change the name of the `event` being passed to the `print_mouse_event` but I see now that it is just the position. Thanks for the cleanup! Will merge when tests pass.

PR #2312, opened 2022-03-22
Fix panzoom camera center setter

alisterburt:
from @jni and myself - here is a fix for napari/napari#3723, we added a center setter to the `Rect` class and use this rather than specific logic in the panzoom camera center setter.

The logic in the previous setter was incorrectly setting the rectangle coordinates to the delta between centers, rather than shifting by that delta.

### DIFF
--- a/vispy/geometry/rect.py
+++ b/vispy/geometry/rect.py
@@ -113,6 +113,12 @@ class Rect(object):
         return (self.pos[0] + self.size[0] * 0.5,
                 self.pos[1] + self.size[1] * 0.5)
 
+    @center.setter
+    def center(self, value):
+        delta_x = value[0] - self.center[0]
+        delta_y = value[1] - self.center[1]
+        self.pos = (self.pos[0] + delta_x, self.pos[1] + delta_y)
+
     def padded(self, padding):
         """Return a new Rect padded (smaller) by padding on all sides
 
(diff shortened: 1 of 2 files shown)


almarklein (approved):

djhoese (approved):
LGTM. I'm very confused at why the camera was doing the logic it was doing. Anyway, could you do me one favor before we merge this and do a search for `.center = ` through vispy to see if there are other locations that could use this new setter?

alisterburt:
definitely a 'did this ever work? how?' one 😆 
had a quick look and can't see any, copied the list below in case any jump out at you
image

djhoese:
So most of these seem like cameras and Visuals that have their own `.center` property. Even the `.rect.center` ones are Rectangle Visuals. I think we're good here. Looks like the BaseCamera has its own center property/setter:

https://github.com/vispy/vispy/blob/acafcb07a0c9947c319e1611d72cf7045670a8e7/vispy/scene/cameras/base_camera.py#L198-L215

So I think that's how all of these various `.camera.center =` usages worked. But it looks like BaseCamera set `self._center` and then called `self.view_changed()`. So confusing, but not necessarily a bug or anything. I think this can be merged. I have a meeting and then I'll triple check everything makes sense.

almarklein:
> I'm very confused at why the camera was doing the logic it was doing.

😕 I don't understand either, and suspect I may have been the author (can't tell for sure because Git blame is confused by a refactoring). Maybe that code was hardly ever used?

Issue #2316, opened 2022-04-01
Setting camera rect on a `PanZoomCamera` not working as expected


dstansby:
In the example below I'm plotting a diagonal line going from (-1, -1) to (1, 1). If I set the camera rectangle to (-2, -2), (2, 2) I would expect the line to be fully contained and in the center of the view. Instead the bottom left corner seems to be set correctly to (-2, -2), but the top right corner of the view seems to be at (0, 0).

```python
import numpy as np
from vispy.scene import SceneCanvas, PanZoomCamera, visuals

canvas = SceneCanvas(size=(400, 400))
view = canvas.central_widget.add_view()
view.camera = PanZoomCamera()
coords = np.array([[-1, -1],
                   [1, 1]])
view.add(visuals.Line(pos=coords, color='white', width=3))

view.camera.rect = (-2, -2), (2, 2)

if __name__ == '__main__':
    canvas.show(run=True)
```
<img width="396" alt="Screenshot 2022-04-01 at 10 54 49" src="https://user-images.githubusercontent.com/6197628/161240933-8a170081-6098-464c-82ca-e8dd61f6e4ee.png">

dstansby:
It turns out I was confused about how to set the rectangle properley - https://github.com/vispy/vispy/pull/2317 is a docfix that would have prevented me doing this wrong 😄

PR #2317, opened 2022-04-01
Expand docs on panzoom rect setter

dstansby:
This adds a reference to the ways the view rectangle can be specified, clearing up potential confusion. Fixes https://github.com/vispy/vispy/issues/2316.

### DIFF
--- a/vispy/scene/cameras/panzoom.py
+++ b/vispy/scene/cameras/panzoom.py
@@ -133,7 +133,8 @@ class PanZoomCamera(BaseCamera):
     @property
     def rect(self):
         """The rectangular border of the ViewBox visible area, expressed in
-        the coordinate system of the scene.
+        the coordinate system of the scene. See `~vispy.geometry.rect.Rect`
+        for different ways this can be specified.
 
         Note that the rectangle can have negative width or height, in
         which case the corresponding dimension is flipped (this flipping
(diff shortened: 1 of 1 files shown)


vispy/scene/cameras/panzoom.py
```diff
@@ -133,7 +133,8 @@ def pan(self, *pan):
     @property
     def rect(self):
         """The rectangular border of the ViewBox visible area, expressed in
-        the coordinate system of the scene.
+        the coordinate system of the scene. See `~vispy.geometry.rect.Rect`
+        for different ways this can be specified.
```
djhoese, line 137: I think you need a `:class:` at the beginning of your Rect reference. Additionally, we should clean up this docstring to meet flake8 standards (current disabled on CI I think). What about...

```
"""Set the rectangular border of the ViewBox visible area.

This is expressed in the coordinate system of the scene. See :class:`~vispy.geometry.rect.Rect`
for different ways this can be specified.
```

Thoughts?

djhoese (changes requested):
Thanks for creating this and cleaning up out documentation! I have a small suggestion/request on reworking this to fit PEP8 conventions.

dstansby:
👍 I've updated as requested. I was lazy and did it through the GH UI though, so hopefully everything is fine? If not I can fork and do it properley locally.

djhoese:
Looks good to me. Thanks!

image

PR #2320, opened 2022-04-13
Add GL_MIN/GL_MAX constants for glBlendEquation

djhoese:
This is a replacement for #2282. This includes major refactoring of the codegen modules, but most importantly it adds a `vispy_ext.h` header as discussed in #2282 to add GL_MIN/GL_MAX. The refactoring wasn't strictly necessary, but the code is a lot easier to follow now and doesn't have at-import side effects. I have a few more cleanups I'd like to do (mostly headerparser.py), but I think this should get reviewed and tested.

CC @psobolewskiPhD @jni


almarklein:
This looks good to me. I did not dive into the details of the changes, but I'm pretty sure that if you screw something up there, you'd see it in the generated code 😄

almarklein (approved):

djhoese:
Thanks Almar. I finished some more cleanup in the header parsing. I'm not sure it is necessarily cleaner in parts, but it got rid of the global `CONSTANTS` that was being cleared outside of its own module. I'd say that's a plus. Once this passes I'll merge it given the simple in-library change it makes. And as you said, if I broke something the files would have changed more. I even tried deleting the files and recreating them after each refactor so I knew they were being regenerated. I think we're good.

kevinyamauchi:
Love it. Thanks @djhoese !

psobolewskiPhD:
Works with my napari layer blending. Thanks for picking this up. Looking at the diffs 🤯

djhoese:
Ha the diffs are mostly just refactoring/rewriting so the code will be easier to maintain/modify in the future.

PR #2327, opened 2022-04-27
Reduce memory copies in LineVisual

christiansandberg:
numpy.ascontiguousarray should not make copies unnecessarily, see https://stackoverflow.com/a/51457275 for example. [astype](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html#numpy-ndarray-astype) will make copies by default.


djhoese (approved):
Awesome! I love the question and answer nature of this PR. The todo comment says "does this result in unnecessary copies?" and you reply "yes, yes it does". Thanks so much for this.

PR #2328, opened 2022-04-29
Enable text depth testing

### LINKED ISSUE #1165: Text isn't occluded properly
When using the Text visual in a 3D scene graph, occlusion of text isn't handled properly. 
All visuals that were created after the text was added to the scene will "occlude" the text even if the text is in front.
All visuals that were created before the text cannot occlude the text and the text remains visible regardless.

It seems that the text is not properly incorporated into the 3D scene and instead it is just painted over.

I created a demo scene to illustrate the issue. Note how the numbers are always visible through the green plane, even when the numbers are behind the green plane. The red plane always hides the numbers, even when numbers are in front of the plane.

``` python
import numpy as np
import vispy
import vispy.scene

canvas = vispy.scene.SceneCanvas(keys='interactive', bgcolor=(1,1,1))
canvas.show()

view.camera = vispy.scene.TurntableCamera(up='z', fov=60)

view.camera.set_range(x=(-5, 5), y=(-5, 5), z=(-5, 5))

x_coords = np.array((-5,5))
y_coords = np.array((-5,5))
z_data = np.array([(3,3),(3,3)])

item = vispy.scene.visuals.SurfacePlot(x=x_coords, y=y_coords, z=z_data, color=(0, 1, 0, 1))
view.add(item)

for i in range(-5, 5):
    item = vispy.scene.Text(text="%i" % i, color='black', bold=True, font_size=200, pos=(0,0,i))
    view.add(item)

item = vispy.scene.visuals.SurfacePlot(x=x_coords, y=y_coords, z=-z_data, color=(1, 0, 0, 1))
view.add(item)

vispy.app.run()
```

almarklein:
Closes #1165

### DIFF
--- a/vispy/visuals/tests/test_text.py
+++ b/vispy/visuals/tests/test_text.py
@@ -82,7 +82,6 @@ def test_face_bold_italic():
 
 
 def test_text_depth_test():
-
     t = Text(depth_test=False)
     assert not t._vshare.gl_state["depth_test"]
 
--- a/vispy/visuals/tests/test_text.py
+++ b/vispy/visuals/tests/test_text.py
@@ -84,13 +84,13 @@ def test_face_bold_italic():
 def test_text_depth_test():
 
     t = Text(depth_test=False)
-    assert t._vshare.gl_state["depth_test"] == False
+    assert not t._vshare.gl_state["depth_test"]
 
     t = Text(depth_test=True)
-    assert t._vshare.gl_state["depth_test"] == True
+    assert t._vshare.gl_state["depth_test"]
 
     t = Text()  # Default is false
-    assert t._vshare.gl_state["depth_test"] == False
+    assert not t._vshare.gl_state["depth_test"]
 
 
 run_tests_if_main()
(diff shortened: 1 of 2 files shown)


djhoese:
With that added `discard` in the shader, does that have an effect on the default no depth testing case? I suppose it just stops transparent pixels that were being drawn before from being drawn?

almarklein:
>  I suppose it just stops transparent pixels that were being drawn before from being drawn?

Exactly. So in the `depth_test=False` case it should not affect the output. I checked that it does indeed not, for the example in #1165

djhoese (approved):

djhoese:
It'd be nice if you could add a test, but otherwise looks good.

almarklein:
I added a simple test to check that the depth mode is correctly set.

vispy/visuals/tests/test_text.py
```diff
@@ -81,4 +81,16 @@ def test_face_bold_italic():
         assert font1 is font4
 
 
+def test_text_depth_test():
+
```
djhoese, line 85:
```suggestion
```

djhoese (approved):
I was hoping for a more involved test (checking rendered content), but this will do.

PR #2338, opened 2022-06-05
Fix link to CHANGELOG.md in README.rst

psobolewskiPhD:
The README.rst is using .md syntax for the link to the CHANGELOG.md.
This fixes it to use .rst syntax, so the link works.


djhoese (approved):
Ah I used markdown syntax instead of restructuredtext. Thanks!

PR #2343, opened 2022-06-08
Removed future warning for network 3.0

cvanelteren:


djhoese:
There's no other way to handle this? Darn. Maybe we should just keep the warnings where they are then. Catching of warnings seems hacky.

Also, seems like you are basing your PR off of an old version of `main`.

djhoese:
Oh and I thought the warning was in the examples, not the layout code itself?

cvanelteren:
The warnings are generated by calling `nx.adjacency` not by the examples directly

djhoese:
Ok, I was hoping there was some keyword argument we could provide or a different method that could be called that would avoid this warning, but it looks like it is pretty hardcoded to always appear until 3.0:

https://github.com/networkx/networkx/blob/5dfd57af2a141a013ae3753e160180b82bec9469/networkx/linalg/graphmatrix.py#L166-L173

I vote for just ignoring the warning for now to avoid the warning catching. It is kind of an odd warning in that it is always going to appear.

djhoese:
Yep, thanks for putting it together regardless.

cvanelteren:
Don't mention it, happy to help. Just ping me again when it comes up ;-)

Issue #2279, opened 2021-12-19
About the viewpoint of turntable camera


nnop:
I'm confused by the viewpoint of the turnable camera.

My test code (vispy 0.9.4):
```python
canvas = scene.SceneCanvas(keys='interactive', show=True)
view = canvas.central_widget.add_view()
axis = scene.visuals.XYZAxis(parent=view.scene)

view.camera = 'turntable'
camera = view.camera
camera.elevation = 0
camera.azimuth = 0
camera.center = (0, 0, 0)
camera.scale_factor = 5
```
And result:
![image](https://user-images.githubusercontent.com/23091984/146663074-fb611415-1d36-4d8a-b38b-ad86dfa4f32e.png)

As the [doc](https://vispy.org/api/vispy.scene.cameras.turntable.html) says:
> Azimuth: Zero degrees places the camera on the positive x-axis, pointing in the negative x direction.

But the result shows, the camera is on the **-y** direction pointing in the **+y** direction.

Is the doc wrong?

almarklein:
Either that, or the XYZAxis is wrong :)

https://github.com/vispy/vispy/blob/ef982591e223fff09d91d8c2697489c7193a85aa/vispy/visuals/xyz_axis.py#L7-L10

djhoese:
Strange that the property for `azimuth` uses different language:

```
The angle of the camera in degrees around the y axis. An angle of 0 places the camera within the (y, z) plane.
```

I read that as being the same thing as being on/along the X axis.

@nnop Could you maybe add something to an `on_draw` or a mouse handler to have it prints out the `azimuth` and other properties to see what the camera thinks these values are as you move to different positions?

harripj:
+1 for this. In my testing adding 45 deg. azimuthal rotation also appears to rotate around the z-axis (blue), not y (green):
```python
import sys

from vispy import scene
from vispy.scene import XYZAxis


def main():
    canvas = scene.SceneCanvas(show=True)
    view = canvas.central_widget.add_view()

    camera = scene.cameras.TurntableCamera(elevation=0, azimuth=45, roll=0, distance=5)
    view.camera = camera

    view.add(XYZAxis())
    image = canvas.render()

    if sys.flags.interactive == 0:
        canvas.app.run()


if __name__ == "__main__":
    main()
```

<img width="912" alt="image" src="https://user-images.githubusercontent.com/16853829/177645844-74d0e85c-db7a-4baa-a133-110e0e014270.png">

PR #2352, opened 2022-07-06
Implement turntable camera `roll` programmatically and clarify transformation docstrings

harripj:
As discussed in #2279 the docstrings of `elevation`, `azimuth`, and `roll` appear to be misleading for `TurntableCamera`. This PR aims to address this and also implements `roll` functionality. The `TurntableCamera._get_rotation_tr()` function has been updated so all three transformations represent consistent right-hand rotations around the respective axes.

This PR may be missing some nuance as to the way camera rotations are defined in VisPy, for example rotations of the camera around a static scene or vice-versa, or as to why `roll` was not implemented in the first place. regarding the former point,I have written this PR in the first sense.

Any help or guidance appreciated! Thanks!


djhoese (changes requested):
Looks good. Thanks for putting this together. I think this makes sense. I'm not really sure why the documentation was so wrong. I have a couple requests before merging this:

1. Could you remove your test script.
2. Could you add a proper test to the unit tests. Probably here: https://github.com/vispy/vispy/tree/main/vispy/scene/cameras/tests
3. Could you restructure the docstrings for the properties to be more PEP8 compliant where the first line is a single sentence with an imperative verb, then a blank line, then any additional description. For example:

```python
def my_method(self, x, y, z):
    """Get some value by doing stuff.

    Further description.
    """
    ... code ...
```

vispy/scene/cameras/tests/test_cameras.py
```diff
@@ -0,0 +1,27 @@
...
+        [0, 0, 0, np.eye(4)],
+        [90, 0, 0, [[1, 0, 0, 0], [0, 0, 1, 0], [0, -1, 0, 0], [0, 0, 0, 1]]],
+        [0, 90, 0, [[0, 1, 0, 0], [-1, 0, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]],
+        [0, 0, 90, [[0, 0, -1, 0], [0, 1, 0, 0], [1, 0, 0, 0], [0, 0, 0, 1]]],
+    ],
+)
+def test_turntable_camera_transform(elevation, azimuth, roll, expected):
+    camera = TurntableCamera(elevation=elevation, azimuth=azimuth, roll=roll)
+    matrix = camera._get_rotation_tr()
+    assert np.allclose(matrix, expected)
```
djhoese, line 24: You should be able to do:

```suggestion
    np.testing.assert_allclose(matrix, expected)
```

djhoese (changes requested):
Beautiful! Thank you. One small change request for better test output on failure.

djhoese (approved):
:tada: 

@brisvag or @almarklein do you want to give one more review and then we can merge this.

brisvag (approved):
Looks good to me too, assuming tests pass!

djhoese:
Looks like two tests failed in at least one of the environments:

```
 =================================== FAILURES ===================================
___________________________ test_perspective_render ____________________________
vispy/scene/cameras/tests/test_perspective.py:53: in test_perspective_render
    max_px_diff=200)
vispy/testing/image_tester.py:138: in assert_image_approved
vispy/testing/image_tester.py:201: in assert_image_match
    assert mask.sum() <= px_count
E   AssertionError
----------------------------- Captured stdout call -----------------------------
Image comparison failed. Test result: (200, 120, 4) uint8   Expected result: (200, 120, 4) uint8
Uploaded to: 
http://data.vispy.org/data/scene/cameras/3cf5aa00cbb7b70110e2afb592671d73472eca89/perspective_test.png
_____________________________ test_rotation_angle ______________________________
vispy/visuals/tests/test_axis.py:53: in test_rotation_angle
    assert_allclose(axis2._rotation_angle, -30)
E   AssertionError: 
E   Not equal to tolerance rtol=1e-07, atol=0
E   
E   Mismatched elements: 1 / 1 (100%)
E   Max absolute difference: 59.99999836
E   Max relative difference: 1.99999995
E    x: array(29.999998)
E    y: array(-30)
```

The last one I suppose could just be a precision error brought on by the additional multiplications with handling `roll`. Allowing more variance in the atol is probably OK. The other one I'm not so sure what's going on. Best case the test is actually incorrect and `roll` was being tested and wasn't actually doing anything and now it is. :man_shrugging:

djhoese:
Ah the tests upload images of the result so it looks like things are being shifted:

image

djhoese:
I have no experience with the 3D cameras so I'm not sure I should make a decision. @almarklein @brisvag ?

harripj:
> Is there a way we can somehow deprecate it and let people "opt in"?

Would a `UserWarning` informing people of the future change on instantiating `TurntableCamera` and a `use_updated_transforms` flag (default is `False`) serve this purpose?

brisvag:
So for one release we do that, and for the next one we have to deprecate `use_updated_transforms` and then remove it?

harripj:
I think that would be a safe way to implement the change and would give users warning about it. Happy to implement this if you like the idea!

djhoese:
I'm curious if @rougier or @almarklein know why the camera's `elevation`, `azimuth`, and `roll` aren't all rotating following the same rules. Maybe there is some other common way of thinking about it where this makes sense?

almarklein:
I'm not sure. If this is only for how a mouse move rotates the camera, than one might view it more as a UX problem than inherent mathematical way of working, and a change would be fine, I suppose?

brisvag:
@harripj do you have a quick code example to test out exactly what the difference entails in UX terms?

harripj:
You can test out the UX differences by running this script (based on the failing test @djhoese mentioned above: https://github.com/vispy/vispy/pull/2352#issuecomment-1177892006):
```python
from types import MethodType

from vispy import io, scene
from vispy.util import transforms


def test_perspective_render():
    canvas = scene.SceneCanvas(size=(120, 200), show=True)

    imdata = io.load_crate().astype('float32') / 255

    for i in range(2):
        v = grid.add_view(row=i, col=0, border_color='white')
        v.camera = 'turntable'
        v.camera.fov = 50
        v.camera.distance = 30
        v.camera.elevation = 30
        v.camera.roll = 0
        v.camera.azimuth = 40

        if not i:
            def _old_rotation_tr(self):
                """Return a rotation matrix based on camera parameters"""
                up, forward, right = self._get_dim_vectors()
                matrix = (
                    transforms.rotate(self.elevation, -right)
                    .dot(transforms.rotate(self.azimuth, up))
                    .dot(transforms.rotate(self.roll, forward))
                )
                return matrix

            v.camera._get_rotation_tr = MethodType(_old_rotation_tr, v.camera)

        else:
            def _new_rotation_tr(self):
                """Return a rotation matrix based on camera parameters"""
                up, forward, right = self._get_dim_vectors()
                matrix = (
                    transforms.rotate(self.elevation, right)
                    .dot(transforms.rotate(self.azimuth, up))
                    .dot(transforms.rotate(self.roll, forward))
                )
                
                return matrix

            v.camera._get_rotation_tr = MethodType(_new_rotation_tr, v.camera)


        image = scene.visuals.Image(imdata, grid=(4, 4))
        image.transform = scene.STTransform(translate=(-12.8, -12.8),
                                            scale=(0.1, 0.1))
        v.add(image)

    return canvas

if __name__ == "__main__":
    canvas = test_perspective_render()
    canvas.app.run()
```
image


After running this I see why the transform matrix is constructed as it is (with `-elevation`): the `elevation` and `azimuth` interaction both give the impression of rotating the object rather than the camera. In which case these interactions are consistent in this sense and should be left as they were, if everyone agrees?

Further, the roll `interaction` is not fully implemented in the UX in this PR, so perhaps the title is misleading, but this PR does allow the `roll` interaction to be implemented programmatically. I think that this UX is suitable for the purpose of `TurntableCamera` but allows the user further flexibility if they want to change `roll` manually. It would be great to get others' opinions on this!

brisvag:
Sounds great! Thanks for the example, very clear. Indeed it seems we want to keep the old implementation (and maybe add a comment or docstring to explain why and point to this PR for future reference).

brisvag:
@harripj do you have time to push over the line? The remaining [test failure](https://github.com/vispy/vispy/runs/7236644850?check_suite_focus=true#step:11:464) should be fixable as you said by increasing the tolerance a bit!

harripj:
@brisvag sorry for the delay, I have been busy lately. I will take another look at this over the weekend!

harripj:
@brisvag I have had another look over this and updated the tests. Effectively this PR clarifies the `TurntableCamera` docstrings to clarify the interaction and allows programmatic adjustment of the `roll` parameter (this parameter is not implemented in the UI, which I think makes sense).

The UI interaction is such that `azimuth` always rotated around the scene `Z` axis, whereas `elevation` rotates not around the scene `X` axis, but around the 'effective' scene `X` (horizontal) axis after the `azimuth` transformation has been performed. I have tried to describe this in the docstring, but realise that a better description may be possible using VisPy reference frames if they are documented somewhere?

vispy/scene/cameras/turntable.py
```diff
@@ -23,14 +23,16 @@ class TurntableCamera(Base3DRotationCamera):
...
-        the center point.
+        Elevation angle in degrees. The elevation angle represents a
+        rotation of the camera around the current scene x-axis. The
+        camera looks along the +y direction when the elevation is 0 degrees.
     azimuth : float
-        Azimuth angle in degrees. Zero degrees places the camera on the
-        positive x-axis, pointing in the negative x direction.
+        Azimuth angle in degrees. The azimuth angle represents a
+        rotation of the camera around the scene z-axis according to the
+        right-hand screw rule.
```
djhoese, line 32: Can/should you mention what the camera "sees" when azimuth is at 0 degrees as it did in the old docstring?

vispy/scene/cameras/turntable.py
```diff
@@ -97,9 +94,7 @@ def azimuth(self, azim):
 
     @property
     def roll(self):
-        """The angle of the camera in degrees around the z axis. An angle of
-        0 places puts the camera upright.
-        """
+        """Get the camera roll angle in degrees."""
```
djhoese, line 97: If the user only sees the docstring for the `getter` part of the property, should this include more information about what setting it to 0 means as it did before? If that information made it more confusing then ignore this.

djhoese (approved):
Looks good to me from the small amount I understand of the conversation here. I had two comments about the docstrings.

vispy/scene/cameras/turntable.py
harripj: The previous information was confusing, but this information can be added for the other two angle getters.

harripj:
@djhoese thanks for reviewing, I have updated the docstrings with your suggestions. Hopefully the information comes across more clearly now.

djhoese:
This looks good to me. Let's merge it. The build website failing is a known issue and I doubt your PR is causing any issues. Thanks again for this @harripj!

PR #2354, opened 2022-07-12
Debug getting DPI from CI environment

djhoese:


.github/workflows/main.yml
```diff
@@ -39,258 +34,6 @@ jobs:
...
-
-  coveralls:
-    needs: [build_0]
-    runs-on: ubuntu-latest
-    steps:
-      - name: Coveralls Finished
-        uses: AndreMiras/coveralls-python-action@develop
-        with:
-          parallel-finished: true
+        xdpyinfo
```
djhoese, line 39:
```suggestion
        export DISPLAY=:99.0
        xrandr
```

djhoese:
Adding `x11-xserver-utils` seems to install `xrandr` just fine, but the resulting execution of the command shows a screen size of 0mm x 0mm which will lead to a warning and a default DPI of 96 on CI. Let's try adding the package for xdpyinfo...

Issue #2369, opened 2022-09-02
vispy/visual/tube.py does not support python 3


tralfaz:
It looks like this is specific to python 3.10 and later.

At line 68:

 if not isinstance(radius, collections.Iterable):

This only works for python 2.6 or 2.7.

For python 3 the line would be

 if not isinstance(radius, collections.abc.Iterable):

Or to work with both python 2 and python 3

import sys
if sys.version_info.major > 2:
   from collections.abc import Iterable
else:
   from collections import Iterable

Then change line 68 to:

 if not isinstance(radius, Iterable):

FYI

Michael

djhoese:
This is...very strange.

This example uses the TubeVisual:

https://vispy.org/gallery/scene/marker_spheres.html#sphx-glr-gallery-scene-marker-spheres-py

and was built in a Python 3.9 environment:

https://github.com/vispy/vispy/runs/7997191272?check_suite_focus=true

And succeeded (as it generated the image). :confused:

djhoese:
Ah, the move happened a long time ago (Python 3.3) but wasn't actually removed until Python 3.10. We don't need to support Python 2 or even old versions of Python 3. @tralfaz how would you feel about making a pull request?

tralfaz:
No problem doing that, but I think it costs next to nothing to support aa wide a range of python versions as possible.  Though a bit ugly.

Either by using sys.version_info check or try import except clause.  Python 2 is rapidly becoming a thing of the past, but its still hard for some to move on.

—mdm

djhoese:
We don't support Python <3.7:

https://github.com/vispy/vispy/blob/main/setup.py#L95

And there are way too many parts of VisPy that would have to be changed to support anything older, but more importantly the maintenance burden is not worth it to support users that want new features without upgrading all parts of their environment.

tralfaz:
OK that makes it a simple one liner, but I’ll grep through the code for other instances.

Ran into issue in my VisPy sandbox https://github.com/tralfaz/GUTS <https://github.com/tralfaz/GUTS> :-)

—mdm

djhoese:
Looks like we forgot to close this when #2370 was merged. This should be fixed now. Thanks again.

PR #2370, opened 2022-09-02
Update tube.py to support Python 3.10 and higher.

tralfaz:
Fix reference to Iterable abstract base class (abc).  It has been moved to collections.abc and removed from collections in Python 3.10.

### DIFF
--- a/vispy/visuals/tube.py
+++ b/vispy/visuals/tube.py
@@ -65,7 +65,7 @@ class TubeVisual(MeshVisual):
         segments = len(points) - 1
 
         # if single radius, convert to list of radii
-        if not isinstance(radius, collections.Iterable):
+        if not isinstance(radius, collections.abc.Iterable):
             radius = [radius] * len(points)
         elif len(radius) != len(points):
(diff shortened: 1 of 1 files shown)


djhoese:
Closes #2369

Thanks for fixing this! I'll merge when tests pass.

djhoese:
The website error is unrelated to your changes. I'll fix it after the US holiday weekend

tralfaz:
Have a little fun away from the keyboard.  Thanks for the help.

djhoese:
So it looks like the website building is failing due to a new version of the pydata-sphinx-theme. I've filed a bug report here:

https://github.com/pydata/pydata-sphinx-theme/issues/913

If it isn't an easy fix (for us or them) then I'll force the version in CI and merge this. Actually...I'll probably just merge this tomorrow since this is clearly not the problem.

tralfaz:
Its always frustrating when your tools are more work than the work.

tralfaz:
Cool.  Hope they get the sphinx stuff fixed.

PR #2287, opened 2022-01-28
I saw there was no option to plot bar graphs

Bliss3d:
So I wrote quickly wrote it. Im not sure Im in compliance with vispy project standarts, probably not. So let me know if there are some changes needed.
Best Regards

Issue #2372, opened 2022-09-09
HistogramVisual orientation backwards


djhoese:
See https://stackoverflow.com/questions/73663631/plot-1d-histogram-data-using-vispy where I first noticed this. After some debugging it looks like specifying an orientation of `h` makes the bars go from bottom to top (bins axis is the bottom or 'x' axis), and `v` makes the bars horizontal (bins axis on the left or 'y' axis).

Am I seeing this wrong? How did this go unnoticed for so long?

PR #2394, opened 2022-10-02
New Visual: Bar

Bliss3d:
This PR replaces #2287, this contains fixes for Bar.meshdata.get_vertices() which error out in the previous PR and is probebly a better solution and contains an example

As ever your input is very welcome :)

https://user-images.githubusercontent.com/13086508/193459506-af7b020a-d512-45c7-b82a-5dc3bc0b34e3.mp4

### DIFF
--- a/examples/scene/bar_plot.py
+++ b/examples/scene/bar_plot.py
@@ -1,6 +1,5 @@
 import numpy as np
 from vispy import scene, app
-from vispy.color import ColorArray
 
 if __name__ == "__main__":
 
@@ -24,8 +23,8 @@ if __name__ == "__main__":
 
     grid_lines1 = scene.visuals.GridLines(parent=vb1.scene)
 
-    bar1 = scene.Bar(height=np.arange(0, 250, 0.5), bottom=np.arange(0,125, 0.25),width=0.8, shift=0.5, color=(0.25, 0.8, 0.),
-                           parent=vb1.scene)
+    bar1 = scene.Bar(height=np.arange(0, 250, 0.5), bottom=np.arange(0, 125, 0.25), width=0.8, shift=0.5, 
+                     color=(0.25, 0.8, 0.), parent=vb1.scene)
 
     vb1.camera.set_range()
 
--- a/examples/scene/bar_plot.py
+++ b/examples/scene/bar_plot.py
@@ -1,0 +1,33 @@
+import numpy as np
+from vispy import scene, app
+from vispy.color import ColorArray
+
+if __name__ == "__main__":
+
+    canvas = scene.SceneCanvas(keys='interactive', vsync=False)
+    canvas.size = 800, 600
+    canvas.show()
+
+    grid = canvas.central_widget.add_grid()
+    grid.padding = 10
+
+    vb1 = grid.add_view(row=0, col=1, camera='panzoom')
+
+    x_axis1 = scene.AxisWidget(orientation='bottom')
+    x_axis1.stretch = (1, 0.1)
+    grid.add_widget(x_axis1, row=1, col=1)
+    x_axis1.link_view(vb1)
+    y_axis1 = scene.AxisWidget(orientation='left')
+    y_axis1.stretch = (0.1, 1)
+    grid.add_widget(y_axis1, row=0, col=0)
+    y_axis1.link_view(vb1)
+
+    grid_lines1 = scene.visuals.GridLines(parent=vb1.scene)
+
+    bar1 = scene.Bar(height=np.arange(0, 250, 0.5), bottom=np.arange(0,125, 0.25),width=0.8, shift=0.5, color=(0.25, 0.8, 0.),
+                           parent=vb1.scene)
+
+    vb1.camera.set_range()
+
+    canvas.measure_fps()
+    app.run()
--- a/vispy/visuals/bar.py
+++ b/vispy/visuals/bar.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+
+import numpy as np
+from .mesh import MeshVisual
+
+
+class BarVisual(MeshVisual):
+    """Visual that calculates and displays a bar graph
+
+    Parameters
+    ----------
+    data : array-like
+        1st column is the height, optional 2nd column is bottom
+    height : array-like
+        Height of the bars
+    bottom : array-like
+        Bottom of the bars
+    width : int | float
+        Width of all bars
+    shift : int | float
+        Shift of all bars along the x-axis
+    color : instance of Color
+        Color of the bar.
+    """
+
+    def __init__(self, height, bottom=None, width=0.8, shift=0.0, color='w'):
+        if bottom is None:
+            bottom = np.zeros(height.shape[0])
+        rr, tris = calc_vertices(height, bottom, width, shift)
+
+        MeshVisual.__init__(self, rr, tris, color=color)
+
+    def update_data(self, height, bottom=None, width=0.8, shift=0.0, color='w'):
+        if bottom is None:
+            bottom = np.zeros(height.shape[0])
+        rr, tris = calc_vertices(height, bottom, width, shift=shift)
+
+        MeshVisual.set_data(self, rr, tris, color=color)
+
+
+def calc_vertices(height, bottom, width, shift):
+
+    y_position = np.zeros((height.shape[0], 2))
+    y_position[:, 0] = height
+    y_position[:, 1] = bottom
+
+    y_position = y_position.flatten().repeat(2)
+
+    stack_n_x2 = np.arange(0, height.shape[0], dtype=int) 
+    stack_n_x2 = np.expand_dims(stack_n_x2, axis=1)
+    stack_n_x2 = np.repeat(stack_n_x2, 1, axis=1)
+    stack_n_x2 = np.repeat(stack_n_x2, 2, axis=0)
+    
+    vertices = np.array([
+        [-width/2 + shift, 1, 0],
+        [width/2 + shift, 1, 0],
+        [width/2 + shift, 0, 0],
+        [-width/2 + shift, 0, 0]])
+
+    vertices = np.tile(vertices, (height.shape[0], 1))
+    vertices[:, 0] = vertices[:, 0] + stack_n_x2[:, 0].repeat(2)
+    vertices[:, 1] = y_position
+
+    base_faces = np.array([[0, 1, 2], [0, 2, 3]])
+
+    base_faces = np.tile(base_faces, (height.shape[0], 1))
+
+    faces = stack_n_x2*4 + base_faces
+
+    return vertices, faces
--- a/vispy/scene/visuals.py
+++ b/vispy/scene/visuals.py
@@ -232,6 +232,7 @@ def generate_docstring(subclass, clsname):
 
 Arrow = create_visual_node(visuals.ArrowVisual)
 Axis = create_visual_node(visuals.AxisVisual)
+Bar = create_visual_node(visuals.BarVisual)
 Box = create_visual_node(visuals.BoxVisual)
 ColorBar = create_visual_node(visuals.ColorBarVisual)
 Compound = create_visual_node(visuals.CompoundVisual)
--- a/vispy/visuals/__init__.py
+++ b/vispy/visuals/__init__.py
@@ -12,6 +12,7 @@ defined in vispy.scene.
 """
 
 from .axis import AxisVisual  # noqa
+from .bar import BarVisual  # noqa
 from .box import BoxVisual  # noqa
 from .cube import CubeVisual  # noqa
 from .ellipse import EllipseVisual  # noqa


djhoese:
Wow, this seems very simple for such a nice feature. I'm wondering how you think this compares to some of the logic in the existing HistogramVisual:

https://github.com/vispy/vispy/blob/main/vispy/visuals/histogram.py

Could/should this new Visual be used by the HistogramVisual? I notice the HistogramVisual allows for orientation (even though it is currently backwards from my experiments, see #2372). How hard would it be to add that here?

djhoese:
Also, don't worry about the `(3.9, full, standard, true)` test that is failing. That is the new PyQt6 causing seg faults.

Otherwise, I forgot to mention, would it be possible to add a unit test for this new visual? Maybe something simple that shows that it is able to draw. You could even check the colors at certain points in the rendered result. For example, if the first bar is should fill the whole SceneCanvas then make sure that a color near the top of the image is the right color. And if the last bar is 50% of the canvas then make sure the colors are correct at the top and bottom half of the image.

Bliss3d:
> Wow, this seems very simple for such a nice feature. I'm wondering how you think this compares to some of the logic in the existing HistogramVisual:
> 
> https://github.com/vispy/vispy/blob/main/vispy/visuals/histogram.py
> 
> Could/should this new Visual be used by the HistogramVisual? I notice the HistogramVisual allows for orientation (even though it is currently backwards from my experiments, see #2372). How hard would it be to add that here?

As I started with vispy ploting I found it very strage that there where no bar visuals even though there is a histogram visual, seems kind assbackwards, no offense. However it familiarized me with vispy ploting and I did use it to guide me to create #2287 - now I'm much more proficient with vispy ploting, I'm personally more interested in the Candle visual this is just a by-product. Short anwser is yes it would probebly make sense to use bars in the histogram. However I dont use histograms so I'm hardly an authority on the matter - people who actually use the histogram component should make the decision. If I remember correctly in a vispy histogram there is no option for spaces between bins and its a single mesh, which in my opinion looks kind of outdated. It might be useful to have a single mesh for a histogram as I sad I dont use histograms. 

Regaarding vertical bars probebly just switch width and height

And for unit tests i've never done them, so I have no experience

examples/scene/bar_plot.py
```diff
@@ -1,3 +1,14 @@
+# -*- coding: utf-8 -*-
+# vispy: gallery 2
+# -----------------------------------------------------------------------------
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+"""
+Bar plot with Axis 
+=====================
```
djhoese, line 9: The underline must be the same number of characters (width) as the title.

Bliss3d: Okay, fixed

Bliss3d:
As per your request vertical bar ploting for a possible usage in histogram is now implemented, easy as pie

python-2022-10-04-00-52-00-39

vispy/visuals/bar.py
```diff
@@ -0,0 +1,86 @@
...
+            bottom = np.zeros(height.shape[0])
+        
+        rr, tris = calc_vertices(height, bottom, width, shift, orientation)
+
+        MeshVisual.__init__(self, rr, tris, color=color)
+
+    def update_data(self, height, bottom=None, width=0.8, shift=0.0, color='w'):
+        if bottom is None:
+            bottom = np.zeros(height.shape[0])
+        rr, tris = calc_vertices(height, bottom, width, shift=shift)
```
brisvag, line 43: Missing orientation here

vispy/visuals/bar.py
```diff
@@ -0,0 +1,86 @@
...
+        Width of all bars
+    shift : int | float
+        Shift of all bars along the x-axis
+    color : instance of Color
+        Color of the bar.
+    orientation : {'h', 'v'}
+        Orientation of the histogram - 'h' is default
+    """
+
+    def __init__(self, height, bottom=None, width=0.8, shift=0.0, color='w', orientation='h'):
```
brisvag, line 32: Would be awesome to be able to pass a per-bar color as well!

vispy/visuals/bar.py
```diff
@@ -0,0 +1,86 @@
...
+    
+    y_position[:, 0] = height
+    y_position[:, 1] = bottom
+
+    y_position = y_position.flatten().repeat(2)
+
+    stack_n_x2 = np.arange(0, height.shape[0], dtype=int) 
+    stack_n_x2 = np.expand_dims(stack_n_x2, axis=1)
+    stack_n_x2 = np.repeat(stack_n_x2, 1, axis=1)
+    stack_n_x2 = np.repeat(stack_n_x2, 2, axis=0)
```
brisvag, line 60: If I understand correctly, this was the goal? Maybe this way is a bit more intuitive.

```suggestion
    stack_n_x2 = np.arange(height.shape[0]).repeat(4).reshape(-1, 1)
```

By setting `repeat(4)` instead of `2` we don't have to repeat again later, which is both faster for memory allocation and clearer in the code.

vispy/visuals/bar.py
```diff
@@ -0,0 +1,86 @@
...
+
+    vertices = np.array([
+        [-width/2 + shift, 1, 0],
+        [width/2 + shift, 1, 0],
+        [width/2 + shift, 0, 0],
+        [-width/2 + shift, 0, 0]])
```
brisvag, line 66: What's the point of the `0` at index 2? I'm having a hard time following the flow.

vispy/visuals/bar.py
```diff
@@ -0,0 +1,86 @@
...
+        vertices[:, 1] = vertices[:, 0] + stack_n_x2[:, 0].repeat(2)
+        vertices[:, 0] = y_position
+    else:
+        raise ValueError('orientation can only be v (vertical) or h (horizontal). You specified: ' + str(orientation))
+
+    base_faces = np.array([[0, 1, 2], [0, 2, 3]])
+
+    base_faces = np.tile(base_faces, (height.shape[0], 1))
+
+    faces = stack_n_x2*4 + base_faces
```
brisvag, line 84: ok, maybe this use of `stack_n_x2` counters my proposal of using `repeat(4)` above, if I'm following.

brisvag (commented):
Looks really good in general! I proposed some small change for clarity's sake, but as far as I can tell it makes sense. I really thionk we need some comments about *what* each part of `calc_vertices` is doing, othwerwise we won't know how to maintain it a month from now, since it's quite cryptic.

Other than that, I think we could merge this, and think about using it as our histogram visual at a later time!

vispy/visuals/bar.py
Bliss3d: I think I implemented something kinda like that in #2388 and those PR's are very samey.
#2388 has also more comments trying to explain whats going on

vispy/visuals/bar.py
Bliss3d: Obviously yes, it was late yesterday

vispy/visuals/bar.py
Bliss3d: As you saw its used later, however i'd say 
`stack_n_x2 = np.arange(height.shape[0]).repeat(2).reshape(-1, 1)`
would be better
I did a quick test: 

`%timeit old(np.arange(0, 250, 0.5))
%timeit new(np.arange(0, 250, 0.5))
`
12.8 µs ± 1.18 µs per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
7.23 µs ± 689 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

So its about 50% faster. 

if I inclue the second repeat(2) from line 71/75 its twice as fast

25.2 µs ± 5.47 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each) # original commit
12.4 µs ± 1.78 µs per loop (mean ± std. dev. of 7 runs, 100,000 loops each) # repeat(2).repeat(2)
7.69 µs ± 784 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each) # repeat(4)

vispy/visuals/bar.py
Bliss3d: Well I thought that vertices would have to be x,y,z and the resulting vertices for the mesh should be in the shape (n,3), since its a 2d plot the z axis is zero. I'm not married to it if you want to change it it would be fine with me

brisvag: Ok, I see why I was confused: this resulted in a deprecation warning from numpy due to a staggered sequence in my testing, because I was passing an array to `width` rather than a single value... My bad :P

brisvag: But yeah, I think you could remove the zero there and it would be fine, meshes work well with 2D data and we do it in several other places as well.

vispy/visuals/bar.py
brisvag: Yeah no need to do this here, I just pointed out it would be cool to add in the future :)

vispy/visuals/bar.py
```diff
@@ -0,0 +1,86 @@
...
+    bottom : array-like
+        Bottom of the bars
+    width : int | float
+        Width of all bars
+    shift : int | float
+        Shift of all bars along the x-axis
+    color : instance of Color
+        Color of the bar.
+    orientation : {'h', 'v'}
+        Orientation of the histogram - 'h' is default
```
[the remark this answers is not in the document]
Bliss3d: What a philosophical question, on the one hand is it called horizontal if the bars got vertical up and are alined along the horizontal axis or is it called vertical.
I was thinking along which axis are the bars alined as opposed to the direction of the bars
I'm have no clue, I''ll use the vispy philosophical bend so its inline with other visuals

vispy/visuals/bar.py
Bliss3d: Well I just looked into it, it should be not to problematic

Bliss3d:
Added individual colors per bar 

vispy_bar

vispy/visuals/bar.py
```diff
@@ -26,23 +26,39 @@ class BarVisual(MeshVisual):
...
+        if height.shape != bottom.shape:
+            raise ValueError("Height and Bottom must be same shape: Height: " + height.shape + " Bottom: " + bottom.shape)
         
         rr, tris = calc_vertices(height, bottom, width, shift, orientation)
 
-        MeshVisual.__init__(self, rr, tris, color=color)
+        if color_array is not None and color_array.shape[0] == height.shape[0] and color_array.shape[1] == 3:
+            MeshVisual.__init__(self, rr, tris, face_colors=np.repeat(color_array, 2, axis=0))
+        else:
+            MeshVisual.__init__(self, rr, tris, color=color)
```
brisvag, line 46: Great! To get rid of this if clause (and get extra perks like string inputs and aplha values) I would use the `ColorArray` just like you did in the `Candle` PR.

brisvag (changes requested):
Still small suggestion left, but otherwise This is looking great for me! @djhoese anything to add?

vispy/visuals/bar.py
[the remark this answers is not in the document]
djhoese: This is why I mentioned https://github.com/vispy/vispy/issues/2372. I assumed we were talking about the direction a single bar "grows". So if the bin dimension (width) is on the x-axis, then the bars grow on the y-axis and the plot is "vertical".

vispy/visuals/bar.py
brisvag: @Bliss3d you think you can get this to work with a single `color` argument? I think it would be more intuitive, kinda like `Markers` does in these few lines:

https://github.com/vispy/vispy/blob/f13d7da790030ef86f2969d04f9462d7d1d35e11/vispy/visuals/markers.py#L607-L613

As you can see, you can use the `ColorArray` object from `vispy.color`, which automatically takes care of all the weird cases (a string like `'w'`, a `(n, 3)` rgb array, or an `rgba`, and so on). You can then simply check if it's a single color, and if so pass it to `color`, otherwise pass it to `face_colors`.

Bliss3d: I'm guessing something like this

vispy/visuals/bar.py
```diff
@@ -33,23 +35,33 @@ class BarVisual(MeshVisual):
...
+                        color_array = np.repeat(color, 2, axis=0)
+                        color = None
+                else:
+                    raise ValueError('If you pass an numpy array as color it needs to be rgb i.e. 3 columns')
+        elif isinstance(color, ColorArray):
+            pass
+        elif isinstance(color, str):
+            pass
+        else:
+            raise ValueError("Color has to be either a vispy ColorArray, numpy array or a string for example: 'w'")
```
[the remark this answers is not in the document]
Bliss3d: I'm not sure how this is supposed to work, color array is not defiened in that context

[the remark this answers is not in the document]
brisvag: sorry, wrote that in a rush. Fixed:
```py
        color = ColorArray(color).rgba
        if len(color):
        	  color = color_array[0]
            color_array = None
        else:
            color = None
			  color_array = color
```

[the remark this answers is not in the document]
Bliss3d: I modified it a bit, but I hope it works for you

djhoese:
Yes @brisvag. Here's what I said early on that I think would be good for tests:

> Otherwise, I forgot to mention, would it be possible to add a unit test for this new visual? Maybe something simple that shows that it is able to draw. You could even check the colors at certain points in the rendered result. For example, if the first bar is should fill the whole SceneCanvas then make sure that a color near the top of the image is the right color. And if the last bar is 50% of the canvas then make sure the colors are correct at the top and bottom half of the image.

Bliss3d:
Well I have pretty much no clue how to run or create those test... :(

djhoese:
> Well I have pretty much no clue how to run or create those test... :(

Well there's no time like the present! VisPy has some basic documentation for writing tests (really glad past-Dave wrote this):

https://vispy.org/dev_guide/contributor_guide.html#adding-tests

In your case, you're writing tests for a Visual object so a new test module should be added here:

https://github.com/vispy/vispy/tree/main/vispy/visuals/tests

You'll find other tests in that directory to use as a starting point, but keep in mind that some of them do things like `assert_image_approved` (the histogram tests do this) where they compare against a known good image. The conversation @brisvag and I are having in the above comments is saying we *don't* want to use this. Creating, uploading, and maintaining these "known truth" images is annoying. Since your Visual is relatively simple as far as what it is drawing, it should be possible to get the result of the rendered canvas and then check a few specific values. Here's a portion of the `test_mesh.py` module where this test creates a `Mesh` object and then checks how many unique values there are in the Red channel (the result of `c.render()` is RGBA). This is very basic but should give you an idea that this rendered result is just a numpy array so if we know what colors to expect in certain parts of the result then we can just check those individual pixels as a "sanity check".

https://github.com/vispy/vispy/blob/859fcaedcc54ebe1ecfd21348cefa37a1cf08ca6/vispy/visuals/tests/test_mesh.py#L194-L213

Let me know if any of this isn't making sense.

Bliss3d:
Will do, also I'll transfer the feedback to #2388 since there is quite a bit of overlap, but at the moment I need a change of pace from vertices and I still got the date axis labels in the works - and it is quite good and I need to flush it out before I dump all that time and date stuff out of my brain. I had this PR and #2388 lying around for mouths...

Issue #2395, opened 2022-10-02
wx backend timer expects integer instead of float


Bliss3d:
wxpython version 4.2.0
vispy version 0.11.0

Traceback (most recent call last):
  File "c:\Users\Joel I. Oelsner\dev\vispy\date_axis_wip\vispy\examples\scene\axis_date_plot.py", line 93, in <module>
    timer.start(0)
  File "C:\Users\Joel I. Oelsner\dev\vispy\date_axis_wip\venv\lib\site-packages\vispy\app\timer.py", line 118, in start
    self._backend._vispy_start(self.interval)
  File "C:\Users\Joel I. Oelsner\dev\vispy\date_axis_wip\venv\lib\site-packages\vispy\app\backends\_wx.py", line 469, in _vispy_start
    self._timer.Start(interval * 1000., False)
TypeError: Timer.Start(): argument 1 has unexpected type 'float'

If you try to replicate pip install attrdict3 because its missing in wxpython 4.2.0 requirments manifest even though its needed.

Its an easy fix just cast the float to an int:

`class TimerBackend(BaseTimerBackend):

    def __init__(self, vispy_timer):
        BaseTimerBackend.__init__(self, vispy_timer)
        assert _wx_app is not None
        parent = _wx_app.GetTopWindow()  # assume it's the parent window
        self._timer = wx.Timer(parent, -1)
        parent.Bind(wx.EVT_TIMER, self._vispy_timeout, self._timer)

    def _vispy_start(self, interval):
        self._timer.Start(interval * 1000., False)  # cast to int: self._timer.Start(int(interval * 1000.), False) 

    def _vispy_stop(self):
        self._timer.Stop()

    def _vispy_timeout(self, evt):
        self._vispy_timer._timeout()
        evt.Skip()`

Its just an annoying bug, and i dont know if changing it to integer break the timer for older versions of wx...

djhoese:
Here's the C++ libraries version of the base timer class from 2001 that looks like it takes integer milliseconds:

https://github.com/wxWidgets/wxWidgets/blob/9e3cb9eefcb4e8d2d1fb45ad028468547ecc6ad9/src/common/timercmn.cpp#L118

I think it is safe to change it to always be `int`. A pull request would be welcome. There is a chance there was a conversion by SIP (which wxPython uses to access the C++ library) that is now gone, but the underlying Wx library seems to have always used `int` or `long` integers.

Bliss3d:
fixed with #2396

[...]

PR #2396, opened 2022-10-03
fix: wx timer from float to int

Bliss3d:
As mentioned in #2395 the wx backend had a bug with the wx timer receiving a float as argument, instead of an integer


djhoese (approved):
Thanks!

djhoese:
Merging as the failure in Qt6 is not your fault.

PR #2404, opened 2022-10-15
Address deprecation warnings with the usage of distuilts for LooseVersion

hmaarrfk:
Gotta love these deprecation warnings. I hope these help!

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -26,7 +26,7 @@ import os
 import sys
 import atexit
 import ctypes
-from distutils.version import LooseVersion
+from packaging.version import Version
 
 from ...util import logger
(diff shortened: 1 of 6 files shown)


vispy/app/backends/_qt.py
```diff
@@ -114,7 +114,7 @@ def _get_event_xy(ev):
     _check_imports('PySide6')
     if not USE_EGL:
         from PySide6.QtCore import __version__ as QT_VERSION_STR
-        if LooseVersion(QT_VERSION_STR) >= '6.0.0':
```
hmaarrfk, line 117: packaging does not support direct comparison to strings. So the hardest part is remembering to do this everywhere.

djhoese:
Awesome! Any idea how many warnings this gets rid of in a typical CI run? I guess I could just look... :wink:

PR #2411, opened 2022-10-22
Improve PySide6 Support: swapBehavior for DoubleBuffer, BUTTONMAP as pyqt6, xfail test_context

### LINKED ISSUE #2407: error with pyside6 6.4.0 backend on macOS arm64
Making a fresh env, python 3.9 or python 3.10, then installing pyside6 `pip install pyside6` and vispy (0.11.0) works, but running
https://vispy.org/gallery/scene/turntable_box.html#sphx-glr-gallery-scene-turntable-box-py
fails with error:
```
╰─ python turntable_box.py                                        (test-env3) ─╯
  File "/Users/piotrsobolewski/Downloads/turntable_box.py", line 17, in <module>
    canvas = scene.SceneCanvas(keys='interactive', size=(800, 600), show=True)
  File "/Users/piotrsobolewski/Dev/miniforge3/envs/test-env3/lib/python3.9/site-packages/vispy/scene/canvas.py", line 135, in __init__
    super(SceneCanvas, self).__init__(
  File "/Users/piotrsobolewski/Dev/miniforge3/envs/test-env3/lib/python3.9/site-packages/vispy/app/canvas.py", line 211, in __init__
    self.create_native()
  File "/Users/piotrsobolewski/Dev/miniforge3/envs/test-env3/lib/python3.9/site-packages/vispy/app/canvas.py", line 228, in create_native
    self._app.backend_module.CanvasBackend(self, **self._backend_kwargs)
  File "/Users/piotrsobolewski/Dev/miniforge3/envs/test-env3/lib/python3.9/site-packages/vispy/app/backends/_qt.py", line 372, in __init__
    self._init_specific(p, kwargs)
  File "/Users/piotrsobolewski/Dev/miniforge3/envs/test-env3/lib/python3.9/site-packages/vispy/app/backends/_qt.py", line 791, in _init_specific
    glformat = _set_config(p.context.config)
  File "/Users/piotrsobolewski/Dev/miniforge3/envs/test-env3/lib/python3.9/site-packages/vispy/app/backends/_qt.py", line 283, in _set_config
    glformat.setSwapBehavior(glformat.DoubleBuffer if c['double_buffer']
AttributeError: 'PySide6.QtGui.QSurfaceFormat' object has no attribute 'DoubleBuffer'
```

(by comparison the same but with `pip install pyqt6` works fine.)

psobolewskiPhD:
Fixes #2407 by using `SwapBehavior` to set DoubleBuffer
PYSIDE6_API is included in the existing PYQT6_API logic:
https://github.com/vispy/vispy/blob/8bf4f9e538dee5cb986779006b84f62eed81ed7c/vispy/app/backends/_qt.py#L281-L287

This should work for PySide6 going back to at least 6.2 (LTS)
I found the Qt 6.2 (LTS) docs. The swap behavior was already present.
https://doc.qt.io/qt-6.2/qsurfaceformat.html#SwapBehavior-enum
https://doc.qt.io/qtforpython-6.2/PySide6/QtGui/QSurfaceFormat.html#PySide6.QtGui.PySide6.QtGui.QSurfaceFormat.SwapBehavior

Edit: Also uses the same BUTTONMAP as PyQT6
Plus, xfails the shared context tests: test_context.py
Fixes https://github.com/vispy/vispy/issues/2410

### DIFF
--- a/vispy/app/tests/test_context.py
+++ b/vispy/app/tests/test_context.py
@@ -17,8 +17,8 @@ def test_context_properties():
         return  # cannot set more than once on Pyglet
     if a.backend_name.lower() == 'osmesa':
         return  # cannot set config on OSMesa
-    if 'pyqt5' in a.backend_name.lower() or 'pyqt6' in a.backend_name.lower() or 'pyside6' in a.backend_name.lower():
-        pytest.xfail("Context sharing is not supported in PyQt5, PyQt6, PySide6 at this time.")
+    if 'pyqt5' in a.backend_name.lower() or 'pyqt6' in a.backend_name.lower() or 'pyside2' in a.backend_name.lower() or 'pyside6' in a.backend_name.lower():
+        pytest.xfail("Context sharing is not supported in PyQt5, PyQt6, PySide2, or PySide6 at this time.")
 
     # stereo, double buffer won't work on every sys
     configs = [dict(samples=4), dict(stencil_size=8),
@@ -71,8 +71,8 @@ def test_context_sharing():
         check()
 
         # pyqt5 does not currently support context sharing, pyside6 seg faults on app tests
-        if 'pyqt5' in c1.app.backend_name.lower() or 'pyqt6' in c1.app.backend_name.lower() or 'pyside6' in c1.app.backend_name.lower():
-            pytest.xfail("Context sharing is not supported in PyQt5, PyQt6, PySide6 at this time.")
+        if 'pyqt5' in c1.app.backend_name.lower() or 'pyqt6' in c1.app.backend_name.lower() or 'pyside2' in c1.app.backend_name.lower() or 'pyside6' in c1.app.backend_name.lower():
+            pytest.xfail("Context sharing is not supported in PyQt5, PyQt6, PySide2, or PySide6 at this time.")
 
         # Tkinter does not currently support context sharing
         if 'tk' in c1.app.backend_name.lower():
--- a/vispy/app/tests/test_context.py
+++ b/vispy/app/tests/test_context.py
@@ -17,6 +17,8 @@ def test_context_properties():
         return  # cannot set more than once on Pyglet
     if a.backend_name.lower() == 'osmesa':
         return  # cannot set config on OSMesa
+    if 'pyqt5' in a.backend_name.lower() or 'pyqt6' in a.backend_name.lower() or 'pyside6' in a.backend_name.lower():
+        pytest.xfail("Context sharing is not supported in PyQt5, PyQt6, PySide6 at this time.")
 
     # stereo, double buffer won't work on every sys
     configs = [dict(samples=4), dict(stencil_size=8),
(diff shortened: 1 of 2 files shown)


djhoese:
So this drops support for older PySide versions (6.1 and 6.2 most likely), this is OK with me. With this single change to the `if` statement, do your tests from the related issue work now @psobolewskiPhD?

psobolewskiPhD:
> So this drops support for older PySide versions (6.1 and 6.2 most likely), this is OK with me. With this single change to the `if` statement, do your tests from the related issue work now @psobolewskiPhD?

pyside6 6.2 is still fine. the only question is 6.0 and 6.1. I can't install those or find any docs on them.
Alas, it turns out this solves the DoubleBuffer issue, but unmasks a new issue. When running turntable example and trying to interact with the object:
```
  File "/Users/piotrsobolewski/Dev/vispy/vispy/app/backends/_qt.py", line 515, in mouseReleaseEvent
    button=BUTTONMAP[ev.button()],
KeyError: <MouseButton.LeftButton: 1>
  File "/Users/piotrsobolewski/Dev/vispy/vispy/app/backends/_qt.py", line 515, in mouseReleaseEvent
    button=BUTTONMAP[ev.button()],
KeyError: <MouseButton.LeftButton: 1>
```

djhoese:
That's probably an update needed here:

https://github.com/vispy/vispy/blob/8bf4f9e538dee5cb986779006b84f62eed81ed7c/vispy/app/backends/_qt.py#L194

psobolewskiPhD:
> That's probably an update needed here:
> 
> https://github.com/vispy/vispy/blob/8bf4f9e538dee5cb986779006b84f62eed81ed7c/vispy/app/backends/_qt.py#L194

Thanks for the tip!
Fixed.

I went through everywhere there is `PYQT6_API` and checked the PySide6 docs. I think everything should be correct now. And I get no errors.

I also added pyside6 to the xfails for shared context, so now the app tests can run without set fault on pyside6 (note that run individually, `test_context.py` does pass on pyside6 if `se_shared_contexts()` is set to TRUE, but run as all app tests still segfaults.
This adresses https://github.com/vispy/vispy/issues/2410

vispy/app/tests/test_context.py
```diff
@@ -68,9 +70,9 @@ def check():
         # Check while c1 is active
         check()
 
-        # pyqt5 does not currently support context sharing
-        if 'pyqt5' in c1.app.backend_name.lower() or 'pyqt6' in c1.app.backend_name.lower():
-            pytest.xfail("Context sharing is not supported in PyQt5 at this time.")
+        # pyqt5 does not currently support context sharing, pyside6 seg faults on app tests
+        if 'pyqt5' in c1.app.backend_name.lower() or 'pyqt6' in c1.app.backend_name.lower() or 'pyside6' in c1.app.backend_name.lower():
+            pytest.xfail("Context sharing is not supported in PyQt5, PyQt6, PySide6 at this time.")
```
djhoese, line 75: Pyside2 also, right?

psobolewskiPhD: I don't know and I can't test, because pyside2 seems borked on arm64.
Pyside6 the context sharing works and the test passes if run individually, but if you run the app suite then you get seg fault.

psobolewskiPhD: Oh, of course you're right. I'll add the xfail for pyside2 then.

djhoese:
Thanks @psobolewskiPhD!

PR #2415, opened 2022-10-26
Provide a way for the user to specify the z locaton of an Ellipse

hmaarrfk:
When drawing overlapping visuals, the ambiguity in z makes it hard to position the visual in the scene.

I guess i should add a test for this.  Which is why its a draft.


djhoese:
Can't this already be done with an `STTransform` translating in the Z axis?

hmaarrfk:
maybe.... this is a good point.

hmaarrfk:
I'm not sure how it all plays out in the backend, but you are likely correct.

hmaarrfk:
You are right. STTransforms are enough

PR #2420, opened 2022-11-07
Fix deprecated save-output usage in CI

djhoese:
According to this (and the warnings in CI) we're using some deprecated functionality:

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This PR replaces it with the new method.


djhoese:
Ah the warnings I was hoping to fix were part of the cache action and fixed in v3.0.11, but it doesn't seem to be working yet.

https://github.com/actions/cache/issues/953

djhoese:
:man_facepalming: there were two of them

djhoese:
Ugh I missed one:

https://github.com/vispy/vispy/blob/main/.github/workflows/main.yml#L227

PR #2429, opened 2022-11-08
Demo for Drafting: On-GPU Data Access (CUDA, CuPy, PyTorch, DLPack)

xloem:
This is just a scratch work showcasins concepts for gpu data access coming together. It's a quick half-working thing.

Demo (watch the text output: attaches a vertex buffer to a program, then exports it to pytorch in-place):
```
pip3 install torch
cd vispy/gloo/dlpack
python3 cuda_.py
```

I copied from the dlpack repository to write a basic dlpack wrapper for vispy's GL buffers. I spent some hours on this, and am sharing it now that it actually runs at all. I'm not sure whether I will return to this or not, but I expect others would find it helpful.

It looks like something on the GL end is getting corrupt, maybe from my quick calls to access the buffer leaving something out, or from an unaddressed mistake managing the data, not sure, haven't looked at it.

Strangely, CuPy appears to be ignoring its from_dlpack implementation and maybe handing off to numpy, which doesn't support CUDA? So CuPy fails right now. Not sure why that is. It has a cuda-oriented from_dlpack implementation in its codebase, though.

But it was pretty exciting for me to see the same data output in the torch tensor as I passed in to vispy's program, once I got this far.

### DIFF
--- a/vispy/gloo/dlpack/LICENSE
+++ b/vispy/gloo/dlpack/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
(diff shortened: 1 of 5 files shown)


jakirkham:
Thanks for sharing! 🙏

> But it was pretty exciting for me to see similar data output in the torch tensor as I passed in to vispy's program, once I got this far.

Could you please share specifically what was run and the error seen?

xloem:
I've updated the post to include the output. The command run is the first code block.

xloem:
It runs without errors now. As added in my edit, current issues are:
- messiness, it was a quick hack
- integrating with vispy's GL pipeline; flushing commands and accessing buffers in a more general way, without breaking anything

PR #2433, opened 2022-11-15
Avoid explicit loop in computing normals in MeshData

soraxas:
Test timing: (assuming you have two folder named as `MeshData__original` and `MeshData__PR`, which corresponds to the **main** branch and this **PR** branch:
```py
from vispy_main_branch.geometry import MeshData as MeshData__original
from vispy_PR_branch.geometry import MeshData as MeshData__PR
import numpy as np
import time

def run(N_vertex, N_faces):
    ###############################################
    # prepare data
    np.random.seed(0)

    vertices = np.random.rand(N_vertex, 3).astype(np.float32)
    faces = np.random.choice(N_vertex, size=(N_faces, 3))

    # remove rows that are repeated
    faces.sort(axis=1)
    faces = np.unique(faces, axis=0)

    # remove rows with repeated vertices indices
    faces = faces[(faces[:, 0] != faces[:, 1]) & (faces[:, 0] != faces[:, 2]) & (faces[:, 1] != faces[:, 2])]
    ###############################################

    mesh_ori = MeshData__original(vertices=vertices, faces=faces)
    mesh_pr = MeshData__PR(vertices=vertices, faces=faces)

    t = time.time()
    vn_ori = mesh_ori.get_vertex_normals()
    elapsed_ori = time.time() - t
    t = time.time()
    vn_pr = mesh_pr.get_vertex_normals()
    elapsed_pr = time.time() - t

    assert np.isclose(vn_ori, vn_pr).all()

    print(f"{N_vertex},{N_faces},{elapsed_ori},{elapsed_pr}")


N_vertex = 10_000
N_faces = 10_000

print("N_vertex,N_faces,original,PR")
while N_vertex < 5_000_000:
    run(N_vertex, N_faces)
    N_vertex = int(N_vertex * 1.1)

# N_vertex = 10_000
# N_faces = 10_000

# while N_faces < 5_000_000:
#     run(N_vertex, N_faces)
#     N_faces = int(N_faces * 1.1)
```

timings_N_vertex
timings_N_faces

(plotted with the following)
```py
import matplotlib.pyplot as plt
import numpy as np
import csv

with open('out.csv', 'r') as f:
    reader = csv.reader(f)
    label = next(reader)
    data = np.array(list(list(map(float, line)) for line in reader))


plt.plot(data[:, 0], data[:, 2], label=label[2])
plt.plot(data[:, 0], data[:, 3], label=label[3])
plt.xlabel(label[0])
plt.ylabel("time (sec)")

plt.legend()
[...]

PR #2434, opened 2022-11-27
Vectorize MeshData vertex normals computation

asnt:
Suggestion to extend https://github.com/vispy/vispy/pull/2433 with tests, clarification of documentation and some formatting.

### DIFF
--- a/vispy/geometry/meshdata.py
+++ b/vispy/geometry/meshdata.py
@@ -327,21 +327,18 @@ class MeshData(object):
             The normals.
         """
         if self._vertex_normals is None:
-            faceNorms = self.get_face_normals()
-            vertFaces = self.get_vertex_faces()
-            self._vertex_normals = np.empty(self._vertices.shape,
+            faceNorms = self.get_face_normals().astype(np.float32)
+            _vertex_normals = np.zeros(self._vertices.shape,
                                             dtype=np.float32)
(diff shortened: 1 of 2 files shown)


djhoese:
Ok so this is #2433 with tests?

@soraxas how does this look to you?

soraxas:
Yep the code LGTM.

Unit test seems straightforward although it didn't quite test much of the edge cases. Should perform sufficient sanity check though.

djhoese:
Great. @asnt do you think these need any other tests?

asnt:
> Great. @asnt do you think these need any other tests?

Not sure.
What are the possible edges cases? What comes to mind:
- Degenerate triangles (i.e. collinear or coinciding vertices) having undefined normals.
- Distance between vertices close to machine epsilon creating numerical errors: inaccurate or wrong normals.
These are not handled in the current code. Not sure we want this handled at this level of abstraction.

asnt:
Same issue: https://github.com/vispy/vispy/pull/2069

djhoese:
I want this PR and all the related ones to stop weighing on me so I'm going to merge this and anything that breaks we can fix later. We've spent too much time trying to get a perfect PR for this optimization and I think this is the closest we've gotten. I think after this then @asnt can work on #2098 to clean up the code a bit (if you want).

PR #2437, opened 2022-12-17
Fix character codec error on Windows Python <3.10

### LINKED ISSUE #2386: character mapping issue from vispy.gloo.gl._constants
It looks like there is a character encoding issue in vispy that is happening in this bug report, please correct me if I am wrong.

https://forum.image.sc/t/try-to-load-napari-library-from-stardist-napari-dock-widget-import-surface-from-polys-and-failed-like-bellow-any-help-would-be-more-than-welcome/71524

haesleinhuepf:
As proposed by @djhoese 
closes #2386

### DIFF
--- a/codegen/createglapi.py
+++ b/codegen/createglapi.py
@@ -116,10 +116,10 @@ class Enum(int):
 
 DEFINE_CONST_MAP = '''
 ENUM_MAP = {}
-for ob in list(globals().values()):
-    if repr(ob).startswith('GL_'):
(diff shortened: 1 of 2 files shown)


djhoese:
Thanks for this. I just pushed a commit to your branch with the regenerated code. Turns out it didn't need to be done by me, it was just `python createglapi.py`.

djhoese:
Thanks!

PR #2440, opened 2022-12-24
New Visual: Circle

Bliss3d:
As ever just a small new visual, since every graphing api should have a circle.
Anyway I just had to have it for pie charts.

Hope you like it :)

### DIFF
--- a/vispy/visuals/circle.py
+++ b/vispy/visuals/circle.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+# -----------------------------------------------------------------------------
+
+import numpy as np
+import math
+
+from .mesh import MeshVisual
+
+
+class CircleVisual(MeshVisual):
+    """Visual that calculates and displays a circle using triangle's
+
+    Parameters
+    ----------
+    corners : int
+        Data to histogram. Currently only 1D data is supported.
+    radius : int
+        Radius of the circle
+    color : ColorArray
+        ColorArray or color letter for example 'w' for white
+    """
+
+    def __init__(self, corners, radius=1, color='w'):
+
+        color_array = None
+        if isinstance(color, (np.ndarray, np.generic)):
+            if color.shape[0] > 1:
+                color_array = []
+                for i in range(len(color)):
+                    color_array.append(color[i])
+                    if i != (len(color) - 1):
+                        color_array.append([0, 0, 0])
+                color = None
+
+        rr, tris = self._calculate(corners, radius)
+
+        MeshVisual.__init__(self, rr, tris, color=color, face_colors=color_array)
+
+    def update_data(self, corners, radius=1, color='w'):
+
+        color_array = None
+        if isinstance(color, (np.ndarray, np.generic)):
+            if color.shape[0] > 1:
+                color_array = []
+                for i in range(len(color)):
+                    color_array.append(color[i])
+                    if i != (len(color) - 1):
+                        color_array.append([0, 0, 0])
+                color = None
+
+        rr, tris = self._calculate(corners, radius)
+
+        MeshVisual.set_data(self, rr, tris, color=color, face_colors=color_array)
+
+    def _calculate(self, corners, radius):
+        last_angle = 0
+        vertices = []
+        for i in range(corners + 1):
+            if i != 0:
+                y = radius * math.cos(last_angle)
+                x = radius * math.sin(last_angle)
+                vertices.append((x, y))
+                vertices.append((0, 0))
+                angle = 2 * math.pi * i / corners
+                y = radius * math.cos(angle)
+                x = radius * math.sin(angle)
+                vertices.append((x, y))
+                last_angle = angle
+        rr = np.array(vertices)
+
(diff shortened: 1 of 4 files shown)


Bliss3d:
image

vispy/visuals/circle.py
```diff
@@ -0,0 +1,82 @@
...
+
+from .mesh import MeshVisual
+
+
+class CircleVisual(MeshVisual):
+    """Visual that calculates and displays a circle using triangle's
+
+    Parameters
+    ----------
+    corners : int
```
djhoese, line 18: I'm not sure I understand, why "corners"? Are these the "wedges" of the pie chart? What units are these in? Degrees?

Oh wait, it says histogram, but do you ever run the data through a histogram? No wait, this is an integer?

Oh, looking at the example so `color` is really what determines the size of the wedges? And `corners` is more of a "resolution"? No...wait...I'm very confused.

vispy/visuals/circle.py
```diff
@@ -0,0 +1,82 @@
...
+            if i != 0:
+                y = radius * math.cos(last_angle)
+                x = radius * math.sin(last_angle)
+                vertices.append((x, y))
+                vertices.append((0, 0))
+                angle = 2 * math.pi * i / corners
+                y = radius * math.cos(angle)
+                x = radius * math.sin(angle)
+                vertices.append((x, y))
+                last_angle = angle
```
djhoese, line 71: Is there a way to do this without the for loop? For a start, you could remove the `if` and do `range(1, corners + 1)` and that would remove an entire indentation level. Next, if `i` (better variable name?) was `np.arange(corners)` could we do the same type of math and then have `vertices = np.zeros((corners * 3, 3))` and do some level of `vertices[::3, :2] = (radius * np.sin(angles), radius * np.cos(angles))` and then something similar for `vertices[2::3, :2]`. I might have an off-by-one on the size of `i` and need to calculate the `angles`, but otherwise I think that makes sense. Thoughts?

djhoese (changes requested):

vispy/visuals/circle.py
Bliss3d: Oh yes corners are a bit confosing I called it corners because its not actually a circle its a "n-gon", like a octogon or a pentagon. For example an octogon would have 8 corners, and a pentagon would have 5. For the pie chart I set it to 100, so its more like a 100-gon. So yes its made up of wedges, and each wedge can be individualy colored.

Regarding the histogram thats my bad I made this while having covid and lying in bed and I copyed the histogram.py Parameters comment and forgot to change the line with the histogram - im sorry for the confusion

Bliss3d:
octogon
This illustrates the wedges

vispy/visuals/circle.py
Bliss3d: I didnt explore it i just needed it quick, and because i'm not very good generationg the verticies arrays i took the easy route. However your suggestion should be a good start,

djhoese:
Interesting, interesting, interesting. Ok, so my instinct is to say `corners` should be called something like `divisions` to indicate some kind of "resolution" of the pie chart and it should be optional with a high enough default to give a decent looking circle. The main required argument (wedge_percentages?) is then a 1D array representing percentage of the circle to divide up. `color` (or maybe `colors`) is then optional with a default generated set of colors. If provided the colors should be the same number of elements as the wedge percentages. I'm sure there is some algorithm for generating RGB colors to be semi-unique and still have enough for all of the wedges.

For the `divisions`, if it started with 100 then you can easily determine the number of wedges that need to be a certain color to make up a single wedge. If there are more wedges provided by the user than divisions then error out, but if divisions is the default then make it equal to `max(len(divisions), 100)`.

```
def __init__(..., divisions=None):
    if divisions is not None and divisions < len(wedges):
        raise ValueError(...)
    if divisions is None:
        divisions = max(len(wedges), 100)
```

Another optional parameter: starting angle of first wedge (default 0 degrees is the x-axis).

`</brainstorm>`

Bliss3d:
Yes that would make sense, but it's not quite clear to me if you what this visual to be a pie chart?, I just created it to be a circle. The pie chart is just an example - I'd say the pie chart should be in plotwidget.py

djhoese:
You can already do circles with Markers so I guess I'm not sure what purpose this serves except as a divided pie chart circle. Let me know if you had other ideas.

brisvag:
I agree that this seems to be a (very useful) n-gon or pie chart addition, rather than a straight up circle. I would reach out to the `Marker` for a straight up circle (though it has limitations on some systems, see #2078), and to this for a pie chart. I would change the name to reflect that!

Bliss3d:
Sorry for the late response I'm busy studying. Anyway, I didnt know that Marker could be used to create circles, and yes I created this visual because I wanted a pie chart, but I wanted it also to be general purpose. However if you prefer it to be just a piechart visual or an n-gon generator, I'd have no issue with that.

djhoese:
> However if you prefer it to be just a piechart visual or an n-gon generator, I'd have no issue with that.

I think the main thing is that have a generic "circle" visual where the main interface is "here's how you split a circle into N wedges", means this isn't a generic circle anymore. A generic circle would be `pos` (x, y, z position), `radius`, a color, and maybe something to describe the orientation. I think that is covered decently by Markers with a circle marker (or sphere) depending on the needs of the user. I don't think there is anything that draws a 2D circle that you can pan around in 3D space. I think Markers always "face" the camera.

PR #2444, opened 2023-01-15
Clean up and test computation of normals in MeshData

asnt:
Supersedes #2098.

Summary:
- Test computation of face normals.
- Extract computation of face normals and vertex normals out of MeshData for clarity.

### DIFF
--- a/vispy/geometry/meshdata.py
+++ b/vispy/geometry/meshdata.py
@@ -34,6 +34,32 @@ def _repeat_face_normals_on_corners(normals):
     return np.repeat(normals, n_corners_in_face, axis=0).reshape(new_shape)
 
 
+def _compute_vertex_normals(face_normals, faces, vertices):
+    assert face_normals.shape[1:] == (3,), "Require Fx3 array of face normals."
+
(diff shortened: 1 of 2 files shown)


djhoese (changes requested):
Looks good to me. Let me know if you want to do any more work on this besides what I describe below.

The only thing I'd like is: could you switch your `assert` statements in the `meshdata.py` code to if statements with regular ValueError exceptions? It is considered by some linting/complexity/security tools to be bad practice to use assert statements in non-testing code. Let me know what you think.

asnt:
> Looks good to me. Let me know if you want to do any more work on this besides what I describe below.

Thanks @djhoese for the review. That would be all from my side for this PR.

> The only thing I'd like is: could you switch your `assert` statements in the `meshdata.py` code to if statements with regular ValueError exceptions? It is considered by some linting/complexity/security tools to be bad practice to use assert statements in non-testing code. Let me know what you think.

Agreed. Change applied.

I also changed the relevant `raise Exception` to `raise ValueError`.

djhoese (approved):
Love it! Thank you!

PR #2457, opened 2023-03-06
Fix f-string missing prefix in gloo/program.py error message

aganders3:
I just noticed this small issue while working on something else, the missing `f` causes the error to be missing the intended debug info.


brisvag (approved):

djhoese (approved):
Awesome! Thanks!

PR #2463, opened 2023-03-15
Fix shader compilation error for MeshVisual when mesh.clim = 'auto'

aganders3:
Rendering a mesh with `vertex_values` produces a shader compilation error when `mesh.clim == "auto"` (the default):
```
RuntimeError: Shader compilation error in GL_VERTEX_SHADER:
    on line 336: Use of undeclared identifier 'a'
      float cmap(float val) { return (val - a) / (u - a); }
    on line 336: Use of undeclared identifier 'u'
      float cmap(float val) { return (val - a) / (u - a); }
    on line 336: Use of undeclared identifier 'a'
      float cmap(float val) { return (val - a) / (u - a); }
```
These "undeclared identifiers"  `a` and `u` are meant to be numbers that come from `mesh.clim`, but this attribute may be a tuple or the string `"auto"`, so here it was taking the first two characters from that string instead of the values.

### DIFF
--- a/vispy/visuals/mesh.py
+++ b/vispy/visuals/mesh.py
@@ -283,8 +283,8 @@ class MeshVisual(Visual):
         clim_func = 'float cmap(float val) { return (val - $cmin) / ($cmax - $cmin); }'
         if colors.ndim == 2 and colors.shape[1] == 1:
             fun = Function(clim_func)
-            fun['cmin'] = self.clim[0]
-            fun['cmax'] = self.clim[1]
+            fun['cmin'] = self._clim_values[0]
+            fun['cmax'] = self._clim_values[1]
             fun = FunctionChain(None, [fun, Function(self.cmap.glsl_map)])
         else:
             fun = Function(null_color_transform)
--- a/vispy/visuals/tests/test_mesh.py
+++ b/vispy/visuals/tests/test_mesh.py
@@ -33,6 +33,23 @@ def test_mesh_color():
 
 @requires_pyopengl()
 @requires_application()
+def test_mesh_with_values():
+    size = (45, 40)
+    with TestingCanvas(size=size) as c:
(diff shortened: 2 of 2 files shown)


brisvag (approved):
Nice!

vispy/visuals/tests/test_mesh.py
```diff
@@ -31,6 +31,23 @@ def test_mesh_color():
     np.testing.assert_allclose(vertices['position'], new_vertices)
 
 
+@requires_pyopengl()
+@requires_application()
+def test_mesh_with_values():
```
djhoese, line 36:
```suggestion
def test_mesh_with_vertex_values():
```

vispy/visuals/mesh.py
```diff
@@ -283,8 +283,8 @@ def _build_color_transform(self, colors):
         clim_func = 'float cmap(float val) { return (val - $cmin) / ($cmax - $cmin); }'
         if colors.ndim == 2 and colors.shape[1] == 1:
             fun = Function(clim_func)
-            fun['cmin'] = self.clim[0]
-            fun['cmax'] = self.clim[1]
+            fun['cmin'] = self._clim_values[0]
+            fun['cmax'] = self._clim_values[1]
```
djhoese, line 287: So this method/property was added and then never used? Wow, good find, but sad that it took this long to find it.

vispy/visuals/tests/test_mesh.py
aganders3: A good change and thanks for saving me the keystrokes!

vispy/visuals/mesh.py
aganders3: Diving into the git history a bit I think it was used when added but lost during a refactor.

PR #2488, opened 2023-05-14
Do not use old style enum for PySide6

hmaarrfk:
It seems to be getting removed

See: https://wiki.qt.io/Qt_for_Python_Development_Notes

> old Enums are being removed, and the startup time will improve a bit (still in progress)

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -284,8 +284,8 @@ def _set_config(c):
         glformat.setSwapBehavior(glformat.DoubleBuffer if c['double_buffer']
                                  else glformat.SingleBuffer)
     elif PYQT6_API or PYSIDE6_API:
-        glformat.setSwapBehavior(glformat.SwapBehavior.DoubleBuffer if c['double_buffer']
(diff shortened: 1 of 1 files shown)


djhoese:
I'm guessing this needs to be handled differently for PyQt6 and Qt5-based backends. Very odd that this pyside6 change means the enum is accessed from the class and not an instance of the class like before.

hmaarrfk:
https://doc.qt.io/qtforpython-6/considerations.html#the-differences-between-old-and-new-enums

I just found this doc too....

djhoese:
Oh good. I was hoping there was some guidance...although the differences and suggested usage really don't make sense to me.

hmaarrfk:
Nor do they to me.

Feel free to wait until the conversation with Czaki finishes in 
https://github.com/napari/napari/pull/5833#issuecomment-1546853596

I found similar bugs in napari.

hmaarrfk:
I think that a recent change was made here that is not in line with 0.11.0

I've decided to use PyQt for napari for now. until the dust settles there.

I think the code as you have it now in master/main is correct

Issue #2495, opened 2023-06-06
Exception with glfw create_window


smyeungx:
Hi,

Vispy is amazing.  

**Exception:**
I have tried to use with with glfw backend (3.3.8) under python 3.9.7 under Microsoft Visual Studio Code.  
In vispy->app->backends->_glfw.py -> CanvasBackend -> __init__ --> class CanvasBackend(BaseCanvasBackend):
In the code around line 264:
        self._id = glfw.create_window(width=size[0], height=size[1],
                                      title=p.title, monitor=monitor,
                                      share=share)
will return an exception with glfw

**Observation:**
In vispy->app->backends->_glfw.py -> CanvasBackend -> __init__ --> class CanvasBackend(BaseCanvasBackend):

In the code around line 254:
use_size = glfw.get_video_mode(monitor)[:2]

Under Python 3 with glfw:
glfw.get_video_mode(monitor) returns a tuple like the following:
glfw.get_video_mode(monitor)
GLFWvidmode(size=Size(width=3840, height=2160), bits=Bits(red=8, green=8, blue=8), refresh_rate=60)

Therefore, the width and height cannot be returned with [:2] directly (becomes returning Size and Bits of the Monitor):
glfw.get_video_mode(monitor)[:2]
(Size(width=3840, height=2160), Bits(red=8, green=8, blue=8))

**Proposal:**
Proposed to modify with
1) import sys
2) use_size = glfw.get_video_mode(monitor)[:2] if (sys.version_info.major==2) else glfw.get_video_mode(monitor)[0][:2]

So that when Python 3 is detected, the use_size will obtain the first tuple [0], and width and height obtained as usual with [:2]

Best,
Simon

djhoese:
Sorry, I'm not sure I understand what is causing this error. Or rather, what changed to cause this error? We've been using Python 3 for a long time, dropped Python 2 a long time ago, and haven't changed anything related to glfw for a long time. Even the `glfw` library (pyglfw) that we depend on in this module has only had bug fix releases for the last year:

https://github.com/FlorianRhiem/pyGLFW/blob/master/CHANGELOG.md

```
pip install glfw
```

At least I think that's the one we use. Did you maybe install `pyglfw`? I have no idea if these are different.

Side note: Could you please paste the original exception that you got (and with proper code formatting) so that users googling for the error can find this issue?

djhoese:
So it definitely looks like our CI tests don't run the tests with glfw:

[...]

PR #2496, opened 2023-06-08
Update _glfw.py

smyeungx:
#2495 
A fix that take care of the 3 namedtuple returned by function glfw.get_video_mode(monitor) in glfw backend, and use only the first tuple (size) as the use_size variable.


djhoese:
After this change are you able to use glfw with your project and/or the vispy examples?

djhoese:
> Further investigation show in single monitor mode, it doesn't have any error in existing implementation. But in dual or multi-monitor configuration, PR modification is need.

Ah what a good find. Very interesting. Thanks for testing.

PR #2500, opened 2023-06-14
Add primitive picking filters for Mesh and Markers visuals

aganders3:
This adds a filter for picking individual faces from a mesh. It works similarly to the picking mode already in use for picking _visuals_ but instead colors the individual faces (triangles) of the mesh based on an index.

I also added an example (heavily borrowing from the mesh shading example) to illustrate how the feature works and how to use it. In the example the mouse will "paint" faces as it moves over a mesh. Pressing 'p' will show the face picking view, that is it will show the colors used to encode the face indices.

https://github.com/vispy/vispy/assets/1231828/4556796d-7bce-4bbf-85eb-697b2b7be940


vispy/visuals/filters/mesh.py
```diff
@@ -766,3 +766,85 @@ def _attach(self, visual):
...
+        else:
+            n_faces = len(self._visual.mesh_data.get_faces())
+
+        # we only care about the number of faces changing
+        if self._n_faces == n_faces:
+            return
+        self._n_faces = n_faces
+
+        # pack the face ids into a color buffer
+        # TODO: consider doing the bit-packing in the shader
```
aganders3, line 832: Passing the ID as the buffer might save a little VRAM? It's sort of unfortunate we need a buffer for this at all - I think in newer GLSL it could use `gl_PrimitiveID` but it's not available in 1.20.

brisvag (commented):
Awesome @aganders3! This is a lot smaller than I expected in terms of actual changes :) And I see how this can be adapted to other visuals!

One thing I noticed in the `picking` mode builtin in the canvas is that you can actually render a specific `region` and/or `crop` of the canvas. Maybe we can use the same here to improve performance, by only rendering a small area around the cursor. Or am I misunderstanding?

aganders3:
Thanks!

Regarding cropping - you're probably right but I have not tested! I can try this and do some profiling. I'll note right now this is pretty much left up to the downstream developer because I didn't add anything like `visual_at` (`face_at`). The picking-by-mouse logic is only in the example.

examples/scene/face_picking.py
```diff
@@ -82,30 +83,42 @@ def on_transform_change(event):
...
     face_picking_filter.enabled = True
     mesh.update_gl_state(blend=False)
-    picking_render = canvas.render(bgcolor=(0, 0, 0, 0), alpha=True)
-
+    picking_render = canvas.render(
+        region=(x_pos - 1, y_pos - 1, 3, 3),
+        size=(3, 3),
+        bgcolor=(0, 0, 0, 0),
+        alpha=True,
+    )
```
aganders3, line 111: Thanks for the idea @brisvag - I think this is something like a 2-4x improvement! Is that in line with what you'd expect?

aganders3: Hm, actually using `crop` instead of `region` seems even a bit faster. This was unexpected for me but seems to agree with the docstring for `SceneCanvas._render_picking`:
>  The crop (x, y, w, h) of the framebuffer to read. For picking the full canvas is rendered and cropped on read as it is much faster than triggering transform updates across the scene with every click.

aganders3:
> As far as I can tell, this filter could be used almost as-is for Markers, you just need to change _update_data :)

I think I see what you mean - same with some of the other visuals. Would you like me to try to add that to this PR? I guess that would need a separate but similar filter (more or less) for each visual type? Or do you have an idea for how it could be a single class -- maybe it could dispatch `_update_data` based on the visual type?

examples/scene/face_picking.py
brisvag: Interesting! I definitely expected `region` to be much faster, but if there's transform updates involved I guess it makes sense. 2-4x is already pretty good though :)

brisvag:
It's actually  slightly more complex, because here you also rely on `mesh_data` and some events that are not present in other places. Still, it shouldn't be too complex to adapt.

> I guess that would need a separate but similar filter (more or less) for each visual type?

Yea, I think the easiest way would be to have a main class with the shader logic, and each subclass has the logic to go from `Visual` to `Number of primitives` and how/when to update it... It's really a pity that `glPrimitiveID` is not available always, it would be so much easier :(

> Would you like me to try to add that to this PR?

Not necessarily other visuals, but if you could set this up so it's extensible in a followup PR it would be great! On the other hand, this is already mergeable on its own, so feel free to choose where to put your energy :)

jni:
Just dropping by to say 😍!

brisvag:
Wait, I just noticed something in the initial video: when you display the picking, quads are colored rather than triangles. Something's off with indexing in that mode?

aganders3:
I think they just look like quads because the triangles are ordered sort of in pairs? If triangles are next to each other (in the index space) they will be very close in color. Here it is with the indices multiplied by 10_000 to show the differences.
Screenshot 2023-06-15 at 6 44 53 AM

It does seem like a strange ordering. Maybe this mesh comes from a quad mesh that was triangulated or something?

aganders3:
It was not too bad to add marker picking. I'm going to clean it up and see if there's a good way to abstract the base class 


https://github.com/vispy/vispy/assets/1231828/831555fb-5649-4f22-8e17-976937055d27

aganders3:
@brisvag this is now refactored into a base class and two picking filters for Markers and Mesh visuals. Please take another look when you have a chance!

vispy/visuals/filters/base_filter.py
```diff
@@ -120,3 +125,98 @@ def _detach(self, visual):
...
+                    return;
+                }
+                gl_FragColor = v_marker_picking_color;
+            }
+        """)
+
+        self._id_colors = VertexBuffer(np.zeros((0, 4), dtype=np.float32))
+        vfunc['ids'] = self._id_colors
+        self._n_primitives = 0
+        super().__init__(vcode=vfunc, fcode=ffunc)
```
brisvag, line 158: Might be a good idea to set `fpos` to a high value (like `1e9` or so), since we *really* want this color change to happen at the end of any other filter.

vispy/visuals/filters/base_filter.py
```diff
@@ -120,3 +125,98 @@ def _detach(self, visual):
 
         self._attached = False
         self._visual = None
+
+
+class PrimitivePickingFilter(Filter, metaclass=ABCMeta):
```
brisvag, line 130: Any reason to use `ABCMeta` other than just inherit from `ABC`? Not really important, just curious.

vispy/visuals/filters/base_filter.py
```diff
@@ -120,3 +125,98 @@ def _detach(self, visual):
...
+
+    @abstractmethod
+    def _update_id_colors(self):
+        """Calculate the colors encoding the picking IDs for the visual.
+
+        Generally, this method should be implemented to:
+            1. Calculate the number of primitives in the visual, stored in
+            `self._n_primitives`.
+            2. Pack the picking IDs into a 4-component float array (RGBA
+            VertexBuffer), stored in `self._id_colors`.
```
brisvag, line 169: Would be awesome if the packing was done here instead of in the subclasses, since it's error prone and anyways it's always the same... Youy probbaly nee dto juggle some extra submethods calling each other, but it would be useful. not blocking though!

vispy/visuals/markers.py
```diff
@@ -573,6 +574,8 @@ def __init__(self, scaling="fixed", alpha=1, antialias=1, spherical=False,
                           blend_func=('src_alpha', 'one_minus_src_alpha'))
         self._draw_mode = 'points'
 
+        self.events.add(data_updated=Event)
```
brisvag, line 577: Should we add this to the base visual to ensure it always exists? @djhoese

brisvag (commented):
Awesome! A few comments here and there, but nothing major. Supert excited to use this in napari, will be a huge speed up and improvement in rensponsiveness!

vispy/visuals/filters/base_filter.py
aganders3: I agree but didn't immediately have a way to do it that felt clean/intuitive so I punted and tried to compensate in the docstring. I'll give it a bit more thought because it does seem possible.

An alternative would be to move the packing into the shader and instead just pass the IDs in the `VertexBuffer`. I avoided this for now just because I'm more comfortable with numpy than I am with GLSL but I don't think it would be too difficult.

Since these are also "just" implementation details I'm also happy to explore both options in a new PR if we want this to (potentially) be available a bit quicker.

vispy/visuals/filters/base_filter.py
aganders3: I think I chose to specify the metaclass directly because it was already inheriting from another class so this felt more explicit. I was a little back-and-forth on this because it does the exact same thing and even seeing the word "metaclass" can be a bit ominous, haha.

vispy/visuals/filters/base_filter.py
aganders3: This is definitely a good idea. One question is how it should be set relative to the existing `PickingFilter`, which only sets `fpos=10`.

aganders3:
Welcome back and thanks for the great comments!

I have another question now that I have looked more closely at the existing `PickingFilter` - should we include something like this in the fragment shader?
```glsl
if( gl_FragColor.a == 0.0 )
    discard;
```

vispy/visuals/markers.py
djhoese: I wouldn't be against that. I'd be a little worried that users seeing it would assume that it is properly emitted/handled by all visuals which won't be true after a lot of work. Updating every `set_data`, right? Regardless, moving it to the base classes and documenting that it is new in version X and may not be used/emitted by all Visuals would be a good start.

vispy/visuals/filters/base_filter.py
aganders3: I just played with this a bit and it's a small problem if `fpos` is higher than that for `PickingFilter` (10), then visual picking (i.e. `visuals_at`) is broken. This doesn't feel like a common use case, but still.

My first instinct is to set `fpos` on the existing `PickingFilter` to something high like `1e9` and something like `1e9-1` below that on the `PrimitivePickingFilter` ABC, but I think this would be a breaking change.

Otherwise maybe just set `fpos` here to `9` but also it to `__init__` in case that's not sufficient.

What do you think?

vispy/visuals/filters/base_filter.py
aganders3: See [36db4fc](https://github.com/vispy/vispy/pull/2500/commits/36db4fc96dbffd8293f6583e43595292542e3079) for one idea

brisvag:
> I have another question now that I have looked more closely at the existing `PickingFilter` - should we include something like this in the fragment shader?

This stuff should already be handled by the visuals themselves, imo, if required. If a visual *does* exist somewhere but is fully transparent, I can imagine cases where you'd still want to detect it with the picking mode... Maybe having a flag for it (`pick_transparent`) would allow this to work both ways?

vispy/visuals/markers.py
brisvag: Yeah, definitely document it. But I guess it's also not *too* much work to simply emit this in every visual's set_data.

vispy/visuals/filters/base_filter.py
brisvag: I was thinking more something like:

```python
@staticmethod
def _pack_ids_into_rgba(ids):
	# pack ids
	...
	return packed

@abstractmethod
def _get_picking_ids(self):
	# this is the method that subclasses will change, which simply
	# calculates the ids and spits them out as normal numpy array
	# and it could for example return None if no change is needed
	# to improve performance
	return None

def _update_id_colors(self):
	# this should remain untouched
	ids = self._get_picking_ids()
	if ids is None:
		return

	id_colors = self._pack_ids_into_rgba(ids)
	self._id_colors.set_data(id_colors)

def _on_data_updated(self, event=None):
	if not self.attached:
		return
	self._update_id_colors()
```

A bit more complicated, but removes the need to think about packing from the sublcasses, and should make it easier to switch to a shader implementation later as well.

brisvag: > Since these are also "just" implementation details I'm also happy to explore both options in a new PR if we want this to (potentially) be available a bit quicker.

To be clear, perfectly happy with this as well!

vispy/visuals/markers.py
djhoese: The problem becomes the Visuals that have other properties for setting data like `.color = new_color` and don't end up calling a `.set_data` method at all.

brisvag:
Exactly, which is why I think it's useful to be able to both include or exclude transparent fragments.

vispy/visuals/markers.py
aganders3: If it's okay I would just leave it as-is for now. I can look into moving this to the base visual (and updating visuals to emit it) in another PR.

brisvag:
Btw, some inspiration for doing this in the shader (here or later): https://stackoverflow.com/questions/28910861/opengl-unique-color-for-each-triangle

aganders3:
I tried moving this into the shader, but I'm stuck unable to pas the ids as uint32 in the VertexBuffer:
```
    on line 339: int can't be an in in the vertex shader
      attribute int a_ids;
```
I tried some workarounds, but I don't think you can do casting in glsl, and some possibly useful features (`floatBitsToUint`) are not available in this version of GLSL. I'm happy to leave this as-is and explore moving more into the shader later.

alisterburt:
also jumping in to say 😍

vispy/visuals/markers.py
brisvag: Sounds good to me!

brisvag (approved):
Ok, I think we're good to go! @djhoese, feel free to take a last look and/or merge!

vispy/visuals/filters/markers.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Note that the ID color uses the alpha channel, so this may not be used
+    with blending enabled.
+
+    Examples
+    --------
+    See
+    `examples/scene/marker_picking.py
+    <https://github.com/vispy/vispy/blob/main/examples/scene/marker_picking.py>`_
+    example script.
```
djhoese, line 17: We should be able to refer to this with a sphinx gallery reference. I think it is something like:

```
:ref:`sphx-glr-gallery-scene-marker-picking-py`
```

But I could be wrong.

aganders3: Thanks! It took me a bit to get the docs to build locally but this is working:
```
:ref:`sphx_glr_gallery_scene_marker_picking.py`
```

aganders3:
Thanks again for the reviews. I was away all last week but am available now to address any remaining issues.

djhoese (approved):
I downloaded the CI artifact (a zip of the rendered HTML) and yeah this looks good. Very cool. Thanks for figuring it out and for getting this all working.

PR #2503, opened 2023-06-22
Add pre-commit config, formatting with black, linting with ruff

### LINKED ISSUE #2291: Add pre-commit config to enable auto-formatter and linting on commit
See https://pre-commit.com/

Related https://github.com/vispy/vispy/issues/2146

A pre-commit config is a great utility/tool to take advantage of to avoid code style, security, and other types of issues before the code is ever committed. With the additional use of `pre-commit.ci`, we can enable the execution of it on PRs.

We could borrow from projects like:

* satpy (https://github.com/pytroll/satpy/blob/main/.pre-commit-config.yaml)
* dask (https://github.com/dask/dask/blob/main/.pre-commit-config.yaml)
* xarray (https://github.com/pydata/xarray/blob/main/.pre-commit-config.yaml)

I think a good starting point would be `black` and `flake8` and see how bad the complaints are. We may not be able to enable `pre-commit.ci` immediately as it has to succeed on the entire repository and VisPy docstrings are just not consistent enough for that.

aganders3:
This closes #2291 and fixes the broken lint check in CI (https://github.com/vispy/vispy/issues/2291#issuecomment-1601356010) by replacing it with pre-commit. This will be a lot to review but hopefully it helps give a sense of the scale of changes. Timing-wise I'm not sure if there are any existing PRs you'd rather merge first (I have two open that would need to be rebased, haha).

Configuration is almost entirely in `pyproject.toml`, with a bit in `.pre-commit-config.yaml`. The main exclusion is `vispy/gloo/gl`, which I believe is all/mostly generated code. I tried to comment the configs to explain how values, but of course I'm open to changes on that as well. I played with line-length a bit to pick a sane-ish value and settled on 100 for black, with ruff errors at 110. Pulling it down to 88 (black's default) or 79 created over 1000 more errors I think. So, a lot more but maybe a drop in the bucket compared to the changes already here.

Finally, this adds a `.git-blame-ignore-revs` file so the one massive commit would be skipped by default in "blame" view on GitHub.


aganders3:
I think the CI failure is the same as this one: https://github.com/vispy/vispy/pull/2502#issuecomment-1600892400

djhoese:
I just merged your other PR. Feel free to rebase/merge with `main` and let's see how CI behaves.

PR #2516, opened 2023-08-18
Dark and Light mode for logo on Navbar

UtkarshMishra12:
I've taken a reference from this [website](https://pradyunsg.me/furo/customisation/logo/#different-logos-for-light-and-dark-model) in order to support the logo in dark mode.

I did not manage to build this locally but hopefully, the CI will correctly.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -203,6 +203,12 @@ html_title = 'VisPy'
 # of the sidebar.
 html_logo = "_static/vispy-teaser-short.png"
 
+html_static_path = ["_static"]
+html_theme_options = {
+    "light_logo": "logo-light-mode.png",
+    "dark_logo": "logo-dark-mode.png",
+}
+
 
 # The name of an image file (within the static path) to use as favicon of the
 # docs.  This file should be a Windows icon file (.ico) being 16x16 or 32x32
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ vispy/static/
 vispy/version.py
 .vscode
 vispy/visuals/text/_sdf_cpu*.pyd
+.venv


rossant:
(Note: this is part of the EuroSciPy 2023 sprint with @rougier and I)

doc/conf.py
```diff
@@ -201,7 +201,12 @@ def _custom_edit_url(github_user, github_repo, github_version, doc_path, file_na
 
 # The name of an image file (relative to this directory) to place at the top
 # of the sidebar.
-html_logo = "_static/vispy-teaser-short.png"
+
+html_static_path = ["_static"]
+html_theme_options = {
+    "light_logo": "logo-light-mode.png",
+    "dark_logo": "logo-dark-mode.png",
+}
```
djhoese, line 209: The documentation you linked to for implementing this was for the "Furo" sphinx theme which we aren't using. We are using the pydata sphinx theme which has documentation for doing light/dark logos here:

https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/branding.html#different-logos-for-light-and-dark-mode

You'll need to change this to match. Additionally you are overwriting the `html_theme_options` dictionary. You'll need to move these change to the existing `html_theme_options` dictionary (starting on line 185). Similarly, the `html_static_path` is already defined a couple lines below your changes.

djhoese (changes requested):
Thanks for trying to tackle this. I've made an inline comment for the changes needed to make this hopefully work.

djhoese:
Also note I think you can turn off building the gallery of examples with:

```
cd doc
make html_dev-pattern PATTERN="abcd"
```

This will only generate the examples matching "abcd" in the filename which should be none of them.

PR #2540, opened 2023-10-23
Change `devicePixelRatio` calls to `devicePixelRatioF` to get a floating point number instead of an integer

### LINKED ISSUE #2538: Using `QT_SCALE_FACTOR` causes a cropped canvas (Qt backend)
Hi, seems like setting the Qt environment variable `QT_SCALE_FACTOR` to a value like `1.8` causes the canvas to appear cropped/without using the space properly. A preview of the behavior:

* Running the [`sphere.py` example](https://github.com/vispy/vispy/blob/main/examples/scene/sphere.py) without setting `QT_SCALE_FACTOR`:

sphere1

* Running the [`sphere.py` example](https://github.com/vispy/vispy/blob/main/examples/scene/sphere.py) setting `QT_SCALE_FACTOR` to `1.8` (`set QT_SCALE_FACTOR=1.8`):

sphere2

Tested the above running (relative to the cloned vispy repo directory) from a Python environment on Windows with vispy 0.14.1 the following:

```cmd
set QT_SCALE_FACTOR=
```
 or

```cmd
set QT_SCALE_FACTOR=1.8
```
and
```cmd
python examples/scene/sphere.py
```

Also, for tracebility, I started investigating this due to https://github.com/napari/napari/issues/6197. A possible fix is explained at https://github.com/napari/napari/issues/6197#issuecomment-1769207938. Let me know if opening a PR with the changes described there makes sense or if more info is needed!

dalthviz:
Closes #2538

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ Announcing:
 """
 
 import os
+import sys
 from os import path as op
 from setuptools import setup, find_packages
 
@@ -59,6 +60,10 @@ extensions = [Extension('vispy.visuals.text._sdf_cpu',
                         ),
               ]
 
+install_requires = ['numpy', 'freetype-py', 'hsluv', 'kiwisolver', 'packaging']
+if sys.version_info <= (3, 8):
+    install_requires.append("importlib-resources")
+
 readme = open('README.rst', 'r').read()
 setup(
     name=name,
@@ -92,7 +97,7 @@ setup(
     platforms='any',
     provides=['vispy'],
     python_requires='>=3.8',
-    install_requires=['numpy', 'freetype-py', 'hsluv', 'kiwisolver', 'packaging'],
+    install_requires=install_requires,
     extras_require={
         'ipython-static': ['ipython'],
         'pyglet': ['pyglet>=1.2'],
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ extensions = [Extension('vispy.visuals.text._sdf_cpu',
               ]
 
 install_requires = ['numpy', 'freetype-py', 'hsluv', 'kiwisolver', 'packaging']
-if sys.version_info <= (3, 8):
+if sys.version_info < (3, 9):
     install_requires.append("importlib-resources")
 
 readme = open('README.rst', 'r').read()
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -169,17 +169,17 @@ jobs:
       matrix:
         include:
           - test_number: 1
-            python_version: "3.8"
+            python_version: "3.9"
             deps: "min"
             test: "standard"
             new_qt: false
           - test_number: 2
-            python_version: "3.8"
+            python_version: "3.9"
             deps: "full"
             test: "standard"
             new_qt: false
           - test_number: 3
-            python_version: "3.8"
+            python_version: "3.9"
             deps: "osmesa"
             test: "osmesa"
             new_qt: false
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -111,7 +111,7 @@ jobs:
         sleep 5
         # export python_version
         echo "python_version=310" >> $GITHUB_OUTPUT
-    - uses: mamba-org/provision-with-micromamba@v16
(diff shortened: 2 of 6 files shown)


djhoese:
Looks like our CI hasn't run for a while. I think I've fixed all the basic issues (hopefully). FYI to @brisvag looks like the networkx stuff was depending on importlib.resources.files which was only added in Python 3.9 so the external `importlib-resources` package is needed.

We should drop Python 3.8 support.

setup.py
```diff
@@ -59,6 +60,10 @@ def set_builtin(name, value):
                         ),
               ]
 
+install_requires = ['numpy', 'freetype-py', 'hsluv', 'kiwisolver', 'packaging']
+if sys.version_info <= (3, 8):
+    install_requires.append("importlib-resources")
```
djhoese, line 65: Apparently this was flawed. Let's try:

```suggestion
install_requires = ['numpy', 'freetype-py', 'hsluv', 'kiwisolver', 'packaging']
if sys.version_info < (3, 9):
    install_requires.append("importlib-resources")
```

djhoese:
Looks like a Python 3.9+ version of networkx was built on conda-forge without updating the python specifier. I'll just run tests on Python 3.9. This is getting ridiculous.

dalthviz:
Yes, I think the problem comes from `networkx` since version 3.2 dropped support for Python 3.8 but the conda feedstock was auto updated without taking into account that (just in case the issue following that: https://github.com/networkx/networkx/issues/7039)

djhoese:
@larsoner I updated testing to use Python 3.9 and now osmesa doesn't want to find the installed lib. Any ideas?

```
 python make test osmesa
WARNING: VISPY_GL_LIB is ignored when using OSMesa. Use OSMESA_LIBRARY instead.
  File "/home/runner/work/vispy/vispy/make/make.py", line 105, in test
/home/runner/micromamba-root/envs/vispy-tests/lib/libOSMesa32.so: cannot open shared object file: No such file or directory
    test(args[0], ' '.join(args[1:]), coverage=True)
  File "/home/runner/work/vispy/vispy/vispy/util/wrappers.py", line 153, in test
    return _test(*args, **kwargs)
  File "/home/runner/work/vispy/vispy/vispy/testing/_runners.py", line 378, in test
    from ..app.backends import BACKEND_NAMES as backend_names
  File "/home/runner/work/vispy/vispy/vispy/app/__init__.py", line 13, in <module>
    from .canvas import Canvas, MouseEvent, KeyEvent  # noqa
  File "/home/runner/work/vispy/vispy/vispy/app/canvas.py", line 16, in <module>
    from ..gloo.context import (GLContext, set_current_canvas, forget_canvas)
  File "/home/runner/work/vispy/vispy/vispy/gloo/__init__.py", line 47, in <module>
    from . import gl  # noqa
  File "/home/runner/work/vispy/vispy/vispy/gloo/gl/__init__.py", line 228, in <module>
    from . import gl2 as default_backend  # noqa
  File "/home/runner/work/vispy/vispy/vispy/gloo/gl/gl2.py", line 26, in <module>
    _lib = ctypes.cdll.LoadLibrary(_lib)
  File "/home/runner/micromamba/envs/vispy-tests/lib/python3.9/ctypes/__init__.py", line 452, in LoadLibrary
    return self._dlltype(name)
  File "/home/runner/micromamba/envs/vispy-tests/lib/python3.9/ctypes/__init__.py", line 374, in __init__
OSError: /home/runner/micromamba-root/envs/vispy-tests/lib/libOSMesa32.so: cannot open shared object file: No such file or directory
```

If I remember right @tlambert03 tried using the 64-bit version a while back and it failed so we have a limit set in https://github.com/dalthviz/vispy/blame/issue-2538/ci/requirements/linux_osmesa_deps_conda.txt

larsoner:
No idea on this one. I would try to get rid of the limit if possible, they eventually always seem to cause problems and now you seem to be hitting one

djhoese:
I'll try, but I think the limit was because conda-forge stopped producing 32-bit builds and for some reason the 64-bit version didn't work. I might make another PR just for this osmesa annoyance.

djhoese:
Merged via #2541. Thanks!

PR #2541, opened 2023-10-23
Update osmesa in CI

djhoese:
A continuation/addition to #2540 to debug osmesa CI issues.


djhoese:
Ok, I'm just dumb. @larsoner FYI the failure in the other PR was because the path changed to the conda environment so the osmesa env var was pointing to the wrong location (one that didn't exist). Since I'm doing this already, let's see how 64-bit osmesa goes...

PR #2545, opened 2023-10-31
Add ivec2, ivec3, ivec4 to ATYPEINFO in glir.py

hmaarrfk:
I was playing around with sampling texels and these were required in my fragment shader.

### DIFF
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -1045,6 +1045,9 @@ class GlirProgram(GlirObject):
         'vec2': (2, gl.GL_FLOAT, np.float32),
         'vec3': (3, gl.GL_FLOAT, np.float32),
         'vec4': (4, gl.GL_FLOAT, np.float32),
+        'ivec2': (2, gl.GL_INT, np.int32),
+        'ivec3': (3, gl.GL_INT, np.int32),
+        'ivec4': (4, gl.GL_INT, np.int32),
         'int': (1, gl.GL_INT, np.int32),
         'bool': (1, gl.GL_BOOL, np.int32)
     }


djhoese:
Note that ivec is not supported by GL ES 2.0 (from my googling). So while I'm fine merging this, just wanted to point out to anyone reading this that if they are targeting old machines there is a small small small chance that their GPU drivers won't understand ivec.

djhoese:
Thanks @hmaarrfk!

hmaarrfk:
what did it get replaced by?

hmaarrfk:
They still seem to be in there...
https://www.khronos.org/opengles/sdk/docs/reference_cards/OpenGL-ES-2_0-Reference-card.pdf

hmaarrfk:
correct, i've been having trouble getting opengl 3.0 to work on OSX. not sure if you were ever able to (i saw some posts on github where it hinted you were able to get it to work on some operating system)

If you have a minimum reproducible code OpenGL 3.0 or 3.2 that would be greatly appreciated.

djhoese:
There are a couple examples that use OpenGL 3.x. One is the geometry shaders:

https://github.com/vispy/vispy/blob/main/examples/gloo/geometry_shader.py

But this is not processed as part of the docs generation because CI virtual GPUs can't handle it from what I remember.

hmaarrfk:
And this one works on OSX as well? I guess they can hit `#version 330` but not `#version 130`? (I was trying to use `#version 130`)

djhoese:
I never got OpenGL versions that required a "core" profile to work on OSX. I haven't had a mac for 5 or so years though so no idea what its like now. Can't imagine it's better than the last time I tried (Apple's feelings towards OpenGL). Put another way, if it wasn't supported with a OpenGL extension module that worked with the compatibility/default profile of the drivers then I didn't get it to work. From what I remember that was mostly "fancier" newer shader functionality.

hmaarrfk:
ok thanks for the tips. 

I'm going to have to write "two shaders" one for the newer OpenGLs. One for the old OpenGL. OSX compatibility isn't a huge deal for us, it is just to have sometimes.

PR #2550, opened 2023-11-20
Higher order transfer functions for volume rendering

aganders3:
As mentioned in https://github.com/vispy/vispy/pull/2539#issuecomment-1792813896, I have been playing around with adding some features to volume rendering to allow the use of more advanced color features, specifically targeting 2D transfer functions (2D TF), mostly motivated by [this paper](https://liu.diva-portal.org/smash/get/diva2:954156/FULLTEXT01) and similar. I think there’s a way this _could_ take the place of that work, but I’m not sure that it _should_. If we want both though, there should be a way to implement that feature in terms of this one (for example with `ComplexMagnitudeTF`, `ComplexPhaseTF`, etc. classes).

I believe the way I have implemented this is backwards compatible for the affected rendering methods (MIP, attenuated MIP, translucent, and additive). I can imaging including some generally useful transfer functions out of the box, but much like the already-scriptable colormaps this is pretty powerful.

I have a couple concerns/questions with my design/implementation so far (hence the Draft PR):
1. I am not sure I am adding snippets to the shader in the most correct/optimal way, but I was having trouble defining functions that could still use other defined functions.
2. I am not sure the implementation for MIP makes semantic sense given the parameters for the TF. I appreciate feedback on what arguments should be passed to the TF function.

I guess this is best illustrated by examples - these come from the included `examples/scene/volume_2dtf.py`.

This shows a volume rendering with an interactive 2D TF used for some segmentation. In the bottom view you see the 2D transfer function overlaid on the 2D (value, gradient magnitude) histogram. Just changing to the 2D TF is pretty subtle for this dataset - I will probably keep looking to try to show something more interesting here.
Screenshot 2023-11-17 at 5 04 27 PM copy

Here is a MIP colored by depth of the found max val (see [this figure](https://www.nature.com/articles/s41592-022-01417-2/figures/5) for a real-world example of this technique):
Screenshot 2023-11-17 at 5 02 37 PM copy

### DIFF
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -364,6 +364,7 @@ _RAYCASTING_SETUP_PLANE = """
     vec3 intersection = intersectLinePlane(v_position.xyz, view_ray,
                                            u_plane_position, u_plane_normal);
     vec3 depth_origin = intersection - normalize(u_plane_normal) * u_plane_thickness / 2.0;
+    vec3 depth_plane_normal = u_plane_normal;
     float max_depth = u_plane_thickness;
 
     // and texture coordinates


examples/scene/volume_2dtf.py
```diff
@@ -0,0 +1,217 @@
...
+    sample a 2D transfer function.
+    """
+
+    glsl_tf = """\
+    vec4 applyTransferFunction(vec4 color, vec3 loc, vec3 step) {
+        vec4 hue = applyColormap(length(loc));
+        vec4 final_color = vec4(color.r * hue.rgb, 1.0);
+        return final_color;
+    }
+    """
```
brisvag, line 92: @royerloic will love this, we couldn't figure it out last time but we just needed to use `loc` :P

examples/scene/volume_2dtf.py
```diff
@@ -0,0 +1,217 @@
...
+    glsl_tf = """\
+    vec4 applyTransferFunction(vec4 color, vec3 loc, vec3 step) {
+        step = step / u_relative_step_size;
+        vec4 prev;
+        vec4 next;
+        // calculate normal vector from gradient
+        vec3 N; // normal
+        prev = $get_data(loc + vec3(-step.x, 0.0, 0.0));
+        next = $get_data(loc + vec3(step.x, 0.0, 0.0));
+        N[0] = colorToVal(prev) - colorToVal(next);
```
brisvag, line 61: just for consistency I'd use `normal.x` and so on instead of `N[0]`, cause I thought it was an array at first.

brisvag (commented):
First pass: I don't quite understand how the gradient one works (I have some reading to do), but I get the gist and this looks awesome :) I think having something built in the `Volume` that allows us to pass a snippet of glsl like the `depthcolorTF` is awesome; however, I'm a bit skeptical about adding the LUT logic in the base class... this seems like it should be handled entirely by the transfer function, like colormaps do.

examples/scene/volume_2dtf.py
aganders3: Thanks - this is a good idea. To be honest I just copied this code from the iso shader (`calculateColor`) but I agree this would be more readable. It also could be just factored out as finding the gradient at a point may be otherwise useful.

examples/scene/volume_2dtf.py
aganders3: Initially I was just working on the 2D LUT stuff, but was inspired by a conversation with him to add this feature. It was very simple on this branch, which felt like a good sign! Note this is what I called out in the PR description though - in the MIP shader it's actually passing in `max_loc_tex - start_loc` for this parameter. I'm not sure if that's the best though. An alternative would be to pass in `start_loc` separately, but I also don't want to pass in a bunch of generally unused parameters.

brisvag: I see... well, I think it's fair to pass as many parameters as we think might be useful :shrug: It's easy to not use them.

Alternatively, glsl simply accessing stuff from the outer scope generally works (though it's uglier).

brisvag:
I do think providing the LUT logic out-of-the-box is nice, so I will try instead to make this a subclass of `BaseTransferFunction` (for example `TextureSamplingTF`). It's easy to just drop it as well and put that in the example instead.

Yep, I like this!

> One other concern I have is that I think there is now perhaps a redundant texture lookup in some of the shaders. I want to see if it can be avoided or make sure that's not a big performance regression.

We are *extremely* inefficient with texture lookups with isosurface. Things get exponentially worse if you use non-nearest, non-linear sampling (cause those are actually re-sampling the texture multiple times and doing calculations on the results).
This is quite bad in 3D raycasting, where the same texture position might be sampled dozens of time in a single render call... Unless the compiler can automatically solve this somehow, to me it seems super hard to solve without rewriting from scratch. I just don't know the hardware and driver internals enough to know ^^'

aganders3:
That sounds like a separate problem I'd be happy to investigate at some time :). What I meant was I want to make sure _this_ feature doesn't make it significantly worse (right now this doesn't touch the iso method).

brisvag (commented):
@aganders3 any plans to push this forward? I dont' remember what was the status exactly...

aganders3:
I'd like to! I think the status was that I need to:
* possibly refactor the gradient calculation into a function (shader)
* set comprehensive inputs for the transfer function (shader)
* move the LUT functionality out of the `BaseTransferFunction` into a subclass (or maybe even just an example?)
* maybe do some rough benchmarking to make sure this is not a performance hit when using the `BaseTransferFunction` only

brisvag:
No pressure. I was just going through PRs and seeing what might be worth getting in for a release before napari so we can get some goodies downstream :)

aganders3:
I picked this up again today and made most of the changes discussed.

One concern I have is that this is pretty similar to what the "snippets" in the volume rendering modes are already doing. Of course with this more exposed to the user. The problem though is that it feels like the transfer function will be closely tied to the rendering mode. So for example you can get some weird results by changing the `mode` without changing (or resetting) the transfer function.

I still like this feature - as you said:
> I think having something built in the Volume that allows us to pass a snippet of glsl like the depthcolorTF is awesome

Maybe it's okay for this to be limited to certain rendering modes? Maybe it's better for user to just subclass the `Volume` visual and write their own snippets for a new mode?

jni:
Super great to see this activity — I missed it the first time around. Will be super-awesome to have this in and use it in napari!!!

aganders3:
There's an issue with the depth-color TF here. Basically the color is set by the depth within the texture, but ignores the distance between the camera and the texture surface. So the edges/corners of the cube can cast a hue on the image.

I have tried to sort this out with mixed success, and I'm open to suggestions. In the end the value needs to be normalized, but using the distance along the full potential ray length (`nearpos` to `farpos`) generally doesn't make sense because the clipping planes are so far apart by default. Maybe there needs to be a guess and a parameter to scale it?

brisvag:
> I have tried to sort this out with mixed success, and I'm open to suggestions. In the end the value needs to be normalized, but using the distance along the full potential ray length (`nearpos` to `farpos`) generally doesn't make sense because the clipping planes are so far apart by default. Maybe there needs to be a guess and a parameter to scale it?

I distinctly remember hitting this problem when playing with this idea a while back with @royerloic... I don't think we came up with a good idea :sweat:

aganders3:
I have an idea I think is working.

I first calculate the intersection of the ray with a bounding sphere of the cube/volume. This is pretty cheap, and can be done in raycasting setup. It might even offer a performance benefit because we can discard if there is no intersection at all (or if intersections are behind the origin) instead of having to find `nsteps < 1`.

With this intersection point we can get the depth of the point inside a _sphere_, instead of the cube, which avoids the uneven shading.

depth-color-attenuated-mip

brisvag:
Wooo that's awesome! Makes perfect sense :)

aganders3:
Test failures are for "plane" mode raycasting. I haven't really tested any of this that way but should be able to fix it.

aganders3:
Okay I think this is ready for some more eyes and discussion.

> One concern I have is that this is pretty similar to what the "snippets" in the volume rendering modes are already doing. Of course with this more exposed to the user. The problem though is that it feels like the transfer function will be closely tied to the rendering mode. So for example you can get some weird results by changing the mode without changing (or resetting) the transfer function.

I'm still concerned this is not the right API so happy to get feedback on that. Instead I could see this just being a couple new rendering modes. I'm at least very happy with the color depth MIP mode.

Also, does anyone know where I can find this "tooth" dataset? I see it in a ton of papers, but rarely cited. I think it would better show off the 2D transfer functions. I think it may originally be from GE?
https://www.researchgate.net/figure/Volume-rendering-of-the-tooth-dataset-with-a-2D-TF_fig9_346226832
Volume-rendering-of-the-tooth-dataset-with-a-2D-TF

vispy/color/transfer_function.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class BaseTransferFunction:
+    """Class representing a 2D transfer function.
+
+    Parameters
+    ----------
+    lut : array-like
+        Lookup table for the transfer function. This may be None (default) if
+        the transfer function can compute colors directly.
```
brisvag, line 13: this should be in the subclass right?

vispy/visuals/volume.py
```diff
@@ -296,6 +333,14 @@
 """  # noqa
 
 _RAYCASTING_SETUP_VOLUME = """
+    vec3 center = u_shape / 2.0;
+    float radius = length(u_shape) / 2.0;
+    bool hit_sphere;
+    vec3 sphere_intersection = intersectLineSphere(nearpos, view_ray, center, radius, hit_sphere);
+    if (!hit_sphere) { discard; }
```
brisvag, line 340: neat!

vispy/visuals/volume.py
```diff
@@ -382,11 +430,11 @@
         // Refine search for max value, but only if anything was found
         if ( maxi > -1 ) {
             // Calculate starting location of ray for sampling
+            vec3 loc = start_loc + step * maxi;
```
brisvag, line 433: can you elaborate on this change?

vispy/visuals/volume.py
```diff
@@ -398,7 +446,9 @@
                 loc += small_step;
             }
             frag_depth_point = max_loc_tex * u_shape;
-            gl_FragColor = applyColormap(maxval);
+            vec4 max_color = $get_data(max_loc_tex);
```
brisvag, line 449: why is this not the same as `maxval`?

examples/scene/volume_depth_color_tf.py
```diff
@@ -0,0 +1,128 @@
...
+    @property
+    def depth_color_scale(self):
+        return self._depth_color_scale
+
+    @depth_color_scale.setter
+    def depth_color_scale(self, value):
+        self._depth_color_scale = value
+
+    def get_uniforms(self):
+        return {"u_depth_color_scale": self.depth_color_scale}
```
brisvag, line 73: I love this approach, makes it intuitive to work with complex stuff!

vispy/color/transfer_function.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np
+
+import vispy.gloo
+
+
+class BaseTransferFunction:
```
brisvag, line 6: Since this is a user-facing object that people are supposed to subclass, I think we need to explain some stuff in the docstring:
- how to modify it (change the glsl code, update uniforms)
- what needs to stay
- explain a bit the arguments passed in to `applyTransferFunction`

brisvag (commented):
I love the way this is shaping up! Very few comments on my side, looks great.

examples/scene/volume_depth_color_tf.py
aganders3: Thanks! Now that I look again we may want to add something like this. It doesn't matter at the moment but would if for some reason the `BaseTransferFunction` ever added a uniform.

`return super().get_uniforms() | {"u_depth_color_scale": self.depth_color_scale}`

vispy/visuals/volume.py
aganders3: The difference is `max_color` is a `vec4`, but `val` is `float`. I think I can refactor this because I don't like this extra texture lookup.

vispy/visuals/volume.py
```diff
@@ -382,11 +430,11 @@
         // Refine search for max value, but only if anything was found
         if ( maxi > -1 ) {
             // Calculate starting location of ray for sampling
+            vec3 loc = start_loc + step * maxi;
             vec3 start_loc_refine = start_loc + step * (float(maxi) - 0.5);
-            loc = start_loc_refine;
 
             // Variables to keep track of current value and where max was encountered
-            vec3 max_loc_tex = start_loc_refine;
+            vec3 max_loc_tex = loc;
```
aganders3, line 437: I think this is an edge case bug in the existing shader and it was causing a test failure when I re-sampled the value below (which, again, I'd like to avoid).

The problem is `start_loc_refine` is not the location where the `maxval` was found in `in_loop`, it's half a step back. Then in the refinement loop, this value is only updated if `val > maxval`. The strict comparison (`>`, not `>=`) means this is possibly never hit, and `max_loc_tex` is never updated. In the end this means `max_loc_tex` (used to set the frag depth) is half a step back from the actual location of `maxval`.

The change here is one way to resolve this. Perhaps the better way is re-initialize `maxval` here.

examples/scene/volume_depth_color_tf.py
aganders3: Hm, I still like having this but I'm realizing I can achieve what I intended by setting `vol.gamma`, since this transfer function is using `applyColormap`.

vispy/visuals/volume.py
aganders3: Weird, my response to this ended up in https://github.com/vispy/vispy/pull/2550/files/f858c92ab13fb572b2ec070e436c08ea5b1f0944#r1681378506

vispy/visuals/volume.py
```diff
@@ -224,6 +231,36 @@
...
+        hit_sphere = false;
+        return vec3(0.0);
+    }
+    // it's okay one of the intersections is behind the camera
+    // we still want the depth from the "front" of the sphere
+    hit_sphere = true;
+    return origin + min(t0, t1) * direction;
+}
+
+$def_tf
```
brisvag, line 262: In principle this is not really needed? You can just use `$apply_transfer_function` in the place where it's needed and the definition will be placed where appropriate... though it might not work if the snippet has multiple functions inside... It's fine like this if this gets too tricky :P But for reference, look at how `get_data` is defined.

vispy/visuals/volume.py
```diff
@@ -366,11 +414,13 @@
 _MIP_SNIPPETS = dict(
     before_loop="""
         float maxval = u_mip_cutoff; // The maximum encountered value
+        vec4 maxcolor = vec4(0.0); // The 'color' of the maximum encountered value
         int maxi = -1;  // Where the maximum value was encountered
         """,
     in_loop="""
         if ( val > maxval ) {
             maxval = val;
+            maxcolor = color;
```
brisvag, line 423: Yeah I like version much better!

vispy/visuals/volume.py
brisvag: I think you're right. If you see how `frag_depth_point` is initialized in `_RYCASTING_SETUP_VOLUME`, you see it's using `front`, which has that 0.5 offset included. Unless it's wrong there? But I think your point makes sense. Is there a way to check this with some test like I did originally in https://github.com/vispy/vispy/pull/2239?

aganders3: I will try - I first need to understand what you did in #2239, haha.

This will only affect the depth buffer, not the color buffer, right? So I think it needs a test with two visuals. Maybe a volume and an image visual where the volume should end up in front, but does not. 🤔

vispy/visuals/volume.py
aganders3: I admit I don't understand all the templating options.

I'm not sure the method you're suggesting would allow you to do things like add new uniforms. Or, like you mention, multiple functions.Maybe these are not such important features though - you could always re-define an update the instance. Maybe that's not quite as performant.

vispy/visuals/volume.py
brisvag: > I will try - I first need to understand what you did in https://github.com/vispy/vispy/pull/2239, haha.

I don't remember much, sorry xD

> This will only affect the depth buffer, not the color buffer, right?

Yep, exactly. I'm assuming they will have to be within 0.5 texels from one another for the error to show up?

vispy/visuals/volume.py
brisvag: Yeah, definitely not blocking :)

vispy/visuals/volume.py
aganders3: Here's the test I came up with, but I can't quite convince myself it's correct.

```python
@requires_pyopengl()
@requires_application()
def test_mip_depth():
    N = 101
    center = ((N - 1) / 2,) * 3
    sphere_volume = np.meshgrid(
        *[np.linspace(-1, 1, N)] * 3,
    )
    sphere_volume = (np.sqrt(3) - np.linalg.norm(sphere_volume, axis=0)) / np.sqrt(3)
    sphere_volume = sphere_volume.astype(np.float32)

    with TestingCanvas(size=(N, N)) as canvas:
        view.camera = "arcball"
        view.camera.fov = 0
        view.camera.depth_value = 1.5 * N
        view.camera.center = center
        view.camera.scale_factor = N

        # render a mip of a soft sphere
        # the maximum of every voxel should be at the center
        scene.visuals.Volume(
            sphere_volume,
            clim=(0, 1),
            cmap="reds",
            method='attenuated_mip',
            attenuation=0.0,
            interpolation='nearest',
            relative_step_size=0.01,
            parent=view.scene,
        )

        # add a solid green plane at the center
        plane = scene.visuals.Plane(
            width=N + 10,
            height=N + 10,
            color=(0, 1.0, 0, 1.0),
            direction="+y",
            parent=view.scene,
        )
        plane.set_transform(
            "st",
            translate=center,
        )

        # no pixels should be all green (showing the plane)
        assert not np.any(
            np.all(canvas.render() == [0, 255, 0, 255], axis=-1)
        )
```

With the current code I get some green pixels and the test fails. I can make the test pass if I change this line:
https://github.com/vispy/vispy/blob/f858c92ab13fb572b2ec070e436c08ea5b1f0944/vispy/visuals/volume.py#L483
to
```
             frag_depth_point = max_loc_tex * u_shape - 0.5;
```

I think this change is what the end of [this comment](https://github.com/vispy/vispy/pull/2239#issuecomment-949960620) was suggesting, but seems to have not been included. 

Which is correct, though, and how can I know? Perhaps we should _expect_ the z-fighting with the plane at the center.

I also was hoping to see the render 100% green if I set the plane forward even a bit, but to make it all green I need to use -10 or so.

brisvag: Something's off, yeah. I made an interactive version of your test, and it's weird to me. Not only that, but I think I'm getting some issues due to the bounding sphere trick you used. Basically, I took the code above and made small changes to make it interactive and more obvious (5 pixels istead of 100):

```py
import numpy as np
from vispy.scene import SceneCanvas, Volume, Plane
N = 5
center = ((N - 1) / 2,) * 3
sphere_volume = np.meshgrid(
    *[np.linspace(-1, 1, N)] * 3,
)
sphere_volume = (np.sqrt(3) - np.linalg.norm(sphere_volume, axis=0)) / np.sqrt(3)
sphere_volume = sphere_volume.astype(np.float32)

canvas = SceneCanvas(size=(N*100, N*100), show=True)
view.camera = "arcball"
view.camera.fov = 0
view.camera.depth_value = 1.5 * N
view.camera.center = center
view.camera.scale_factor = N

# render a mip of a soft sphere
# the maximum of every voxel should be at the center
Volume(
    sphere_volume,
    clim=(0, 1),
    cmap="reds",
    method='attenuated_mip',
    attenuation=0.0,
    interpolation='nearest',
    relative_step_size=0.01,
    parent=view.scene,
)

# add a solid green plane at the center
plane = Plane(
    width=N + 1,
    height=N + 1,
    color=(0, 1.0, 0, 1.0),
    direction="+y",
    parent=view.scene,
)
plane.set_transform(
    "st",
    translate=center,
)
```

My idea was that this way you should see some z-fighting (as you said, expected). But when you move the plane even slightly, the fighting should be gone. Instead I see some weird glimpes of the plane between the voxels, which I can't quite explain.

Also, you see that round "cut" at the edges of the cube? Suspicious!
image

aganders3: That is a bit suspicious, and to me very surprising!

I'm leaving today for vacation but will try to pick this up again next week.

brisvag: No rush! Enjoy your holidays :)

aganders3: I'm tempted to push this `frag_depth_point` into another issue/PR as in the end it's related but not caused by this work.

brisvag: Yep, good idea. Sorry, sometimes I get stuck in "solve everything" mode :P The sphere fix loos great now!

brisvag (commented):
Great work! I was about to say "ready to merge", but I decided to test in napari and I'm seeing some stuff:


https://github.com/user-attachments/assets/2df28971-9a4b-47f9-a028-461e7d5f856e

There's something wrong happening with discarding it seems?

aganders3:
Ooh yikes! 🤦

aganders3:
Okay I can reproduce without napari at least - it's caused by using an orthographic camera. Any FOV > 0 seems to be okay. Now I am stretching my projection geometry knowledge a bit, but should be fixable.

aganders3:
I have fixed the severe discard artifacts by changing the origin of the intersection ray to `front` instead of `nearpos`. I think `nearpos` is somehow divergent with the orthographic projection, and was causing issues when finding the sphere intersection.

Looking at the cell data in napari I also now notice an artifact from the sphere, particularly in orthographic projection (left). It's much less noticeable with FOV=60 (right).


The depth-color effect can also be a little awkward when the data is not a cube, but I still think it's pretty good. Perhaps it could be improved with a bounding ellipsoid? I think that still presents an issue with normalizing the depth.

Another note - are you very familiar with the 'plane' raycasting mode? I have tried playing with it in my example and in napari, but in napari it looks quite "flat" by comparison, so I will probably try to debug that a bit more.

brisvag:
> I have fixed the severe discard artifacts by changing the origin of the intersection ray to `front` instead of `nearpos`. I think `nearpos` is somehow divergent with the orthographic projection, and was causing issues when finding the sphere intersection.

Interesting... would be good to know what's wrong with nearpos, but I guess it makes sense that `front` works.

> The depth-color effect can also be a little awkward when the data is not a cube, but I still think it's pretty good. Perhaps it could be improved with a bounding ellipsoid? I think that still presents an issue with normalizing the depth.

I don't think ellipsoid would be better... but I'm also not super sure what I think about the sphere. This effect on the left is unexppected; at least the "line" effects if we don't use a sphere seem more intuitive to me... but ugly ^^'

> Another note - are you very familiar with the 'plane' raycasting mode? I have tried playing with it in my example and in napari, but in napari it looks quite "flat" by comparison, so I will probably try to debug that a bit more.

Not *very* familiar anymore, but yeah I worked with @alisterburt on it. The big difference with normal raycasting is that we cast the ray from a point above the plane to a point below, perpendicularly to the plane, instead of casting the ray from the eye.

aganders3:
No worries - unfortunately my time for this is going to start running a bit thin. 😢 

I tried one more idea for calculating depth by finding the intersection with a plane placed tangent to the bounding sphere at the front of the volume, parallel to the near clipping plane.


This generally works! It looks pretty good to me. Funnily enough with the membrane volume it now kind of swaps the artifact for ortho (left) and perspective (right) projections 😆. I think that ends up being more intuitive though, especially once you start interacting with the data. Maybe there's a way to put this plane into the projected space so it is also warped accordingly?


The changes so far are a little hacky but I can clean them up and push them so you can try it (I also have a script so you can play with it in napari).

brisvag:
The reason it swaps is that you need to use the projection of that distance onto the normal of the plane, I think. These 2 are the same in ortho, but different in perspective.

aganders3:
Yeah I'll play with it a bit more as it seems like this should work, so maybe I've got the math wrong somewhere.

aganders3:
Hopefully nth time is the charm here. I calculate the normal for the depth plane now in the vertex shader, but I'm sure there's a way to do it on the CPU and pass it as a uniform.

Here's a script if you want to play with this in napari - of course you need to install vispy from this branch first.

```python
"""
Depth color rendering
=========

This example demonstrates how to create a custom transfer function that colors a MIP based on the
depth of the maximum value.

.. tags:: visualization-advanced
"""


from skimage import data
from vispy.color import BaseTransferFunction

import napari


class DepthColorTF(BaseTransferFunction):
    """Transfer function that colors the volume based on the depth of the maximum value."""

    glsl_tf = """\
    vec4 applyTransferFunction(vec4 color, vec3 loc, vec3 step, vec3 depth_origin, vec3
                               depth_plane_normal, float max_depth) {
        float depth = dot(loc - depth_origin, depth_plane_normal);
        depth = smoothstep(0.0, 1.0, depth / max_depth);
        depth = depth * (clim.y - clim.x) + clim.x;
        vec4 hue = applyColormap(depth);
        vec4 final_color = vec4(color.r * hue.rgb, 1.0);
        return final_color;
    }
    """


cells3d = data.cells3d()

viewer = napari.view_image(
    cells3d, channel_axis=1, name=['membranes', 'nuclei']
)

viewer.dims.ndisplay = 3
qt_viewer = viewer.window.qt_viewer

for layer in viewer.layers:
    layer.colormap = 'jet'
    layer.rendering = 'attenuated_mip'
    visual = qt_viewer.layer_to_visual[layer]
    visual.node.transfer_function = DepthColorTF()

if __name__ == '__main__':
    napari.run()
```

brisvag:
Nice! It looks great :)

Sorry for continuously being the party pooper... but playing with this now makes me realize that the main reason why people usually wanted this is to have an at-a-glance idea of what's front and what's back, which is completely lost when looking at this volume through the thin side, because of the sphere being so big compared to that side's thickness.

I think either going the ellipse route (hard, and probably has multiple solutions?) or adding some way to adjust the "depth contrast limits" (a bit more finnicky for users) would be necessary to make this really work. What do you think?

aganders3:
Yeah, I noticed this as well. You can already adjust a bit by changing gamma, but I don't think it's sufficient. A previous implementation I had included a uniform in the `DepthColorTF` to scale it. I have some half-ideas to try to adjust this automatically I can play with.

Adding all this specifically for depth coloring still brings me back to the idea that perhaps this should just be an additional rendering mode instead of the flexible transfer-function approach here.

Either way it's good to work all this out and I think something valuable can come from it. You're not party-pooping at all I'm happy to iterate on this!

aganders3:
Thinking about it a bit more - it seems like any attempt to solve this issue (auto-scaling the depth contrast limits) could lead to similar artifacts it's taken all this work to avoid. I think then the best option may be to make it something the user can configure. I don't think that should be too hard, and we might be able to wire up (in napari, for example) a way to make a good guess for a given view.

brisvag:
Actually, deriving this thing but for 3D might just not be too hard and solve our problem?
https://math.stackexchange.com/questions/623922/ellipse-bounding-rectangle

aganders3:
Hm, one thing I like in the current version is that while manipulating the camera, everything at the same depth stays the same color. This means you lose depth perception when viewing through the thin-axis, so it's not the best for a still frame, but it feels intuitive when manipulating live. The still-frame appearance could then be tweaked with a control (depth contrast limits).

I think normalizing to the "current" maximum depth as the view changes might lose some sense of the absolute depth of the volume as you move the camera.

> Actually, deriving this thing but for 3D might just not be too hard and solve our problem?

I promise I am not (only) trying to avoid actually doing this (because I do think it will be tricky), but I am trying to think through how it would look and a little skeptical this will solve the problems without introducing new ones!

I guess the solution is perhaps to attempt the math and see how it looks/feels, and perhaps offer both as options. Unfortunately as I said I will have limited time for this right now, but will tinker on it in my free time :).

brisvag:
You might be right, yeah... I think that maybe having a way to (manually) control an offset for this max/min depth would be the easiest way to ensure the shape of the volume does not limit in what cases this can be used!

aganders3:
Okay I hacked an automatic depth-limit adjustment into a napari script to at least see what it feels like. This could surely be more efficiently implemented in vispy itself.

```python
"""
Depth color rendering
=========

This example demonstrates how to create a custom transfer function that colors a MIP based on the
depth of the maximum value.

.. tags:: visualization-advanced
"""


import magicgui
import numpy as np
from skimage import data
from vispy.color import BaseTransferFunction

import napari


class DepthColorTF(BaseTransferFunction):
    """Transfer function that colors the volume based on the depth of the maximum value."""

    glsl_tf = """\
    uniform vec2 depth_clim;
    vec4 applyTransferFunction(vec4 color, vec3 loc, vec3 step, vec3 depth_origin, vec3
                               depth_plane_normal, float max_depth) {
        float depth = dot(loc - depth_origin, depth_plane_normal);
        depth = smoothstep(0.0, 1.0, depth / max_depth);
        depth = (depth - depth_clim.x) / (depth_clim.y - depth_clim.x);
        depth = depth * (clim.y - clim.x) + clim.x;
        vec4 hue = applyColormap(depth);
        float intensity = colorToVal(color);
        intensity = clamp(intensity, min(clim.x, clim.y), max(clim.x, clim.y));
        intensity = (intensity - clim.x) / (clim.y - clim.x);
        vec4 final_color = vec4(intensity * hue.rgb, 1.0);
        return final_color;
    }
    """

    def __init__(self, depth_clim=(0.3, 0.6)):
        self.depth_clim = depth_clim

    def get_uniforms(self):
        return {'depth_clim': self.depth_clim}


cells3d = data.cells3d()

viewer = napari.view_image(
    cells3d, channel_axis=1, name=['membranes', 'nuclei']
)

viewer.dims.ndisplay = 3
qt_viewer = viewer.window.qt_viewer

for layer in viewer.layers:
    layer.colormap = 'jet'
    layer.rendering = 'attenuated_mip'
    visual = qt_viewer.layer_to_visual[layer]
    visual.node.transfer_function = DepthColorTF()


def update_depth_clim(clim: tuple[float, float]):
    for layer in viewer.layers:
        visual = qt_viewer.layer_to_visual[layer]
        visual.node.transfer_function.depth_clim = (
            clim[0] / 100,
            clim[1] / 100,
        )
        visual.node._need_tf_update = True
        visual.node.update()


@viewer.camera.events.angles.connect
def auto_depth_clim():
    # get the min and max depth values for each layer
    minmax = np.dot(
        viewer._sliced_extent_world_augmented,
        np.abs(viewer.camera.view_direction),
    )
    depth = np.ptp(minmax)
    max_depth = np.linalg.norm(
        viewer._sliced_extent_world_augmented[1]
        - viewer._sliced_extent_world_augmented[0]
    )
    print(minmax, depth, max_depth)
    width = 100 * depth / max_depth
    center = 50
    update_depth_clim((center - width / 2, center + width / 2))


@magicgui.magicgui(auto_call=True)
def enable_auto_depth(enable: bool = True):
    if not enable and auto_depth_clim in viewer.camera.events.angles.callbacks:
        viewer.camera.events.angles.disconnect(auto_depth_clim)
    elif enable and auto_depth_clim not in viewer.camera.events.angles.callbacks:
        auto_depth_clim()
        viewer.camera.events.angles.connect(auto_depth_clim)


viewer.window.add_dock_widget(enable_auto_depth)


if __name__ == '__main__':
    napari.run()
```

brisvag:
Awesome :)

aganders3:
I re-wrote the auto-depth-contrast to be part of the transfer function, so at least I know it can be done. A few thoughts here:
1. The calculation could actually go in the vertex shader, but in fact it doesn't need to be on the GPU at all (it only needs to change when the camera moves).
2. As a drawback from the previous version, you can't get the auto-calculated limits back out of the shader.
3. This doesn't know anything about what's _in_ the volume, for example if you have a volume that is half zeros, I think you'll want manual control over these limits anyway. This could be another reason to keep the calculation on the CPU.
4. If it's in the shader, there's no way to set it consistently across multiple volumes (in the above example I use the world extent rather than the extent of each layer).

So I think it may be okay to keep this part out of vispy?

```python
"""
Depth color rendering
=========

This example demonstrates how to create a custom transfer function that colors a MIP based on the
depth of the maximum value.

.. tags:: visualization-advanced
"""


import magicgui
from skimage import data
from vispy.color import BaseTransferFunction

import napari


class DepthColorTF(BaseTransferFunction):
    """Transfer function that colors the volume based on the depth of the maximum value."""

    glsl_tf = """\
    uniform vec2 u_depth_clim;
    uniform bool u_auto_depth_clim;

    vec4 applyTransferFunction(vec4 color, vec3 loc, vec3 step, vec3 depth_origin, vec3
                               depth_plane_normal, float max_depth) {

        // get the current limits of the volume in the depth direction
        float volume_limit_min = dot(vec3(-0.5), abs(depth_plane_normal));
        float volume_limit_max = dot(u_shape - 0.5, abs(depth_plane_normal));
        float current_total_depth = abs(volume_limit_max - volume_limit_min);

        float width = current_total_depth / max_depth;
        float center = 0.5;
        float depth = dot(loc - depth_origin, depth_plane_normal);
        depth = smoothstep(0.0, 1.0, depth / max_depth);

        vec2 depth_clim;
        if (u_auto_depth_clim) {
            depth_clim = vec2(center - width / 2, center + width / 2);
        } else {
            depth_clim = u_depth_clim;
        }

        depth = (depth - depth_clim.x) / (depth_clim.y - depth_clim.x);
        depth = depth * (clim.y - clim.x) + clim.x;
        vec4 hue = applyColormap(depth);
        float intensity = colorToVal(color);
        intensity = clamp(intensity, min(clim.x, clim.y), max(clim.x, clim.y));
        intensity = (intensity - clim.x) / (clim.y - clim.x);
        vec4 final_color = vec4(intensity * hue.rgb, 1.0);
        return final_color;
    }
    """

    def __init__(self, depth_clim=(0.0, 1.0), auto_depth_clim=False):
        super().__init__()
        self._depth_clim = depth_clim
        self._auto_depth_clim = auto_depth_clim

    @property
    def depth_clim(self):
        return self._depth_clim

    @depth_clim.setter
    def depth_clim(self, clim):
        inverse = clim[1] < clim[0]
        lo = max(min(clim), 0.0)
        hi = min(max(clim), 1.0)
        if lo == hi:
            raise ValueError(
                f'clim values must be distinct after clipping: {clim=}'
            )
        self._depth_clim = (lo, hi) if not inverse else (hi, lo)

    @property
    def auto_depth_clim(self):
        return self._auto_depth_clim

    @auto_depth_clim.setter
    def auto_depth_clim(self, enable):
        self._auto_depth_clim = enable

    def get_uniforms(self):
        return {
            'u_depth_clim': self._depth_clim,
            'u_auto_depth_clim': self._auto_depth_clim,
        }


cells3d = data.cells3d()

viewer = napari.view_image(
    cells3d, channel_axis=1, name=['membranes', 'nuclei']
)

viewer.dims.ndisplay = 3
qt_viewer = viewer.window.qt_viewer

for layer in viewer.layers:
    layer.colormap = 'jet'
    layer.rendering = 'attenuated_mip'
    visual = qt_viewer.layer_to_visual[layer]
    visual.node.transfer_function = DepthColorTF()


@magicgui.magicgui(auto_call=True)
def start_stop_auto_depth(enable: bool = False):
    for layer in viewer.layers:
        visual = qt_viewer.layer_to_visual[layer]
        visual.node.transfer_function.auto_depth_clim = enable
        visual.node._need_tf_update = True
        visual.node.update()


viewer.window.add_dock_widget(start_stop_auto_depth, name='auto depth clim')


if __name__ == '__main__':
    napari.run()
```

------

> PS: fun fact: I just re-learned about np.ptp, but apparently it's somehow slower than arr.max() - arr.min()... Though for big arrays the difference is small...

Ahh, that's annoying but I guess because it's not strictly the same and the subtraction is a ufunc. I like the brevity of `ptp` but maybe `arr.max() - arr.min()` is actually clearer anyway.

brisvag:
Yeah I agree on not having autocontrast in the shader, but just the limits as uniforms, especially because of the last 2 points!

brisvag:
Btw I'm gonna try to get to this one in the coming days, I hope I remember enough to understand it 😅 might need your help still @aganders3 !

aganders3:
That's great! Let me know if you want to arrange some time to talk through it. We might be able to come up with some less-disruptive way to get most of the value. I think just adding a depth-color-MIP (+attenuated MIP) would be huge!

PR #2554, opened 2023-12-01
Fix SurfacePlotVisual not allowing 2D x and y inputs

ullmannJan:
# Bug
The test for len(y) != z.shape[1] does not work for 2-dimensional y.

# Fix
It is now possible to use 2-dimensional x and y data where x.shape[0] != y.shape[1] by differantiating the ndim==1 and ndim==2 cases in the error handling.


djhoese:
So does that mean it never worked with 2D data at all?

ullmannJan:
As far as I see it, yes. Except you you have n x n matrices

djhoese:
Would it be possible for you to add a test case for this in the unit tests?

ullmannJan:
I did create a test. I, however, do not understand why the other tests (e.g. test_spectogram.py) do not need a view to generate an image. My installation only creates empty canvases when not adding a view. Nevertheless, I hope this works. 

Btw. Here is the image for the test.
surface_plot

ullmannJan:
Thanks, I think it makes way more sense to use a view, too.

On the other note. Basically all tests in vispy.visuals.tests use a test image to assert a results (e.g. assert_image_approved()). The test pictures can be found in the repo vispy/test-data. One example would be `visuals/spectrogram.png` for instance. In my opinion one should treat all the tests for visuals the same. This results in an improved structure for tests.

ullmannJan:
I am sorry. I did not mean to offend you. Just trying to help. I just like the idea of comparing to test_images, somehow :D. And the assert_image_approved function is quite smart only checking the correlation of two images. So there is room for future variability in rendering. But in the end you should decide on what causes less trouble.

brisvag:
> @brisvag feel free to weigh in on this. I'm 50/50. I've just had too many annoyances with `test-data` and binary blobs in git and maintaining multiple repositories for a single project that I'm not a huge fan of this. The comparison with a real result would be nice though.

I also lean for no image, those are indeed a pain to maintain. Thanks for the PR @ullmannJan, and I agree that comparing to images is "better" in principle, but with the limited amount of maintenance we can currently provide it's probably a bad idea, knowing how this stuff was a pain to deal with in the past ^^'

ullmannJan:
Ok, so maybe it is best if you implement a `vispy.testing.image_tester.assert_image_reasonable()` function or whatever you want to call it. That way there is a new standard for testing renders from now on where one does not need comparable images. But it is probably best if you implement such a feature. That way you can decide which criteria should raise an error (e.g. uniform color of render, wrong size of image data, etc.)

djhoese:
Good idea. I can think of a few shared checks, but what I've noticed is that a lot of the checks are specific to the test. The shared cases would be that the result is not all white and not all black. Otherwise, I know I've done checks like number of unique colors in the result or checking that there are blocks of colors in expected places in the result. Most of my experience is with 2D images though so not the best example.

Other ideas for how we can do comparisons that are generic enough to be shared between tests but also not require a hard coded image array in the tests?

djhoese:
Yeah I've also seen stuff along the lines of ffts and frequency of colors or something...I won't pretend to understand it. But in those cases it was still comparison of two images to see if they were "close enough", not "is this image pretty close to this encoded scalar value". Even so, I think the test can end up being structured in a way (orienting the view/canvas/camera) to give a specific and unique result where checking for a set of unique colors would be feasible. For a future PR I guess.

ullmannJan:
ok I have created a placeholder function called `assert_image_reasonable` that you can improve in the future. It only tests whether the image is completely white or black for now on.

djhoese:
Well I thought this would happen in a different future PR but I guess this is a start and is in a logical place for others to find it and improve it in the future. Thanks.

I guess there are no other ways to restructure your visualization to make sure it is valid?

ullmannJan:
One option would be to test for colors that are used in the image e.g. values of the used colormap. But this is probably a project for a future PR. I think for now testing for black and white renders is sufficient.

djhoese:
> But this is probably a project for a future PR. I think for now testing for black and white renders is sufficient.

You went from wanting an exact image comparison to only checking for not all black and not all white is sufficient? Lol

I was originally hoping for the "unique colors" check which is why I mentioned it as an option in a previous comment, but I'm not sure it is worth the work at this point. To be clear and for future reference to anyone reading this, a unique colors check wouldn't mean "the Visual used this colormap so make sure only the colors from this colormap exist". It would be using a simple colormap (ex. black to red), structuring your fake data with a specific maximum region and a specific minimum region, position the camera (if any) to limit hidden or at-an-angle regions, and comparing the colors of the expected min regions for very low color values and max regions for high reds and low green/blues (in the RGB). Like I said, that's a lot of work for this particular change which is a simple fix.

ullmannJan:
Yes, this is exactly what I meant. But I am not sure how good this will work for 3D structures with reflections and other lighting caveats. 

I added the TODO at the beginning of the `assert_image_reasonable` function to make it clear that this is still work in progress. That's what I meant with sufficient, of course it is not optimal. But I don't have time to work on this right now. I already spent way too much time on this "simple" bug fix xD. Maybe in a future PR.

djhoese:
Along the lines of making the test more testable like the view being a specific orientation, the shading can be made flat too which would minimize unexpected colors or variations of colors. Anyway...sounds good.

djhoese (approved):
Thanks for this and dealing with all the back and forth conversation.

ullmannJan:
Thank you for maintaining this amazing project!

PR #2555, opened 2023-12-04
Make HistogramVisual updateable after instantiation

tlambert03:
adds `set_raw_data` method to `HistogramVisual` to allow data to be updated after the initial instantiation

### DIFF
--- a/vispy/plot/plotwidget.py
+++ b/vispy/plot/plotwidget.py
@@ -162,7 +162,7 @@ class PlotWidget(scene.Widget):
 
         self._configured = True
 
-    def histogram(self, data, bins=10, color='w', orientation='h'):
+    def histogram(self, data=None, bins=10, color='w', orientation='h'):
         """Calculate and show a histogram of data
 
         Parameters
--- a/vispy/visuals/histogram.py
+++ b/vispy/visuals/histogram.py
@@ -3,6 +3,7 @@
 # Copyright (c) Vispy Development Team. All Rights Reserved.
 # Distributed under the (new) BSD License. See LICENSE.txt for more info.
 # -----------------------------------------------------------------------------
+from typing import Callable
 
 import numpy as np
 
@@ -22,37 +23,62 @@ class HistogramVisual(MeshVisual):
         Color of the histogram.
     orientation : {'h', 'v'}
         Orientation of the histogram.
+    calc_hist : callable
+        Function that computes the histogram. Must accept (data, bins) and
+        return (hist_data, bin_edges). Default is numpy.histogram.
     """
 
-    def __init__(self, data, bins=10, color='w', orientation='h'):
-        #   4-5
-        #   | |
-        # 1-2/7-8
-        # |/| | |
-        # 0-3-6-9
-        data = np.asarray(data)
-        if data.ndim != 1:
-            raise ValueError('Only 1D data currently supported')
-        if not isinstance(orientation, str) or \
-                orientation not in ('h', 'v'):
-            raise ValueError('orientation must be "h" or "v", not %s'
-                             % (orientation,))
-        X, Y = (0, 1) if orientation == 'h' else (1, 0)
+    def __init__(
+        self,
+        hist_data=None,
+        bins=10,
+        color="w",
+        orientation="h",
+        calc_hist: Callable = np.histogram,
+    ):
+        self.orientation = orientation
+        if not callable(calc_hist):
+            raise TypeError("calc_hist must be a callable that accepts (data, bins).")
+        self.calc_hist = calc_hist
+        self._bins = bins
+        MeshVisual.__init__(self, color=color)
+        if hist_data is not None:
+            self.set_raw_data(hist_data, bins)
 
+    @property
+    def orientation(self):
+        return self._orientation
+
+    @orientation.setter
+    def orientation(self, orientation: str = "h"):
+        if orientation not in ("h", "v"):
+            raise ValueError('orientation must be "h" or "v", not %s' % (orientation,))
+        self._orientation = orientation
+
+    def set_raw_data(self, data, bins=None) -> None:
+        # update bins if provided
+        if bins is None:
+            bins = self._bins
+        else:
+            self._bins = bins
         # do the histogramming
-        data, bin_edges = np.histogram(data, bins)
+        hist_data, bin_edges = self.calc_hist(data, bins)
         # construct our vertices
+        verts, faces = self.bins2mesh(hist_data, bin_edges)
+        super().set_data(verts, faces)
+
+    def bins2mesh(self, hist_data, bin_edges):
+        X, Y = (0, 1) if self.orientation == "h" else (1, 0)
         rr = np.zeros((3 * len(bin_edges) - 2, 3), np.float32)
         rr[:, X] = np.repeat(bin_edges, 3)[1:-1]
-        rr[1::3, Y] = data
-        rr[2::3, Y] = data
+        rr[1::3, Y] = hist_data
+        rr[2::3, Y] = hist_data
         bin_edges.astype(np.float32)
         # and now our tris
         tris = np.zeros((2 * len(bin_edges) - 2, 3), np.uint32)
-        offsets = 3 * np.arange(len(bin_edges) - 1,
-                                dtype=np.uint32)[:, np.newaxis]
+        offsets = 3 * np.arange(len(bin_edges) - 1, dtype=np.uint32)[:, np.newaxis]
         tri_1 = np.array([0, 2, 1])
         tri_2 = np.array([2, 0, 3])
         tris[::2] = tri_1 + offsets
         tris[1::2] = tri_2 + offsets
-        MeshVisual.__init__(self, rr, tris, color=color)
+        return rr, tris
--- a/vispy/visuals/histogram.py
+++ b/vispy/visuals/histogram.py
@@ -55,7 +55,7 @@ class HistogramVisual(MeshVisual):
             raise ValueError('orientation must be "h" or "v", not %s' % (orientation,))
         self._orientation = orientation
 
-    def set_raw_data(self, data, bins=None) -> None:
+    def set_raw_data(self, data, bins=None, color=None) -> None:
         # update bins if provided
         if bins is None:
             bins = self._bins
@@ -65,7 +65,7 @@ class HistogramVisual(MeshVisual):
         hist_data, bin_edges = self.calc_hist(data, bins)
         # construct our vertices
         verts, faces = self.bins2mesh(hist_data, bin_edges)
-        super().set_data(verts, faces)
+        super().set_data(verts, faces, color=color)
 
     def bins2mesh(self, hist_data, bin_edges):
         X, Y = (0, 1) if self.orientation == "h" else (1, 0)
--- a/vispy/visuals/histogram.py
+++ b/vispy/visuals/histogram.py
@@ -24,8 +24,8 @@ class HistogramVisual(MeshVisual):
     orientation : {'h', 'v'}
         Orientation of the histogram.
     calc_hist : callable
-        Function that computes the histogram. Must accept (data, bins) and
-        return (hist_data, bin_edges). Default is numpy.histogram.
+        Function that computes the histogram. Must accept two positional arguments
+        (data, bins) and return (hist_data, bin_edges). Default is numpy.histogram.
     """
 
     def __init__(
@@ -56,6 +56,7 @@ class HistogramVisual(MeshVisual):
         self._orientation = orientation
 
     def set_raw_data(self, data, bins=None, color=None) -> None:
+        """Set the data underlying the histogram, optionally update bins or color."""
         # update bins if provided
         if bins is None:
             bins = self._bins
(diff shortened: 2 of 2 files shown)


vispy/plot/plotwidget.py
```diff
@@ -162,7 +162,7 @@ def _configure_3d(self):
 
         self._configured = True
 
-    def histogram(self, data, bins=10, color='w', orientation='h'):
+    def histogram(self, data=None, bins=10, color='w', orientation='h'):
```
djhoese, line 165: Do you want to add the other new kwarg or a `**kwargs` (I'm usually no a fan but this might have to be an exception) too?

Also, could you update the docstring to mention None?

vispy/visuals/histogram.py
```diff
@@ -22,37 +23,63 @@ class HistogramVisual(MeshVisual):
         Color of the histogram.
     orientation : {'h', 'v'}
         Orientation of the histogram.
+    calc_hist : callable
```
djhoese, line 26: `callable` or `Callable`? If you or napari or some other project are OK with only type annotations I'm OK slowly migrating to that.

vispy/visuals/histogram.py
```diff
@@ -22,37 +23,63 @@ class HistogramVisual(MeshVisual):
...
+        return self._orientation
+
+    @orientation.setter
+    def orientation(self, orientation: str = "h"):
+        if orientation not in ("h", "v"):
+            raise ValueError('orientation must be "h" or "v", not %s' % (orientation,))
+        self._orientation = orientation
+
+    def set_raw_data(self, data, bins=None, color=None) -> None:
+        """Set the data underlying the histogram, optionally update bins or color."""
```
djhoese, line 59:
```suggestion
        """Set the data underlying the histogram.
        
        Optionally update bins and color. Provided data will be passed
        to the histogram function (``calc_hist``).
        
        """
```

vispy/visuals/histogram.py
```diff
@@ -22,37 +23,63 @@ class HistogramVisual(MeshVisual):
...
+        calc_hist: Callable = np.histogram,
+    ):
+        self.orientation = orientation
+        if not callable(calc_hist):
+            raise TypeError("calc_hist must be a callable that accepts (data, bins).")
+        self.calc_hist = calc_hist
+        self._bins = bins
+        MeshVisual.__init__(self, color=color)
+        if hist_data is not None:
+            self.set_raw_data(hist_data, bins)
```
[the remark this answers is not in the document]
tlambert03: it would be the same as a MeshVisual when no data had been provided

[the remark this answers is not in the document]
djhoese: I'm not sure what that is, but good point. Got it.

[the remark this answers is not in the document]
tlambert03: yeah, I think it will just show a blank canvas?

tlambert03:
thanks for the comments @djhoese!  Updated the typing and docs.  haven't added a test yet.  Technically, it should have the same coverage as before, since it's mostly just refactoring existing functionality into separate methods.  but I'll have a look at the existing tests and see if I can extend them easily

tlambert03:
ok, updated the test here to instantiate without data, and use the set_raw_data method to update the plot.

vispy/plot/plotwidget.py
```diff
@@ -162,27 +162,36 @@ def _configure_3d(self):
...
-            Number of bins, or bin edges.
+        data : array-like, optional
+            Data to histogram.  May be `None` on initialization, use `set_raw_data`
+            on returned object to set data after initialization.
+        bins : int | array-like | str
+            If `bins` is an int, it defines the number of equal-width
+            bins in the given range (10, by default). If `bins` is a
+            sequence, it defines a monotonically increasing array of bin edges,
+            including the rightmost edge, allowing for non-uniform bin widths.
+            May also be a string if the calc_hist function supports it.
```
djhoese, line 178: Should this, instead of being detailed here, just point to the HistogramVisual object?

vispy/visuals/histogram.py
```diff
@@ -3,56 +3,114 @@
...
+        May also be a string if the calc_hist function supports it.
+    color : str | Color
+        Color of the faces in the histogram mesh.
     orientation : {'h', 'v'}
         Orientation of the histogram.
+    calc_hist : Callable
+        Function that computes the histogram. Must accept two positional arguments
+        (data, bins) and return (hist_data, bin_edges). Default is numpy.histogram.
+    **kwargs : dict
+        Keyword arguments to pass to `MeshVisual`.
```
djhoese, line 48: Thoughts on removing the type definitions in the docstring? We'd have to check the generated docs, but I'm pretty sure sphinx should handle this fine and include the types in the API docs. Not sure it is needed in the docstring otherwise, right?

tlambert03: i don't have any strong thoughts on that. I'd say it's a project-specific decision, and I can see arguments for and against.  but, I'm happy to do whatever you want!  if vispy is moving in a particular direction with docstrings and you want to use this PR to update this particular module, that's fine, just let me know how you want it to look :)

djhoese: @brisvag Thoughts? @tlambert03 what have you seen other projects do? It seems redundant to have you put this much work into defining the type annotations with TypeAliases and then have to type them again in the docstrings. Could you try removing them and generating the docs locally? Or if you want, push here and we can download the website artifact from CI and view it.

```
cd doc
make html_dev-pattern PATTERN=abcd
```

^ this should generate the docs but skip all the examples.

tlambert03: I agree it's redundant.  but I do still think it's slightly a matter of opinion.  I think you'll find some holdouts who feel like docstrings are for humans and type annotations are for machines, and that docstrings are "sacred" in a sense... people in that camp would probably say "go ahead and use type annotations if you want, but leave my docstrings out of it".  They might additionally argue that strictly accurate type annotations can become verbose at the expense of human readability.  then there's strict DRY camp that says it should only ever be in one place, even if it possibly requires someone to look in a different place than they might be accustomed to in interactive usage. 

(*both* might argue that they don't care how your rendered HTML docs look, or whether you're using sphinx type hints to achieve the same final result on your webpage)

i kinda like the modern type-hints only approach... but more than anything this feels like a repo-wide discussion, and not something I would try to change incrementally in each PR that happens to touch a docstring.  I added the "regular" docs to match the rest of vispy, I added the type hints after your comment suggested to add them ... I think this general discussion belongs somewhere outside of this PR :)

djhoese: Sure. I think my thing is that I typically either am looking at API docs on the website or I'm telling my IDE to pull them up for me. I rarely open the docstring version of the docs in my IDE and only pull up the "show me the parameters and their types" view inline as I'm typing. I think PyCharm is smart enough to put type annotations where it needs to in these cases. Anyway...I don't think vispy is maintained well enough to do this repo-wide so I personally would be OK with it being done here as a test...assuming it renders properly in sphinx.

brisvag: No strong opinions, frankly ^^' I would only say, if you do add them to the docstrings keep them human-radable, in which case that's probably my preferred way.

djhoese: @tlambert03 Maybe that leaves it up to you then since we're asking you to do the work. I think I'd prefer type annotations only, @brisvag seems to have a slight preference for docstrings (human-readable). I don't think NOT having type annotations is a reasonable option so the choice is both or annotations only. Again, I vote for annotations only and we see how it goes with user complaints (if any).

tlambert03: ok, so just so I'm clear.  you want me to add type annotations to parts of the code didn't previously have them? (like that one method in plotwidget)?  You're ok with just that single method having type annotations?  or you want me to add it to the whole module?  (this is why i find this all a bit confusing... i imagined this would be a quick PR that just moves a little existing code around to allow it to be called as a method 😄 )

brisvag: Then it looks like what we have here already will do just fine (better than anything we had before anyways :P)

brisvag (approved):
Nice!

vispy/visuals/histogram.py
tlambert03: ok, will do.  I believe I added type annotations in the `__init__` in response to a previous comment ... it wasn't in the original PR, which largely aimed to match the existing code patterns.  
I'll go ahead and remove docstring types in HistogramVisual

vispy/visuals/histogram.py
```diff
@@ -28,23 +28,23 @@ class HistogramVisual(MeshVisual):
 
     Parameters
     ----------
-    data : array-like, optional
+    data
```
djhoese, line 31: I think you still need the colons to be numpy docs compatible...right?

djhoese:
Looks pretty good:

image

djhoese:
Wait...those docstrings don't match the current state of your PR. WTF

djhoese:
I'm going to rerun the documentation building job to see if the docstrings make more sense. Otherwise, @tlambert03 if you have time you could generate the docs locally and see if the API docs docstrings match for you.

djhoese:
@tlambert03 I seem to not have push permissions on your fork. Any idea why?

tlambert03:
hmm, I'm not sure. I do have the box checked to allow edits by maintainers, and it is from a branch in my personal fork... but I have had this happen before in PRs to vispy... so I wonder if there's something wrong about my fork.  sorry about that.  will try to build docs locally

djhoese:
I did end up making a PR in your fork. If you merge that there should be some better handling of type annotations by sphinx.

djhoese:
Did you change something? I was able to make a commit now. This time I did it in github's basic editor, before I was doing it in a "github codespace".

djhoese:
Ok so issues:

1. The building of the website failing doesn't cause the CI to fail. Even if it doesn't produce anything the artifact upload doesn't even fail.
2. Error from a recently merged PR:

```
/home/runner/micromamba/envs/vispy-tests/lib/python3.9/site-packages/numpydoc/docscrape.py:455: UserWarning: Unknown section Parameters: in the docstring of assert_image_reasonable in /home/runner/work/vispy/vispy/vispy/testing/image_tester.py.
  warn(msg)
```

3. Error from this PR:

```
WARNING: error while formatting signature for vispy.scene.visuals.Histogram: Handler <function mangle_signature at 0x7f3f90cad040> for event 'autodoc-process-signature' threw an exception (exception: The section Parameters appears twice in  Visual that calculates and displays a histogram of data

    This class inherits from visuals.HistogramVisual and scene.Node, allowing the visual to be placed inside a scenegraph.
looking for now-outdated files... none found


    Parameters
    ----------
    data
        Data to histogram.  May be `None` on initialization, use `set_raw_data`
        to set data after initialization.
    bins
        If `bins` is an int, it defines the number of equal-width
        bins in the given range (10, by default). If `bins` is a
        sequence, it defines a monotonically increasing array of bin edges,
        including the rightmost edge, allowing for non-uniform bin widths.
        May also be a string if the calc_hist function supports it.
    color
        Color of the faces in the histogram mesh.
    orientation
        Orientation of the histogram.
    calc_hist
        Function that computes the histogram. Must accept two positional arguments
        (data, bins) and return (hist_data, bin_edges). Default is numpy.histogram.
    **kwargs
        Keyword arguments to pass to `MeshVisual`.

    parent : Node
        The parent node to assign to this node (optional).
    name : string
        A name for this node, used primarily for debugging
        (optional).
Visual that calculates and displays a histogram of data

    Parameters
    ----------
    data
        Data to histogram.  May be `None` on initialization, use `set_raw_data`
        to set data after initialization.
    bins
        If `bins` is an int, it defines the number of equal-width
        bins in the given range (10, by default). If `bins` is a
        sequence, it defines a monotonically increasing array of bin edges,
        including the rightmost edge, allowing for non-uniform bin widths.
        May also be a string if the calc_hist function supports it.
    color
        Color of the faces in the histogram mesh.
    orientation
        Orientation of the histogram.
    calc_hist
        Function that computes the histogram. Must accept two positional arguments
        (data, bins) and return (hist_data, bin_edges). Default is numpy.histogram.
    **kwargs
        Keyword arguments to pass to `MeshVisual`. in the docstring of Histogram in /home/runner/work/vispy/vispy/vispy/scene/visuals.py.)
WARNING: Failed guarded type import with ImportError("cannot import name 'TypeAlias' from 'typing' (/home/runner/micromamba/envs/vispy-tests/lib/python3.9/typing.py)")
WARNING: Cannot resolve forward reference in type annotations of "vispy.visuals.histogram.HistogramVisual": name 'npt' is not defined
WARNING: Cannot resolve forward reference in type annotations of "vispy.visuals.histogram.HistogramVisual.orientation": name 'Orientation' is not defined
WARNING: Cannot resolve forward reference in type annotations of "vispy.visuals.histogram.HistogramVisual.set_raw_data": name 'npt' is not defined
```

My understanding is that the `create_visual_node` function that creates a SceneCanvas Visual object from the basic Canvas Visual class is concatenating the docstrings in some way that is producing two sets of "Parameters" sections. I'm a little worried this is going to be annoying to fix.

djhoese:
https://github.com/vispy/vispy/blob/66675999571de784e736d4ffa645fbcabe385649/vispy/scene/visuals.py#L146-L226

tlambert03:
Can we go back to my original PR that simply moves the init logo into methods, using the existing patterns, and consider possible changes to how vispy build the docs and how type hints play a role in another issue?

djhoese:
Sure. Could you please copy this existing branch to a new name and make a new PR and describe it as adding the ability to use type annotations in sphinx docs or something like that? I'd like to not lose what's here already. I suppose the best solution here would be to add the typing back into the docstring and then in `doc/conf.py` comment out the type annotations extension. Right?

PR #2561, opened 2023-12-19
Radius of ellipse can now be specified as numpy.array

ullmannJan:
It was only possible to parse in the parameters as list or tuple. It is useful to accept numpy arrays

### DIFF
--- a/vispy/visuals/ellipse.py
+++ b/vispy/visuals/ellipse.py
@@ -69,9 +69,8 @@ class EllipseVisual(PolygonVisual):
             if len(radius) == 2:
                 xr, yr = radius
             else:
-                raise ValueError("radius must be float or 2 value tuple/list"
-                                 " (got %s of length %d)" % (type(radius),
-                                                             len(radius)))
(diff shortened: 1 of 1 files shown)


djhoese (changes requested):
While I'm OK merging this, I'm curious what your use case is? Why not just convert your numpy array to a list before passing it? I'm worried about this being error prone with users providing any numpy array of any shape and then getting a weird error back. Oh I guess it does check the length before it tries extracting the elements.

Otherwise, could you also update the `ValueError` error message to mention numpy arrays as well? Feel free to update the formatting of that string. For example, you could use f-strings and that might clean it up.

ullmannJan:
Of course you can convert it to a list at any point. But the center should be given as an array. So it is sensible to accept radii as arrays as well.

djhoese:
> But the center should be given as an array.

How strange. Good enough for me. I have no idea why center is expected to be an array. I wonder if the person who originally wrote that docstring really meant "Sequence" or "Iterable" in the sense that it is a container or items. NO big deal. I'll merge when CI passes.

PR #2563, opened 2023-12-27
Fix typo in plot widget error message

jeertmans:
Hello, seems like there is a typo from copy/pasting the error message. I fixed it.


djhoese (approved):

djhoese:
Thanks!

Issue #2564, opened 2023-12-28
Errorous behavior in Vispy.Visuals.Markers.set_data


c40zAtGitHub:
In line 606-608 of markers.py, the following statement is written:

`if (edge_width is not None) + (edge_width_rel is not None) != 1:
            raise ValueError('exactly one of edge_width and edge_width_rel '
                             'must be non-None')`

while in line 583-585, the method is defined as

`def set_data(self, pos=None, size=10., edge_width=1., edge_width_rel=None,
                 edge_color='black', face_color='white',
                 symbol='o'):`

An error occurs when only edge_width_rel is supplied to the set_data method while the edge_width is untouched. The sum in the if statement in this case will be 2, causing the ValueError raised instead of normal executation of the method (which is expected since the edge_width_rel is supplied).

Current issue can be dodged by explicitly supplying edge_width=None in the input.

Apparent logic for the ValueError on line 607 to be raised should be
`if (edge_width is None) and (edge_width_rel is None):`

Personally, I guess that the usage of edge_width should be surpressed when edge_width_rel is supplied. This can be achieved by changing code in line 610-617 to something like
`        if edge_width_rel is not None:
            edge_width_rel = np.asarray(edge_width_rel)
            if np.any(edge_width_rel < 0):
                raise ValueError('edge_width_rel cannot be negative')
        else:
           edge_width = np.asarray(edge_width)
            if np.any(edge_width < 0):
                raise ValueError('edge_width cannot be negative')`

djhoese:
Looks like some of this logic has been there for 9 years or so. In the last couple years @brisvag had updated some of it. I think your suggestion to essentially swap the `i edge_width_rel is not None` and `edge_width` if statements makes sense. Would you mind submitting a pull request and then we can get more feedback from other vispy maintainers.

c40zAtGitHub:
Hi, David. Thanks for the reply. I have setup the pull request: https://github.com/vispy/vispy/pull/2565. It has a slight deviation from what I had written originally, but I believe the problem is still there.
________________________________

PR #2565, opened 2023-12-28
Correction to Vispy.Visuals.Markers.set_data

c40zAtGitHub:
This pull request follows issue 2546 (https://github.com/vispy/vispy/issues/2564). After further examining the original code, I found I misunderstood its purpose when I was writing the post. However, there is a bug in the original code, and I would like to restate it:

The (inferred) purpose of the original code:
In the original code, it is assumed that either edge_width or edge_width_rel is passed to set_data. 
If both are supplied, the method yields a ValueError. 
If neither is supplied, the method simply uses a default edge_width of 1.0. 
If either edge_width or edge_width_rel is supplied, it will be applied.

However, due to the default value of edge_width given in the definition of set_data, the function ends up with a ValueError when only edge_width_rel is supplied, which deviates from the last purpose, and I personally believe this is a bug.

The solution is to set both edge_width and edge_width_rel to None in the definition of set_data, and assign 1.0 to edge_width when both variables are None.

### DIFF
--- a/vispy/visuals/markers.py
+++ b/vispy/visuals/markers.py
@@ -579,7 +579,7 @@ class MarkersVisual(Visual):
 
         self.freeze()
 
-    def set_data(self, pos=None, size=10., edge_width=1., edge_width_rel=None,
+    def set_data(self, pos=None, size=10., edge_width=None, edge_width_rel=None,
                  edge_color='black', face_color='white',
                  symbol='o'):
         """Set the data used to display this visual.
@@ -602,9 +602,13 @@ class MarkersVisual(Visual):
         symbol : str or array
             The style of symbol used to draw each marker (see Notes).
         """
-        if (edge_width is not None) + (edge_width_rel is not None) != 1:
-            raise ValueError('exactly one of edge_width and edge_width_rel '
-                             'must be non-None')
+        if (edge_width is not None) and (edge_width_rel is not None):
+            #both edge_width and edge_width_rel are supplied
+            raise ValueError('either edge_width or edge_width_rel '
+                             'should be provided, not both')
+        elif (edge_width is None) and (edge_width_rel is None):
+            #neither the two widths is supplied
+            edge_width = 1.0
 
         if edge_width is not None:
             edge_width = np.asarray(edge_width)
--- a/vispy/visuals/markers.py
+++ b/vispy/visuals/markers.py
@@ -497,8 +497,10 @@ class MarkersVisual(Visual):
     edge_width : float or array or None
         The width of the symbol outline in screen (or data, if scaling is on) px.
     edge_width_rel : float or array or None
-        The width as a fraction of marker size. Exactly one of
-        `edge_width` and `edge_width_rel` must be supplied.
+        The width as a fraction of marker size.
+        Note on edge_width and edge_width_rel:
+            By default, edge_width = 1.0 if neither edge_width nor edge_width_rel is provided
+            A ValueError will be raised if both edge_width and edge_width_rel are supplied.
     edge_color : Color | ColorArray
         The color used to draw each symbol outline.
     face_color : Color | ColorArray
--- a/vispy/visuals/markers.py
+++ b/vispy/visuals/markers.py
@@ -595,10 +595,8 @@ class MarkersVisual(Visual):
         edge_width : float or array or None
             The width of the symbol outline in screen (or data, if scaling is on) px.
         edge_width_rel : float or array or None
-            The width as a fraction of marker size.
-            Note on edge_width and edge_width_rel:
-                By default, edge_width = 1.0 if neither edge_width nor edge_width_rel is provided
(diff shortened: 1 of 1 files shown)


vispy/visuals/markers.py
```diff
@@ -602,9 +602,13 @@ def set_data(self, pos=None, size=10., edge_width=1., edge_width_rel=None,
...
-        if (edge_width is not None) + (edge_width_rel is not None) != 1:
-            raise ValueError('exactly one of edge_width and edge_width_rel '
-                             'must be non-None')
+        if (edge_width is not None) and (edge_width_rel is not None):
+            #both edge_width and edge_width_rel are supplied
+            raise ValueError('either edge_width or edge_width_rel '
+                             'should be provided, not both')
+        elif (edge_width is None) and (edge_width_rel is None):
+            #neither the two widths is supplied
+            edge_width = 1.0
```
djhoese, line 611: The extra parentheses aren't needed. The existing code was flawed. I'm also deleting the comments as they seem redundant given the simplicity of the conditions now.

```suggestion
        if edge_width is not None and edge_width_rel is not None:
            raise ValueError("either edge_width or edge_width_rel "
                             "should be provided, not both")
        elif edge_width is None and edge_width_rel is None:
            edge_width = 1.0
```

djhoese (changes requested):
I like this solution. Nice job. This makes it more clear. Do you think you could mention in the docstring for `edge_width` that if defaults to `1.0` if neither edge_width nor edge_width_rel are provided?

c40zAtGitHub:
I have modified the doc string in my branch, do I need additional operations? Sorry, I'm not quite farmiliar with github.

vispy/visuals/markers.py
c40zAtGitHub, line 1: I append information on `edge_width` and `edge_width_rel` to the edge_with_rel entry as a note, since the feature involves both parameters. Both the doc string of the MarkersVisual class and the MarkersVisuals.set_data method have been modified

c40zAtGitHub (commented):
The doc string has been updated.

vispy/visuals/markers.py
```diff
@@ -497,8 +497,10 @@ class MarkersVisual(Visual):
     edge_width : float or array or None
         The width of the symbol outline in screen (or data, if scaling is on) px.
```
djhoese, line 498:
```suggestion
        The width of the symbol outline in screen (or data, if scaling is on) px.
        Defaults to 1.0 if None or not provided and ``edge_width_rel`` is not
        provided.
```

vispy/visuals/markers.py
```diff
@@ -497,8 +497,10 @@ class MarkersVisual(Visual):
...
     edge_width_rel : float or array or None
-        The width as a fraction of marker size. Exactly one of
-        `edge_width` and `edge_width_rel` must be supplied.
+        The width as a fraction of marker size.
+        Note on edge_width and edge_width_rel:
+            By default, edge_width = 1.0 if neither edge_width nor edge_width_rel is provided
+            A ValueError will be raised if both edge_width and edge_width_rel are supplied.
```
djhoese, line 503:
```suggestion
        The width as a fraction of marker size. Can not be specified along with
        edge_width. A ValueError will be raised if both are provided.
```

vispy/visuals/markers.py
```diff
@@ -593,18 +595,22 @@ def set_data(self, pos=None, size=10., edge_width=1., edge_width_rel=None,
         edge_width : float or array or None
             The width of the symbol outline in screen (or data, if scaling is on) px.
         edge_width_rel : float or array or None
-            The width as a fraction of marker size. Exactly one of
-            `edge_width` and `edge_width_rel` must be supplied.
+            The width as a fraction of marker size.
+            Note on edge_width and edge_width_rel:
+                By default, edge_width = 1.0 if neither edge_width nor edge_width_rel is provided
+                A ValueError will be raised if both `edge_width` and `edge_width_rel` are supplied.
```
djhoese, line 601:
```suggestion
            The width as a fraction of marker size. Can not be specified along with
            edge_width. A ValueError will be raised if both are provided.
```

vispy/visuals/markers.py
```diff
@@ -593,18 +595,22 @@ def set_data(self, pos=None, size=10., edge_width=1., edge_width_rel=None,
         edge_width : float or array or None
             The width of the symbol outline in screen (or data, if scaling is on) px.
```
djhoese, line 596:
```suggestion
            The width of the symbol outline in screen (or data, if scaling is on) px.
            Defaults to 1.0 if None or not provided and ``edge_width_rel`` is not
        provided.
```

djhoese (commented):
Looks good. I just reworded a few things as I'd like the default for edge_width to be part of that arguments documentation.

djhoese (approved):

PR #2573, opened 2024-02-14
Fix a few typos in the docs

kraasch:


doc/getting_started/modern-gl.rst
```diff
@@ -273,18 +273,18 @@ and not the fixed one.
...
 <http://duriansoftware.com/joe/An-intro-to-modern-OpenGL.-Table-of-Contents.html>`_
-walks through the parts that are still relevant today.
+that walks through the parts that are still relevant today.
 
 
 Learning Modern 3D Graphics Programming
 ---------------------------------------
 
-This `book <https://paroj.github.io/gltut/>`_ is intended to teach you how
+`This book <https://paroj.github.io/gltut/>`_ is intended to teach you how
```
djhoese, line 287: Is there a reason for this change?

doc/getting_started/modern-gl.rst
```diff
@@ -273,18 +273,18 @@ and not the fixed one.
 An intro to modern OpenGL
 -------------------------
 
-OpenGL has been around a long time, and from reading all the accumulated layers
+OpenGL has been around a long time and from reading all the accumulated layers
```
djhoese, line 276: If we're being picky/pedantic, is this grammatically correct? The "and" after the comma throws me off as to whether the comma is appropriate.

djhoese (approved):

doc/getting_started/modern-gl.rst
kraasch: The comma seemed not threw me a bit off while I was reading the text. I don't see a reason to put a comma there since the word 'and' already does its work and connects the two sentences. 

It cannot be a serial comma either, because there is nothing being enumerated.

doc/getting_started/modern-gl.rst
djhoese: Good point! Do you want to put the real name of the online book there instead?

doc/getting_started/modern-gl.rst
djhoese: Ha, sorry, I thought this was the opposite change and you were adding the comma. I agree with this. Thanks.

djhoese (approved):
Nice job. Thanks for fixing these things.

PR #2592, opened 2024-05-28
Add extra for glfw

### LINKED ISSUE #2591: What's the status of GLFW?
I see there's a `glfw` backend, but in `setup.py` it is neither listed as a dependency nor defined as an extra. Is that an oversight or deliberately?

I'm packaging `vispy` for Fedora and with [glfw](https://pypi.org/project/glfw/) added as a dependency, I'm seeing `glfw` listed in the output of `vispy.sys_info()`. Tests using `glsw` as a backend also appear to work fine.

penguinpee:
The backend is already there. Close #2591.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -106,6 +106,7 @@ setup(
         'pyside': ['PySide'],
         'pyside2': ['PySide2'],
         'pyside6': ['PySide6'],
+        'glfw': ['glfw'],
         'sdl2': ['PySDL2'],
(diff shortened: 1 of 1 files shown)


djhoese (approved):
Since I'm on leave and may not be in the right mindset to make sure we're not missing something simple here, @brisvag could you review this too?

brisvag (approved):
Seems straight-forward to me!

PR #2595, opened 2024-06-04
Raise ValueError when setting relative_step_size 0 or negative

aganders3:
See discussion in #2589 

This changes the the `relative_step_size` setter to:
* raise a `ValueError` when setting to 0 or a negative value
* warn when setting small values between 0 and `side_len / (2 * MAX_CANVAS_SIZE)`
* silently comply otherwise

This is closer to the previous behavior, and I think makes intuitive sense (step size <=0 is non-sensical and could cause degenerate behavior in the shader).

A bit confusing - there is a test for this and it passed on #2589 checks:
https://github.com/vispy/vispy/blob/4f061ba23ddac3131eb0d1d955573dd71c9a3d2f/vispy/visuals/tests/test_volume.py#L46-L48

### DIFF
--- a/vispy/visuals/volume.py
+++ b/vispy/visuals/volume.py
@@ -1174,17 +1174,21 @@ class VolumeVisual(Visual):
 
         https://github.com/vispy/vispy/pull/2587
 
-        For this reason, this setter raises a ValueError when the value is
+        For this reason, this setter issues a warning when the value is
         smaller than ``side_len / (2 * MAX_CANVAS_SIZE)``, where ``side_len``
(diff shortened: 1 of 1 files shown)


jni (approved):

brisvag (approved):
Mhm... weird with the tests. Have you checked if the test passes if a value < 0 is given instead of 0?

jni:
@brisvag see #2596

brisvag (approved):
Ok, let's get this in so the test is passing in the other PR!

PR #2601, opened 2024-06-14
Set self._data_dtype in scale_and_set_data for CPU scaled textures

aganders3:
I noticed this warning when running tests, and I think it traces back to the change in this PR.

```
vispy/visuals/tests/test_colormap.py: 6 warnings
vispy/visuals/tests/test_image.py: 39 warnings
vispy/visuals/tests/test_scalable_textures.py: 4 warnings
  /Users/aandersoniii/src/vispy/vispy/gloo/texture.py:18: DeprecationWarning: finfo() dtype cannot be None. This behavior will raise an error in the future. (Deprecated in NumPy 1.25)
    info = np.finfo(dtype)
```

See https://github.com/vispy/vispy/pull/2350#discussion_r1638869690 for some context.

I am partly guessing what this code meant to be doing, but here the `_data_dtype` is set the first time `scale_and_set_data` is called. @brisvag at your suggestion I am trying to think of a test that could have caught this, but need to think about it a little more.

### DIFF
--- a/vispy/visuals/_scalable_textures.py
+++ b/vispy/visuals/_scalable_textures.py
@@ -311,7 +311,7 @@ class CPUScaledTextureMixin(_ScaledTextureMixin):
     def scale_and_set_data(self, data, offset=None, copy=True):
         """Upload new data to the GPU, scaling if necessary."""
         if self._data_dtype is None:
-            data.dtype == self._data_dtype
+            self._data_dtype = data.dtype
 
         # ensure dtype is the same as it was before, or funny things happen
         # no copy is performed unless asked for or necessary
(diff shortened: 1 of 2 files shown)


brisvag (commented):
So, looking back at it:

- this should happen when the texture is first initialized with no data, resulting in `self._data_dtype == None`
- then, if new data is set via `scale_and_set_data`, a new dtype is saved. However, before this PR, the `self._data_dtype` remains `None`.
- the *next time* data is set, if the data has a different dtype, things should get messed up because the conversion/clipping will go awry.

I tried to spin up a simple test but things are getting in the way and I don't have the time now to dive deeper, so if you do please take a stab!

aganders3:
@brisvag does this test make sense to you?

Here is a failed run without the proposed change:
https://github.com/vispy/vispy/actions/runs/9549721978/job/26320015919

aganders3:
Of course I left a bad debug print in there though. 🤦

I believe tests will pass now with b1692b4.

Issue #2542, opened 2023-10-24
in (/vispy/visuals /surface_plot.py) line 129


Abbas-Yasir:
if colors.ndim == 3:
            # convert (width, height, 4) to (num_verts, 4)
            vert_shape = self.__vertices.shape
            num_vertices = vert_shape[0] * vert_shape[1]
            colors = colors.reshape(num_vertices, **_3_**)
        return colors

Shouldn't:
colors = colors.reshape(num_vertices, 3)
be
colors = colors.reshape(num_vertices, 4) ????

djhoese:
FYI you can link to exact lines and line ranges with a github URL. You can even use a "permalink" and github will show a preview of the code in your comment. Here is what you are referencing:

https://github.com/vispy/vispy/blob/c366456b66938577c449137a30331859eb52c8d4/vispy/visuals/surface_plot.py#L121-L130

Let me review the code and see what I think. I feel like this came up in another issue recently.

Issue #2517, opened 2023-08-19
issue about SurfacePlotVisual


JintaoLee-Roger:
Hi!

When I use `SurfacePlot` to draw a surface and seed `color` parameter, an error occurs:

```
File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/site-packages/vispy/visuals/surface_plot.py", line 129, in _prepare_mesh_colors
    colors = colors.reshape(num_vertices, 3)
ValueError: cannot reshape array of size 960000 into shape (240000,3)
```

I note that the `_prepare_mesh_colors` function in `vispy/visuals/surface_plot.py` is:
```
    def _prepare_mesh_colors(self, colors):
        if colors is None:
            return
        colors = np.asarray(colors)
        if colors.ndim == 3:
            # convert (width, height, 4) to (num_verts, 4)
            vert_shape = self.__vertices.shape
            num_vertices = vert_shape[0] * vert_shape[1]
            colors = colors.reshape(num_vertices, 3)
        return colors
```

The comment and document state that I should pass in a matrix of dimensions `(width, height, 4)`, which will then be transformed into a matrix of dimensions `(num_verts, 4)`. However, the code does not follow this approach; instead, it transforms it into a matrix of dimensions `(num_verts, 3)`.

djhoese:
At first glance this does seem like a bug, but also what colors are you providing? RGB? RGBA?

Skimming through the code it does definitely seem like RGBA (4 elements) is the more performant way of providing the colors. Otherwise, VisPy adds the alpha channel itself which requires a copy of the color array. I'm not really sure how or why this is reshaping to `3` (RGB). If you feel up to it, it would be great if you could make a pull request fixing this.

PR #2645, opened 2025-02-17
Fix mesh colors reshaping

BenZickel:
# Bug Description

Setting `colors` of `SurfacePlot` fails (see below code snippet).

```
from vispy import scene
from numpy.random import rand
p = scene.visuals.SurfacePlot(z=rand(6,6))
p.set_data(colors=rand(6,6,4))
```

# Solution

Fix reshape of colors array to be with a length four last dimension.

### DIFF
--- a/vispy/visuals/surface_plot.py
+++ b/vispy/visuals/surface_plot.py
@@ -126,7 +126,7 @@ class SurfacePlotVisual(MeshVisual):
             # convert (width, height, 4) to (num_verts, 4)
             vert_shape = self.__vertices.shape
             num_vertices = vert_shape[0] * vert_shape[1]
-            colors = colors.reshape(num_vertices, 3)
+            colors = colors.reshape(num_vertices, 4)
         return colors
 
     def set_data(self, x=None, y=None, z=None, colors=None):
--- a/vispy/visuals/surface_plot.py
+++ b/vispy/visuals/surface_plot.py
@@ -126,7 +126,7 @@ class SurfacePlotVisual(MeshVisual):
             # convert (width, height, 4) to (num_verts, 4)
             vert_shape = self.__vertices.shape
             num_vertices = vert_shape[0] * vert_shape[1]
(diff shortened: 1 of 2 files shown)


vispy/visuals/surface_plot.py
```diff
@@ -126,7 +126,7 @@ def _prepare_mesh_colors(self, colors):
             # convert (width, height, 4) to (num_verts, 4)
             vert_shape = self.__vertices.shape
             num_vertices = vert_shape[0] * vert_shape[1]
-            colors = colors.reshape(num_vertices, 3)
+            colors = colors.reshape(num_vertices, 4)
```
djhoese, line 129: There is another merge request with a fix for this that was never merged and has a similar fix:

https://github.com/vispy/vispy/pull/2620/files#diff-e962a5a20e008c521c34955c7cc159ff74e96a8c2edaf8a7bf22733c402c28f9

Mainly, instead of `4` it does `colors.shape[-1]` to handle RGB and RGBA colors.

```suggestion
            colors = colors.reshape(num_vertices, colors.shape[-1])
```

djhoese:
Thanks @BenZickel. Do you think you'd be able to add a test for this?

BenZickel:
I've added a test of setting colors via a grid.

djhoese (approved):
Thanks!

BenZickel:
@djhoese, there seem to be CI failures which are not due to this pull request (these failures have also occured in previous pull requests):
- XFAIL vispy/geometry/tests/test_triangulation.py::test_triangulate_collinear_path_with_repeat - See https://github.com/vispy/vispy/issues/2247
- XFAIL vispy/util/fonts/tests/test_font.py::test_font_glyph[Noto Color Emoji] - reason:

brisvag (approved):
Makes sense to me!

Btw, this also fixes #2542 and #2517, right? And it partly being addressed in #2620 (but it's good to tackle it separately here).

PR #2647, opened 2025-02-21
Fix type of vertice passed to function for adding triangles

Czaki:
Since numpy 2.0 The `arr[ind]` is `np.int(64)`, not `int` that lead to not finding edge later when construct triangulation. 

This PR fixes it by converting it to python int.


Czaki:
CI fail is fixed in #2636

brisvag (approved):
Makes sense!

Czaki:
Will be nice to release this change in the next few days to reduce failures of triangulation.

PR #2674, opened 2025-05-15
change cancelation policy of wheel workflow to base on ref

Czaki:
implements https://github.com/vispy/vispy/pull/2672#issuecomment-2882016826

github ref contains branch trigger name, so fon main it will be `refs/heads/main` but for tagged commit something like `refs/heads/tag` 

Each PR has a unique ref, but I left `event_type` for in-repo PRs

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -2,7 +2,7 @@ name: cibuildwheel
 # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency
 # https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pullrequestevent
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.type }}
+  group: ${{ github.ref }}-${{ github.event.type }}
   cancel-in-progress: true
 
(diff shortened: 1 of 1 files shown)


brisvag (approved):

.github/workflows/wheels.yml
```diff
@@ -2,7 +2,7 @@ name: cibuildwheel
 # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency
 # https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pullrequestevent
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.type }}
+  group: ${{ github.ref }}-${{ github.event.type }}
```
djhoese, line 5: By getting rid of `workflow` aren't you making `wheels.yml` workflow and `main.yml` workflow in the same group?

In another project I've successfully been using:

```
group: ${{ github.workflow }}-${{ github.event.number || github.event.ref }}-${{ github.event.type }}
```

djhoese:
Ok. I guess the "workflow" property would have given you the same thing and is more copy/paste-able, but all good. Thanks.

PR #2692, opened 2025-08-05
Add doc about Qt backend vsync limitation

### LINKED ISSUE #2690: Framerate drops when using PyQt background threads
I have adjusted the Qt example from the example gallery for my use case and compared it to PyQtGraph
https://vispy.org/gallery/scene/realtime_data/ex03c_data_sources_threaded_events.html#sphx-glr-gallery-scene-realtime-data-ex03c-data-sources-threaded-events-py

What I noticed is that there are some random framerate drops occuring from time to time. On my machine (mobile RTX 3060) the FPS is 350, but drops to 75 from time to time.

I compared it to the mentioned library, making the code as similiar as I could and there I get 250 FPS, but it is a stable framerate. I also checked the Windows resource monitor and I have plenty of computing power left, except that for pyqtgraph one CPU core is fully used.

I don't know how to debug it further, so any advice on GPU and VisPy profiling would be helpful. The 75 value is interesting, because this is the refresh rate of my monitor. Could this be related?

Here is a reproduction repo:
https://github.com/krokosik/vispy-minimal-example-repro/tree/2688

krokosik:
Closes #2690

### DIFF
--- a/vispy/scene/canvas.py
+++ b/vispy/scene/canvas.py
@@ -45,7 +45,13 @@ class SceneCanvas(app.Canvas, Frozen):
         Whether to create the widget immediately. Default True.
     vsync : bool
         Enable vertical synchronization.
-        Note for Qt backend: this setting might not work on every platform [src](https://doc.qt.io/qt-6/qsurfaceformat.html#setSwapInterval).
+        Note for some backends
+        (ex. `Qt-based backends <https://doc.qt.io/qt-6/qsurfaceformat.html#setSwapInterval>`_)
+        and some platforms this setting may not work as expected. Some GPU
+        drivers may require modifying vsync settings at a system or driver
+        level (ex. NVIDIA control panel). See documentation for the specific
+        backend engine and your system's GPU in the case of unexpected
+        behavior.
     resizable : bool
         Allow the window to be resized.
     decorate : bool


vispy/app/canvas.py
```diff
@@ -48,6 +48,7 @@ class Canvas(object):
         Whether to create the widget immediately. Default True.
     vsync : bool
         Enable vertical synchronization.
+        Note for Qt backend: this setting might not work on every platform [src](https://doc.qt.io/qt-6/qsurfaceformat.html#setSwapInterval).
```
djhoese, line 51: Does this render properly in the HTML? This seems like a markdown link but most docstrings should be restructuredtext, unless we have pluguins parsing markdown too. Let's see...

djhoese:
Ok that's better:

image

@krokosik what do you think about what text should have the hyperlink?

krokosik:
Looks good to me :)
However, I had an afterthought that maybe we should also mention that vsync settings would have to be changed in the GPU settings, like the NVIDIA control panel?

djhoese:
How about:

```
Enable vertical synchronization.
        Note for some backends (ex. `Qt-based backends <https://doc.qt.io/qt-6/qsurfaceformat.html#setSwapInterval>`_) and some platforms this setting may not work as expected. Some GPU drivers may require modifying vsync settings at a system or driver level (ex. NVIDIA control panel). See documentation for the specific backend engine and your system's GPU in the case of unexpected behavior.
```

krokosik:
Awesome, I was hesitant if you don't want shorter docstrings, but this is very reasonable to me :)

vispy/app/canvas.py
```diff
@@ -48,6 +48,7 @@ class Canvas(object):
         Whether to create the widget immediately. Default True.
     vsync : bool
         Enable vertical synchronization.
+        Note for Qt backend: this setting might not work on `every platform <https://doc.qt.io/qt-6/qsurfaceformat.html#setSwapInterval>`_.
```
djhoese, line 51:
```suggestion
        Note for some backends
        (ex. `Qt-based backends <https://doc.qt.io/qt-6/qsurfaceformat.html#setSwapInterval>`_)
        and some platforms this setting may not work as expected. Some GPU
        drivers may require modifying vsync settings at a system or driver
        level (ex. NVIDIA control panel). See documentation for the specific
        backend engine and your system's GPU in the case of unexpected
        behavior.
```

djhoese (approved):

PR #2738, opened 2026-03-25
Update for new jupyter_rfb and add experimental Marimo support

almarklein:
Jupyter_rfb underwent some massive refactoring lately to (finally) use anywidget!

### Compatibility

This PR updates to the new version, although it is still compatible with older versions of jupyter_rfb.

Jupyter_rfb has a flag to help in the transition process (`_event_compatibility`). This PR sets it to get the new-style events. This means that once Vispy uses this for say a year, jupyter_rfb can remove its compatibility logic for the old-style events.

Once the new jupyter_rfb has been tested in the field for a while, we can pin in to 1.x, and remove some of the compatibility code in *our* module.

### Titlebar

The new widget supports a titlebar, so we can show titles too. Currently I set it up to show the titlebar if there is a title, and hide it when the title is an empty string.

### Marimo

With anywidget, we can also use the marimo notebook! There are two open issues though. I just created/commented them, so we can wait for an answer.

* I'm not sure what the recommended way to detect Marimo is. I implemented something, but maybe a better method exists. See https://github.com/marimo-team/marimo/discussions/8865
* The `canvas._backend` is an anywidget, the `canvas` itself is not. In Marimo, the `_repr_mimetype_` delegation is not enough. See https://github.com/manzt/anywidget/issues/792

### DIFF
--- a/vispy/app/application.py
+++ b/vispy/app/application.py
@@ -6,6 +6,7 @@
 
 from __future__ import division
 
+import __main__ as main_module
 import builtins
 import os
 import sys
@@ -116,7 +117,20 @@ class Application(object):
             return False
 
     def is_notebook(self):
-        """Determine if the user is executing in a Jupyter Notebook"""
+        """Determine if the user is executing in a Jupyter/Marimo Notebook"""
+
+        # Detect Marimo: https://github.com/marimo-team/marimo/discussions/8865
+        #
+        # For now, anywidgets that are wapped don't work in Marimo;
+        # our _repr_mimebundle_ delegation does not work. So users will have
+        # to display ``canvas._backend`` until that is fixed.
+        # See https://github.com/manzt/anywidget/issues/792
+        try:
+            main_module.__marimo__
+            return True
+        except AttributeError:
+            pass
+
         try:
             # 'get_ipython' is available in globals when running from
             # IPython/Jupyter
--- a/vispy/app/application.py
+++ b/vispy/app/application.py
@@ -6,7 +6,6 @@
 
 from __future__ import division
 
-import __main__ as main_module
 import builtins
 import os
 import sys
@@ -125,11 +124,9 @@ class Application(object):
         # our _repr_mimebundle_ delegation does not work. So users will have
         # to display ``canvas._backend`` until that is fixed.
         # See https://github.com/manzt/anywidget/issues/792
-        try:
-            main_module.__marimo__
-            return True
-        except AttributeError:
-            pass
+        if "marimo" in sys.modules:
+            if sys.modules["marimo"].running_in_notebook():
+                return True
 
         try:
             # 'get_ipython' is available in globals when running from
--- a/vispy/app/application.py
+++ b/vispy/app/application.py
@@ -119,11 +119,6 @@ class Application(object):
         """Determine if the user is executing in a Jupyter/Marimo Notebook"""
 
         # Detect Marimo: https://github.com/marimo-team/marimo/discussions/8865
-        #
-        # For now, anywidgets that are wapped don't work in Marimo;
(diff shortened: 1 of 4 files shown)


vispy/app/application.py
```diff
@@ -116,7 +117,20 @@ def is_interactive(self):
...
+        #
+        # For now, anywidgets that are wapped don't work in Marimo;
+        # our _repr_mimebundle_ delegation does not work. So users will have
+        # to display ``canvas._backend`` until that is fixed.
+        # See https://github.com/manzt/anywidget/issues/792
+        try:
+            main_module.__marimo__
+            return True
+        except AttributeError:
+            pass
```
brisvag: An reason to not do `return hasattr(main_module, '__marimo__')`?

brisvag (commented):
I'm not really familiar with using vispy in jupyter. Do you have some snippet to test things out?

vispy/app/application.py
almarklein: Ah, good point, we could do:
```py
if hasattr(main_module, '__marimo__'):
    return True
```

Let's give it a day or two to see if someone responds to https://github.com/marimo-team/marimo/discussions/8865 for a better solution

almarklein: I replaced it with the recommended approach; there's actually a `marimo.running_in_notebook()` function 🚀

almarklein:
> I'm not really familiar with using vispy in jupyter. Do you have some snippet to test things out?

You can run this in a notebook (this is the vispy1.ipynb from the jupyter_rfb examples):

```py
from vispy import scene
from vispy.visuals.transforms import STTransform

canvas = scene.SceneCanvas(
    keys="interactive", bgcolor="white", size=(500, 400), show=True, resizable=True
)

view.camera = "arcball"

sphere1 = scene.visuals.Sphere(
    radius=1, method="latitude", parent=view.scene, edge_color="black"
)

sphere2 = scene.visuals.Sphere(
    radius=1, method="ico", parent=view.scene, edge_color="black"
)

sphere3 = scene.visuals.Sphere(
    radius=1,
    rows=10,
    cols=10,
    depth=10,
    method="cube",
    parent=view.scene,
    edge_color="black",
)

sphere1.transform = STTransform(translate=[-2.5, 0, 0])
sphere3.transform = STTransform(translate=[2.5, 0, 0])

view.camera.set_range(x=[-3, 3])
canvas.title = "Three spheres"
canvas
```

On Marimo: when I started this PR, anywidgets that are wapped didn't work in Marimo; vispy's `_repr_mimebundle_` delegation did not work, see https://github.com/manzt/anywidget/issues/792. This has been fixed. That said, we'd need a new release of Marimo for the above fix to be available, and also another fix which fixed that streaming to Marimo was super-slow.

Also worth noting that there are problems with Vispy and jupyter_rfb (also v0.5.4), see https://github.com/vispy/jupyter_rfb/issues/151. Not sure when that started. It seems related to an invalid context.

brisvag:
This looks good on marimo for me! But I can't check whether everything is fine on jupyter because I get a kernel crash. If I try to interact with the crashed notebook, I get a keyerror (but maybe unrelated)

```
  File "/home/brisvag/git/vispy/.venv/lib/python3.14/site-packages/ipykernel/subshell_manager.py", line 92, in get_shell_channel_to_subshell_pair
    return self._cache[subshell_id].shell_channel_to_subshell
           ~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: '213ffb3a-902d-40ff-8bd9-4503e62407f0'
```

It does not seem related to #2742, as it happens the same way there. I just see a gray square as canvas, and then it crashes.

brisvag:
Ok testing with `ipykernel<7` works; All seems good, merging!

PR #2742, opened 2026-04-09
Fix issues with context creation on Jupyter notebooks

almarklein:
Fixes https://github.com/vispy/jupyter_rfb/issues/151

This isn't particularly pretty, but much better then users (and downstream libs) creating extra vispy canvases to workaround the issue.

cc @gselzer

### DIFF
--- a/vispy/app/backends/_jupyter_rfb.py
+++ b/vispy/app/backends/_jupyter_rfb.py
@@ -81,7 +81,7 @@ class CanvasBackend(BaseCanvasBackend, RemoteFrameBuffer):
         self._context = OffscreenContext()  # OffscreenContext.get_global_instance()
         # Briefly create a new window to trigger something context related.
         # Not sure why this works, but it's part of the fix for  https://github.com/vispy/jupyter_rfb/issues/151
-        if sys.platform.stratswith('win'):
+        if sys.platform.startswith('win'):
             self._context.make_current()
             OffscreenContext().make_current().close()
         self._helper = FrameBufferHelper()
--- a/vispy/app/backends/_jupyter_rfb.py
+++ b/vispy/app/backends/_jupyter_rfb.py
@@ -78,6 +78,10 @@ class CanvasBackend(BaseCanvasBackend, RemoteFrameBuffer):
         # Use a context per canvas, because we seem to make assumptions
         # about OpenGL state being local to the canvas.
         self._context = OffscreenContext()  # OffscreenContext.get_global_instance()
+        # Briefly create a new window to trigger something context related.
+        # Not sure why this works, but it's part of the fix for  https://github.com/vispy/jupyter_rfb/issues/151
+        self._context.make_current()
+        OffscreenContext().make_current().close()
         self._helper = FrameBufferHelper()
         self._loop = asyncio.get_event_loop()
         self._logical_size = 1, 1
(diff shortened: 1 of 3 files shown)


djhoese (commented):
Do we know if `OffscreenContext` is used elsewhere in vispy or is it mostly for Jupyter-based workflows? Is there any chance that the issues being seen are related to the version of glfw?

This seems "fine" to me in a "get it to work" kind of way, but I haven't been a very active maintainer so I'm maybe not the most important opinion here (@brisvag ?).

almarklein:
> Do we know if `OffscreenContext` is used elsewhere in vispy or is it mostly for Jupyter-based workflows?

The jupyter_rfb backend is the only internal code that uses `OffscreenContext`, but we don't know what's used downstream. I refactored it a bit to make the change confined to the jupyter_rfb backend.

>  Is there any chance that the issues being seen are related to the version of glfw?

It also happens when I disbale the gflw path in OffscreenContext, so it uses Qt as a context provider. Sadly, the proposed fix only seems to work with glfw.

djhoese:
Technically the utility is in a "private" module so external uses aren't required to be supported. I'm OK with this PR being merged in a "I trust you" kind of way, not in a "I understand" kind of way. I'd like Lorenzo to comment before merging though.

brisvag (approved):
Eh, this makes as muich sense to me as it does to you xD I agree that "get it to work" here is more important thn holding this up.. it seems like a tricky bug to debug. I wouldn't worry about the uses of the private api, it still mostly untouched and shouldn't cause issues.

almarklein:
> it seems like a tricky bug to debug

It sure was 😅 

There is one (very long-running) CI job that keeps failing. Funny enough it fails with exactly the error that this PR attempts to fix. That ... sucks.

almarklein:
The CI builds spend 3 hours on installing dependencies with conda, what's up with that?

brisvag:
I don't know... I haven't been on top of things the last couple months so there's some CI/dependabot stuff to take care of, but I'm not sure if it's related. I'll try to take a look at it soon!

djhoese:
Ah mamba isn't as widely used. We could try using miniconda which is equivalent to mamba performance in modern versions. Here's what I have for my Satpy project:

```yaml
      - name: Setup Conda Environment
        uses: conda-incubator/setup-miniconda@v3
        with:
          miniforge-version: latest
          python-version: ${{ matrix.python-version }}
          activate-environment: test-environment
          channels: conda-forge
          conda-remove-defaults: true
          channel-priority: strict
```

djhoese:
I'm not saying that will fix things but :man_shrugging:

vispy/app/backends/_jupyter_rfb.py
```diff
@@ -78,6 +79,11 @@ def __init__(self, vispy_canvas, **kwargs):
         # Use a context per canvas, because we seem to make assumptions
         # about OpenGL state being local to the canvas.
         self._context = OffscreenContext()  # OffscreenContext.get_global_instance()
+        # Briefly create a new window to trigger something context related.
+        # Not sure why this works, but it's part of the fix for  https://github.com/vispy/jupyter_rfb/issues/151
+        if sys.platform.stratswith('win'):
```
brisvag:
```suggestion
        if sys.platform.startswith('win'):
```

almarklein: Woops!

brisvag:
Since this is fixing a problem on windows, and my issue (same as the one from https://github.com/vispy/jupyter_rfb/issues/121) already exists on main and is unchanged from this PR, I'll say let's merge. We can tackle the remaning issue separately.

PR #2726, opened 2026-02-23
Delay importing networkx to NetworkxCoordinates.__init__

psobolewskiPhD:
The main motivation is napari, where we use a lot of `vispy.visuals` imports which pulls along networkx, but we don't use `vispy.visuals.graphs.layouts.networkx_layout` so it's a waste of time for our users. Using import timings of napari, networkx is a ~100 ms import that is eagerly triggered by any `vispy.visuals` import, e.g. on main:
`python -X importtime -c 'from vispy.visuals import VolumeVisual' 2>&1 | grep networkx`:
`import time:      6083 |      56805 |           networkx`
(I'm not sure why it's 60 ms in a pure vispy env vs napari import)

And the total import (on main, 2nd run): `python -X importtime -c 'from vispy.visuals import VolumeVisual'`
`import time:       194 |     230860 | vispy.visuals`

In this PR the import is moved into the class, so it's not eagerly imported:
```
python -X importtime -c 'from vispy.visuals import VolumeVisual' 2>&1 | grep networkx
import time:       847 |        847 |         vispy.visuals.graphs.layouts.networkx_layout
```
And the total import 2nd run): `python -X importtime -c 'from vispy.visuals import VolumeVisual'`
`import time:       195 |     172709 | vispy.visuals`

PR #2744, opened 2026-04-18
[maint] Defer importing scipy.sparse

psobolewskiPhD:
Followup to: https://github.com/vispy/vispy/pull/2726
part of the efforts to improve napari startup time: https://github.com/napari/napari/issues/8689

Importing a visual also imports scipy.sparse via
vispy.visuals ->
  vispy.visuals.graphs -> vispy.visuals.graphs.layouts.random -> vispy.visuals.graphs.util -> scipy.sparse

In this PR I remove the duplicated code in `force_directed.py` and defer the import of scipy.sparse.

main, second run:
`python -X importtime -c 'from vispy.visuals import VolumeVisual' 2>&1 | grep scipy`
`import time:       250 |      73988 |             scipy.sparse`

this branch same command returns nothing.

### DIFF
--- a/vispy/visuals/graphs/util.py
+++ b/vispy/visuals/graphs/util.py
@@ -14,7 +14,7 @@ def issparse(obj):
     try:
         from scipy.sparse import issparse as _issparse
         return _issparse(obj)
-    except ImportError:
+    except ModuleNotFoundError:
         return False
 
 
--- a/vispy/visuals/graphs/util.py
+++ b/vispy/visuals/graphs/util.py
@@ -10,11 +10,11 @@ A module containing several graph utility functions.
 
 import numpy as np
 
-try:
-    from scipy.sparse import issparse
-    from scipy import sparse
-except ImportError:
-    def issparse(*args, **kwargs):
+def issparse(obj):
+    try:
+        from scipy.sparse import issparse as _issparse
+        return _issparse(obj)
+    except ImportError:
         return False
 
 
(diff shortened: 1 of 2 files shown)


vispy/visuals/graphs/util.py
```diff
@@ -10,11 +10,11 @@
...
-try:
-    from scipy.sparse import issparse
-    from scipy import sparse
-except ImportError:
-    def issparse(*args, **kwargs):
+def issparse(obj):
+    try:
+        from scipy.sparse import issparse as _issparse
+        return _issparse(obj)
+    except ImportError:
```
brisvag, line 17: I think in general it's preferred to use `ModuleNotFoundError` for this, because it's more specific. We still want to surface the error if something else goes wrong during scipy import.

psobolewskiPhD: I thought it was better the other way around since ImportError is broader it will catch more potential issues and ensure the code works correctly?
Given that scipy is not a dependency, it feels better to lean on the safer side rather than more specific side?

Happy to make the change, so this is mostly for my edification.

djhoese: I think I agree with @brisvag on this. If there is a `.so` issue in scipy while being imported or the `sparse` sub-package is renamed then I think we'd want to know about that as soon as possible. Although maybe the `sparse` rename is still a `ModuleNotFoundError` but `issparse` being renamed would be a plain `ImportError`.

psobolewskiPhD:
```suggestion
    except ModuleNotFoundError:
```

djhoese (approved):

brisvag (approved):

Issue #2746, opened 2026-04-24
Unable to change alpha value in vispy.util.svg.Color


akahana06:
There is no way to set the alpha value of a utility.svg.color object as content.strip only assigns r, g and b.

From vispy.util.svg.color.py
`class Color(object):

    def __init__(self, content):

        color = content.strip()
        if color.startswith("#"):
            rgb = color[1:]
            if len(rgb) == 3:
                r, g, b = tuple(ord((c + c).decode('hex')) for c in rgb)
            else:
                # r,g,b = tuple(ord(c) for c in rgb.decode('hex'))
                r, g, b = tuple(c for c in _rgb(rgb))
        elif color.startswith("rgb("):
            rgb = color[4:-1]
            r, g, b = [value.strip() for value in rgb.split(',')]
            if r.endswith("%"):
                r = 255 * int(r[:-1]) // 100
            else:
                r = int(r)
            if g.endswith("%"):
                g = 255 * int(g[:-1]) // 100
            else:
                g = int(r)
            if b.endswith("%"):
                b = 255 * int(b[:-1]) // 100
            else:
                b = int(r)
        elif color in _keyword_colors:
            r, g, b = _keyword_colors[color]
        else:
            # text = "Unknown color (%s)" % color
            r, g, b = 0, 0, 0
        self._rgb = r / 255., g / 255., b / 255.

    @property
    def rgb(self):
        r, g, b = self._rgb
        return r, g, b

    @property
    def rgba(self):
        r, g, b = self._rgb
        return r, g, b, 1

    def __repr__(self):
        r, g, b = self._rgb
        r, g, b = int(r * 255), int(g * 255), int(b * 255)
        return "#%02x%02x%02x" % (r, g, b)
`

brisvag:
Hi @akahana06! Yes, this is very old code that seems to only handle rgb data, no alpha. I think the whole svg machinery is like this. If you would like for this to work with alpha, it's quite a bit of work, and unfortunately no one is really working on this, so you would have to open a pull request and make the changes yourself. I'd be happy to help with reviews, but this is not a part of the codebase that I have time to look into myself, sorry!

PR #2747, opened 2026-04-26
fix: util/svg/color rgba functionality + bug fix

akahana06:
I ran into this issue when trying to find which "Color" class is used in vispy.visuals. There was no way to set the alpha value in ._rgba + 'g' and 'b' values were incorrectly assigned "int(r)".

### DIFF
--- a/vispy/util/svg/color.py
+++ b/vispy/util/svg/color.py
@@ -191,17 +191,37 @@ class Color(object):
             if g.endswith("%"):
                 g = 255 * int(g[:-1]) // 100
             else:
-                g = int(r)
+                g = int(g)
             if b.endswith("%"):
                 b = 255 * int(b[:-1]) // 100
             else:
-                b = int(r)
+                b = int(b)
+        elif color.startswith("rgba("):
+            rgba = color[5:-1]
+            r, g, b, a = [value.strip() for value in rgba.split(',')]
+            if r.endswith("%"):
+                r = 255 * int(r[:-1]) // 100
+            else:
+                r = int(r)
+            if g.endswith("%"):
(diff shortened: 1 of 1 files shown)


vispy/util/svg/color.py
```diff
@@ -191,17 +191,37 @@ def __init__(self, content):
...
+                a = 255 * int(a[:-1]) // 100
+            else:
+                a = int(a)
         elif color in _keyword_colors:
             r, g, b = _keyword_colors[color]
         else:
             # text = "Unknown color (%s)" % color
             r, g, b = 0, 0, 0
         self._rgb = r / 255., g / 255., b / 255.
+        self._rgba = r / 255., g / 255., b / 255., a / 255. if 'a' in locals() else 1
```
djhoese, line 224: I'm not a big fan of this `locals()` usage. How about defining `a = 1` at the top. Then what about only having `self._rgba` defined and just drop the alpha everywhere `self._rgb` is used currently (ex. `self._rgba[:3]`)?

djhoese (changes requested):
Wow, how was this never caught before? Nice catch. Just a small change.

vispy/util/svg/color.py
brisvag: Yep, +1 on having always rgba with a=1 unless specified!

brisvag (commented):
Nice, thanks @akahana06! Definitely agree with @djhoese on the `locals()` use, the rest is great! For context, this closes #2746.

brisvag:
@akahana06 are you still interested in pushing this forward to get it merged? :)

PR #2748, opened 2026-05-18
Set depth buffer for gridlines

brisvag:
Currently the gridlines visual does not set the depth buffer. This is not important in 2D, buut when using it in 3D the depth is set to a plane parallel to the screen (cause of the impostor rendering as a screen-covering quad). This PR sets the depth buffer correctly, so this is no longer a problem.

Here's an example:

```py
from vispy import scene
import numpy as np

canvas = scene.SceneCanvas(show=True)

view.camera = 'arcball'
view.camera.scale_factor = 30

grid = scene.GridLines(parent=view.scene)
grid.transform = scene.STTransform(translate=(0, 0, -5))
grid.update_gl_state(preset='translucent')

cube = scene.Cube(size=10, parent=view.scene)
cube.transform = scene.STTransform(translate=(0, 0, 5))
cube.update_gl_state(preset='translucent')
```

On main, you'll see no cube, regardless of rotation, cause the grid wind the depth test all the time. On this pr, the grid and cube exist in space as expected.

image

### DIFF
--- a/vispy/visuals/gridlines.py
+++ b/vispy/visuals/gridlines.py
@@ -18,6 +18,7 @@ uniform float u_border_width;
 vec4 grid_color(vec2 pos) {
     vec4 px_pos = $map_to_doc(vec4(pos, 0, 1));
     px_pos /= px_pos.w;
+    gl_FragDepth = px_pos.z * 0.5 + 0.5;
 
     // Compute vectors representing width, height of pixel in local coords
     vec4 local_pos = $map_doc_to_local(px_pos);
--- a/vispy/visuals/tests/test_gridlines.py
+++ b/vispy/visuals/tests/test_gridlines.py
@@ -8,7 +8,7 @@ Tests for GridLines
 
 import numpy as np
 
-from vispy.scene import GridLines, STTransform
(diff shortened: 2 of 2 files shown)


djhoese (approved):
A test would be nice, but looks good.

vispy/visuals/gridlines.py
```diff
@@ -18,6 +18,7 @@
 vec4 grid_color(vec2 pos) {
     vec4 px_pos = $map_to_doc(vec4(pos, 0, 1));
     px_pos /= px_pos.w;
+    gl_FragDepth = px_pos.z * 0.5 + 0.5;
```
djhoese, line 21: Actually, what is with the 0.5s?

brisvag: The position is in [-1, 1] clip/ndc range, but we need it in frag depth world which goes from 0 to 1!

djhoese: Perfect. Thanks.

brisvag:
Test added!

brisvag:
Looks like the switch to miniconda worked! Merging :muscle:

PR #2749, opened 2026-05-19
Use miniconda in ci

brisvag:
Context: https://github.com/vispy/vispy/pull/2742#issuecomment-4254340527

Replacing micromamba (which is often hanging) with miniconda.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -126,7 +126,7 @@ jobs:
         auto-activate: false
     - name: Install conda-dependencies
       run: |
-        conda install -y python=3.9 -f ci/requirements/linux_full_deps_conda.txt -c conda-forge
+        conda install -y python=3.9 -f ci/requirements/linux_full_deps_conda.txt
     - name: Install pip dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


brisvag:
doesn't seem to solve the hang in the website build :/

djhoese:
Ugh. I've seen some weird things in other projects where it complained about "Cython" not being available and it turned out we had "Cython" and not "cython" in our environment YAML file. But I don't think I've seen this...ugh

djhoese:
So one thing is it seems that the website building using Python 3.10 to create the environment, but then the `install` of the dependencies specified `python=3.9`. Maybe remove that?

.github/workflows/main.yml
```diff
@@ -113,12 +113,20 @@ jobs:
...
+        channels: conda-forge
+        conda-remove-defaults: true
+        channel-priority: strict
+        activate-environment: vispy-tests
         environment-file: ./ci/requirements/py${{ steps.vars.outputs.python_version }}.yml
+        auto-activate: false
     - name: Install conda-dependencies
       run: |
-        micromamba install -y python=3.9 -f ci/requirements/linux_full_deps_conda.txt -c conda-forge
+        conda install -y python=3.9 -f ci/requirements/linux_full_deps_conda.txt
```
djhoese:
```suggestion
        conda install -y -f ci/requirements/linux_full_deps_conda.txt
```

brisvag:
Seems to maybe have worked! LEt's merge and see if things are better on the other PR!

PR #2757, opened 2026-07-02
Use np.reshape instead of setting shape attr

gselzer:
NumPy deprecated setting the shape attribute with [`v2.5.0`](https://github.com/numpy/numpy/releases/tag/v2.5.0). This PR fixes the warnings arising from doing so within the vispy codebase, or at least the ones I found by running `git grep "\.shape\s=[^=]"`

The current calls to `reshape` only copy the underlying data if a copy-free reshape is impossible. I do not fully understand when that happens, but if a copy would be a problem we could explicitly pass `copy=False` - let me know what you think!

### DIFF
--- a/vispy/geometry/calculations.py
+++ b/vispy/geometry/calculations.py
@@ -149,8 +149,12 @@ def resize(image, shape, kind='linear'):
 
         c_1 = np.minimum(c_1, image.shape[1] - 1)
         r_1 = np.minimum(r_1, image.shape[0] - 1)
-        for arr in (top, bot, lef, rig):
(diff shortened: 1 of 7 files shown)


djhoese (approved):
This looks good to me. Thanks for making the changes. I think we'd rather have a copy when necessary rather than error out. If a user ends up diving into performance and discovers the copy then it is likely something they have to change on their end anyway. As a test, it could be interesting to force `copy=False` and see what fails in our tests, but I don't think that's something we *need* to do before merging this.

brisvag (approved):

PR #2758, opened 2026-07-02
Add RGB volume rendering support

kephale:
VolumeVisual was hardcoded to single-channel (luminance) data. This adds support for RGB volumes (z, y, x, 3) by:

- Removing the hardcoded format='luminance' in _create_texture, letting the texture infer format from data shape
- Adding a u_rgb_mode uniform to branch the shader between scalar (colormap lookup) and RGB (direct color with per-channel clim/gamma)
- Changing colorToVal to use perceptual luminance for RGB gradient calcs
- Updating applyColormap to accept vec4 and handle both paths
- Updating all rendering snippets (MIP, MINIP, translucent, additive, ISO, average) to pass vec4 color through the pipeline
- Seeding the MIP/MINIP refinement's maxcolor/mincolor by re-sampling the coarse pass' extremum location (not a fix — main only needed the scalar `maxval`, which already carried that value; the color has to be re-sampled because it can't be recovered from the luminance)

Scalar rendering is unchanged: renders are pixel-identical to main across all seven methods at gamma 1.0 and 2.2.

Includes an RGB volume example and tests for RGB data acceptance and rendering across multiple methods. The volume example looks like:
image


psobolewskiPhD:
Exciting!

vispy/visuals/volume.py
```diff
@@ -123,14 +124,23 @@
 
 float colorToVal(vec4 color1)
 {
-    return color1.r; // todo: why did I have this abstraction in visvis?
+    if (u_rgb_mode == 1)
+        return dot(color1.rgb, vec3(0.2126, 0.7152, 0.0722));
+    return color1.r;
```
brisvag, line 131: I love how this PR answers the comment by deleting it :P Adding some comments for future us!


```suggestion
    if (u_rgb_mode == 1)
    	// perceptual luminance, a good approach to transform rgb into a single
    	// value for things like gradient calculations in isosurface
        return dot(color1.rgb, vec3(0.2126, 0.7152, 0.0722));
    return color1.r;
```

vispy/visuals/volume.py
```diff
@@ -123,14 +124,23 @@
...
+        color.rgb = pow(color.rgb, vec3(gamma));
+        color.a = 1.0;
+        return color;
+    }
+    float data = color.r;
     data = clamp(data, min(clim.x, clim.y), max(clim.x, clim.y));
     data = (data - clim.x) / (clim.y - clim.x);
-    vec4 color = $cmap(pow(data, gamma));
-    return color;
+    return $cmap(pow(data, gamma));
```
brisvag: I found this a bit confusing to understand where the differences actually lie. What about this? Or are we sacrificing performance by operating on `rgb` when unnecessary?

```suggestion
vec4 applyColormap(vec4 color) {
    color.rgb = clamp(color.rgb, min(clim.x, clim.y), max(clim.x, clim.y));
    color.rgb = (color.rgb - clim.x) / (clim.y - clim.x);
    color.rgb = pow(color.rgb, vec3(gamma));
    color.a = 1.0;
    if (u_rgb_mode == 1) {
        // no colormapping, just gamma and clim
        return color;
    } else {
        return $cmap(color.r);
    }
}
```

vispy/visuals/volume.py
```diff
@@ -389,16 +399,19 @@
             vec3 max_loc_tex = start_loc_refine;
 
             vec3 small_step = step * 0.1;
+            vec4 maxcolor = $get_data(start_loc + step * float(maxi));
```
brisvag, line 406: I guess this is what you refer to here:

> Fixing MIP/MINIP refinement to initialize maxcolor/mincolor from the coarse-pass max location rather than the refinement start

Are you saying this is an actual bug in main and unrelated to this? Like, on main this line should be added and should be:

```glsl
float maxval = $get_data(start_loc + step * float(maxi));
```
?

brisvag (approved):
The rgb volume is so trippy! Very cool :) In general this looks good!

I left a couple suggestions and one question. Other than that, I think this can go in!

vispy/visuals/volume.py
kephale: Done, comment added (with spaces instead of tabs to match the rest of the shader).

vispy/visuals/volume.py
kephale: Agreed, much clearer — took it, with the shared math hoisted above the branch so the *only* difference is the colormap lookup:

```glsl
vec4 applyColormap(vec4 color) {
    // clim and gamma are applied identically in both modes: per channel for rgb
    // data, and to the single (red) channel for scalar data.
    color.rgb = clamp(color.rgb, min(clim.x, clim.y), max(clim.x, clim.y));
    color.rgb = (color.rgb - clim.x) / (clim.y - clim.x);
    color.rgb = pow(color.rgb, vec3(gamma));
    if (u_rgb_mode == 1) {
        // no colormapping, the rgb values are used directly
        return vec4(color.rgb, 1.0);
    }
    return $cmap(color.r);
}
```

On perf: yes, scalar mode now does the clamp/scale/`pow` on 3 components instead of 1, so ~2 extra `pow` per `applyColormap` call. That is per-sample for `translucent`/`additive`, so not free, but it is a couple of ALU ops next to a 3D texture fetch per sample, which dominates. I think readability wins here; happy to split it back if you would rather not pay it at all.

vispy/visuals/volume.py
kephale: Good question — no, this is not a bug on main, and my PR description was misleading. Sorry, I have reworded it.

On main the coarse loop already carries the winning *value* in `maxval`, and the refinement only overwrites it when it finds something higher, so `applyColormap(maxval)` was correct as-is. The line only becomes necessary here because we now need the winning **color**, and for rgb data you cannot reconstruct it from the luminance. So `maxcolor` has to be seeded from the coarse maximum's location (the refinement then replaces it only if it beats it), which is exactly the initialization `maxval` already had implicitly. I added a comment saying so.

While double-checking this, though, I did find a real regression **in this PR**: for `attenuated_mip` I had used `applyColormap($get_data(max_loc_tex))`, which silently dropped the `scale` factor that main applied via `maxval = val * scale`. Fixed by tracking the attenuated color in the loop instead:

```glsl
} else if( val * scale > maxval ) {
    maxval = val * scale;
    maxcolor = color * scale;
    ...
```

which also avoids the extra texture fetch. I verified renders are now pixel-identical to main (max abs diff 0) across all seven methods at gamma 1.0 and 2.2; before the fix `attenuated_mip` differed by up to 105/255. Also added `minip` and `attenuated_mip` to the parametrized RGB render test.

kephale:
demo_grid_rgbvolume_vispy


okey and with an updated screenshot

vispy/visuals/volume.py
brisvag: Ugh, github adds tabs instead of spaces when typing in a comment here? sad :/

vispy/visuals/volume.py
brisvag: Yeah I had a feeling it was negligible. Let's keep it like this!

brisvag (approved):
Ok, LGTM!

djhoese:
@brisvag I know very little about volume rendering so if you're good with these changes then feel free to merge and don't wait for my approval.

brisvag:
Alright, let's get this in! I'll just wait for the CI to be fixed via the other PRs, just in case :P

PR #2759, opened 2026-07-03
Implement _compute_bounds for InstancedMeshVisual

kephale:
Wrapping up this TODO for instanced meshes! Test included.


djhoese (commented):
Thanks for trying to get this completed.

It's been a while since I've looked at how `_compute_bounds` is used, but would it be worth it to cache this in an instance variable? Or is that already done internally? I'm mostly worried about all the numpy operations and list comprehensions being very slow if someone is simply panning/zooming around.

kephale:
Awesome, thanks @djhoese 🙏 . Some of the caching does seem to already work internally. There was an extra caching across axis that I added though!

brisvag (approved):
Finally got around to this! Looking great, and from some quick manual tests it looks ok.

@kephale did you have a specific use case where this was failing? I'd like to be able to tests these bounds calculations but I'm always unsure on when they exactly come into play...

PR #2761, opened 2026-07-15
Add GaussianSplat visual and example

aganders3:
I'm at SciPy this week and met a user in civil engineering provide me with some data of a bridge they'd like to visualize. This was a .ply file, but I was surprised to find it contained Gaussian splat data, not a standard mesh. Unfortunately I can't share this user's data, but I was able to munge it a bit and display it as a colored point cloud, which actually looked pretty good.

This helped me understand Gaussian splats in a way I had not before. Combined with my recent experience implementing instanced-rendering for the `Markers` visual, this gave me a very clear idea for implementing Gaussian splatting in vispy.

I relied on Claude Code for a few key bits of math in the shader, but overall this is pretty understandable and readable.

Note that for now this only supports a single color per-splat. Gaussian splats often provide colors as _spherical harmonics_ to create orientation-dependent colors. Think specular highlights, pearlescent or color-shifting pigment, etc. The current example just extracts the 0th spherical harmonic term and uses that as the color. This is another place I don't fully understand the math, but I don't think it's a huge lift to add it later.

For the example here, I uploaded some data to a fork of the vispy demo data repo. You can find more demo data at https://superspl.at/.

This PR is draft at least until it points to the real demo data repo. I'd also like to add some tests.

#### ~26k splats (6 MB):

https://github.com/user-attachments/assets/d426d7c0-5ab7-47d8-9be0-16a8f3877be8

#### ~3.5M splats (827 MB):

https://github.com/user-attachments/assets/97cd1369-1239-4682-ba1d-31bf7fc4ea78

NB @nclack @royerloic


vispy/util/fetching.py
```diff
@@ -42,7 +42,7 @@ def load_data_file(fname, directory=None, force_download=False):
     fname : str
         The path to the file on the local system.
     """
-    _url_root = 'https://raw.githubusercontent.com/vispy/demo-data/main/'
+    _url_root = 'https://raw.githubusercontent.com/aganders3/vispy-demo-data/gaussian-splatting/'
```
brisvag, line 45: I guess this is temporary :P

brisvag (commented):
Finally looking into this, so fun! What's your suggestion going forward @aganders3? We just merge it like this as a new visual?

I get the general gist; I definitely don't understand all the math regarding the covariance and projection stuff, but it seems to Work™, and it's a new visual so it won't break anything. So IMO we can totally get it in and iterate 🤷

aganders3:
Thanks!

Merge and iterate is pretty much what I'm thinking. We could make a note or change the name to indicate it's experimental or something if you'd like.

The one thing that needs to change is the URL you pointed out. We need to merge the sample data into the vispy data repo first.

The only other thing I'd consider adding here would be the orientation-dependent color. I got it working with LLM-assistance but I need to think about the API for providing the higher-order colors, etc.

brisvag:
I'll let you think on it and leave this parked for now! :rocket:

PR #2764, opened 2026-07-25
Optimizing Axis Performance

baoshuang:
Optimize axis._get_ticks_talbot functions with Numba decorators


djhoese:
Correct me if I'm wrong, but vispy doesn't currently depend on numba. I've seen improvements in some of my past projects involving vispy where I numba jit'd some coordinate transformations so this improvement doesn't surprise me. However, I think numba is too big of a dependency (@brisvag feel free to overrule me) to be a hard requirement of vispy. My preference would be to have a `_compat.py` module or something that attempts to import the jit function and if it doesn't exist then provide a passthrough/no-op function that doesn't modify the function being decorated. This way we get the benefit of numba as an optional dependency, but users who can't install it for their environment don't lose out on vispy entirely.

djhoese:
...and do you have any performance numbers on how things improved?

brisvag:
Definitely should be optional, and we need to see some numbers :)

djhoese:
Oh you could even have both versions exist since I see some of the implementation may differ for numba versus numpy-only where you could do something like:

```python
try:
    from numba import jit
except ImportError:
    def jit(*args, **kwargs):
        def _wrapper(func):
            return func
        return _wrapper

@jit(fastmath=True)
def _some_func_numba(...):
    ...

def _some_func_numpy(...):
    ...

some_func = _some_func_numpy if jit is None else _some_func_numba
```

PR #2765, opened 2026-07-25
Fix KeyError when releasing extra mouse buttons on Qt backend

### LINKED ISSUE #2751: Bug: App crashes when releasing extra mouse button
This line should also use `BUTTONMAP.get(ev.button(), 0)`, otherwise you get a key error for any of the extra mouse buttons.

https://github.com/vispy/vispy/blob/9c3fa07d0832d1d7c2a945ae113db5f4d3189d56/vispy/app/backends/_qt.py#L545

Phoenix0531-sudo:
The original crash was a hard `BUTTONMAP[ev.button()]` lookup in the Qt `mouseReleaseEvent` path. Press and double-click already used a safe lookup, so releasing an unmapped extra mouse button could raise `KeyError` and crash the app. This PR makes release use the same fallback path (`0`, VisPy’s no/unknown button) and keeps the Qt button-state helper driven from the same map.

While testing the fix, the PR now also maps Qt `ExtraButton3..24` when those enum members exist, preserving the existing VisPy IDs 1–5 for Left/Right/Middle/Back/Forward and assigning extra buttons 6–27. This keeps side-button events distinguishable instead of collapsing mapped extra buttons to `0`, while still falling back to `0` for truly unknown/unmapped buttons.

The regression tests now exercise real Qt backend mouse handlers for press, release, and double-click, including standard buttons, extra buttons, and the unknown-release fallback. The PyQt6 imports are inside the PyQt6-guarded tests/helpers so documentation/test collection still works when PyQt6 is not installed.

Closes #2751

### DIFF
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -381,6 +381,29 @@ def _get_qpoint_pos(pos):
     return pos.x(), pos.y()
 
 
+def _buttonmap_to_list(buttons) -> list[int]:
+    """Map Qt mouse button flags to VisPy button ids.
+
+    Module-level (no ``self``) so tests can call it without a backend
+    instance. Preserves BUTTONMAP order: standard 1-5, then
+    ExtraButton3..24 as 6-27 when present on this Qt build.
+    """
+    if buttons == QtCore.Qt.MouseButton.NoButton:
+        return []
+
+    mouse_buttons = []
+    if PYQT6_API or PYSIDE6_API:
+        for qt_button, vispy_button in BUTTONMAP.items():
+            if vispy_button and qt_button in buttons:
+                mouse_buttons.append(vispy_button)
+    else:
+        for qt_button, vispy_button in BUTTONMAP.items():
+            if vispy_button and (buttons & qt_button) != QtCore.Qt.MouseButton.NoButton:
+                mouse_buttons.append(vispy_button)
+
+    return mouse_buttons
+
+
 class QtBaseCanvasBackend(BaseCanvasBackend):
     """Base functionality of Qt backend. No OpenGL Stuff."""
 
@@ -511,23 +534,6 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
     def sizeHint(self):
         return self.size()
 
-    def _buttonmap_to_list(self, buttons) -> list[int]:
-        if buttons == QtCore.Qt.MouseButton.NoButton:
-            return []
-
-        # Preserve BUTTONMAP order (standard 1-5, then ExtraButton3..24 as 6-27).
-        mouse_buttons = []
-        if PYQT6_API or PYSIDE6_API:
-            for qt_button, vispy_button in BUTTONMAP.items():
-                if vispy_button and qt_button in buttons:
-                    mouse_buttons.append(vispy_button)
-        else:
-            for qt_button, vispy_button in BUTTONMAP.items():
-                if vispy_button and (buttons & qt_button) != QtCore.Qt.MouseButton.NoButton:
-                    mouse_buttons.append(vispy_button)
-
-        return mouse_buttons
-
     def mousePressEvent(self, ev):
         if self._vispy_canvas is None:
             return
@@ -536,7 +542,7 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
             native=ev,
             pos=_get_event_xy(ev),
             button=BUTTONMAP.get(ev.button(), 0),
-            buttons=self._buttonmap_to_list(ev.buttons()),
+            buttons=_buttonmap_to_list(ev.buttons()),
             modifiers=self._modifiers(ev),
         )
         # If vispy did not handle the event, clear the accept parameter of the qt event
@@ -550,7 +556,7 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
             native=ev,
             pos=_get_event_xy(ev),
             button=BUTTONMAP.get(ev.button(), 0),
-            buttons=self._buttonmap_to_list(ev.buttons()),
+            buttons=_buttonmap_to_list(ev.buttons()),
             modifiers=self._modifiers(ev),
         )
         # If vispy did not handle the event, clear the accept parameter of the qt event
@@ -564,7 +570,7 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
             native=ev,
             pos=_get_event_xy(ev),
             button=BUTTONMAP.get(ev.button(), 0),
-            buttons=self._buttonmap_to_list(ev.buttons()),
+            buttons=_buttonmap_to_list(ev.buttons()),
             modifiers=self._modifiers(ev),
         )
         # If vispy did not handle the event, clear the accept parameter of the qt event
@@ -578,7 +584,7 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
         vispy_event = self._vispy_mouse_move(
             native=ev,
             pos=_get_event_xy(ev),
-            buttons=self._buttonmap_to_list(ev.buttons()),
+            buttons=_buttonmap_to_list(ev.buttons()),
             modifiers=self._modifiers(ev),
         )
         # If vispy did not handle the event, clear the accept parameter of the qt event
@@ -607,7 +613,7 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
             native=ev,
             delta=(deltax, deltay),
             pos=_get_event_xy(ev),
-            buttons=self._buttonmap_to_list(ev.buttons()),
+            buttons=_buttonmap_to_list(ev.buttons()),
             modifiers=self._modifiers(ev),
         )
         # If vispy did not handle the event, clear the accept parameter of the qt event
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -193,16 +193,29 @@ KEYMAP = {
     qt_keys.Key_Tab: keys.TAB,
 }
 if PYQT6_API or PYSIDE6_API:
+    # VisPy buttons 1-5 are Left/Right/Middle/Back/Forward.
+    # Qt also exposes ExtraButton1..24 (ExtraButton1/2 alias Back/Forward;
+    # ExtraButton3 is TaskButton). Map ExtraButton3..24 to VisPy 6..27 so
+    # unmapped side/extra buttons stay distinguishable instead of collapsing
+    # to 0 via BUTTONMAP.get(..., 0).
     BUTTONMAP = {
         QtCore.Qt.MouseButton.NoButton: 0,
         QtCore.Qt.MouseButton.LeftButton: 1,
         QtCore.Qt.MouseButton.RightButton: 2,
         QtCore.Qt.MouseButton.MiddleButton: 3,
         QtCore.Qt.MouseButton.BackButton: 4,
-        QtCore.Qt.MouseButton.ForwardButton: 5
+        QtCore.Qt.MouseButton.ForwardButton: 5,
     }
+    for _extra_i in range(3, 25):
+        BUTTONMAP[getattr(QtCore.Qt.MouseButton, f'ExtraButton{_extra_i}')] = (
+            _extra_i + 3
+        )
 else:
+    # Qt5 bit values: No=0, Left=1, Right=2, Mid=4, X1/Back=8, X2/Fwd=16,
+    # then ExtraButton3..24 as 1 << (n + 1) for n=3..24.
     BUTTONMAP = {0: 0, 1: 1, 2: 2, 4: 3, 8: 4, 16: 5}
+    for _extra_i in range(3, 25):
+        BUTTONMAP[1 << (_extra_i + 1)] = _extra_i + 3
 
 
 # Properly log Qt messages
@@ -494,30 +507,16 @@ class QtBaseCanvasBackend(BaseCanvasBackend):
         if buttons == QtCore.Qt.MouseButton.NoButton:
             return []
 
+        # Preserve BUTTONMAP order (standard 1-5, then ExtraButton3..24 as 6-27).
         mouse_buttons = []
-
         if PYQT6_API or PYSIDE6_API:
-            if QtCore.Qt.MouseButton.LeftButton in buttons:
-                mouse_buttons.append(1)
-            if QtCore.Qt.MouseButton.RightButton in buttons:
-                mouse_buttons.append(2)
-            if QtCore.Qt.MouseButton.MiddleButton in buttons:
-                mouse_buttons.append(3)
-            if QtCore.Qt.MouseButton.BackButton in buttons:
-                mouse_buttons.append(4)
-            if QtCore.Qt.MouseButton.ForwardButton in buttons:
-                mouse_buttons.append(5)
+            for qt_button, vispy_button in BUTTONMAP.items():
+                if vispy_button and qt_button in buttons:
+                    mouse_buttons.append(vispy_button)
         else:
-            if buttons & QtCore.Qt.MouseButton.LeftButton != QtCore.Qt.MouseButton.NoButton:
-                mouse_buttons.append(1)
-            if buttons & QtCore.Qt.MouseButton.RightButton != QtCore.Qt.MouseButton.NoButton:
-                mouse_buttons.append(2)
-            if buttons & QtCore.Qt.MouseButton.MiddleButton != QtCore.Qt.MouseButton.NoButton:
-                mouse_buttons.append(3)
-            if buttons & QtCore.Qt.MouseButton.BackButton != QtCore.Qt.MouseButton.NoButton:
-                mouse_buttons.append(4)
-            if buttons & QtCore.Qt.MouseButton.ForwardButton != QtCore.Qt.MouseButton.NoButton:
-                mouse_buttons.append(5)
+            for qt_button, vispy_button in BUTTONMAP.items():
+                if vispy_button and (buttons & qt_button) != QtCore.Qt.MouseButton.NoButton:
+                    mouse_buttons.append(vispy_button)
 
         return mouse_buttons
 
--- a/vispy/app/backends/_qt.py
+++ b/vispy/app/backends/_qt.py
@@ -194,10 +194,12 @@ KEYMAP = {
 }
 if PYQT6_API or PYSIDE6_API:
     # VisPy buttons 1-5 are Left/Right/Middle/Back/Forward.
-    # Qt also exposes ExtraButton1..24 (ExtraButton1/2 alias Back/Forward;
-    # ExtraButton3 is TaskButton). Map ExtraButton3..24 to VisPy 6..27 so
-    # unmapped side/extra buttons stay distinguishable instead of collapsing
-    # to 0 via BUTTONMAP.get(..., 0).
+    # Qt 5.0+ also exposes ExtraButton1..24 (ExtraButton1/2 alias Back/Forward;
+    # ExtraButton3 is TaskButton). They were not present in Qt 4. Map any
+    # ExtraButton3..24 that exist on this Qt build to VisPy 6..27 so extra
+    # buttons stay distinguishable instead of collapsing to 0.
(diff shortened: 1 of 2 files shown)


djhoese:
Thanks for this fix and I'm glad it was a simple work around. Looking at modern Qt docs it looks like there are a lot of "Extra" buttons in the enum:

https://doc.qt.io/qtforpython-6/PySide6/QtCore/Qt.html#PySide6.QtCore.Qt.MouseButton

What do you think about adding those to the button map so that if someone wanted to map those to a purpose they could? Otherwise with the way you have it it would always be "0" for all those buttons, right?

vispy/app/tests/test_qt_mouse_buttons.py
```diff
@@ -0,0 +1,50 @@
...
+def test_mouse_press_and_double_click_keep_get():
+    src = _qt_backend_source()
+    assert 'BUTTONMAP.get(' in _method_body(src, 'mousePressEvent')
+    assert 'BUTTONMAP.get(' in _method_body(src, 'mouseDoubleClickEvent')
+
+
+def test_buttonmap_get_fallback_semantics():
+    sample = {1: 1, 2: 2, 4: 3}
+    assert sample.get(8, 0) == 0
+    assert sample.get(1, 0) == 1
```
djhoese: These tests don't actually test anything. The last one is checking that dictionaries work the way dictionaries work. The other two are checking the source and making sure it doesn't change. We need to test functionality, not source code.

Phoenix0531-sudo:
Good catch — with only `.get(..., 0)`, every unmapped extra button would indeed collapse to VisPy button `0`.

I pushed a follow-up that maps Qt `ExtraButton3..24` (including `TaskButton` / `ExtraButton3`) to distinct VisPy ids `6..27`. `ExtraButton1`/`2` stay as the existing Back/Forward aliases (`4`/`5`). `_buttonmap_to_list` now walks the same `BUTTONMAP`, so multi-button state reports extras too. The safe `.get(..., 0)` path remains for anything still outside the map.

Local check: `pytest vispy/app/tests/test_qt_mouse_buttons.py` — 5 passed.

vispy/app/tests/test_qt_mouse_buttons.py
Phoenix0531-sudo: Agreed — those were source/dict checks, not behavior. Replaced them with functional tests that construct real Qt mouse events, run them through `mousePressEvent` / `mouseReleaseEvent` / `mouseDoubleClickEvent`, and assert the VisPy event button ids (standard, ExtraButton3/4/24, and an unmapped value → 0 without KeyError). Also covers `_buttonmap_to_list` multi-button state.

`pytest vispy/app/tests/test_qt_mouse_buttons.py` → 4 passed.

djhoese:
Are you an AI and is there any human in the loop on this pull request? Regardless, if you could find documentation from Qt for Qt5 and Qt6 on when these extra buttons were added to the enum that would be nice. My assumption is that they haven't always existed and that your `for` loop's `getattr` on the enum should check for existence first and only add the entry to button maps if it exists in the enum rather than assume it exists in all versions of Qt.

Phoenix0531-sudo:
Yes — there is a human in the loop on this PR. I review the diffs and local test output before pushing and before replying here; tooling helps with drafting and verification, but the changes and responses are checked by me.

On the ExtraButton history and existence checks:

- Qt 4.x `MouseButton` only had up through `XButton1`/`XButton2` (no `ExtraButton3..24`). See the Qt 4.8 docs: https://doc.qt.io/archives/qt-4.8/qt.html#MouseButton-enum
- Qt 5.0 introduced the extended set (`ExtraButton1..24`, with `ExtraButton1`/`2` as aliases of Back/Forward and `ExtraButton3` as `TaskButton`). They remain in Qt 5 and Qt 6 docs:
  - https://doc.qt.io/qt-5/qt.html#MouseButton-enum
  - https://doc.qt.io/qt-6/qt.html#MouseButton-enum
- Confirmed in qtbase headers: present from the `v5.0.0` tag onward; absent on Qt 4 tags such as `v4.8.7`.

I pushed a follow-up that only adds an ExtraButton entry when the enum member actually exists on the running Qt (`getattr(..., None)` on Qt 6, and the same existence check on the Qt 5 integer-key path). That avoids assuming every Qt version exposes the full set.

Local: `pytest vispy/app/tests/test_qt_mouse_buttons.py` → 4 passed; `py_compile` on `_qt.py` ok.

vispy/app/backends/_qt.py
```diff
@@ -494,30 +515,16 @@ def _buttonmap_to_list(self, buttons) -> list[int]:
...
-                mouse_buttons.append(2)
-            if QtCore.Qt.MouseButton.MiddleButton in buttons:
-                mouse_buttons.append(3)
-            if QtCore.Qt.MouseButton.BackButton in buttons:
-                mouse_buttons.append(4)
-            if QtCore.Qt.MouseButton.ForwardButton in buttons:
-                mouse_buttons.append(5)
+            for qt_button, vispy_button in BUTTONMAP.items():
+                if vispy_button and qt_button in buttons:
+                    mouse_buttons.append(vispy_button)
```
djhoese: I'm so confused why the code wasn't always this way.

vispy/app/backends/_qt.py
djhoese: Correct me if I'm wrong, but `self` isn't used here. What if we made this a `staticmethod` or even better moved it outside the backend class. Then the tests could run against this without doing any of the weird `Dummy` class stuff. Thoughts?

@brisvag Do you anticipate any problems with this? It is a private method so people shouldn't be using it directly anyway.

vispy/app/tests/test_qt_mouse_buttons.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+@requires_application('pyqt6')
+def test_buttonmap_to_list_includes_extra_buttons():
+    """Multi-button state must report Extra buttons, not only 1-5."""
+    QtCore, QtGui, qt_backend = _qt_modules()
+    mb = QtCore.Qt.MouseButton
+    helper = qt_backend.QtBaseCanvasBackend._buttonmap_to_list
+
+    class _Dummy:
```
djhoese: See other comment in backend code. Extracting the `_buttonmap_to_list` makes this test much simpler if possible.

vispy/app/tests/test_qt_mouse_buttons.py
```diff
@@ -0,0 +1,179 @@
...
+from vispy.app import Canvas, use_app
+from vispy.testing import run_tests_if_main, requires_application
+
+
+def _qt_modules():
+    """Return (QtCore, QtGui, backend_module) after loading PyQt6."""
+    use_app('pyqt6')
+    from PyQt6 import QtCore, QtGui
+    import sys
+    qt_backend = sys.modules['vispy.app.backends._qt']
```
djhoese: Can't we just import the module directly? Why are we accessing the sys.modules cache?

vispy/app/tests/test_qt_mouse_buttons.py
```diff
@@ -0,0 +1,179 @@
...
+    return QtCore, QtGui, qt_backend
+
+
+def _make_mouse_event(QtCore, QtGui, event_type, button, buttons=None):
```
djhoese: Could this instead just import QtCore and QtGui directly? Passing modules around as arguments is a bit confusing for anyone coming to this code later.

vispy/app/tests/test_qt_mouse_buttons.py
```diff
@@ -0,0 +1,179 @@
...
+        # Must not raise KeyError (the original bug).
+        canvas._backend.mouseReleaseEvent(fake)
+        assert len(seen) == 1
+        assert seen[0].button == 0
+    finally:
+        canvas.close()
+
+
+@requires_application('pyqt6')
+def test_press_and_double_click_extra_buttons():
```
djhoese: This seems like two separate tests or maybe a parametrized test (using `pytest.mark.parametrize`.

vispy/app/tests/test_qt_mouse_buttons.py
```diff
@@ -0,0 +1,179 @@
...
+    app = use_app('pyqt6')
+    canvas = Canvas(app=app, size=(80, 80), show=False, create_native=True)
+    try:
+        seen = []
+        canvas.events.mouse_release.connect(lambda e: seen.append(e))
+        # Choose a value that is not a MouseButton flag in BUTTONMAP.
+        unknown = object()
+        assert unknown not in qt_backend.BUTTONMAP
+        fake = _UnknownButtonEvent(unknown, QtCore)
+        # Must not raise KeyError (the original bug).
```
djhoese:
```suggestion
        # Must not raise KeyError (the original bug in GH #2765).
```

vispy/app/tests/test_qt_mouse_buttons.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def modifiers(self):
+        return self._QtCore.Qt.KeyboardModifier.NoModifier
+
+    def ignore(self):
+        self.ignored = True
+
+
+@requires_application('pyqt6')
+def test_standard_and_extra_button_release_events():
```
djhoese: Let me know if I'm misunderstanding but this test and `test_press_and_double_click_extra_buttons` seem like they could be combined and parametrized. Yes, it requires making the `Canvas` object over and over again, but I think the simpler test code is better in the long run for maintainability. @brisvag opinions welcome.

djhoese (changes requested):
Nice job. I had a couple suggestions and a couple concerns that I noted in inline comments.

Phoenix0531-sudo:
Thanks for the detailed review — I applied the suggestions:

1. **`_buttonmap_to_list`** — moved to module level (it never used `self`). Call sites now use `_buttonmap_to_list(...)` instead of `self._buttonmap_to_list(...)`. Tests call it directly; no Dummy needed.
2. **Tests** — combined press/release/double-click into one `pytest.mark.parametrize` case table; each case builds a Canvas and drives the real handler.
3. **Imports** — after `use_app('pyqt6')`, import `from vispy.app.backends import _qt as qt_backend` and `from PyQt6 import QtCore, QtGui` directly (no `sys.modules` cache, no passing modules as args).
4. **Comment** — unmapped-button test now references GH #2751 (the issue; #2765 is this PR).

Local: `pytest vispy/app/tests/test_qt_mouse_buttons.py` → 12 passed.

vispy/app/backends/_qt.py
Phoenix0531-sudo: Agreed — `self` was unused. Extracted to a module-level `_buttonmap_to_list(buttons)` (private helper, same behavior). Call sites updated; tests import and call it directly.

vispy/app/tests/test_qt_mouse_buttons.py
Phoenix0531-sudo: Done — helper is module-level now, so this test is just `qt_backend._buttonmap_to_list(mixed) == [1, 5, 7]` with no Dummy.

vispy/app/tests/test_qt_mouse_buttons.py
Phoenix0531-sudo: Switched to `from vispy.app.backends import _qt as qt_backend` after `use_app('pyqt6')`. Direct import still needs the backend selected first (otherwise `_qt` raises), but we no longer poke `sys.modules`.

vispy/app/tests/test_qt_mouse_buttons.py
Phoenix0531-sudo: Agreed — tests now `from PyQt6 import QtCore, QtGui` at module level and build events without passing modules around.

vispy/app/tests/test_qt_mouse_buttons.py
Phoenix0531-sudo: Combined press / double-click into the same parametrized table as release (see `test_mapped_button_events`).

vispy/app/tests/test_qt_mouse_buttons.py
Phoenix0531-sudo: Updated the comment to reference GH #2751 (the issue). Thanks.

vispy/app/tests/test_qt_mouse_buttons.py
Phoenix0531-sudo: Done — release + press + double-click are one parametrized test now. Canvas is created per case for isolation; the case table is much easier to read/extend.

vispy/app/backends/_qt.py
Phoenix0531-sudo: Yeah — once Extra buttons are in `BUTTONMAP`, walking the map is the natural way to keep multi-button state in sync. Left it as the BUTTONMAP-driven loop (now at module level).

djhoese:
Thanks for the updates. Just curious what AI agent and model are you using?

vispy/app/backends/_qt.py
```diff
@@ -193,16 +193,37 @@ def _get_event_xy(ev):
...
+    # exist from Qt 5.0; Qt 4 only had XButton1/2. Add them only when present.
     BUTTONMAP = {0: 0, 1: 1, 2: 2, 4: 3, 8: 4, 16: 5}
+    _mouse_button = getattr(QtCore.Qt, 'MouseButton', QtCore.Qt)
+    for _extra_i in range(3, 25):
+        _qt_btn = getattr(_mouse_button, f'ExtraButton{_extra_i}', None)
+        if _qt_btn is None:
+            _qt_btn = getattr(QtCore.Qt, f'ExtraButton{_extra_i}', None)
+        if _qt_btn is not None:
+            _code = int(getattr(_qt_btn, 'value', _qt_btn))
+            BUTTONMAP[_code] = _extra_i + 3
```
djhoese, line 226: Where is this `+ 3` coming from?

djhoese: Oh it's in the upper loop too.

djhoese: 🤦 Ah it is to get the next valid button value, the last being 5.

vispy/app/tests/test_qt_mouse_buttons.py
```diff
@@ -0,0 +1,153 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+"""Functional regression tests for Qt mouse button mapping (#2751)."""
+
+import pytest
+from PyQt6 import QtCore, QtGui
```
djhoese: Hhhmmm importing this at a module level may be a problem for the tests. If we do this in other test modules then forget I said anything. Let's see how CI does.

djhoese: Yeah, CI is failing when Qt6 isn't available:

```

WARNING: autodoc: failed to import module 'test_qt_mouse_buttons' from module 'vispy.app.tests'; the following exception was raised:
No module named 'PyQt6' [autodoc.import_object]
```

Note the cibuildwheel jobs can be ignored. They are complaining about something else.

Phoenix0531-sudo: Fixed — moved the PyQt6 / QtCore / QtGui imports inside the tests/helpers that are already guarded by `@requires_application('pyqt6')`, and changed the parametrization table to store enum names instead of Qt objects.

I also checked the no-PyQt6 collection path by monkeypatching imports locally: importing `vispy/app/tests/test_qt_mouse_buttons.py` now succeeds without PyQt6 (`import_without_pyqt6_ok`). The targeted Qt test still passes locally with PyQt6: `12 passed`.

djhoese:
@brisvag This osmesa CI failure is expected right?

brisvag:
Is it? :thinking: Do you mean "it's not caused by this PR", or do you mean "it's fine if it fails"? I don't recognize it, but at first glance I'm guessing we have some outdated env variablle setings. Have you seen it before?

djhoese:
I think I've only seen it in other recent PRs. I was trying to say "not caused by this PR".

Phoenix0531-sudo:
Yeah I checked — the same `OSError` shows up on other recent PRs too. "Optimizing Axis Performance" (run 30141507581, opened 7-25) and the dependabot cibuildwheel one (run 30264290161) both fail with the byte-identical error at the same path. For reference, the last few push-triggered runs on main (7-20 and earlier) were still green, but there hasn't been a push on main since then so that's just before whatever broke started.

The error is `ctypes.cdll.LoadLibrary` in `vispy/gloo/gl/gl2.py` trying to load `~/miniconda3/envs/vispy-tests/lib/libOSMesa.so`, which the workflow hardcodes in `main.yml`. The conda env installs `mesalib` 17.3.9 (from `ci/requirements/linux_osmesa_deps_conda.txt`) and that package doesn't actually ship a `libOSMesa.so` to that path — so it's a CI packaging / runner-image thing, not anything in this PR's diff. `_qt.py` and the mouse tests never get imported.

Happy to open an issue for it tomorrow, but I'm calling it a night now. Good night.

Phoenix0531-sudo:
Opened #2767 for the osmesa CI failure — tldr: `mesalib` isn't pinned in `ci/requirements/linux_osmesa_deps_conda.txt` and conda resolved to 17.3.9 (which doesn't ship `libOSMesa.so`) on every PR since 7-25, while the last green main push got 25.0.5. Pinning `mesalib >=25.0.5,<25.1.0` should unblock everyone.

Phoenix0531-sudo:
Just rebased this onto current main — it also picks up the mesalib pin from #2777, so the osmesa job shouldn't fail anymore on this branch. All 14 app tests and py_compile are green locally. The new CI run is sitting in "action required", so it needs an approve-and-run whenever you get a chance. Thanks!

brisvag:
Thanks for following up on this @Phoenix0531-sudo! @djhoese I think you had a clearer idea on this PR, and marked as changes requested; could you take another look?

vispy/app/tests/test_qt_mouse_buttons.py
```diff
@@ -0,0 +1,163 @@
...
+        ev = _make_mouse_event(event_type, qt_button, buttons=buttons_flag)
+        getattr(canvas._backend, _HANDLER[event_name])(ev)
+        assert len(seen) == 1
+        assert seen[0].button == vispy_button
+        if event_name == 'release':
+            assert tuple(seen[0].pos) == (12.0, 34.0)
+        if buttons_flag is not None:
+            assert vispy_button in seen[0].buttons
+    finally:
+        canvas.close()
```
djhoese: It's been a while since I've done actual vispy coding but doesn't the `Canvas` have a context manager functionality so you can do `with Canvas(...) as canvas:`? This could avoid this try/except pattern and then these tests would match what we do/recommend elsewhere.

djhoese (changes requested):
I made one small request that affects 2 tests and that's commented inline. I think this PR is good otherwise. I have a feeling some of the main logic could be cleaned up but when I tried it myself I wasn't able to figure anything out. I also find the tests not obvious, but I'm not sure there is much that can be done about that.

vispy/app/tests/test_qt_mouse_buttons.py
Phoenix0531-sudo: Done — switched both tests to `with Canvas(...) as canvas:` in 9c33707. Thanks!

djhoese (approved):

djhoese:
@brisvag Did you change the repository's settings to not allow rebasing or merge commits as options? I only have "Squash and merge" as a merge option for this PR.

djhoese:
Hm I wonder if Github misconfigured based on a change by rossant on the vispy2 repository: https://github.com/organizations/vispy/settings/audit-log?q=action%3Arepo.change_merge_setting

brisvag:
I did that years ago, yes (I think we talked about it at the time). We found squash&merge to be way better in napari for figuring out the history on `main`.

djhoese:
Maybe I should have warned you that my 2.5 year old still doesn't consistently sleep through the night and my memory and decisions may be...flawed.

Phoenix0531-sudo:
Ha, no worries at all — thanks for pushing this through! Enjoy the coffee. ☕

PR #2769, opened 2026-08-09
Update build to use uv

### LINKED ISSUE #2767: CI: osmesa test job fails — mesalib resolved to 17.3.9 (no libOSMesa.so) instead of 25.0.5
## What happened

The `vispy tests - linux (3, 3.9, osmesa, osmesa, false)` job has been failing on every PR opened since ~7-25 with:

```
OSError: /home/runner/miniconda3/envs/vispy-tests/lib/libOSMesa.so:
  cannot open shared object file: No such file or directory
```

It's raised in `vispy/gloo/gl/gl2.py` at `ctypes.cdll.LoadLibrary(OSMESA_LIBRARY)`, before `vispy.app` is imported — so no PR code is involved.

## It's not the PRs

Same failure on three unrelated PRs in the same window, all with the *exact same* error at the same path:

- #2765 (this PR, run [30450737360](https://github.com/vispy/vispy/actions/runs/30450737360))
- "Optimizing Axis Performance" (run [30141507581](https://github.com/vispy/vispy/actions/runs/30141507581), opened 7-25)
- dependabot cibuildwheel-4.1.1 (run [30264290161](https://github.com/vispy/vispy/actions/runs/30264290161), 7-27)

The last `push`-triggered run on `main` (7-20, run [29776758152](https://github.com/vispy/vispy/actions/runs/29776758152)) had the osmesa job green. There hasn't been a push on `main` since.

## Root cause: `mesalib` isn't pinned and conda resolved to a different version

`ci/requirements/linux_osmesa_deps_conda.txt` lists `mesalib` with no version. The resolver picked:

| Run | When | Result | mesalib version |
|---|---|---|---|
| 29776758152 (main push) | 7-20 | ✅ osmesa loaded | **25.0.5** |
| 30141507581 (Axis PR) | 7-27 | ❌ `OSError` | **17.3.9** |
| 30264290161 (dependabot PR) | 7-27 | ❌ `OSError` | **17.3.9** |
| 30450737360 (this PR) | 7-29 | ❌ `OSError` | **17.3.9** |

`mesalib 25.0.5` ships `libOSMesa.so` to `$CONDA_PREFIX/lib`. `mesalib 17.3.9` does not — that version is from 2018 and predates the modern conda-forge osmesa packaging.

`OSMESA_LIBRARY` in the workflow (`main.yml:275-276`) is hardcoded to `~/miniconda3/envs/vispy-tests/lib/libOSMesa.so`, so whenever the resolver lands on a mesalib build that doesn't provide that file, the job fails before any test runs.

It's just luck that it worked on the 7-20 main push and broke on every PR after.

## Related: OSMesa is being removed from Mesa upstream

conda-forge `mesalib` 25.1.0+ drops `libOSMesa` entirely ([conda-forge/mesalib-feedstock#109](https://github.com/conda-forge/mesalib-feedstock/issues/109)). The short-term workaround mentioned there is to pin `mesalib < 25.1.0` — which `25.0.5` happens to satisfy. So pinning to `>=25.0.5,<25.1.0` gets the working `libOSMesa.so` back, but the long-term fix is migrating the osmesa test job off OSMesa (e.g. EGL with `EGL_PLATFORM=headless`, per the mesa MR linked in that feedstock issue) or packaging osmesa separately.

## Suggested fix

In `ci/requirements/linux_osmesa_deps_conda.txt`, pin:

```
mesalib >=25.0.5,<25.1.0
```

This unblocks all currently-failing PRs without changing any product code. The EGL migration can happen as a follow-up.

## Environment

- Runner: `ubuntu-24.04` (image `20260720.247.2`)
- Conda env: `vispy-tests`, Python 3.9
- The `OSMESA_LIBRARY` / `VISPY_GL_LIB` exports are set by `.github/workflows/main.yml:271-276` and `:294-297`

cc @djhoese @brisvag (from the discussion on #2765)

melonora:
closes #2767 

I wnated to check into non-linear transforms, but before doing that just wanted to update the build so uv can be used instead. In addition I did the following:

- Add ci jobs for python 3.12 and 3.14 using full dependencies and newer qt. Reason for using full and newer qt is that I expect that people using newer python versions would migrate.
- Changed the python version for building the wheels to 3.12. This was 3.9
- Updated the python classifiers to include python versions up to and including python 3.14

One thing to note is that for the docs build pyqt5 is used which does not have a suitable wheel on windows. Pyqt6 does, but for now I didn't look into that migration. Several examples and also test modules seem to use pyqt5.

Another thing, I did not update the documentation yet. I first would like an ok for moving to uv or pixi before I update the documentation.


djhoese:
I see a readthedocs config, we don't use readthedocs.

melonora:
ah sorry will remove it, got a bit confused with the old `rtd_requirements` and also http://vispy.readthedocs.org in readme.rst

melonora:
@djhoese one question, is `stdeb.cfg` still necessary for Debian? Based on the python version it seems stale.

melonora:
Also, if you are in general ok with making the change to uv / pixi I will adjust the documentation accordingly. Just did not want  to do it before I get an ok on it.

pyproject.toml
```diff
@@ -1,27 +1,183 @@
...
+
+[tool.setuptools_scm]
 write_to = "vispy/version.py"
 local_scheme = "dirty-tag"
 
+[tool.uv]
+cache-keys = [
+    { file = "pyproject.toml" },
+    { git = { commit = true, tags = true } },
+    { file = "vispy/visuals/text/_sdf_cpu.pyx" },
```
brisvag, line 164: Why this line?

pyproject.toml
```diff
@@ -1,27 +1,183 @@
...
+    "wxPython",
+]
+tk = [
+    "pyopengltk",
+]
+io = [
+    "meshio",
+    "Pillow",
+]
+
```
brisvag: Maybe we should leave the `[doc]` and `[test]` extras in, since vispy's been a stable part of people's project's CI for a long time, and we don't want to break stuff just because we renamed it. Anyways we're unlikely to change any of these deps any time soon :P

pyproject.toml
```diff
@@ -1,27 +1,183 @@
...
+    "Framework :: IPython",
+]
+dynamic = ["version"]
+dependencies = [
+    "numpy",
+    "freetype-py",
+    "hsluv",
+    "kiwisolver",
+    "packaging",
+    "importlib-resources; python_version < '3.9'",
```
brisvag: I guess this is no longer needed since we're 3.9+ only.

.github/workflows/main.yml
```diff
@@ -291,11 +318,34 @@ jobs:
...
+              -quit
+          )"
+
+    
+          if [ -z "${OSMESA_LIBRARY}" ]; then
+            echo "Could not find libOSMesa in ${CONDA_PREFIX}"
+            conda list | grep -Ei 'mesa|osmesa' || true
+            find "${CONDA_PREFIX}" -iname '*osmesa*' -print || true
+            exit 1
+          fi
```
brisvag, line 388: What about the comment in https://github.com/vispy/vispy/issues/2767 that osmesa is straight up dropped after a certain version? How is this working around that?

brisvag (changes requested):
Generally makes sense to me! Left some comments.

pyproject.toml
melonora: this is because this file is cython. If this file changes it means the uv project cache is still and vispy should be rebuild. Maybe too much here given that vispy is in maintenance mode.

.github/workflows/main.yml
melonora: ah read over that. So that wouldn't be fixed, but then again I see on CI that v17.3.9 of mesalib gets installed. I added the pin now in addition. Will see if the ci job passes, but we could at least have it double to prevent file name issues later on in general as these files seem to change name from time to time and this will capture that in any case.

melonora: test passes for this, and v25.0.5 is now used in the ci job.

pyproject.toml
melonora: ah right should not have copied that hehe

melonora: removed

pyproject.toml
melonora: done

brisvag:
That hanging test is worrying :/ @melonora you think you have some conda magic to make that work?

melonora:
Will check when I am at the airport

melonora:
checking the cause of the segfault. So far ruled out:
- pyqt6 version
- python version

melonora:
So @brisvag and @djhoese (have a good holiday btw, if you see this, I don't expect a reply before you are back:)) I had quite a back and forth trying to troubleshoot a segfault when adding ci jobs for testing pyqt6 with python 3.12 and 3.14. I ultimately localized the problem to be in the test script itself. In particular the app was quit synchronously during a draw event, causing a `QPaintEvent` to be send after there was an event indicating that the surface platform was about to be destroyed. The particular bit was:
```python
@c.events.draw.connect
            def draw(event):
                nonlocal quit_timer
                print(event)  # test event __repr__
                if c.app.backend_name.lower() == "pyqt6":
                    from vispy.app.backends._qt import QtCore
                    QtCore.QTimer.singleShot(0, c.app.quit)
                else:
                    # This was the only way the app was quit before.
                    c.app.quit()
```
I found out that the `singleShot` here enforces that the quit happens after returning back to the event loop, thereby preventing the segfault. I tried to use the vispy `Timer` for this but didn't manage to fix it that way. I would have prefered that in order to not bring backend specifics in, but I hope this fix is still ok.

brisvag:
> I would have prefered that in order to not bring backend specifics in, but I hope this fix is still ok.

Mmmm... if it's just to fix a test, I guess maybe it's fine? But we should definitely split that out into a separate PR just to see if that works on its own.

djhoese:
If this was discussed before all the work was done I would have said I don't see the benefit. I don't use uv or pixi in my local workflow because it doesn't fit with how I usually do things. That said, if this simplifies the CI configs then :+1: but right now it seems like a lot of changes that I'm not sure I follow.

brisvag:
2 main benefits are:
- slightly simpler configs
- significantly faster

I agree it's not strictly necessary; @melonora was mostly wanting to fix CI and iterating with uv is faster :P I think we can revisit once #2777 is merged, and it should require as much fiddling.

melonora:
@brisvag with pixi we could simplify the whole CI as it combines both uv pip and conda. I will create a PR for it after the initial drop python 3.9 PR and support for python 3.14 PR.

brisvag:
Will close this for now then, since either way it will be superseded.

Issue #2771, opened 2026-08-10
OME-NGFF non-linear transforms


melonora:
With OME-NGFF rfc 5, non-linear transforms by means of displacement and coordinate fields will be introduced. Currently, vispy does not support this. In order for napari to be OME-NGFF compliant it would be necessary to introduce this in vispy. 
Both should share a common  transform implementation where possible.

Reading up and looking into the codebase, I was broadly thinking of something like this:
1. Allow BaseTransform implementations without an inverse. Currently the BaseTransform depends on map / imap etc, but there is no guaranteed inverse for a non-linear transform.
2. Add an in-memory 2D DisplacementFieldTransform with CPU + GLSL forward mapping.
3. Support forward-only rendering through ImageVisual(method="subdivide").
4. Add explicit inverse fields and method="impostor".
5. Add method="auto" and adaptive subdivision. For adaptive subdivision a cheap coarse way for calculating the error would be required.
6. Add optional numerical inverse estimation.
7. Add lazy/chunked field loading for very large datasets.
8. Extend to 3D fields, slices, and eventually volume rendering.

Things might change as I go through it.

@brisvag @jni @joshmoore

djhoese:
I have not looked at your PR yet, but wouldn't this prevent most (all?) mouse interactions with the Visuals to break? I could be misremembering but isn't the inverse mapping needed to implement picking? That is, going from the mouse location on screen to the buffer location and then back to the Visual's coordinate system?

melonora:
Sorry for not replying here sooner! Now with the github capability of stacking PRs I will create the series of PRs this way so I can explicitly check this as part of the larger work well while keeping PRs relatively small for review. I will close the open PR regarding this for now.

melonora:
@djhoese To note, I know that at least inverse coordinate based picking is not possible in this case. In the PR I opened I meant to only allow for the possibility of having a transform without inverse. Since we do not support that yet, I did not expect that to be an issue. 

As stated in the comment above, I will use github stacked PRs to approach this work. This because otherwise I am afraid that there would be one big PR that is too big to review. I am exploring other options than inverse coordinate based picking but not certain yet about whether that would work. The intention though is that if I can get something working to have that be one of the subsequent PRs in the stacked PR.

PR #2772, opened 2026-08-11
Allow forward only transform

melonora:
As a first step for #2771, this PR allows for transforms to be forward only. The transform does not require `imap`, `shader_imap` and `inverse`. An attribute `inverse_kind` is added which as a value has an enum. By default this is `exact` indicating that there is an exact inverse transform. This is the case for all the linear transforms. The other potential values are `approximate` and `None`.


melonora:
closing this as mentioned in #2771

PR #2774, opened 2026-08-11
Fix perspective for text

### LINKED ISSUE #8872: Setting perspective makes text tiny in 3D view
### 🐛 Bug Report

I wanted to add text labels with points to my 3D figure. The layer is added without errors, but the text is absolutely tiny (first I thought it didn't render at all). If I switch to 2D mode, the text is the expected size. Also, if I set the camera perspective to 0, the text is once again normal sized.

### 💡 Steps to Reproduce

Minimal code I could come up with:

```py
import napari

viewer = napari.Viewer(ndisplay=3, show=True)

viewer.camera.perspective = 5  # Setting perspective to a small number >0 makes text tiny

viewer.add_points(
    [[1, 0, 0],
     [0, 1, 0],
     [0, 0, 1]],
    size=0.1,
    face_color='white',
    text={
        'string': ['A', 'D', 'R'],
        'color': 'white',
        'size': 20,
        'anchor': 'center',
        'translation': [0, 0, 0.5],
    }
)

viewer.show(block=True)
```

### 💡 Expected Behavior

I expect the text to be similar sized to the case when perspective is set to 0.

### 🌎 Environment

napari: 0.7.0
Platform: Windows-10-10.0.26200-SP0
Python: 3.11.9 (tags/v3.11.9:de54cf5, Apr 2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)]
Qt: 6.9.0
PyQt6: 6.9.1
NumPy: 2.2.6
SciPy: 1.16.0
Dask: 2025.5.1
VisPy: 0.16.1
magicgui: 0.10.1
superqt: 0.8.1
in-n-out: 0.2.1
app-model: 0.5.1
psygnal: 0.15.1
npe2: 0.8.2
pydantic: 2.12.5

OpenGL:
- PyOpenGL: 3.1.10
- GL version: 4.6.0 - Build 27.20.100.8729
- MAX_TEXTURE_SIZE: np.int32(16384)
- GL_MAX_3D_TEXTURE_SIZE: 2048

Screens:
- screen 1: resolution 1536x864, scale 1.25

Optional:
- numba: 0.61.2
- triangle not installed
- napari-plugin-manager not installed
- bermuda not installed
- PartSegCore not installed

Experimental Settings:
- Async: False
- Autoswap buffers: False
- Triangulation backend: Fastest available

Settings path:
- C:\Users\Abel\AppData\Local\napari\.venv_9b85e2a1a525325a24835e9c68138c82c04b8803\settings.yaml

Launch command:
- C:\Users\Abel\PycharmProjects\fus\fusanalysis\users\Abel\user_scripts\napari_text_bug_reproduction.py

Plugins:
- napari: 0.7.0 (88 contributions)
- napari-animation: 0.0.9 (2 contributions)
- napari-console: 0.1.4 (0 contributions)
- napari-svg: 0.2.1 (2 contributions)

### 💡 Additional Context

_No response_

brisvag:
Similar issue to #2773. To test, try out the example at `examples/scene/text.py`. If you change FOV using shift+right click drag, you'll see the text rescale in weird ways, and then *pop* to the correct font_size when FOV=0. This PR makes it behave as you instead would expect (constant screen size, no weirdnesses).

Closes napari/napari#8872.

### DIFF
--- a/examples/scene/text.py
+++ b/examples/scene/text.py
@@ -36,8 +36,8 @@ t1.font_size = 24
 t1.pos = canvas.size[0] // 2, canvas.size[1] // 3
 
 t2 = Text('Text in viewbox (18 pt)', parent=vb.scene, color='green',
-          rotation=30)
-t2.font_size = 18
+          rotation=30, scaling=True)
+t2.font_size = .2  # scaling True means this size is in scene coordinates
 t2.pos = 0.5, 0.3
 
 # Add a line so you can see translate/scale of camera
--- a/vispy/visuals/text/text.py
+++ b/vispy/visuals/text/text.py
@@ -449,7 +449,7 @@ class TextVisual(Visual):
     def __init__(self, text=None, color='black', bold=False,
                  italic=False, face='OpenSans', font_size=12, line_height=1.1, pos=[0, 0, 0],
                  rotation=0., anchor_x='center', anchor_y='center',
-                 method='cpu', font_manager=None, depth_test=False, scaling=True):
+                 method='cpu', font_manager=None, depth_test=False, scaling=False):
         Visual.__init__(self, vcode=self._shaders['vertex'], fcode=self._shaders['fragment'])
         # Check input
         valid_keys = ('top', 'center', 'middle', 'baseline', 'bottom')
--- a/vispy/visuals/text/text.py
+++ b/vispy/visuals/text/text.py
@@ -21,7 +21,6 @@ from ...gloo import (TextureAtlas, IndexBuffer, VertexBuffer)
 from ...gloo import context
 from ...gloo.wrappers import _check_valid
 from ...util.fonts import _load_glyph
-from ..transforms import STTransform
 from ...color import ColorArray
 from ..visual import Visual
 from ...io import load_spatial_filters
@@ -150,6 +149,9 @@ class FontManager(object):
 # The visual
 
 _VERTEX_SHADER = """
+    uniform bool u_scaling;
+    uniform vec2 u_px_scale;
+    uniform float u_npix;
     attribute float a_rotation;  // rotation in rad
     attribute vec2 a_position; // in point units
     attribute vec2 a_texcoord;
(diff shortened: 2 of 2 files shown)


djhoese:
Looks good. If you have a video/screencast of before/after that'd be amazing, but otherwise :+1:

brisvag:
Mmmm... as you see, one thing that changed is that now text is fixed in screen space even when zooming. Before, by default it had a real phisical size, it was just mangled by FOV. 

I guess the proper way would be to fix the FOV so it doesn't act like in main, while still maintaining zooming, and add a new `scaling` parameter (like Markers) to enable or disable the "scale with zoom" behaviour...

Main:

https://github.com/user-attachments/assets/9984de18-fe0a-4886-b3b6-03db9c75d165


This PR:

https://github.com/user-attachments/assets/a63b11f3-134a-4845-a404-6d5c86dbb413


---

djhoese:
Yeah I mean we want the text to respect the transforms of the Visual/Node graph that it's a part of.

brisvag:
Ok I think I figured out the solution; will push something better soon :muscle:

brisvag:
Thsi was a doozie! LLM helped me get out of a slump, but after that kept running in circles; finally figured out all the pieces :P Now we have a scaling parameter like for markers, and it behave similarly!

brisvag:
Damn, no, I forgot the zoom along the way...

brisvag:
Ok I think I finally got it :melting_face: 

With scaling=True:

https://github.com/user-attachments/assets/288dd5d6-5709-4bcd-bd78-0cf4f9c7ca1e

With scaling=False

https://github.com/user-attachments/assets/d9783615-5fb2-4560-8fc4-4dc7a70ea9d0

---

There is still something to decide: the default.

Currently on main we have sort of a "hybrid":
- if FOV>0: text does not scale with zoom, and its size is the "screen size". This is equivalent to the new `scaling=False`.
- if FOV==0: text scales with zoom (like the new `scaling=True`); however, its size is affected by FOV in a buggy way, which means that the meaning of `font_size` is kinda arbitrary and depends on the FOV. In this PR, instead, `font_size` determines the absolute size of the font in world units, which never changes.

What's the least disruptive default we can pick?

IMO we should go with `scaling=False`, because at least text is more or less guaranteed to show up (whereas scaling=True might make it invisibly tiy or huge depending on FOV).

This way, users that had `FOV=0` will have no changes, and users with `FOV>0` will have to adjust `scaling` potentially `font_size`. If instead we default `scaling=True`, `FOV=0` folks will have to switch `scaling` AND `FOV>0` folks will have to adjust `font_size` to get back what they had.

Anyone in between is screwed :sweat:

brisvag:
@djhoese would love your opinion on the default value!

djhoese:
This is so confusing, especially when you haven't been looking at it for a long time...

Let me see if I've got this. So we have the two usual cases in `main` right now:

1. you want text that doesn't change size so you set it to be a child of `canvas.scene`. This is `t1` in the example code.
2. you want text to scale with the viewbox so you set it to be a child of `viewbox.scene`. This is `t2` in the example code.

You are saying that FOV and zoom aren't properly handled in case 2 so you fixed that. Also, none of your changes affect case 1, right? Or does a user maybe need to play with font size depending on scaling=True/False?

I'm confused by the description of what the default would mean. You say that currently on main `FOV==0` are affected by zoom (buggy with FOV changes) similar to `scaling=True` and then later say that setting the default to `scaling=False` would mean no changes for these `FOV==0` people.

brisvag:
I'm not sure I get what you mean with the parenting; this has nothing to do with the parent of the scene (it shouldn't matter?).

The use cases are:
- I want text to scale with the scene (meaning if I zoom, the text stays the same size in scene space)
- I want text to not scale with the scene (meaning if I zoom, the text looks the same in screen space)

Both of these can be legitimately used for 2D and 3D scenes. Example, in napari our points can have text associated with it, and if you enable FOV you may want the text closer to you to actually appear bigger; you also may want to be able to zoom into a point and the text becomes readable, without overlapping with everyuthing else. But you also may want to be able to always read the text well regardless of zoom and FOV, and zooming out should not make the text smaller in that case.

So we want this to be settable (same behaviour as the markers).


> I'm confused by the description of what the default would mean. You say that currently on main `FOV==0` are affected by zoom (buggy with FOV changes) similar to `scaling=True` and then later say that setting the default to `scaling=False` would mean no changes for these `FOV==0` people.

Whoops, I literally swapped the cases in my comment, sorry. `FOV==0` does not scale with zoom, and `FOV>0` does.

djhoese:
Sorry I was confusing a similar, but completely different use case. I mentioned parenting because I was confusing scaling with users wanting the texts not to move on the Canvas. Typically users following examples will add text to the node graph like all the other Visuals, but those follow transforms and the camera. They need to set the parent to the `canvas.scene` to skip over that.

I'm OK with the `scaling=False` default. Does all of this new python code get called on every draw or is it cached or how often is it called?

brisvag:
Ah ok, got it. Yes, that's the use case of the top widget in the example, and it still works as intended! And the nice thing is that that use case is fully unaffected (scaling true or false are the same).

Unfortunately it gets called on every draw... I could add a cache that only runs it if FOV or zoom have changed, but it still runs every time. The alternative would be to implement this more like markers did, but I tried that path and got stuck :/

djhoese:
What's the major difficulty with doing it in the shader? I'm sure it is difficult I'm mostly just curious what you ran into.

brisvag:
The way that transforms are affecting text is weird and not as simple as markers, because the text has the extra billboarding + builtin rotation that complicates things. I kept ending up with insane sizes or losing the billboard in funky ways :sweat_smile: 

But now that I dug deeper into the whole logic in python, maybe I can do it. I'll give it another shot :muscle:

brisvag:
Woo I think it works!

djhoese:
And it seems to perform well? Looks clean even if I don't understand it. I restarted the failed build website job. If you feel good with this then I'm good with it. `scaling=False` still needs to be updated though (currently True default).

brisvag:
Yeah I think this is good! Updated to False.

brisvag:
I tested with 1M text entries; performance seems comparable to main (=blazing fast) in every mode!

djhoese (approved):

PR #2775, opened 2026-08-13
prevent quitting app while draw event in progress

melonora:
Quitting the app while a draw is in progress, as done in test_app.py might lead to events being send after the app is already destroyed. While not having been a problem so far, this became a problem in #2769 as it led to segfaults. In order to test whether this solution works on its own, without being convoluted by other changes, I create this separate PR.

The main thing is that the use of `SingleShotTimer` here forces the quit to only happen after we return to the event loop, thereby preventing teardown of the app until the draw event has completed. This is similar to the use of `SingleShotTimer` in @kephale progressive loading PR.

### DIFF
--- a/vispy/app/tests/test_app.py
+++ b/vispy/app/tests/test_app.py
@@ -155,7 +155,19 @@ def test_run():
             @c.events.draw.connect
             def draw(event):
                 print(event)  # test event __repr__
-                c.app.quit()
+                if c.app.backend_name.lower() in (
+                        "pyqt4",
+                        "pyqt5",
+                        "pyqt6",
+                        "pyside",
+                        "pyside2",
+                        "pyside6",
+                ):
+                    from vispy.app.backends._qt import QtCore
+                    QtCore.QTimer.singleShot(0, c.app.quit)
+                else:
+                    c.app.quit()
+
             c.update()
             c.app.run()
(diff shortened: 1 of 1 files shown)


vispy/app/tests/test_app.py
```diff
@@ -152,10 +152,25 @@ def test_run():
     """Test app running"""
     for _ in range(2):
         with Canvas(size=(100, 100), show=True, title='run') as c:
+            quit_timer = None
```
brisvag: This is unused?

melonora: whoops yeah, forced push change also to remove the uv.lock that got added by accident

brisvag (approved):

djhoese:
Doesn't the `canvas.app.quit()` eventually call a `quit` method on the low-level backend's Application object? Couldn't the singleshot be done there? Ah wait, this is only in the tests? Before noticing that, my other concern was going to be what if the event loop is being held up by a user event that isn't returning control back to the event loop. I guess that's something that is bad in and of itself so maybe it is fine if it breaks things.

brisvag:
Yeah, I agree with the worries in general, but indeed this is just to prevent tests from failing, so if this bug didn't cause issues in the wild until now, I think we can avoid touching it and potentially making things worse xD

brisvag:
Superseded by #2777

PR #2768, opened 2026-08-04
Fix osmesa failure

brisvag:
Closes #2767.

As pointed out in the issue, this might be a temporary fix due to mesalib dropping libosmesa. This unblocks CI for now.

PR #2776, opened 2026-08-13
Fix osmesa ci job

### LINKED ISSUE #2767: CI: osmesa test job fails — mesalib resolved to 17.3.9 (no libOSMesa.so) instead of 25.0.5

melonora:
closes #2767 

This PR does the following to fix the mesa ci job
- adds suggested dependency pin
- makes looking for the mesalib library more robust as the filename changes dependent on version

In addition this PR fixes the timeout of the python 3.9 ci job due to conda resolving taking too long.
Some of the depdendencies are moved to pip. PyQt5 stays in conda, but is pinned on the exact version
that was used in the last succesful ci job. Scipy was dropped as it is included in scikit-image. This is perhaps 
not ideal, but the dependencies here are specifically only for ci job and not for the ordinary user.


brisvag (changes requested):
I'm confused by this PR...

Are you saying that the mesalib pin is needed *on top of* the smarter library search? AFAICT my pr #2768 already fixed the osmesa fail, without all the extra stuff. Are those needed to cover some other issues/cases?

Also, I don't think hard-pinning deps in tests is a good way to get around conda resolution slowness. We want to test the latest stuff if possible.

And I guess you're moving sphinkx-gallery and `scikit-image` to pip to get around the conda resolution slowness, but is that wise? Is it slow because they actually have conflicts? Can we use mamba instead for speeding up things?

melonora:
> I'm confused by this PR...
> 
> Are you saying that the mesalib pin is needed _on top of_ the smarter library search? AFAICT my pr #2768 already fixed the osmesa fail, without all the extra stuff. Are those needed to cover some other issues/cases?
> 
> Also, I don't think hard-pinning deps in tests is a good way to get around conda resolution slowness. We want to test the latest stuff if possible.
> 
> And I guess you're moving sphinkx-gallery and `scikit-image` to pip to get around the conda resolution slowness, but is that wise? Is it slow because they actually have conflicts? Can we use mamba instead for speeding up things?

so last time I checked the osmesa PR that you created had one failing test. I see now that, that is resolved. Making it more robust with the naming was my initial fix in the UV PR. I do see that your PR also has the timeout problem with python 3.9 though.

I understand that we want to test the latest possible, but if we are keeping the ci job for 3.9 despite it being passed its end of lifetime we get timeouts due to resolving. In my uv pr I did try with explicitly having the libmamba solver, but it still timed out. So using mamba does not fix anything. I made the choice for the newer python versions to test with pyqt6 and not 5. However, if you want to keep testing with pyqt5 I would suggest to split out all the environment files based on python version and not pyqt5 versus pyqt6.

For sphinx, it added a solid 20 minutes to resolving time. I think a general problem here is that conda-forge already dropped support for 3.9 ago and a lot of the dependencies we have here have a large dependency graph. If we do not pin anything, the intersection with old libraries for 3.9 becomes makes it so that the intersection of compatible builds is becoming smaller and smaller as more libraries drop support for 3.9. Hard pinning in that case is the only solution.

There are a couple of options:
- I remove the osmesa fix from this pr so this pr purely fixes the time out for python 3.9. We merge PRs sequentially accepting that the first PR to be merged won't have a complete green CI
- We merge this PR as is which includes fixes for both osmesa and the timeout. Here, I could remove the more robust lookup of the mesa library given that we already pin it now (the more robust lookup was because without the pin I found the name of the libary to have changed, causing the lookup to fail.
- We drop support for python 3.9. Currently the lifetime of python 3.9 has ended. Conda-forge has dropped support for it and the intersection of builds that would be compatible will become relatively smaller and smaller.
- If we want to keep the ci job for python 3.9 we could also hard pin everything given the smaller and smaller intersection of possible builds.

@djhoese thoughts?

brisvag:
> We drop support for python 3.9. Currently the lifetime of python 3.9 has ended. Conda-forge has dropped support for it and the intersection of builds that would be compatible will become relatively smaller and smaller.

I'd go with this as it's EOL.

melonora:
note that this would also remove the osmesa job unless we would now make it so that, that job uses python 3.10

djhoese:
I am OK removing 3.9. I think bumping osmesa to a newer python then makes sense. As a side-line maintainer these days I'd also be OK dropping 3.10 and 3.11 support and going more the numpy/scipy route of maintaining the latest 3 versions of Python. I'm not sure I have followed the possible solutions for the osmesa stuff specifically so not sure I can say much on that at this time.

melonora:
Ok I will do accordingly tomorrow and will follow the python version lifecycle for versions supported. In that case I would prefer closing this PR and also the PR with the fix @brisvag had.

brisvag:
Since in napari we follow SPEC-0 but tend to be a bit slower, and we depend on vispy, I suggest we don't go too aggressive here and rather remove older version tests only after we did it there :P So I'd say let's not drop 3.11 yet!

melonora:
it is why I said I would follow the python versions lifecycle: https://devguide.python.org/versions/. Python 3.10 is nearing its end and 3.11 is then also still there

melonora:
closing in favor of #2777 when tests pass

PR #2778, opened 2026-08-19
Support up to py314

melonora:
Adds official support for python 3.14. Classifiers have been updated and ci jobs added.


.github/workflows/main.yml
```diff
@@ -263,9 +273,9 @@ jobs:
     - name: Install conda-dependencies
       run: |
         if [ "${{ matrix.new_qt }}" == 'true' ]; then
-          conda install -y python=${{ matrix.python_version }} -f ci/requirements/linux_${{ matrix.deps }}_newqtdeps_conda.txt
+          conda install -y python=${{ matrix.python_version }} python-gil -f ci/requirements/linux_${{ matrix.deps }}_newqtdeps_conda.txt
```
djhoese: `python` doesn't install the GIL python anymore?

melonora: it is to enforce for python 3.14 that you don't get the freethreaded version as this does not necessarily have the proper Pyqt wheels

djhoese: But does plain `python` via conda (conda-forge) have the free-threaded interpreter binary?

melonora: Yes, but sometimes the solver seems to have an issue with choosing 3.14t depending on the dependency graph. There is an open issue about it: https://github.com/conda-forge/python-feedstock/issues/861

melonora: so I am now explicitly asking for python 3.14 instead of 3.14t

melonora: this also does not solv ethe issue completely though, look at python 3.14 with pyqt5

melonora: attempting now by constraining ABI

melonora: in the UV PR I did not encounter this problem as I did not test PyQt 5 there

melonora: @djhoese why is pydsl2 in the requirements?

brisvag (approved):
LGTM. Let's hear if @djhoese knows about the dep!

.github/workflows/main.yml
djhoese: No idea. It could have been that it was used for other environments or was in some other old requirements files. It has been so long for me and I probably copied it from some other location.

brisvag (changes requested):
Mmm, well it looks like pysdl2 is one of the backends that we're supposed to be testing. I guess by removing it from there we're skipping it? I'm remembering https://github.com/vispy/vispy/pull/2702 and how sdl was indeed being tested in the full deps, so I'd say we probably need to bring it back in.

melonora:
> Mmm, well it looks like pysdl2 is one of the backends that we're supposed to be testing. I guess by removing it from there we're skipping it? I'm remembering #2702 and how sdl was indeed being tested in the full deps, so I'd say we probably need to bring it back in.

ah thought I would find it then in the logs but didn't. Should it not say something like pydsl2 was not installed so skipping?
@brisvag similar to this for example `Skipping backend pyqt5, not installed or working properly:`

In any case it seems like this backend is not supporting python 3.14 at all. I would also see this as argument that we should move to something like pixi where we can have all these differences just in one file and specify things per python version when required.

melonora:
@brisvag @djhoese I added pydsl2 back for those python versions other than python 3.14. Pydsl2 was last released in 2024 and does not officially support python 3.14.

brisvag:
I don't get it... SDL2 is definitely in the deps; but I can't find it in the logs of the conda install section (maybe it's collapsed though) so I can't tell if the install worked out. And then it definitely says that `sdl2` (the correct import name) is not found or is not working so the backend is skipped... it's annoying cause it's nopt marked as a failure even though clearly the intention is to test for it.

brisvag:
Also, I'm on python 3.14 and I have install pysdl2 just fine on my machine!

melonora:
> Also, I'm on python 3.14 and I have install pysdl2 just fine on my machine!

through pip?

brisvag:
Yep, tru! But then, maybe we should do that here too?

melonora:
yeah so pip does not resolve combined with conda as far as I am aware.
I can check and then if tests pass we do it like that for now

melonora:
`_VISPY_TESTING_APP=sdl2 /home/runner/miniconda3/envs/vispy-tests/bin/python -m pytest --tb=short -m vispy_app_test --cov vispy --cov-report= -ra /home/runner/work/vispy/vispy/vispy`

So it does get checked, but not with newqt as it was never added as a dependency there.

melonora:
`_VISPY_TESTING_APP=sdl2 /home/runner/miniconda3/envs/vispy-tests/bin/python -m pytest --tb=short -m vispy_app_test --cov vispy --cov-report= -ra /home/runner/work/vispy/vispy/vispy`

now also with python 3.14, but not with newqt. I followed the earlier policy of not combining with pyqt6

brisvag (commented):
So, to having multiple files per version, can't we just always grab pysdl2 from pip and ditch the py314 files?

melonora:
> So, to having multiple files per version, can't we just always grab pysdl2 from pip and ditch the py314 files?

can I just please follow this up in the pixi pr where I would get rid of all these files all together? haha

brisvag (commented):
I think in the force push you added some stuff that wasn't there? the py314yml... 

But we're nmot sure what the pixi will look like and if it's gonna be good to merge, so let's get the best version of this in that we can :P

melonora:
> I think in the force push you added some stuff that wasn't there? the py314yml...
> 
> But we're nmot sure what the pixi will look like and if it's gonna be good to merge, so let's get the best version of this in that we can :P

the py314.yml was there already, otherwise the conda install would not work properly because of this line in setting up the conda environment:
`environment-file: ./ci/requirements/py${{ steps.vars.outputs.python_version }}.yml`

I will move pysdl2 to pip then and get rid of the .14 files

melonora:
force push was due to me testing out the new gh-stack feature to keep the PRs / branches in sync, but when you look at the compare you see that the rebase did not change the files:)

melonora:
@brisvag I think we are ready

brisvag (approved):
Love it, thanks @melonora 💪

brisvag:
Confirm that sdl tests are running as expected. Merge!

PR #2780, opened 2026-08-21
Change to pixi as manager

melonora:
This PR switches to a pixi build while allowing people to install the old way, such as `pip install -e .[test]`. Furthermore, the following has changed:

- CI build uses pixi now, resolve is quicker and there are no environment files needed anymore except for apt. Pixi is cached, but if the lock file is not up to date, the job can still run and pass.
- build website and other tasks previously defined in Makefile are now pixi tasks
- Most of setup.py has migrated to pyproject.toml, only thing is that np.include_all() cannot be specified in extensions
in pyproject.toml with the current build. If we were to switch to meson-py we could include it and get rid of setup.py, but might be overkill
- updated specific examples since newer numpy is used for building website. `np.row` is deprecated and remove so this is updated.
- readthedocs requirements file is removed as well as unmaintained debian config
- A monthly cronjob checks whether the lock file is stale and creates a PR to main if it isn't
- docs have added instructions for using pixi as an alternative.

@djhoese @brisvag


brisvag:
Do we actually want the pixi lock?

melonora:
> Do we actually want the pixi lock?

I would say yes as with it we can exactly resolve the versions both on CI and local in case something goes wrong and it also allows for the caching in ci resulting in the setup of the environment being a lot quicker. With a fresh resolution all the time we can run into things as a slightly different version number giving a different result. The lock file does give a reproducible set here.

djhoese:
I'm of a decently strong opinion that we shouldn't use lock files in CI so we test against the most recent upstream versions. Also, that we shouldn't suggest a lock file for users as it may conflict with what they need or be slow to update which means they'd be slow on security patches from certain libraries *or* that we lock them into versions of dependencies that are incompatible with their other tools. Having it for developers/contributors as a "this is what we're all using" could be nice for consistency but also is unlikely to be used by everyone (I don't use pixi) or updated by users when it is updated.

It may be better to cache the python environment using github's caching functionality in CI.

melonora:
I agree with not suggesting the lock file for users. This PR still preserves the functionality of just doing `pip install -e .[test]` and other extras. 

I am not certain whether github caching is a better idea. When exactly would the cache expire? We need a policy for that. Based on time? Or would you check whether a dependency updated? It might be easier in case of the latter to just add a bot similar to dependabot that updates the lock file periodically if updating it results in changes. This gives you the behaviour you want I think, but without forcing a developer / user to even use pixi. We could also add a check in the PR itself to update the lock in case the lock is stale, to ensure developers without pixi can still do the work without installing pixi.

One benefit I’ve found with pixi is that the environment definitions are all in pyproject.toml. With the python version changes in the previous PRs I had to update and cross-check several separate conda and pip requirement files and add new ones. With Pixi, the environments are all defined in one file and the solver also checks conda and pypI dependencies together (which previously did not happen). That makes incompatibilities between the two much easier to catch.

djhoese:
What I've done in other projects is have the github CI cache expire after a day. With pixi it'd also be nice to use `exclude-newer` (https://pixi.prefix.dev/latest/security/#2-delay-fresh-uploads-with-exclude-newer). @brisvag your feelings about all this?

djhoese:
Ah you already have caching enabled through the github action and that requires pixi.lock: https://pixi.prefix.dev/latest/integration/ci/github_actions/#caching

Otherwise we'd be making our own using the "more raw" github actions caching. Darn it. Maybe a cronjob to update the pixi.lock? Or yeah maybe a pre-commit task?

melonora:
> Maybe a cronjob

yeah cronjob was indeed what I was thinking, but if you also want to have a solution that a developer is not forced to use pixi I think the update needs to be happen in the ci job as well in case the manifest has changed. Otherwise during the step of setting up pixi it will detect that the lock is not in sync in case the manifest has changed and the developer does not use pixi.

melonora:
> > Maybe a cronjob
> 
> yeah cronjob was indeed what I was thinking, but if you also want to have a solution that a developer is not forced to use pixi I think the update needs to be happen in the ci job as well in case the manifest has changed. Otherwise during the step of setting up pixi it will detect that the lock is not in sync in case the manifest has changed and the developer does not use pixi.

ah wait for the developer case, we do not necessarily have permission to update the lock, so I take it that if it is stale we don't use the cache and then let the periodic job update the lock file.

Pre-commit would force the developer to use pixi.

djhoese:
Pixi would have to be installable, but if a pre-commit hook already exists to do this then it would happen in a sandbox that is also cached or docker based. But yes, not pretty. If it happens in a CI job then it would be an additional job that has permission to update the users branch (which should happen anyway since it would act as a maintainer of this project). But yeah, not great. :thinking:

melonora:
with the last commit I allow on the ci for the lock to be updated when installing if it is not up to date. I believe it is supposed to use the cache if the lock is up to date. This without the developer having to do a thing.

The only other thing that would be required then is the cron job for periodically updating the lock file. If you agree with this, would you want a separate PR for the cron job?

melonora:
added cronjob now as well. I think that should work.

brisvag:
Mmm so a summary of the above discussion would be:
- advantage of lock is caching and reproducibility of CI
- if we want some kind of caching, we anyways need some custom action, and pixi at least has it already solved
- what's the thing with pixi in pre-commit? Why is that needed? Just to update the lock? If so, is that really necessary?

In general, I'm not that comfortable with pixi cause I don't use it much, but we have several folks like @melonora here at napari that are quite comfy with it, so if any issue arise it's easy to fix :shrug:

melonora:
> Mmm so a summary of the above discussion would be:
> 
> * advantage of lock is caching and reproducibility of CI
> * if we want some kind of caching, we anyways need some custom action, and pixi at least has it already solved
> * what's the thing with pixi in pre-commit? Why is that needed? Just to update the lock? If so, is that really necessary?
> 
> In general, I'm not that comfortable with pixi cause I don't use it much, but we have several folks like @melonora here at napari that are quite comfy with it, so if any issue arise it's easy to fix 🤷

I did not put a pre-commit. The idea was that if the manifest is updated the pre-commit would check whether the lock was up to date with the manifest. While possible, this would force every developer to have pixi installed. For that reason I opted for allowing a fresh resolve if the lock is not up to date (`locked: false` in ci). Then the cron job periodically updates the lock in case of new versions of dependencies. It does so by opening a PR so it wouldn't get straight away merged in main. I set the frequency to once per week now, maybe that is a bit too much though.

brisvag:
Ah, cool, ok! Yeah let's maybe do once a month :P Then I guess this PR works for me.

Looking through the changes, I was trying to figure out exactly where the code for making the website comes from, cause it looks different from the one in Makefile... am I getting it right that is' also (supposed to be?) replacing the makefile in `doc`? If so, should that be deleted?

melonora:
> Ah, cool, ok! Yeah let's maybe do once a month :P Then I guess this PR works for me.
> 
> Looking through the changes, I was trying to figure out exactly where the code for making the website comes from, cause it looks different from the one in Makefile... am I getting it right that is' also (supposed to be?) replacing the makefile in `doc`? If so, should that be deleted?

I changed it to a month now. Regarding the website, this part was originally reponsible for it:

```
export DISPLAY=:99.0
export XDG_RUNTIME_DIR=/tmp/runtime-runner
source ./ci/build_website.sh
```

the shell script basically contained this:
```
export TRAVIS_TAG="main"

cd doc

make clean
# TODO: Switch to sphinx-multiversion
make html SPHINXOPTS="-W --keep-going"
touch _build/html/.nojekyll

# move back to source directory so we don't mess with future steps
cd ..
```

This pixi task does essentially the same:
```
build-website = """
rm -rf doc/_build &&
sphinx-build -M html doc doc/_build -W --keep-going &&
touch doc/_build/html/.nojekyll
"""
```
Combined with changes in the CI it is equivalent I believe:
```
env:
        DISPLAY: ":99.0"
        XDG_RUNTIME_DIR: /tmp/runtime-runner
        TRAVIS_TAG: main
      run: pixi run -e website build-website
```
The makefile in docs contains more things, I am not certain whether developers use it or not, but I think if the website is correct that it can be removed. @djhoese thoughts?

djhoese:
The Makefile in the doc directory is the one generated by sphinx when you init a project. I'm always nervous to delete it and I never remember the exact order of the sphinx-build arguments and it is easier to do `make clean; make html` locally. I use it locally or at least I used to when I was actually working on vispy. I don't know if it hurts to keep it. I have modified it over time to include the other directories that need to be cleaned (in the `clean` target) and additional options for `html_dev-pattern` to only run examples matching a pattern. Actually...yeah don't delete the Makefile.

Removing the `ci/build_website.sh` script seems reasonable given the rest of the changes. If I remember correctly the XDG directory was because of how many images we generate it was filling up whatever work directory was being defaulted to. Or at least I think that's why that is set.

melonora:
Then we can keep it and developers can use what they prefer. I think for now there would not necessarily have to be an update on the docs as I think people can still use the old way, but that might be something you want to test out.

melonora:
@djhoese @brisvag what do you think regarding this point?:

Most of setup.py has migrated to pyproject.toml, only thing is that np.include_all() cannot be specified in extensions
in pyproject.toml with the current build. If we were to switch to meson-py we could include it and get rid of setup.py, but might be overkill

brisvag:
I'd love to get rid of the setup.py, but maybe let's leave that as a followup!

brisvag (approved):
I checked on coveralls, and It looks like we lost a tiny bit of coverage, somehow. It's not much, but I wonder why? 

https://coveralls.io/builds/81358074 

image 

Other than that, I think this is good to merge, IMO! @djhoese?

melonora:
hmm strange, `image.py` the only change is that `imageio` and `pillow` always seem to be installed so we don't run into the exceptions. I don't know how that has changed. For the other file it is also just an except import error.

melonora:
@brisvag for the py310-min test I noticed that the old requirements file contains this:

```
coveralls
cython
meshio
numpy
numpydoc
pillow
pytest
pytest-cov
pytest-sugar
scipy
networkx
hsluv
kiwisolver
```
so no imageio there which could have led to the changes. With all these environment files, what used to be in setup.py was not necessarily corresponding with these environment files and the tests did not install the test extras. Is this an issue?

brisvag:
I guess we should reintroduce it, either here or in another test branch, just to recover the lost lines.

melonora:
@brisvag ok will do that. Other question should I add docs for pixi as alternative?

brisvag:
Yeah some docs will be nice! No need for anything big, but at least some guidelines.

djhoese:
Can numpy's includes be used in pyproject.toml-defined extensions if we switched to hatch/hatchling?

melonora:
> Can numpy's includes be used in pyproject.toml-defined extensions if we switched to hatch/hatchling?

seems like it from (https://pypi.org/project/hatch-cython/):

```
[build-system]
requires = ["hatchling", "hatch-cython", "Cython", "setuptools", "numpy"]
build-backend = "hatchling.build"

[tool.hatch.build.targets.wheel.hooks.cython]
dependencies = ["hatch-cython"]

[tool.hatch.build.targets.wheel.hooks.cython.options]
include_numpy = true
define_macros = [["NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"]]
```

melonora:
@djhoese it seems to be bringing extra complexity though as sdf_cpu does not get compiled. Trying to find a fix

melonora:
ok, I think I did what could, but please check build also locally

.github/workflows/main.yml
```diff
@@ -58,11 +58,11 @@ jobs:
         cache: true
         locked: false
     - name: Check line endings
-      run: pixi run -e lint lineendings
+      run: pixi run lineendings
     - name: Lint with flake8
       run: |
         # FIXME: deliberately ignore linting errors for now
-        pixi run -e lint flake || true
+        pixi run flake || true
```
brisvag, line 65: You removed the `-e lint` here but it's still there in the contributor guide; which is right?

pyproject.toml
```diff
@@ -119,47 +121,50 @@ test = [
...
+[tool.hatch.version]
+source = "vcs"
 
-[[tool.setuptools.ext-modules]]
-name = "vispy.visuals.text._sdf_cpu"
-sources = ["vispy/visuals/text/_sdf_cpu.pyx"]
-define-macros = [
-    ["NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"],
+[tool.hatch.version.raw-options]
+local_scheme = "dirty-tag"
```
brisvag, line 128: I can't find any information about this. Where did you find the docs? What does this do?

.github/workflows/main.yml
melonora: Ah yes, I noticed that I didn't assign the tasks to the particular environments for which they are valid so I changed that.

As a result, if a task is only valid for one environment pixi just lets you run it without specifying. So either is fine, but will change it to the more user friendly option

pyproject.toml
melonora: This is hatch-vcs configuration, specifically when building locally. It passes this on to setuptools-scm (found it here: https://github.com/ofek/hatch-vcs). I needed a way to get the equivalent of this line that we had: https://github.com/melonora/vispy/blob/cd03c92b43460e4923dcdfd9ca2883c8bd6d92ca/setup.py#L70

melonora: what it does if building locally adding the tag dirty if there are changes

melonora:
@djhoese are you now also ok with the PR after the recent changes? Would you be ok with merging?

djhoese:
I am off today watching my kids. I won't have time to review until tonight at the earliest.

melonora:
> I am off today watching my kids. I won't have time to review until tonight at the earliest.

Enjoy your day!

pyproject.toml
```diff
@@ -1,27 +1,349 @@
 [build-system]
 requires = [
-    "setuptools>=69.4.0",
+    "hatchling",
+    "hatch-vcs",
+    "hatch-cython",
+    "setuptools",
```
djhoese, line 6: Why do we still need to depend on setuptools?

djhoese: Ah found it in the hatch-cython docs:

> Important: Cython and setuptools must be included in build-system.requires because the build hook imports Cython modules at load time, before hook-specific dependencies are resolved.

pyproject.toml
```diff
@@ -1,27 +1,349 @@
...
+]
+
+[tool.hatch.build.targets.wheel.hooks.cython.options.files]
+targets = [
+    "./vispy/visuals/text/_sdf_cpu.pyx",
+]
+
+# Specifically needed because of hatch-cython hook otherwise including make
+[tool.setuptools]
+packages = ["vispy"]
```
djhoese: I know the comment says hatch-cython but I don't see this section being needed in their docs. I don't understand the "otherwise including make" in the comment?

pyproject.toml
```diff
@@ -1,27 +1,349 @@
...
 before-all = [
     "yum install -y fontconfig || apk add fontconfig-dev jpeg-dev; pip install freetype-py",
 ]
 
 [tool.cibuildwheel.windows]
 environment = {FREETYPEPY_BUNDLE_FT=1}
+
+[tool.pixi.workspace]
+name = "vispy"
+channels = ["conda-forge"]
```
djhoese, line 181: How does pixi handle strict channel resolution? That is, does it include the defaults channel from anaconda by default? In the past we had to be careful to say "don't you dare get packages from defaults, strictly get them only from conda-forge".

djhoese (approved):
I just had a couple small comments but otherwise I think I'm OK with this being merged and to make the leap into pixi-land.

pyproject.toml
melonora: oof, not certain now whether this was actually just hatch. In any case what I remember is that when I did specify this, it would include all the make files etc which caused a build failure. This specifically specifies to only include vispy package

pyproject.toml
melonora: This is also nice with pixi, it actually defaults to ONLY using conda-forge, but just being explicit here. See https://pixi.prefix.dev/v0.43.0/reference/pixi_configuration/#configuration-options

djhoese: Can we add `exclude-newer` (https://pixi.prefix.dev/latest/security/#2-delay-fresh-uploads-with-exclude-newer) to be like 7 days? Or 14 days? Or did we decide that wasn't useful enough?

pyproject.toml
melonora: I double checked and think the comment is outdated. I remember running into it but it could be that this was with hatch instead of setuptools which allows the ordinary `pip install -e .`. I removed the comment.

pyproject.toml
melonora: I added one for 7 days now. I think 14 days would be a bit too long. I think it is still useful if a compromised release happens just before the cron job.

melonora:
@djhoese build website is failing, but not due to changes related to the PR. Lately a lot of the times there have been time outs when preparing system environments. Not only in this repo.

pyproject.toml
```diff
@@ -1,27 +1,349 @@
...
+    ["NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"],
+]
+
+[tool.hatch.build.targets.wheel.hooks.cython.options.files]
+targets = [
+    "./vispy/visuals/text/_sdf_cpu.pyx",
+]
+
+[tool.setuptools]
+packages = ["vispy"]
```
djhoese, line 161: What I was saying in the previous thread or what I meant is that I don't think this entire section should be needed at all. Hatch is not setuptools. Yes, hatch-cython seems to depend on it. You had mentioned that having this or not having this made all the extra files be included. I think that is something that setuptools decided a while back. And if hatch is following that philosophy then I wouldn't be surprised that it includes everything. I would have thought though that your `tool.hatch.build.targets.wheel` section with `packages = ["vispy"]` would have handled that.

Any chance you're willing to lose this section and see what happens?

brisvag: Yesh probably worth trying removing!

melonora: well I can add the comment back in hahaha. Now I remember why I added it in the first place and I can show you as well. When removing this, you get the following:

```
[1/1] Cythonizing ./vispy/visuals/text/_sdf_cpu.pyx
      
      error: Multiple top-level packages discovered in a flat-layout: ['make', 'vispy', 'codegen'].
      
      
      
      To avoid accidental inclusion of unwanted files or directories,
      
      setuptools will not proceed with this build.
```

brisvag: Well, LGTM then!

djhoese: Thanks for checking.

djhoese:
Good to merge from me then. @brisvag I'll let you do the final button click.

brisvag:
Alrgith, let's click! Thanks for the big push @melonora :muscle:

PR #2781, opened 2026-08-24
Fix clipping for direct children of ViewBox

brisvag:
This issue cropped (HAH!) up in several places in the past:

- https://github.com/napari/napari/pull/9083#issuecomment-4795114872
- https://github.com/napari/napari/issues/8802

I finally tracked down the issue:
- clip_children should be True (I don't see why it shouldn't default to True)
- since it was never True, we never even noticed the fact that the clipper setter was setting the wrong attribute name (`rect` instead of `bounds`).

To see the difference, check out this code on main vs here; note how on main how the ellipse "spills" into the neighbouring viewbox when approaching the middle line (on left or right depending which viewbox is parented).

```py
from vispy import scene, app
import numpy as np

canvas.size = 600, 600
canvas.show()

b1 = grid.add_view(row=0, col=0)
b1.border_color = (0.5, 0.5, 0.5, 1)
b1.camera = scene.PanZoomCamera(rect=(-0.5, -5, 11, 10))

b2 = grid.add_view(row=0, col=1)
b2.camera = 'turntable'
b2.border_color = (0.5, 0.5, 0.5, 1)

ellipse = scene.Ellipse(center=(0, 0), color='white', border_width=5, radius=20, parent=b1)

@canvas.events.mouse_move.connect
def on_mouse_move(event):
    viewbox_size = np.array(canvas.size) / (2, 1)
    viewbox_pos = event.pos % viewbox_size
    ellipse.center = viewbox_pos

def on_key(event):
    if event.text == '1':
        # switch viewbox
        ellipse.parent = b2 if ellipse.parent is b1 else b1

if __name__ == '__main__':
    app.run()
```

main:

image

this PR:

image

### DIFF
--- a/vispy/scene/tests/test_viewbox.py
+++ b/vispy/scene/tests/test_viewbox.py
@@ -3,15 +3,17 @@ from vispy.testing import requires_application, TestingCanvas
 
 
 @requires_application()
-def test_viewbox_clips_direct_children():
-    """Direct children of a viewbox should be clipped."""
+def test_viewbox_clips_direct_children_only():
     with TestingCanvas(size=(200, 200)) as c:
         grid = c.central_widget.add_grid()
         view = grid.add_view(row=0, col=0)
         view.size = (100, 100)  # simulate grid layout
         ell = scene.Ellipse(center=(0, 0), radius=20, parent=view)
 
-        # the direct child must inherit the ViewBox clipper
         assert view in ell._clippers
-        # bounds should track the viewbox rect
         assert ell._clippers[view].bounds is not None
+
+        line = scene.Line(pos=[[0, 0], [1, 1]], parent=view.scene)
+
+        assert view._scene in line._clippers
+        assert view not in line._clippers
--- a/vispy/scene/tests/test_viewbox.py
+++ b/vispy/scene/tests/test_viewbox.py
@@ -1,0 +1,17 @@
+from vispy import scene
+from vispy.testing import requires_application, TestingCanvas
+
+
+@requires_application()
+def test_viewbox_clips_direct_children():
+    """Direct children of a viewbox should be clipped."""
+    with TestingCanvas(size=(200, 200)) as c:
+        grid = c.central_widget.add_grid()
+        view = grid.add_view(row=0, col=0)
+        view.size = (100, 100)  # simulate grid layout
+        ell = scene.Ellipse(center=(0, 0), radius=20, parent=view)
+
+        # the direct child must inherit the ViewBox clipper
+        assert view in ell._clippers
+        # bounds should track the viewbox rect
+        assert ell._clippers[view].bounds is not None
--- a/vispy/scene/widgets/viewbox.py
+++ b/vispy/scene/widgets/viewbox.py
@@ -56,6 +56,11 @@ class ViewBox(Widget):
         self._scene.clip_children = True
         self.transforms.changed.connect(self._update_scene_clipper)
 
+        # we also want to clip direct children of viewbox (not just its scene),
+        # such as screen-space overlays
+        self.clip_children = True
+        self._update_clipper()
+
         # Camera is a helper object that handles scene transformation
         # and user interaction.
         if camera is None:
--- a/vispy/scene/widgets/viewbox.py
+++ b/vispy/scene/widgets/viewbox.py
@@ -58,7 +58,7 @@ class ViewBox(Widget):
 
         # we also want to clip direct children of viewbox (not just its scene),
         # such as screen-space overlays
-        self.clip_children = True
+        self._clipper = Clipper()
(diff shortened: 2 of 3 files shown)


djhoese:
Where is `clip_children` actually checked and used?

brisvag:
There were still some issues; now it works as intended.

`clip_children` is machinery from the Widget class. Unmfortunately it operates under the assumption that *all* children should inherit the clipping from their parents, which leads to the `Scene` inside it also inheriting it (which doesn't work because the coordinates are all wrong in screen space instead of scene space).

So, since the Scene is special anyways (it has its own clipping inside it), I ended up kinda side-stepping the normal inheritance, so now it still operates the same for anything *inside the scene* (i.e: uses the scene transform to decide whether a specific fragment should be dropped), but uses its own clipper in screen space to clip anything else.

brisvag:
Note btw that basically this machinery was unused, otherwise the `rect` vs `bounds` typo would have popped up... which means the vast majority of uses simply use the scene clipper and never parent stuff directly to the viewbox.

djhoese:
I don't understand the special casing in the viewbox. I understand that it should maybe be True all the time and then the sub-scene would handling clipping it's children, but then your comment says:

```
# if we used clip_children it would clip all the way down to the scene
```

Isn't that what we want? If you say clip around these bounding coordinates, then shouldn't everything down the scene graph decide if it is clipped or not?

brisvag:
> Isn't that what we want? If you say clip around these bounding coordinates, then shouldn't everything down the scene graph decide if it is clipped or not?

This already happens, because of the `self._scene._clipper` a few lines above; the scene itself has a clipper and handles its own clipping using its own transform, for its own children.

The issue is that if we do `clip_children=True` on the viewbox itself, this applies to all its children, INCLUDING the scene. AFAICT this doesn't work well because the clipping coordinates are based on the tranform of the viewbox (which lives in canvas space), so you end up clipping the actual world in the scene but using 2D screeb coordinates, which results in nonsense.

I'm not sure if there's a way to make it use the proper transform from the viewbox to the scene itself, in which case the two clippers would overlap and just be identical in scene space...

djhoese:
> The issue is that if we do clip_children=True on the viewbox itself, this applies to all its children, INCLUDING the scene.

So why is the current bad behavior happening? I mean, `clip_children` is currently `False` for the viewbox right? In this PR you're adding it. Is that because there are Visuals/Nodes that are children of the Viewbox but not the SubScene? In your example code why is the Ellipse a child of `b1` and not `b1.scene` (a `SubScene` object)? Wouldn't that fix this?

djhoese:
Side note: Some of the examples have deprecation warnings:

```

generating gallery for gallery/gloo... [ 96%] textured_cube.py
/home/runner/work/vispy/vispy/examples/gloo/textured_cube.py:53: DeprecationWarning: `row_stack` alias is deprecated. Use `np.vstack` directly.
  Z = np.row_stack(grid_num // 2 * (row_even, row_odd)).astype(np.uint8)
generating gallery for gallery/gloo... [100%] textured_quad.py
/home/runner/work/vispy/vispy/examples/gloo/textured_quad.py:38: DeprecationWarning: `row_stack` alias is deprecated. Use `np.vstack` directly.
```

djhoese:
Oh and

```
 /home/runner/miniconda3/envs/vispy-tests/lib/python3.10/site-packages/numpydoc/docscrape.py:420: UserWarning: Unknown section Parameters: in the docstring of assert_image_reasonable in /home/runner/work/vispy/vispy/vispy/testing/image_tester.py.
```

brisvag:
> Is that because there are Visuals/Nodes that are children of the Viewbox but not the SubScene? In your example code why is the Ellipse a child of `b1` and not `b1.scene` (a `SubScene` object)? Wouldn't that fix this?

This intentional, and is exactly the kind of use case you were referring to in #2774 when talking about different types of parenting, and what this PR is trying to fix!

When you set the parent to the `viewbox` instead of the `scene`, you get the "screen overlay" behaviour, so it jus sits on top of everything and is unaffected by scene transforms. Very useful and a behaviour we want to support. However, we still want to clip those so they don't spill out into the next viewbox. This viewbox-level clipping is the one that is currently broken.

> So why is the current bad behavior happening? I mean, `clip_children` is currently `False` for the viewbox right? In this PR you're adding it

No, we specifically *don't* set `clip_children=True` here, because it would misbehave as described above. Instead, we have to manage the clipper ourselves with some extra custom methods (otherwise it would go through the normal machinery of the Widget which would end up clipping the scene wrongly).

---

Those warnings are being addressed as part of the bump in #2780 it seems!

djhoese:
Sounds good. Maybe a more descriptive title for the PR? "Fix viewbox widget clipping"? Or...I don't know.

PR #2782, opened 2026-08-25
Add opt-in GLIR texture upload metering

kephale:
## Summary

Add native, opt-in GLIR texture-upload metering for applications that update large textures incrementally.

- split large 2-D and 3-D texture `DATA` commands into bounded slabs
- enforce a configurable per-draw upload budget while preserving per-object GLIR ordering
- retain deferred commands in a per-context carry queue and schedule redraws until it is empty
- support explicit upload holds, weak drain callbacks, and exemptions for textures that must be initialized atomically
- deduplicate redundant GL state commands
- leave default queue behavior and remote GLIR parsers unchanged

Metering is dispatched natively by `_GlirQueueShare.flush`; enabling it does not replace or monkey-patch VisPy methods. The public entry point is `vispy.gloo.glir_metering.install(...)`.

## Motivation

A regular VisPy flush executes every queued texture upload during the next draw. Applications can accumulate tens of megabytes of `glTexSubImage2D` or `glTexSubImage3D` work between draws, turning that draw into a long blocking operation and delaying input handling, particularly on macOS when OpenGL calls are translated to Metal. A per-draw byte budget spreads that work over multiple draws.

## Validation

- focused GLIR suite: 25 passed, 6 skipped
- complete gloo suite: 153 passed
- isolated no-backend suite: 444 passed, 64 skipped, 1 expected failure
- Flake8 and `git diff --check` pass

The tests use fake parsers and require no GL context. They cover slab reconstruction, byte bounds, command ordering, SIZE/DELETE cancellation, remote-parser fallback, upload holds, and callbacks.

## Related work

This API is intended to replace the GLIR monkey-patch currently carried by napari in napari/napari#9067.

GLIR object-deletion pacing was removed from this PR to keep upload metering focused. It is preserved provisionally in draft #2790 and should be evaluated only after this PR merges and downstream napari testing can determine whether it provides a measurable benefit.

### DIFF
--- a/vispy/gloo/context.py
+++ b/vispy/gloo/context.py
@@ -153,13 +153,17 @@ class GLContext(BaseGlooFunctions):
         """The OpenGL capabilities"""
         return deepcopy(self.shared.parser.capabilities)
 
-    def flush_commands(self, event=None):
+    def flush_commands(self, event=None, force=False):
         """Flush
 
         Parameters
         ----------
         event : instance of Event
             The event.
+        force : bool
+            Execute all currently executable queued work. Automatic draw
+            flushes leave this false so optional command scheduling can apply
+            a per-draw budget.
         """
         if self._do_CURRENT_command:
             self._do_CURRENT_command = False
@@ -169,7 +173,11 @@ class GLContext(BaseGlooFunctions):
             else:
                 fbo = 0
             self.shared.parser.parse([('CURRENT', 0, fbo)])
-        self.glir.flush(self.shared.parser)
+        if force:
+            self.glir.flush(self.shared.parser, force=True)
+        else:
+            # Preserve the long-standing one-argument call for normal draws.
+            self.glir.flush(self.shared.parser)
 
     def set_viewport(self, *args):
         BaseGlooFunctions.set_viewport(self, *args)
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -501,8 +501,11 @@ class _GlirQueueShare(object):
     def flush(self, parser, force=False):
         """Flush current commands to the GLIR interpreter.
 
-        When *force* is true, optional command scheduling must execute all
-        currently executable work instead of applying a per-draw budget.
+        When ``force=False``, any configured command scheduling policy, such
+        as texture-upload metering, is respected. When ``force=True``, such
+        scheduling limits are bypassed and all currently executable commands
+        are processed. If no scheduling policy is enabled, *force* has no
+        effect and commands are flushed immediately as usual.
         """
         from . import glir_metering
 
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -499,7 +499,7 @@ class _GlirQueueShare(object):
         return commands
 
     def flush(self, parser, force=False):
-        """Flush all current commands to the GLIR interpreter.
+        """Flush current commands to the GLIR interpreter.
 
         When *force* is true, optional command scheduling must execute all
         currently executable work instead of applying a per-draw budget.
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -498,15 +498,19 @@ class _GlirQueueShare(object):
         self._commands = []
         return commands
 
-    def flush(self, parser):
-        """Flush all current commands to the GLIR interpreter."""
+    def flush(self, parser, force=False):
+        """Flush all current commands to the GLIR interpreter.
+
+        When *force* is true, optional command scheduling must execute all
+        currently executable work instead of applying a per-draw budget.
+        """
         from . import glir_metering
 
         if (
             glir_metering.is_installed()
             and getattr(parser, 'supports_texture_upload_metering', False)
         ):
-            glir_metering._flush(self, parser)
+            glir_metering._flush(self, parser, force=force)
             return
         if self._verbose:
             show = self._verbose if isinstance(self._verbose, str) else None
@@ -587,9 +591,9 @@ class GlirQueue(object):
             self._shared._associations[ch] = None
         queue._shared = self._shared
 
-    def flush(self, parser):
+    def flush(self, parser, force=False):
         """Flush all current commands to the GLIR interpreter."""
-        self._shared.flush(parser)
+        self._shared.flush(parser, force=force)
 
 
 def _convert_es2_shader(shader):
@@ -758,6 +762,9 @@ class GlirParser(BaseGlirParser):
 
         if cmd == 'CURRENT':
             # This context is made current
+            from . import glir_metering
+
+            glir_metering._on_parser_current(self)
             self.env.clear()
             self._gl_initialize()
             self.env['fbo'] = args[0]
--- a/vispy/gloo/glir.py
+++ b/vispy/gloo/glir.py
@@ -500,6 +500,14 @@ class _GlirQueueShare(object):
 
     def flush(self, parser):
         """Flush all current commands to the GLIR interpreter."""
+        from . import glir_metering
+
+        if (
+            glir_metering.is_installed()
+            and getattr(parser, 'supports_texture_upload_metering', False)
+        ):
+            glir_metering._flush(self, parser)
+            return
         if self._verbose:
             show = self._verbose if isinstance(self._verbose, str) else None
             self.show(show)
@@ -671,6 +679,8 @@ def as_es2_command(command):
 class BaseGlirParser(object):
     """Base class for GLIR parsers that can be attached to a GLIR queue."""
 
+    supports_texture_upload_metering = False
+
     def __init__(self):
         self.capabilities = dict(
             gl_version='Unknown',
@@ -703,6 +713,8 @@ class GlirParser(BaseGlirParser):
     dictionary so that commands like ACTIVATE and DATA can easily
     be executed on the corresponding objects.
     """
+
+    supports_texture_upload_metering = True
 
     def __init__(self):
         super(GlirParser, self).__init__()
--- a/vispy/gloo/glir_metering.py
+++ b/vispy/gloo/glir_metering.py
@@ -29,6 +29,13 @@ shared OpenGL context that receives the uploads. The budget resets at the
 start of each canvas draw. Offscreen users without draw events receive a
 time-based reset so a carry queue cannot remain stalled indefinitely.
 
+While metering is enabled, two related sources of GL stalls are also paced.
+Repeated idempotent ``FUNC`` state setters with unchanged arguments are
+skipped; this cache is cleared whenever a ``CURRENT`` command makes the
+context current. ``DELETE`` commands are held until a flush with no upload
+work, then executed in small batches. These behaviors apply only to the local
+parser path selected by :func:`install`.
+
 Metering is process-wide and opt-in through :func:`install`. It applies only
 to local OpenGL parsers; remote GLIR parsers and the default disabled path are
 unchanged.
@@ -98,10 +105,11 @@ _OBJECT_COMMANDS = frozenset(
 # the same arguments is semantically a no-op, but on a busy GPU each
 # call can block on pipeline synchronization (profiled at ~6ms average
 # on macOS GL-over-Metal; 271 calls cost 1.6s in one short session).
-# Consecutive duplicates are skipped. glEnable/glDisable are handled
-# per capability. Deliberately NOT listed: glClear*, glViewport,
-# glScissor, glFinish, glFlush (must always run or are
-# framebuffer-dependent).
+# Repeated duplicates are skipped. glEnable/glDisable are handled per
+# capability. State setters such as glViewport, glScissor, and glClearColor
+# are included because repeating them with identical arguments leaves the
+# context state unchanged. Commands that perform work, including glClear,
+# glFinish, and glFlush, are deliberately excluded and must always run.
 _IDEMPOTENT_FUNCS = frozenset(
     {
         'glBlendFunc',
@@ -141,15 +149,21 @@ _hooked_canvases: weakref.WeakSet = weakref.WeakSet()
 # carried without execution. Non-upload commands continue to execute.
 _upload_hold_until = 0.0
 
-# GLIR texture ids whose DATA commands bypass the byte budget. This is useful
-# when an application writes a complete staging texture before atomically
-# making it visible: splitting that write across draws could expose partially
-# initialized texture contents if the application switches textures too soon.
+# GLIR texture ids whose DATA commands bypass the byte budget. A typical use is
+# double buffering: an application fills a back texture, then makes it the
+# front texture only after the upload completes. The exemption makes the DATA
+# command synchronous; it neither swaps the textures nor coordinates when the
+# application makes the back texture visible.
 _unmetered_ids: set = set()
 
 
 def add_unmetered_texture(glir_id) -> None:
     """Make DATA commands for one GLIR texture bypass the byte budget.
+
+    This is useful for a staging or back texture that must be initialized in
+    one synchronous upload before an application makes it visible. VisPy does
+    not perform or schedule that visibility change; the application remains
+    responsible for it.
 
     Parameters
     ----------
--- a/vispy/gloo/glir_metering.py
+++ b/vispy/gloo/glir_metering.py
@@ -1,0 +1,654 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+"""Per-frame metering of vispy GLIR texture uploads.
+
+vispy drains its entire GLIR command queue inside whichever draw happens
+next — including interaction frames. With progressive loading, dozens of
+megabytes of ``glTexSubImage3D`` traffic can accumulate between draws and
+the drain then blocks the main thread for hundreds of milliseconds to
+seconds (macOS GL-over-Metal is the worst case: ~125 MB/s effective with
+multi-second outliers for large single uploads).
+
+When enabled, VisPy's GLIR queue dispatches through this module so that
+
+- a single large texture ``DATA`` command is split into small slab
+  sub-uploads (contiguous views along the leading axes, no copies), and
+- each frame spends at most ``frame_budget_bytes`` on texture uploads;
+  the remainder is carried to the *next* frame and a redraw is scheduled
+  so the carry keeps draining even without interaction.
+
+GLIR ordering semantics are preserved per object: once a command for a
+texture is deferred, every later command for that same texture id is
+deferred behind it. ``SIZE`` (re-specification) and ``DELETE`` commands
+cancel any earlier carried uploads for their id, mirroring vispy's own
+queue filtering.
+
+Metering is opt-in through :func:`install`; the default GLIR behavior is
+unchanged.
+
+Examples
+--------
+Enable a 4 MiB per-frame budget with 1 MiB upload slabs::
+
+    from vispy.gloo import glir_metering
+
+    glir_metering.install(
+        frame_budget_bytes=4 * 2**20,
+        slab_bytes=1 * 2**20,
+    )
+
+Run ``python -m vispy.gloo.glir_metering`` for a standalone
+vispy-only benchmark of draw time vs. texture upload size,
+with and without metering.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import logging
+import time
+import weakref
+
+import numpy as np
+
+LOGGER = logging.getLogger(__name__)
+
+__all__ = [
+    'DEFAULT_FRAME_BUDGET_BYTES',
+    'DEFAULT_SLAB_BYTES',
+    'add_drain_callback',
+    'add_unmetered_texture',
+    'discard_unmetered_texture',
+    'hold_uploads_until',
+    'install',
+    'is_installed',
+    'pending_upload_bytes',
+    'remove_drain_callback',
+    'uninstall',
+]
+
+_FACTORY_FRAME_BUDGET_BYTES = 4 * 2**20
+_FACTORY_SLAB_BYTES = 1 * 2**20
+DEFAULT_FRAME_BUDGET_BYTES = _FACTORY_FRAME_BUDGET_BYTES
+DEFAULT_SLAB_BYTES = _FACTORY_SLAB_BYTES
+#: Deferred GL object deletions executed per quiet flush. Deletion can
+#: sync the GPU pipeline, so the backlog drains a few per frame instead
+#: of all at once.
+DELETE_DRAIN_PER_FLUSH = 4
+#: 2D texture DATA at or above this size is metered like 3D uploads.
+#: Small 2D textures (colormap LUTs, interpolation kernels) MUST stay
+#: synchronous: deferring them leaves a shader sampling an unwritten
+#: texture for however long the carry takes to drain.
+TEX2D_MIN_METERED_BYTES = 256 * 1024
+
+# Commands that operate on a specific GLIR object id and therefore must
+# stay ordered behind any deferred command for the same id.
+_OBJECT_COMMANDS = frozenset(
+    {'DATA', 'SIZE', 'WRAPPING', 'INTERPOLATION', 'DELETE'},
+)
+
+# GL state setters that are idempotent: re-issuing the same call with
+# the same arguments is semantically a no-op, but on a busy GPU each
+# call can block on pipeline synchronization (profiled at ~6ms average
+# on macOS GL-over-Metal; 271 calls cost 1.6s in one short session).
+# Consecutive duplicates are skipped. glEnable/glDisable are handled
+# per capability. Deliberately NOT listed: glClear*, glViewport,
+# glScissor, glFinish, glFlush (must always run or are
+# framebuffer-dependent).
+_IDEMPOTENT_FUNCS = frozenset(
+    {
+        'glBlendFunc',
+        'glBlendFuncSeparate',
+        'glBlendEquation',
+        'glBlendEquationSeparate',
+        'glBlendColor',
+        'glDepthFunc',
+        'glDepthMask',
+        'glDepthRange',
+        'glCullFace',
+        'glFrontFace',
+        'glLineWidth',
+        'glPolygonOffset',
+        'glColorMask',
+        'glStencilFunc',
+        'glStencilFuncSeparate',
+        'glStencilMask',
+        'glStencilMaskSeparate',
+        'glStencilOp',
+        'glStencilOpSeparate',
+        'glHint',
+        'glSampleCoverage',
+        # value-state setters: global GL state, idempotent by args.
+        # glClear itself must always run and is never cached.
+        'glViewport',
+        'glScissor',
+        'glClearColor',
+        'glClearDepth',
+        'glClearStencil',
+    }
+)
+
+_enabled = False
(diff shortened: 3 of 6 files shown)


kephale:
@melonora over here!

melonora:
Great! Will check it out tomorrow morning my time:)

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,701 @@
...
+)
+
+# GL state setters that are idempotent: re-issuing the same call with
+# the same arguments is semantically a no-op, but on a busy GPU each
+# call can block on pipeline synchronization (profiled at ~6ms average
+# on macOS GL-over-Metal; 271 calls cost 1.6s in one short session).
+# Consecutive duplicates are skipped. glEnable/glDisable are handled
+# per capability. Deliberately NOT listed: glClear*, glViewport,
+# glScissor, glFinish, glFlush (must always run or are
+# framebuffer-dependent).
```
melonora: hmm you mention that these are deliberately not listed, but except for `glFinish` and `glFlush` they are. What are the implications of having them added?

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,701 @@
...
+_enabled = False
+_hooked_canvases: weakref.WeakSet = weakref.WeakSet()
+# Absolute ``time.monotonic`` deadline before which metered DATA commands are
+# carried without execution. Non-upload commands continue to execute.
+_upload_hold_until = 0.0
+
+# GLIR texture ids whose DATA commands bypass the byte budget. This is useful
+# when an application writes a complete staging texture before atomically
+# making it visible: splitting that write across draws could expose partially
+# initialized texture contents if the application switches textures too soon.
```
melonora: Does this particularly mean here when for example a back buffer would be swapped into view? e.g. does this relate texture buffering in the progressive loading implementation?

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,701 @@
...
+                if gl_state.get(key) == command[2:]:
+                    continue
+                gl_state[key] = command[2:]
+        elif cmd == 'CURRENT':
+            # context switch: cached GL state is no longer trustworthy
+            gl_state.clear()
+        elif cmd == 'DELETE':
+            # run deletes only in quiet flushes (see _ParserState):
+            # ordering is safe because anything queued for this id
+            # earlier either already executed or was dropped with it
```
melonora, line 437: I think it would maybe be nice to add this info to the docstring. As I understand `DELETE` commands are sort of scheduled. I believe I also saw a `gl_state` cache elsewhere that was not yet mentioned in the module docstring. I think these 2 things go beyond the mentioned scope of the module, which on itself I think is ok, but just for clarity would be good to add:)

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,701 @@
...
+
+    # position='first': run before the scene draw issues any flushes
+    try:
+        canvas.events.draw.connect(_on_draw, position='first')
+    except RuntimeError:
+        return
+    _hooked_canvases.add(canvas)
+
+
+def _flush(queue, parser):
```
melonora: Do you envision that this could be integrated with the existing vispy `flush`? Would it be possible to make the metering an optional behaviour there?

vispy/gloo/glir_metering.py
kephale: Updated!

vispy/gloo/glir_metering.py
kephale: Yeah, this is for the double buffering we have in the progressive loading implementation in the napari PR.

vispy/gloo/glir_metering.py
kephale: Agreed, clarified in the module docstring.

vispy/gloo/glir_metering.py
kephale: Arg, so we should probably do this to have better context scoping on the GLIR metering...

vispy/gloo/glir.py
```diff
@@ -498,8 +498,20 @@ def clear(self):
         self._commands = []
         return commands
 
-    def flush(self, parser):
-        """Flush all current commands to the GLIR interpreter."""
+    def flush(self, parser, force=False):
+        """Flush all current commands to the GLIR interpreter.
```
melonora:
```suggestion
        """Flush current commands to the GLIR interpreter.
```

Very minor, but I think this is now the more correct docstring

vispy/gloo/glir.py
```diff
@@ -498,8 +498,20 @@ def clear(self):
...
+        currently executable work instead of applying a per-draw budget.
+        """
+        from . import glir_metering
+
+        if (
+            glir_metering.is_installed()
+            and getattr(parser, 'supports_texture_upload_metering', False)
+        ):
+            glir_metering._flush(self, parser, force=force)
+            return
```
melonora, line 517: If `force=False` and glir_metering is not installed, should be throw an error or fallback to default warning?

vispy/gloo/glir.py
```diff
@@ -498,8 +498,20 @@ def clear(self):
         self._commands = []
         return commands
 
-    def flush(self, parser):
-        """Flush all current commands to the GLIR interpreter."""
+    def flush(self, parser, force=False):
+        """Flush all current commands to the GLIR interpreter.
+
+        When *force* is true, optional command scheduling must execute all
+        currently executable work instead of applying a per-draw budget.
```
melonora:
```suggestion
        When `force=False`, any configured command scheduling policy, such as texture-upload 
        metering, is respected. When `force=True`, such scheduling limits are bypassed and all 
        currently executable commands are processed. 
```
Question, should we be more explicit that even when this is set to `False` this does not necessarily mean that metering is applied? Similar to my question elsewhere, if this is `False` but there is no `glir_metering` installed I take it this should be the same as the old behaviour. In line of that, do we want `False` to be the default or `True`? Just raising it to discuss.

For the part here I would just be a little bit more explicit already and we can decide together on the default.

Edit: ah I see that in the wrappers the default is `True`.

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,758 @@
...
+# Absolute ``time.monotonic`` deadline before which metered DATA commands are
+# carried without execution. Non-upload commands continue to execute.
+_upload_hold_until = 0.0
+
+# GLIR texture ids whose DATA commands bypass the byte budget. A typical use is
+# double buffering: an application fills a back texture, then makes it the
+# front texture only after the upload completes. The exemption makes the DATA
+# command synchronous; it neither swaps the textures nor coordinates when the
+# application makes the back texture visible.
+_unmetered_ids: set = set()
```
melonora, line 160: hmm is this really meant to be global actually? Because from reading other parts of the code and the PR description this is part of a state and the scheduling state belongs to the shared GL context no?

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,758 @@
...
+    'is_installed',
+    'pending_upload_bytes',
+    'remove_drain_callback',
+    'uninstall',
+]
+
+_FACTORY_FRAME_BUDGET_BYTES = 4 * 2**20
+_FACTORY_SLAB_BYTES = 1 * 2**20
+DEFAULT_FRAME_BUDGET_BYTES = _FACTORY_FRAME_BUDGET_BYTES
+DEFAULT_SLAB_BYTES = _FACTORY_SLAB_BYTES
```
melonora: I think the DEFAULT variables are misleading. These can be changed at runtime correct? If so, then I think this is better:
```suggestion
_DEFAULT_FRAME_BUDGET_BYTES = 4 * 2**20
_DEFAULT_SLAB_BYTES = 1 * 2**20
_frame_budget_bytes = _DEFAULT_FRAME_BUDGET_BYTES
_slab_bytes = _DEFAULT_SLAB_BYTES
```
and then change the used variables elsewhere.

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,758 @@
...
+    cannot affect other object types (notably shaders, whose DATA must
+    survive even though vispy DELETEs them right after LINK).
+    """
+    deleted = {c[1] for c in new_commands if c[0] == 'DELETE'}
+    if not deleted:
+        return carry
+    return [c for c in carry if c[1] not in deleted]
+
+
+def _metered_parse(
```
melonora, line 386: not certain whether this can me more integrated with `GlirParser.parse()` as the comment mentions it is mirroring. I would perhaps document this for an issue to investigate. I can be assigned for that, and I don't think that this PR should be blocked for that. We can follow up

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,758 @@
...
+        commands,
+        state,
+        force_defer=holding,
+        force_uploads=force,
+        defer_deletes=not force,
+    )
+    if force:
+        # A forced drain is outside the per-draw allowance. Start the next
+        # automatic draw with a fresh budget rather than a negative balance.
+        state.reset_budget()
```
melonora, line 590: Not entirely certain here, but one of the `state.reset_budget()` feels redundant. Am I correct?

melonora: If not, maybe add a comment as to why:)

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,758 @@
...
+def _drain_deletes(parser, state, limit=None):
+    """Execute up to *limit* previously deferred DELETE commands."""
+    deletes = state.deferred_deletes[:limit]
+    del state.deferred_deletes[:len(deletes)]
+    for command in deletes:
+        parser._parse(command)
+        state.dead_ids.add(command[1])
+
+
+def _flush(queue, parser, force=False):
```
melonora, line 553: maybe modularizing this function a bit more might make it easier, e.g. putting `carry, state.carry...` until `had_carry` in a function called `collect_glir_commands` and then there is a part executing the commands. Lastly a smaller function for the canvas updates or notify_drained.

I think that would make the flow directly clear when going through this.

melonora:
@kephale one general note, I find `deferred_commands` a bit easier to directly understand than `carry`, but maybe that is just me.

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,758 @@
...
+    had_carry = bool(carry)
+    state.executed_metered = False
+    state.carry = _metered_parse(
+        parser,
+        commands,
+        state,
+        force_defer=holding,
+        force_uploads=force,
+        defer_deletes=not force,
+    )
```
melonora: I think there are 3 combinations now here. If force is `True`, then `force_defer=False`, `force_uploads=True` and `defer_deletes=False`. When `force=False`, `defer_deletes` is now always `False` and holding is either `True` or `False`. 

I am not certain if there would be a use case for forcing uploads while at the same time deferring deletes, but if not I think we could have a `FlushMode` enum here replacing the 3 arguments. Maybe?:
```
_SCHEDULED
_UPLOAD_HOLD
_FORCE_DRAIN
```

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,758 @@
...
+        # multi-second stall after a burst of object creation, when the
+        # deletion queue is deepest. The remainder drains over subsequent
+        # redraws.
+        _drain_deletes(parser, state, DELETE_DRAIN_PER_FLUSH)
+        if state.deferred_deletes and canvas is not None:
+            with contextlib.suppress(RuntimeError):
+                canvas.update()
+
+
+def install(
```
melonora, line 614: Maybe update function def, also for uninstall as it is more enable now

vispy/gloo/glir_metering.py
melonora: I like the changes so far! I think as a follow up PR we can look at moving more of the `metered_parse` machinery so there is no need for mirroring.

vispy/gloo/glir_metering.py
melonora: Perhaps modularizing the `metered_parse` function would make it easier to understand the logic:)

kephale:
Addressed the review round in the latest commits:

- metering now integrates through the existing GLIR queue flush path and remains opt-in
- scheduling state is per shared parser/context
- runtime settings are distinct from defaults
- flush behavior uses explicit scheduled/upload-hold/force-drain modes
- command collection, execution, redraw, and drain notification are separated
- module and API docstrings describe deferred deletion and GL-state caching

Fresh focused validation: 30 passed, 6 skipped in the GLIR suites. All 20 hosted checks are green.

vispy/gloo/context.py
```diff
@@ -169,7 +173,11 @@ def flush_commands(self, event=None):
             else:
                 fbo = 0
             self.shared.parser.parse([('CURRENT', 0, fbo)])
-        self.glir.flush(self.shared.parser)
+        if force:
+            self.glir.flush(self.shared.parser, force=True)
+        else:
+            # Preserve the long-standing one-argument call for normal draws.
+            self.glir.flush(self.shared.parser)
```
brisvag, line 180: I don't get why this is needed: `flush()` now has two arguments, so it should always work to just pass `force=force`? Or am I missing something?

vispy/gloo/glir.py
brisvag: I think we should do nothing. All existing use cases (if anyone is manually using `glir` outside of vispy) would be with no `force` argument, so we don't want to spam warnings there.

vispy/gloo/glir.py
```diff
@@ -671,6 +686,8 @@ def as_es2_command(command):
 class BaseGlirParser(object):
     """Base class for GLIR parsers that can be attached to a GLIR queue."""
 
+    supports_texture_upload_metering = False
```
brisvag, line 689: Why this distinction?

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+shared OpenGL context that receives the uploads. The budget resets at the
+start of each canvas draw. Offscreen users without draw events receive a
+time-based reset so a deferred queue cannot remain stalled indefinitely.
+
+While metering is enabled, two related sources of GL stalls are also paced.
+Repeated idempotent ``FUNC`` state setters with unchanged arguments are
+skipped; this cache is cleared whenever a ``CURRENT`` command makes the
+context current. ``DELETE`` commands are held until a flush with no upload
+work, then executed in small batches. These behaviors apply only to the local
+parser path selected by :func:`install`.
```
brisvag, line 37: This explanation is very dense: can you elaborate a bit (not necessarily in the docstring, just here for me to understand).

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+
+Metering is process-wide and opt-in through :func:`install`. It applies only
+to local OpenGL parsers; remote GLIR parsers and the default disabled path are
+unchanged.
```
brisvag, line 41: Any reason to do this process-wide instead of doing it per-Queue? Not sure *which* cases use/require multiple queues, but if there's no specific reason, I guess that would be safer.

However, I don't know how that would look like in terms of usage...

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+_slab_bytes = _DEFAULT_SLAB_BYTES
+#: Deferred GL object deletions executed per quiet flush (a flush with no
+#: texture uploads and no active upload hold). Deletion can synchronize the
+#: GPU pipeline, so only a small batch executes in each such flush.
+DELETE_DRAIN_PER_FLUSH = 4
+#: 2D texture DATA at or above this size is metered like 3D uploads.
+#: Small 2D textures (colormap LUTs, interpolation kernels) MUST stay
+#: synchronous: deferring them leaves a shader sampling an unwritten
+#: texture for however long deferred uploads take to drain.
+TEX2D_MIN_METERED_BYTES = 256 * 1024
```
brisvag, line 99: This comment makes me wonder if we should rather add a way to mark a buffer/texture as "undeferred", to communicate to the glir "this should always be flushed synchronously"? Then we could flag colormaps and stuff like that with this and ensure that this never happens. I'm thinking of something similar to when I implemented instanced rendering in https://github.com/vispy/vispy/pull/2378.

On the other hand, this might be a lot of work, where a heuristic like the above might be already pretty good 🤔 

What's your reasoning on the number chosen? Is it conservative? And what's the donwnside of sampling an unwritten texture, just a temporary visual glitch?

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+
+# Commands that operate on a specific GLIR object id and therefore must
+# stay ordered behind any deferred command for the same id.
+_OBJECT_COMMANDS = frozenset(
+    {'DATA', 'SIZE', 'WRAPPING', 'INTERPOLATION', 'DELETE'},
+)
```
brisvag, line 105: Maybe I'm misunderstanding this, but shouldn't at least `DELETE` simply erase all preceding pending commands on the same `ID`?

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+# GL state setters that are idempotent: re-issuing the same call with
+# the same arguments is semantically a no-op, but on a busy GPU each
+# call can block on pipeline synchronization (profiled at ~6ms average
+# on macOS GL-over-Metal; 271 calls cost 1.6s in one short session).
+# Repeated duplicates are skipped. glEnable/glDisable are handled per
+# capability. State setters such as glViewport, glScissor, and glClearColor
+# are included because repeating them with identical arguments leaves the
+# context state unchanged. Commands that perform work, including glClear,
+# glFinish, and glFlush, are deliberately excluded and must always run.
+_IDEMPOTENT_FUNCS = frozenset(
```
brisvag, line 116: I don't know how to review this list; I understand the idea, but I don't know where to look up if indeed this are idempotent, and whether this list is exhaustive (though I guess that's less important). @kephale how did you get to these? @djhoese any suggestions? 😅

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+        'glStencilFunc',
+        'glStencilFuncSeparate',
+        'glStencilMask',
+        'glStencilMaskSeparate',
+        'glStencilOp',
+        'glStencilOpSeparate',
+        'glHint',
+        'glSampleCoverage',
+        # value-state setters: global GL state, idempotent by args.
+        # glClear itself must always run and is never cached.
```
brisvag, line 140: Why are these singled out? Aren't all of these "state setters" in some way?

vispy/gloo/glir_metering.py
brisvag: If I'm understanding this right, this seems like another point in favour of explicitly marking some things as "unbuffered" as I mentioned above, instead of having this global id.

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+    glir_id : int
+        The texture object's GLIR identifier.
+    """
+    _unmetered_ids.discard(glir_id)
+
+
+# Weakly referenced bound methods invoked on the drawing thread after a flush
+# executes metered DATA commands and leaves no deferred upload commands. A
+# consumer can use this notification instead of polling pending_upload_bytes().
+_drain_callbacks: list = []
```
brisvag, line 193: This of course should also be non-global, if we go that route.

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+        # have been carried
+        self.executed_metered = False
+
+    def reset_budget(self):
+        self.budget_left = self.frame_budget
+        self.last_reset = time.perf_counter()
+
+
+# keyed by GlirParser so canvases sharing a context share one budget
+_states: weakref.WeakKeyDictionary = weakref.WeakKeyDictionary()
```
brisvag, line 299: Is this a trick just to keep the code self-contained here, or is there a reason to have this be a global dictionary instead of just `GlirParser._state` or something like that?

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+    """Drop deferred uploads whose texture is deleted in this flush.
+
+    The deferred queue only ever holds commands for metered textures, so this
+    cannot affect other object types (notably shaders, whose DATA must
+    survive even though vispy DELETEs them right after LINK).
+    """
+    deleted = {c[1] for c in new_commands if c[0] == 'DELETE'}
+    if not deleted:
+        return deferred_commands
+    return [c for c in deferred_commands if c[1] not in deleted]
```
brisvag, line 377: Mhm, this does seem to do what I mentioned in https://github.com/vispy/vispy/pull/2782/changes#r3934242061, but then I don't understand what `DELETE` was doing in that frozenset there.

vispy/gloo/glir_metering.py
brisvag: This also would likely become easier to follow if we just had `parser` hold the state itself and we could just have an if/else clause in there that makes it clear *which* commands are hijacked by these and which are not.

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+    )
+    slab_bytes = int(_slab_bytes if slab_bytes is None else slab_bytes)
+    if frame_budget_bytes <= 0:
+        raise ValueError('frame_budget_bytes must be positive')
+    if slab_bytes <= 0:
+        raise ValueError('slab_bytes must be positive')
+
+    # re-parameterize existing states (and set defaults for new ones)
+    _frame_budget_bytes = frame_budget_bytes
+    _slab_bytes = slab_bytes
```
brisvag, line 651: I'm not sure about this; these are global defaults that are overriden by a specific install call? I guess it's also due to the fact that we're using all these global states, and not necessarily something that needs to be like this?

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+
+
+def _benchmark():  # pragma: no cover - manual profiling tool
+    """Pure-vispy benchmark: draw time vs. 3D texture upload size.
+
+    Renders a vispy Volume and times the draw immediately after
+    ``set_data`` calls of increasing size, with metering off and on.
+    This is the minimal repro for the macOS GL-over-Metal upload stalls
+    (run on Apple Silicon and compare the two curves).
+    """
```
brisvag, line 701: Should be moved to a separate module, maybe `examples/benchmark`!

Would be great if it actually showed something, so we can see in person what the effects of this are. Currently, when running this code I see no appreciable difference from the unemetered and metered, but I don't know if I'm doing something wrong. It would be great both to see the effects on the drawing, and to be able to switch between metered and unmetered manually to quickly test the effects,

Also, using `__doc__` as the description is not quite working well, it dumps all in a single line when doing `python -m vispy.gloo.glir_metering --help`.

vispy/gloo/glir_metering.py
```diff
@@ -0,0 +1,776 @@
...
+        state.reset_budget()
+
+    _request_redraw_or_notify(canvas, state, had_deferred)
+
+    if (
+        not force
+        and state.deferred_deletes
+        and not holding
+        and not state.deferred_commands
+        and state.budget_left >= state.frame_budget
```
melonora, line 599: force and not holding is now in `_FlushMode` I think:) can combine the 2 by just using `mode is _FlushMode.SCHEDULED`

brisvag (commented):
@kephale this is looking great, thanks for pushing this upstream!

A general vibe that comes out of many comments below: IMO if possible we should be less worried about isolating the code and integrate it a bit more with the existing one, to avoid these hard-to-follow patterns like the global `_states` weakdict, stuff like the external call to `_on_parser_current`, etc. A lot of the code can still live in this module, but I think we can explicitly import it in `glir` and instantiate a state per-parser in the init and save ourselves some headache.

vispy/gloo/tests/test_glir_metering.py
```diff
@@ -0,0 +1,757 @@
+# -*- coding: utf-8 -*-
```
melonora, line 1: I think mostly this test module is still missing some tests. I could not find any test related to `add_unmetered_texture` for example. Also `discard_unmetered_texture`. Another is that not all `FlushMode` are tested. The default is set to `Scheduled` and I don't see `mode` specified in any of the tests

vispy/gloo/glir_metering.py
djhoese: No idea. This whole PR is pretty beyond my understanding and I don't think I have the spare brain space (and I might never have it given the size of this PR) to understand this well enough to help review this.

vispy/gloo/glir.py
melonora: From the tests it seemed implied that a remote parser might not support metering. Personally, I haven't worked with this, but I can imagine the parser might not be able to inspect textures if it is a remote one in which case even if glit_metering is enabled, it might not be possible.

melonora: @kephale correct me if I am off with my explanation

vispy/gloo/glir_metering.py
melonora: Because the budget is not per queue, but per shared GL context. I think therefore making it per queue could lead to weird behaviour.

brisvag: I'm asking *why* is it per-process :P Is there a single GL context per process? And even if so, why could it not be per-queue? What weird behaviour?

vispy/gloo/glir_metering.py
brisvag: No worries, I pinged you just in case but I imagined :) thanks anyways for answering!

vispy/gloo/glir_metering.py
melonora: Maybe saying something wrong here, but the way I see this in context of this PR is that it allows global enumeration. If you were to move this to `GlirParser`, how would you ensure that the budget is shared between parsers? The way I understand it is that you potentially end up with multiplying the upload budget to the GPU so more will be uploaded than intended if you have more than 1 parser and they do not share the budget.

vispy/gloo/glir_metering.py
melonora: is there currently something that would prevent a shared context to have more than 1 queue? Because if that is a possiblity, the budget would multiply with the amount of queues you have. A global budget is easier to control and reason about in my opinion than making this a per queue option. In that case if you still want to keep to the budget you still need something on top of the per queue budget to ensure the queues together don't exceed the intended budget.

melonora: So I guess the main question is whether multiple queues can share a GL context. If yes, I don't think the budget should be per queue.

melonora: I also think process wide here just means regarding the install

vispy/gloo/glir_metering.py
kephale: wip over here https://github.com/vispy/vispy/pull/2788

vispy/gloo/glir_metering.py
kephale: Pulling this out over here https://github.com/vispy/vispy/pull/2789

kephale:
okey doke, i'm trying to slim this down, and these two PRs can stand alone pretty well:

- https://github.com/vispy/vispy/pull/2788 : texture upload benchmarking
- https://github.com/vispy/vispy/pull/2789 : deduping to reduce redundant draw commands

If we can get those merged in, then I'll pull out the matching pieces in this PR and this should be much more manageable for reviewing.

PR #2788, opened 2026-09-12
Add 2D and 3D texture upload benchmark

kephale:
## Summary

- add a standalone benchmark for synchronous `Texture2D` and `Texture3D` sub-uploads
- exercise configurable upload sizes against preallocated textures while showing the sampled results
- report upload shape, byte count, median/min/max latency, effective throughput, and reproducibility metadata
- call `glFinish` after each measured draw so asynchronous driver behavior does not hide upload work

## Measurements

Command:

```console
pixi run -e py312-full python examples/benchmark/texture_upload.py
```

Environment:

- macOS 26.6.2 arm64
- Apple M4 Pro
- OpenGL 2.1 Metal 90.5
- PyQt5 backend
- Python 3.12.13
- VisPy 0.16.3.dev21; NumPy 2.5.2
- uint8, one channel, five recorded uploads per size after two warmups

| Dim | Requested (MiB) | Actual (MiB) | Upload shape | Median (ms) | Min (ms) | Max (ms) | Throughput (GiB/s) |
|---:|---:|---:|:---|---:|---:|---:|---:|
| 2D | 1 | 1.00 | 128 x 8192 | 0.83 | 0.74 | 0.90 | 1.17 |
| 2D | 2 | 2.00 | 256 x 8192 | 0.72 | 0.68 | 0.79 | 2.73 |
| 2D | 4 | 4.00 | 512 x 8192 | 0.81 | 0.68 | 0.95 | 4.81 |
| 2D | 8 | 8.00 | 1024 x 8192 | 0.74 | 0.68 | 1.14 | 10.54 |
| 2D | 16 | 16.00 | 2048 x 8192 | 0.62 | 0.57 | 1.11 | 25.40 |
| 2D | 32 | 32.00 | 4096 x 8192 | 0.80 | 0.60 | 1.02 | 39.10 |
| 2D | 64 | 64.00 | 8192 x 8192 | 1.59 | 0.64 | 2.03 | 39.27 |
| 3D | 1 | 1.00 | 4 x 512 x 512 | 0.65 | 0.59 | 0.79 | 1.51 |
| 3D | 2 | 2.00 | 8 x 512 x 512 | 0.62 | 0.59 | 0.68 | 3.17 |
| 3D | 4 | 4.00 | 16 x 512 x 512 | 0.68 | 0.65 | 0.73 | 5.74 |
| 3D | 8 | 8.00 | 32 x 512 x 512 | 0.70 | 0.67 | 0.93 | 11.23 |
| 3D | 16 | 16.00 | 64 x 512 x 512 | 0.81 | 0.70 | 1.02 | 19.22 |
| 3D | 32 | 32.00 | 128 x 512 x 512 | 0.88 | 0.71 | 0.90 | 35.34 |
| 3D | 64 | 64.00 | 256 x 512 x 512 | 0.85 | 0.83 | 0.90 | 73.30 |

The timings include drawing both sampled texture quads and a final `glFinish`; effective throughput is therefore a comparative benchmark value rather than isolated transfer bandwidth.

## Validation

- `pixi run -e lint flake8 examples/benchmark/texture_upload.py`
- CLI help smoke test
- reduced 2D/3D OpenGL smoke run
- full default 1-64 MiB benchmark run
- `git diff --check`


examples/benchmark/texture_upload.py
```diff
@@ -0,0 +1,306 @@
...
+            )
+            self._payload_key = key
+
+        # Change a texel so each iteration supplies distinct data.
+        self._payload.flat[0] ^= 1
+        texture_shape = self._texture_shapes[ndim]
+        offset = ((texture_shape[0] - shape[0]) // 2,) + (0,) * (ndim - 1)
+        self._textures[ndim].set_data(
+            self._payload, offset=offset, copy=False
+        )
```
brisvag, line 206: So, do I understand correctly that the actual upload called here is deferred until `program.draw` is called?

brisvag (commented):
I think It maakes sense to me. Just a question to make sure we are measuring the right thing.

examples/benchmark/texture_upload.py
kephale: Yes, exactly. Texture.set_data() only queues a GLIR DATA command. On the next on_draw, the first program.draw() flushes the canvas’s GLIR queue, executing the texture upload before the draw. The timer starts before that flush, and gloo.finish() waits for GPU completion. So the measurement includes the upload, both draws, clear, and finish, excluding the payload preparation + set_data() queueing.

PR #2789, opened 2026-09-12
Deduplicate redundant GLIR state commands

kephale:
## Summary

- remove repeated GLIR `FUNC` commands that set an already-requested OpenGL state within the same queued batch
- keep the cache queue-local so no state survives a flush or becomes stale after direct external OpenGL calls
- track enabled state separately for each enum passed to `glEnable` or `glDisable`
- group overlapping APIs such as `glBlendFunc` and `glBlendFuncSeparate` so transitions between them are preserved
- clear observed state at `CURRENT` context barriers
- always preserve commands that perform work, including `glClear`, `glFlush`, and `glFinish`

This extracts the GL-state suppression behavior from #2782 into an independent GLIR queue optimization with focused semantics and tests.

## Motivation

Scene drawing can enqueue the same context-state setup before many visuals. Reissuing those setters is semantically redundant and can add driver overhead or synchronization cost, particularly on macOS OpenGL over Metal.

The filtering is deliberately limited to one GLIR batch. The first setter in every flush is retained, so state changed through direct OpenGL calls between flushes is restored correctly. `FUNC` arguments are expected to follow their documented OpenGL signatures.

## Command reduction

A synthetic batch representing 1,000 draws, each preceded by the same four state setters, is reduced as follows:

```text
commands:      5000 -> 1004
state setters: 4000 -> 4
```

All 1,000 draw commands remain. This measures command reduction rather than GPU timing; the runtime benefit depends on the backend and scene.

## Validation

- `pixi run -e lint flake8 vispy/gloo/glir.py vispy/gloo/tests/test_glir.py`
- `pixi run -e lint pytest -q vispy/gloo/tests/test_glir.py vispy/gloo/tests/test_wrappers.py`
- `pixi run -e lint pytest -q vispy/gloo/tests`
- `git diff --check`

Tests cover redundant setters across draws, enable/disable transitions, changed values, overlapping state APIs, context barriers, and work commands.


vispy/gloo/glir.py
```diff
@@ -518,7 +554,42 @@ def _filter(self, commands, parser):
...
+                    key = None
+
+                if key is not None:
+                    try:
+                        hash(command)
+                    except TypeError:
+                        # Non-scalar arguments are outside the normalized
+                        # wrapper path; preserve them rather than guessing at
+                        # equality semantics.
+                        state.pop(key, None)
```
brisvag: What is "the normalized wrapper path"? And is `hash` enough to do this discrimination? One can make hashable things that are not scalars AFAICT.

brisvag (commented):
Very cool!

One question cause I don't quite understand the explanation.

kephale:
Okey updated to check for scalars + comment clarified.

vispy/gloo/glir.py
```diff
@@ -577,17 +579,19 @@ def _filter_redundant_state(commands):
...
+                        # commands from other callers when their equality
+                        # semantics are unknown.
                         state.pop(key, None)
                     else:
                         if state.get(key) == command:
                             continue
                         state[key] = command
+                elif (function in ('glEnable', 'glDisable') and
+                      len(command) == 3):
+                    # The capability cannot safely be used as a cache key.
```
brisvag, line 593: Why was this elif added with the last commit? In general you're uysing terminology that I'm struggling to follow... what's "capability" referring to here? And "emit" above?

PR #2792, opened 2026-09-15
Invalidate cached bounds when visual data changes

### LINKED ISSUE #1899: camera.set_range() does not respect image resizing
`camera.set_range()` correctly fits images to the scene upon it's initial call, but fails there after when `set_data` is called with with image data in a differing shape.

```
import numpy as np
from vispy import app, scene, io
import cv2

# Image
def func(x, y):
    return (1-x/2+x**5+y**3)*np.exp(-x**2-y**2)
x = np.linspace(-3.0, 3.0, 512).astype(np.float32)
y = np.linspace(-3.0, 3.0, 512).astype(np.float32)
X, Y = np.meshgrid(x, y)
data = func(X, Y)

    def __init__(self):
        scene.SceneCanvas.__init__(self, keys='interactive')

    # set view2's image data to the region displayed in view1
    def update_display(self) :
        # map the entirety of view1
        x1, y1, _, __ =vb1.camera._transform.imap((0,0))
        x2, y2, _, __ =vb1.camera._transform.imap((vb1.size))

        # bounds check
        x1 = int(max(0, min(x1, data.shape[0])))
        y1 = int(max(0, min(y1, data.shape[1])))
        x2 = int(max(0, min(x2, data.shape[0])))
        y2 = int(max(0, min(y2, data.shape[1])))

        # copy image data and set data
        selection = data[y2:y1, x1:x2].copy()
        image2.set_data(selection)
        image2.update()
        vb2.camera.set_range()  # fails to transfrom

    def on_mouse_release(self, e) :
        self.update_display()

    def on_mouse_wheel(self, e) :
        self.update_display()

canvas = Canvas()
canvas.size = 800, 600
canvas.show()

# Create two ViewBoxes, place side-by-side
vb1 = scene.widgets.ViewBox(parent=canvas.scene)
vb2 = scene.widgets.ViewBox(parent=canvas.scene)
#
grid.padding = 6
grid.add_widget(vb1, 0, 0)
grid.add_widget(vb2, 0, 1)

# Create the image
image1 = scene.visuals.Image(data, parent=vb1.scene)
image2 = scene.visuals.Image(data, parent=vb2.scene)

# Set 2D camera (PanZoomCamera, TurnTableCamera)
vb1.camera= scene.PanZoomCamera()
vb2.camera= scene.PanZoomCamera(interactive=False)

vb2.camera.aspect = 1  # no auto-scale

# Set the view bounds to show the entire image
vb1.camera.set_range()
vb2.camera.set_range()

if __name__ == '__main__':
    app.run()
```

Here there are 2 views, view1 is used to provide a region of interest of the image data. As view1's camera is transformed view2 receives new image data matching the region of the image displayed in view1.

The subsequent calls to `vb2.camera.set_range()` fail to re-align the transform based on the new data.

**_Quick Fix_**
For now, _my understanding of your code is lacking_, I am calling `view._bounds_changed()`  `if self._need_vertex_update `https://github.com/vispy/vispy/blob/bb2cd96eb16dde1ecac3c70f9e572791bb5699e9/vispy/visuals/image.py#L528-L533

`self._need_vertex_update` is set to `True` upon `set_data` with differing shape, so perhaps this is appropriate. 

For me though, _as I haven't had time to look over this `view` object_, it still isn't clear that this is the best place to address this issue. So for now I'll leave this as an open issue.

If you have any feedback or would like a PR of this fix just let me know.

Thanks for your time.

zephyr4289:
Fixes #1899

BaseVisual.bounds() caches computed per-axis bounds in `_vshare.bounds`, but nothing cleared that cache when a visual's data changed. As a result, calls to `camera.set_range()` or ViewBox autoscaling would keep using the first computed extent even after `set_data()` supplied new data with different dimensions (issue #1899).

`VisualShare` already declared a `bounds_change` event, but it was never emitted. Rather than adding a new caching or invalidation mechanism, this PR finishes wiring up that intended path.

### What changed

- **`BaseVisual._bounds_changed()`**: Clears `_vshare.bounds` and emits `events.bounds_change()`. It now accepts an optional `event=None` argument so it can be connected directly as an event callback.
- **`CompoundVisual` & `CompoundVisualView`**: Connected each subvisual's `bounds_change` event to `self._bounds_changed`, mirroring the existing `_subv_update` pattern so compound and view aggregates invalidate when subvisual data changes.
- **Bounds-affecting paths**: Added `_bounds_changed()` calls across visuals computing bounds from stored data: `Line`, `Markers`, `Image`, `Mesh`, `Volume`, `Text` (pos setter), `InfiniteLine`, `LinearRegion`, `Windbarb`, `ScrollingLines`, `Isocurve`, `Isoline`, and `Isosurface`.
- **Targeted invalidation**: Calls are placed only on code paths modifying geometry, coordinates, or data shapes. Cosmetic updates (such as color changes) bypass bounds invalidation.

### Related fixes

- **`BaseVisualView._compute_bounds`**: Returned the result of the viewed visual's `_compute_bounds` (previously discarded, causing views to report `None`).
- **`ScrollingLinesVisual._compute_bounds`**: Fixed an `AttributeError` referencing `self.pos_data` instead of `self._pos_data`.

### Testing

- Added `vispy/visuals/tests/test_bounds_invalidation.py` with 14 headless regression tests covering individual visuals, subvisual/compound propagation, view bounds, and event firing.
- Ran local test suites across `visuals`, `scene`, `gloo`, `util`, `color`, and `geometry` with zero new regressions.
- Verified against the full GitHub Actions CI matrix (OpenGL checks, lint/style checks, and test matrices across Python 3.10 through 3.14 on Linux/OSMesa/Qt), all passing clean.
- Verified the #1899 scenario end-to-end: calling `camera.set_range()` after `image.set_data()` correctly updates the camera rect to the new data shape instead of retaining the initial bounds.

```python
image.set_data(np.zeros((100, 100), dtype=np.float32))
camera.set_range()  # now respects the updated 100x100 extent
```


brisvag:
From the account history, this is clearly a freshly created account ran by an automated LLM agent opening hundreds of PRs in wildly unrelated projects within a few days.

I don't know if this PR contains anything useful (it might well be the right solution) but I'm going to assume it is at least as much un-reviewed slop as the PR description is.

We don't have an explicit AI policy here to point to @djhoese, but I hope you're with me here? Feel free to reopen if you disagree, but I don't want to waste my time on this :sweat_smile:

zephyr4289:
Fair heuristic honestly — there's a flood of AI slop right now and a fresh account with a polished description pattern-matches straight to it. One factual correction though: this account has opened exactly one PR ever, this one 
Screenshot_20260915_154009
https://github.com/zephyr4289. I think you've got me mixed up with someone else.

What's true from your read: I did use an LLM to draft the description. In hindsight that was a bad call, it reads exactly like the slop you're reacting to, and I should have disclosed it. What's also true: the code is mine. I hit #1899 in actual use  camera.set_range() kept the old extent after set_data() — and traced it to VisualShare.bounds_change existing but never being emitted. The diff finishes that wiring, nothing more exotic than that.

Not asking anyone to burn an afternoon re-reviewing. But if you @brisvag  or @djhoese want to spot-check whether there's a human behind this: ask me anything about the diff , design choices, what I deliberately didn't touch, perf implications of emitting on every set_data and I'll answer straight. If it still smells, keep it closed, and I'll refile the diagnosis as a plain issue with repro + the 14 regression tests so it's useful regardless of who reviews it.

djhoese:
I'm torn. Some of this can easily be explained with an inexperience with English. The PR description is AI, sure, but much better than the wall of text Claude usually wants me to put into a PR description when I use it. @brisvag I'm not getting the same exact vibe from their profile as you so I'm not sure I feel quite as strongly (hundreds of commits are not hundreds of PRs). However, your comment @zephyr4289 talks about where you used AI (the PR description), but you clearly also wrote your last comment with AI. I'll reopen this for now as it does seem like a useful set of fixes although possibly leading to a performance hit. I have at least one suggestion I'll make in a bit.

vispy/visuals/tests/test_bounds_invalidation.py
djhoese, line 1: Do we have other visuals test modules like this where a single module tests a ton of individual modules? If so, then OK. But it feels like maybe these tests should go in the individual test modules for each Visual class.

brisvag:
My bad, I saw commit count, not PR count.

Even so, my point still stands, and is further confirmed by _so many things_ in the writing style of the above comment, the bot's account history, description, previous activity, that I don't think there's a point in listing them all and I'd be talking to the wind or any human who happens to be passing by :)

EDIT: I was about to send but you were faster @djhoese. Feel free to take this on, I have no patience for this right now, sorry :sweat_smile:

zephyr4289:
> I'm torn. Some of this can easily be explained with an inexperience with English. The PR description is AI, sure, but much better than the wall of text Claude usually wants me to put into a PR description when I use it. @brisvag I'm not getting the same exact vibe from their profile as you so I'm not sure I feel quite as strongly (hundreds of commits are not hundreds of PRs). However, your comment @zephyr4289 talks about where you used AI (the PR description), but you clearly also wrote your last comment with AI. I'll reopen this for now as it does seem like a useful set of fixes although possibly leading to a performance hit. I have at least one suggestion I'll make in a bit.

Thanks @djhoese for your time in this. You're right, I've also used AI assistance to generate my last comment, not only for the PR description. My English is not the best and I wanted to be clear, but clearly, it backfired and made it appear as if there's no human intellect behind the comment at all, which is not the case. I'm sorry for that and from now I will write my comments here manually, even if my English is not perfect.

As for the fix itself, that's my own code and I'm open to any questions regarding the implementation. Looking forward to your suggestions. As for the performance issue you've mentioned, I'll make some benchmarks and report the concrete values.

vispy/visuals/tests/test_bounds_invalidation.py
zephyr4289: Yes, . The test code for individual visual bounds invalidation is now in the relevant test modules (test_image.py, test_mesh.py, test_text.py, etc.). The remaining generic tests for BaseVisual, CompoundVisual, view forwarding, and events are in vispy/visuals/tests/test_visual.py.

PR #2793, opened 2026-09-16
Fix Agg-based line vertex shader (end caps/joins behave strangely at some scales)

nathancartlidge:
Hi! This PR should resolve an issue with 'agg'-mode line rendering where at some zoom levels the joins/caps of lines can behave strangely.

I haven't tested it very thoroughly (and am not great at GLSL!), but thought this was worth contributing back.

Before:
before

After:
after


brisvag (commented):
Ah, thanks @nathancartlidge, I was trying to figure this out myself recently and finally gave up 😅 This looks great!

I'm gonna circle back to this soon and play with various use cases to make sure it doesn't do anything weird, but I'm excited to see these screenshots 💪

vispy/glsl/lines/agg.vert
```diff
@@ -90,9 +90,26 @@ void main()
 
     // Attributes to varyings
     v_angles  = a_angles;
-    //v_segment = a_segment * u_scale.x * tr_scale.x;  // TODO: proper scaling
-    //v_length  = v_length * u_scale * tr_scale;  // TODO: proper scaling
```
brisvag, line 94: You basically address these 2 todos (from 2014 😱) with this PR, as I understand it?

brisvag (approved):
I tried it with the code from my recent PR at https://github.com/vispy/vispy/pull/2734, and it puts those screenshots to shame... So _that_'s what they're supposed to look like!

Image

Image

I tried a few examples that use agg lines (bezier, line, arrows), they all look good/better.

Composite visuals also improved: if you add `border_method='agg'` to each visual inside `examples/scene/polygon.py`, on main it's awful, but on this PR it looks great.

So, looks good to me!

vispy/glsl/lines/agg.vert
nathancartlidge: wow, I didn't check the blame on these so didn't see how much history this bug had! 12 years later, finally (hopefully) fixed :)

PR #2794, opened 2026-09-17
Enhance set_data with offset parameter

### LINKED ISSUE #2787: VolumeVisual's set_data function doesn't accept the offset value.
I was working with the VolumeVisual and needed to update only a specific portion of the volume. I looked for solutions on the internet and found that there is a technique to perform sub-buffer uploads for partial texture updates using an offset value.

Afterwards, I casually started to explore how VisPy's VolumeVisual sets data. I was aware that it updated the entire volume on each set_data call; however, I observed that the VolumeVisual's set_data function calls the "scale_and_set_data" functions from either the GPUScaledTextured3D or the CPUScaledTextured3D. These functions accept the offset value, which I reckon is for a partial texture update, though I'm not sure if it's being used or not; either way, it's hardcoded to None.

Is there any solution for this? I can't understand how the scale_and_set_data function works in the _ScaledTextureMixin class.

@brisvag

PythoneerSamurai:
Closes https://github.com/vispy/vispy/issues/2787

Added an optional offset parameter to set_data method for specifying the starting point in the texture for data copying. Restructured the set_data method to prevent errors that would be caused by setting _last_data, clim, and u_shape based on a data sub-block rather than the entire volume, in case of a partial texture upload.

@brisvag can you look into this?


vispy/visuals/volume.py
```diff
@@ -882,19 +883,20 @@ def set_data(self, vol, clim=None, copy=True):
...
+        if offset is None:
+            self._last_data = vol
+            self.shared_program['clim'] = self._texture.clim_normalized
+            self.shared_program['u_shape'] = (vol.shape[2], vol.shape[1],
+                                              vol.shape[0])
+            shape = vol.shape[:3]
+            if self._vol_shape != shape:
+                self._vol_shape = shape
+                self._need_vertex_update = True
             self._vol_shape = shape
```
brisvag, line 901: I think this is not quite right. We need this to be a bit smarter and actually update the `_last_data[...]` based on offset.

The only part that should be guarded behind `if offset is None` is the stuff regarding `shape`, but clims should be recomputed IMO.

Also, we might have to add an extra guard at the top that ensures that the input data fits in the remainder of the volume starting from `offset` (though maybe this guard already exists in `Texture`?). Did you test some edge cases?

brisvag (commented):
Looking good, thanks @PythoneerSamurai! Some suggestions :)

PythoneerSamurai:
> Looking good, thanks @PythoneerSamurai! Some suggestions :)

Thanks for checking it out.
Yeah, you're right. I'll look into this, God willing.

PythoneerSamurai:
> Looking good, thanks @PythoneerSamurai! Some suggestions :)

Below are my observations and fixes based upon your suggestions.
1. Addressing the last point regarding the addition of a guard to check if the data sub-block fits in the actual data in case an offset is provided. This check is already present in the BaseTexture's _set_data function (which is the one that's eventually called).  The following code is used to verify this.
```

for i in range(len(data.shape)-1):
            if offset[i] + data.shape[i] > self._shape[i]:
                raise ValueError("Data is too large")
```
In case an offset is provided, I've simply moved the self._texture's scale_and_set_data function call above the computation of _last_data and clim. I've checked the said function's definition (and the functions it calls), to the best of my observation, and they don't seem to be making use of _last_data. Thus, this function call can possibly be moved without causing issues.

2. You're right regarding the _last_data and clim computations. In my latest commit, I've added the logic to update the _last_data with the data sub-block based on the offset. The clim is computed afterwards.

Kindly look into these.
Thanks.

@brisvag

brisvag (commented):
Nice, I think this is good, thanks @PythoneerSamurai!

Would you like to give the same treatment to `Image` since we're here?  Feel free to say no, In that case I'll folow up with that.

It would be nice to add some test for this as well inside `test_volume.py`!

PythoneerSamurai:
> Nice, I think this is good, thanks @PythoneerSamurai!
> 
> Would you like to give the same treatment to `Image` since we're here?  Feel free to say no, In that case I'll folow up with that.
> 
> It would be nice to add some test for this as well inside `test_volume.py`!

Sure, I'll look into these soon. Currently, I'm occupied in some other tasks, hopefully they'll be done in a week or two. While I'm at it, I'll also have a look at markers, perhaps they can also take sub data uploads.

